PR #4, opened 2020-12-01
Use relative submodule paths

hclarke:
This allows easier cloning from mirrored repos and through other protocols (ssh)


qingling128 (approved):
LGTM

qingling128:
Tested locally that this worked.

PR #6, opened 2020-12-15
Support windows event log in fluent-bit

sophieyfang:
This PR will 

- add a new receiver type as [windows_event_log](https://docs.fluentbit.io/manual/pipeline/inputs/windows-event-log)
- add a default windows configuration file (only logging section, metrics sections will be there when i start on OTEL work)
- add fluent-bit test for this new feature
- NOT add any tests in confgenerator_test.go yet as the structure is for testing both logging and metrics. Though i can supply an empty golden. Also considering now the default metrics is in collectd. I need to think better on how to decouple that and add otel as the replacement. And this work will come around when i start on OTEL work.

Note that i use winlog as the name in many places as that's the name of the fluent-bit windows event log input plugin.

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -212,7 +212,6 @@ func extractReceiverFactories(receivers map[string]*receiver) (map[string]*fileR
 	fileReceiverFactories := map[string]*fileReceiverFactory{}
 	syslogReceiverFactories := map[string]*syslogReceiverFactory{}
 	winlogReceiverFactories := map[string]*winlogReceiverFactory{}
-	fmt.Print(receivers)
 	for n, r := range receivers {
 		if strings.HasPrefix(n, "lib:") {
 			return nil, nil, nil, fmt.Errorf(`receiver id prefix 'lib:' is reserved for pre-defined receivers. Receiver ID %q is not allowed.`, n)
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -157,12 +157,12 @@ func generateFluentBitConfigs(logging *logging, logsDir string, stateDir string)
 	fbFilterRemoveLogNames := []*conf.FilterModifyRemoveLogName{}
 
 	if logging.Service != nil {
-		fileReceiverFactories, syslogReceiverFactories, winlogReceiverFactories, err := extractReceiverFactories(logging.Receivers)
+		fileReceiverFactories, syslogReceiverFactories, wineventlogReceiverFactories, err := extractReceiverFactories(logging.Receivers)
 		if err != nil {
 			return "", "", err
 		}
 		extractedTails := []*conf.Tail{}
-		extractedTails, fbSyslogs, fbWinlogs, err = generateFluentBitInputs(fileReceiverFactories, syslogReceiverFactories, winlogReceiverFactories, logging.Service.Pipelines, stateDir)
+		extractedTails, fbSyslogs, fbWinlogs, err = generateFluentBitInputs(fileReceiverFactories, syslogReceiverFactories, wineventlogReceiverFactories, logging.Service.Pipelines, stateDir)
 		if err != nil {
 			return "", "", err
 		}
@@ -204,14 +204,14 @@ type fileReceiverFactory struct {
 	ExcludePaths []string
 }
 
-type winlogReceiverFactory struct {
+type wineventlogReceiverFactory struct {
 	Channels []string
 }
 
-func extractReceiverFactories(receivers map[string]*receiver) (map[string]*fileReceiverFactory, map[string]*syslogReceiverFactory, map[string]*winlogReceiverFactory, error) {
+func extractReceiverFactories(receivers map[string]*receiver) (map[string]*fileReceiverFactory, map[string]*syslogReceiverFactory, map[string]*wineventlogReceiverFactory, error) {
 	fileReceiverFactories := map[string]*fileReceiverFactory{}
 	syslogReceiverFactories := map[string]*syslogReceiverFactory{}
-	winlogReceiverFactories := map[string]*winlogReceiverFactory{}
+	wineventlogReceiverFactories := map[string]*wineventlogReceiverFactory{}
 	for n, r := range receivers {
 		if strings.HasPrefix(n, "lib:") {
 			return nil, nil, nil, fmt.Errorf(`receiver id prefix 'lib:' is reserved for pre-defined receivers. Receiver ID %q is not allowed.`, n)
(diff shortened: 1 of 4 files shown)


confgenerator/confgenerator.go
```diff
@@ -147,18 +150,19 @@ func generateFluentBitConfigs(logging *logging, logsDir string, stateDir string)
...
 	fbSyslogs := []*conf.Syslog{}
+	fbWinlogs := []*conf.WindowsEventlog{}
 	fbFilterParsers := []*conf.FilterParser{}
 	fbFilterAddLogNames := []*conf.FilterModifyAddLogName{}
 	fbFilterRewriteTags := []*conf.FilterRewriteTag{}
 	fbFilterRemoveLogNames := []*conf.FilterModifyRemoveLogName{}
 
 	if logging.Service != nil {
-		fileReceiverFactories, syslogReceiverFactories, err := extractReceiverFactories(logging.Receivers)
+		fileReceiverFactories, syslogReceiverFactories, winlogReceiverFactories, err := extractReceiverFactories(logging.Receivers)
```
qingling128, line 160: Since there are also windows file based logs, can we include "event" in the naming to help distinguish them? e.g. "winevent"

confgenerator/confgenerator.go
```diff
@@ -200,53 +204,84 @@ type fileReceiverFactory struct {
...
 	fileReceiverFactories := map[string]*fileReceiverFactory{}
 	syslogReceiverFactories := map[string]*syslogReceiverFactory{}
+	winlogReceiverFactories := map[string]*winlogReceiverFactory{}
 	for n, r := range receivers {
 		if strings.HasPrefix(n, "lib:") {
-			return nil, nil, fmt.Errorf(`receiver id prefix 'lib:' is reserved for pre-defined receivers. Receiver ID %q is not allowed.`, n)
+			return nil, nil, nil, fmt.Errorf(`receiver id prefix 'lib:' is reserved for pre-defined receivers. Receiver ID %q is not allowed.`, n)
 		}
 		switch r.Type {
 		case "files":
```
qingling128, line 220: [Not directly related to this PR] We don't have to deal with this in this PR.

It may be more convenient to do an `inclusive` check than the `exclusive` check here for future convenience. With `inclusive` checks, we do not have to touch every case when adding new params.

fluentbit/conf/conf_test.go
```diff
@@ -798,12 +850,155 @@ func TestGenerateFluentBitMainConfigErrors(t *testing.T) {
...
+		tails   []*Tail
+		winlogs []*WindowsEventlog
+		want    string
+	}{
+		{
+			name: "multiple tail and winlog plugins",
+			tails: []*Tail{{
+				Tag:  "test_tag1",
+				DB:   "test_db1",
+				Path: "test_path1",
```
qingling128, line 871: Curious: Was there any special logic we need to adjust to make Windows style file path work? e.g. `C:\Program Files...`

sophieyfang: I don't think it has some special logic. any `/var/log` will go under C:\ i think. For example, for google-fluentd, we have this `buffer_path /var/log/google-fluentd/buffers` https://github.com/GoogleCloudPlatform/google-fluentd/blob/master/windows-installer/fluent-template.conf#L32 and it just creates in `C:\var\log\google-fluentd\buffers`

StevenYCChou: It's different template engine so you might need to confirm first.

It's possible that you might need to use `filepath.Dir` to render the path correctly on Windows, including the code for other paths.

sophieyfang:
Several things found out after i submit the PR.

1. current release including 1.6.8 doesn't have support for storage.type filesystem. Impact is we will have to buffer into memory. Thus waiting for 1.6.9 to come out
2. windows event log plugin doesn't have tag as a field so it has trouble to get caught by a designated stackdriver output plugin
3. tail plugin seems not working right in windows.

fluentbit/conf/conf_test.go
sophieyfang: I verified that in_tail plugin and in_winlog plugin running on Windows understand file paths  "/var/log/file1" and "C:\var\log\file1". Therefore, I think we are good with the current code. The only diff i noticed is that when tailing multiple files, for  syntax - "/var/log_dir/file1", "/var/log_dir/file2", you have to list out all file name one by one. To allow wildcard, you can only use "C:\var\log_dir\*".

confgenerator/confgenerator.go
sophieyfang: Done

confgenerator/confgenerator.go
sophieyfang: ACK

sophieyfang:
ling, dave and yen, you can review the PR with assuming that below two issues will be fixed. As those two though the blockers and we have to ship with them fixed which means fluent-bit on Windows conf will stay the same as this PR has.
- storage in filesystem will be there: https://github.com/fluent/fluent-bit/issues/1694
- the fix for the breaking bug in the stackdriver exporter will be there: https://github.com/fluent/fluent-bit/issues/2880

sophieyfang:
I managed to build .exe out of the master branch which includes supporting filesystem on Windows and a bug fix on the failing to connect to the metadata server. I have just tested that both issues are resolved with the code in master branch✌️

confgenerator/confgenerator.go
```diff
@@ -147,18 +150,19 @@ func generateFluentBitConfigs(logging *logging, logsDir string, stateDir string)
...
 
 	if logging.Service != nil {
-		fileReceiverFactories, syslogReceiverFactories, err := extractReceiverFactories(logging.Receivers)
+		fileReceiverFactories, syslogReceiverFactories, wineventlogReceiverFactories, err := extractReceiverFactories(logging.Receivers)
 		if err != nil {
 			return "", "", err
 		}
 		extractedTails := []*conf.Tail{}
-		extractedTails, fbSyslogs, err = generateFluentBitInputs(fileReceiverFactories, syslogReceiverFactories, logging.Service.Pipelines, stateDir)
+		extractedTails, fbSyslogs, fbWinlogs, err = generateFluentBitInputs(fileReceiverFactories, syslogReceiverFactories, wineventlogReceiverFactories, logging.Service.Pipelines, stateDir)
```
qingling128, line 165: s/fbWinlogs/fbWinEvents?

fluentbit/conf/conf.go
```diff
@@ -260,6 +272,7 @@ const (
 type mainConfigSections struct {
 	TailConfigSections                      []string
 	SyslogConfigSections                    []string
+	WineventlogConfigSections                    []string
```
qingling128, line 275: There seems to be some extra spaces here.

fluentbit/conf/conf.go
```diff
@@ -668,6 +690,39 @@ func (s Syslog) renderConfig() (string, error) {
 	return renderedSyslogConfig.String(), nil
 }
 
+// A WindowsEventlog represents the configuration data for fluentbit's winlog input plugin
+type WindowsEventlog struct {
+        Tag string
+	Channels string
```
qingling128, line 696: Missing 2 spaces it seems. May want to set up `Run goimports` as instructed in the developer guide to use the tool to automatically adjust formatting.

fluentbit/conf/conf_test.go
```diff
@@ -530,6 +530,70 @@ func TestSyslogErrors(t *testing.T) {
...
+func TestWinlogErrors(t *testing.T) {
+	tests := []struct {
+		name   string
+		wineventlog WindowsEventlog
+	}{
+		{
+			name: "empty channels",
+			wineventlog: WindowsEventlog{
+				Tag:          "test_tag",
+				Channels:     "",
```
qingling128, line 575: Do we error out as expected when `Channels` is not a list of strings separated by `,`? e.g. "System;Applicationn", "System@Application"

qingling128 (approved):
Some minor comments. The rest LGTM

fluentbit/conf/conf_test.go
sophieyfang: Channels in winlog plugin or Path in tail plugin both takes comma as separator.  I don't see we have any @ or ; check for the existing tail plugin. If we are going to have such check, we better do it with all fields that only allow comma separated list. I will add a ToDO.

PR #7, opened 2020-12-15
Update fluent-bit to v1.6.8

hsmatulisgoogle:


sophieyfang:
Before we have integrated github kokoro test, we shall always have a bug ticket that link to the PR.
Within the bug ticket:

- create the bug ticket under the right component
- add it to hotlist if there is one
- describe what's the bug ticket for
- link the PR to the bug ticket
- document down the steps you took
- attach testing evidence
- finally link the bug ticket here in format of b/1823238 as an example

sophieyfang:
Have you tested e2e tests across all distros that it supports?

qingling128:
This is obsolete now. https://github.com/GoogleCloudPlatform/ops-agent/pull/24 has updated Fluent Bit to 1.7-rc2+

PR #10, opened 2021-01-04
Support OTEL

sophieyfang:
Adding support for converting upper-level metrics configuration to otel understandable configuration.
It includes:
- A separate folder called as "otel"
- A unit test under folder otel/

Also as i added "otel" in main.go, i've tested generating the otel.conf manually by executing: `go run -mod=mod cmd/generate_config/main.go --service=otel --in=confgenerator/windows-default-config.yaml --out=/tmp/google-cloud-ops-agent`. See b/176757501#comment2 for generated otel conf

It doesn't include:
-  tests under confgenerator/ yet 
This will come up in a separate PR as the test is for both logging and metrics.There are things i need to clean up before writing tests for both. I will need to first submit a PR to get rid of the tail collectd part for Windows. As windows uses OTEl and OTEL metrics agent emits its own log via windows event log so no need for tail any log files for its own log. After that PR, i will write the tests under confgenerator folder)
[](url)

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -434,10 +434,10 @@ func generateOtelReceivers(hostmetricsReceiverFactories map[string]*hostmetricsR
 	for _, pID := range pipelineIDs {
 		p := pipelines[pID]
 		for _, rID := range p.Receivers {
-			if _, ok := hostmetricsReceiverFactories[rID]; ok {
+			if h, ok := hostmetricsReceiverFactories[rID]; ok {
 				hostMetrics := otel.HostMetrics{
 					HostMetricsID:      rID,
-					CollectionInterval: "1",
+					CollectionInterval: h.CollectionInterval,
 				}
 				hostMetricsList = append(hostMetricsList, &hostMetrics)
 				receiveNameMap[rID] = "hostmetrics/" + rID
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -95,8 +95,13 @@ type otelReceiver struct {
         CollectionInterval string `yaml:"collection_interval"`
 }
 
+type otelExporter struct {
+        Type string `yaml:"type"`
+}
+
 type otelMetrics struct {
-        Receivers map[string]*otelReceiver `yaml:"receivers"`
+    Receivers map[string]*otelReceiver `yaml:"receivers"`
+    Exporters map[string]*otelExporter `yaml:"exporters"`
 	Service *otelService `yaml:"service"`
 }
 
@@ -168,20 +173,23 @@ func unifiedConfigWindowsReader(input []byte) (unifiedConfigWindows, error) {
 
 func generateOtelConfig(metrics *otelMetrics, logsDir string) (string, error){
 	hostMetricsList := []*otel.HostMetrics{}
+	stackdriverList := []*otel.Stackdriver{}
 
 	if metrics.Service != nil {
 		hostmetricsReceiverFactories, err := extractOtelReceiverFactories(metrics.Receivers)
 		if err != nil {
 			return "", err
 		}
-		fmt.Print(hostmetricsReceiverFactories)
 		hostMetricsList, err = generateOtelReceivers(hostmetricsReceiverFactories, metrics.Service.Pipelines)
 		if err != nil {
 			return "", err
 		}
-		//exporters
+		stackdriverList, err = generateOtelExporters(metrics.Exporters, metrics.Service.Pipelines)
+		if err != nil {
+			return "", err
+		}
 	}
-	otelConfig, err := otel.GenerateOtelConfig(hostMetricsList)
+	otelConfig, err := otel.GenerateOtelConfig(hostMetricsList, stackdriverList)
 	if err != nil {
 		return "", err
 	}
@@ -390,6 +398,30 @@ func generateOtelReceivers(hostmetricsReceiverFactories map[string]*hostmetricsR
 		}
 	}
 	return hostMetricsList, nil
+}
+
+func generateOtelExporters(exporters map[string]*otelExporter, pipelines map[string]*otelPipeline) ([]*otel.Stackdriver, error) {
+	stackdriverList := []*otel.Stackdriver{}
+	var pipelineIDs []string
+	for p := range pipelines {
+		pipelineIDs = append(pipelineIDs, p)
+	}
+	sort.Strings(pipelineIDs)
+	for _, pID := range pipelineIDs {
+		p := pipelines[pID]
+		for _, rID := range p.Receivers {
+			if _, ok := exporters[rID]; ok {
+				stackdriver := otel.Stackdriver{
+					UserAgent: "$USERAGENT",
+					Prefix: "agent.googleapis.com/",
+				}
+				stackdriverList = append(stackdriverList, &stackdriver)
+				continue
+			}
+			return nil, fmt.Errorf(`receiver %q of pipeline %q is not defined`, rID, pID)
+		}
+	}
+	return stackdriverList, nil	
 }
 
 func generateFluentBitInputs(fileReceiverFactories map[string]*fileReceiverFactory, syslogReceiverFactories map[string]*syslogReceiverFactory, wineventlogReceiverFactories map[string]*wineventlogReceiverFactory, pipelines map[string]*loggingPipeline, stateDir string) ([]*conf.Tail, []*conf.Syslog, []*conf.WindowsEventlog, error) {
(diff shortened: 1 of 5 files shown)


otel/conf.go
```diff
@@ -0,0 +1,584 @@
...
+package otel
+
+import (
+	"fmt"
+	//	"net"
+	"strings"
+	"text/template"
+)
+
+const (
```
davidbtucker, line 25: Question: Would it be possible to use OTel's YAML structs and then just serialize those directly, instead of defining these text templates?

confgenerator/confgenerator.go
```diff
@@ -31,6 +32,11 @@ type unifiedConfig struct {
 	Metrics collectd.Metrics `yaml:"metrics"`
 }
 
+type unifiedConfigWindows struct {
```
davidbtucker, line 35: I think we should be using the same config format for all platforms.  Can we just generate the OTel config from the existing unifiedConfig struct?

davidbtucker (commented):
Hi, a couple of high-level questions.

confgenerator/confgenerator.go
sophieyfang: Existing UnifiedConfig uses collectd in "yaml:metrics"..i am not sure how i can reuse it. Any ideas?

sophieyfang: Basically otel metrics and collectd metrics, they use the same interface. So there is only one set of Metrics definition needed. I can send a refactor PR afterwards. WDYT?

sophieyfang: Sync'ed offline that i will send a follow-up PR for the refactoring

otel/conf.go
sophieyfang: I can give it a try. For the default configuration, e.x. metrics agent itself's metrics which is constant string blob now, How it can be fitted inside the yaml structs. I will let you know how it goes

sophieyfang: Thinking this can be quite an important discussion; thus i opened a section in the Design Doc discussing it. Also cite the same discussion below as i don't want to link the internal doc bookmark in public.

```
First, OTEL configuration is yaml formatted. It's definitely great that if we can use yaml struct for the implementation which just is more natural way of implementing it. 

However, the[ default 500+ lines windows configuration](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/master/config/config-windows.yaml) just makes it quite cumbersome to convert each single OTEL plugin(receiver, processor, exporter and extension) as a struct. 

To give an example, the complicated way that we have for processing the system metrics in the default config: https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/master/config/config-windows.yaml#L92. Converting this 200+ lines of metricstransformer into a yaml struct will need quite some work (Create each metric its own sub struct type variable then append all of it together under the parent struct variable). Since processes metrics are default and free; also we are not exposing it to the end-user for re-configuring, so do we want to spend time on converting it towards a yaml struct that might/might not be worth it.  

As it's impossible to have the yaml partially made by string blob and paritally made by yaml struct, we need to decide whether we want to use string blob or yaml structs. Thinking about the long term (when customer is able to customize more things), it might be better to spend sometime and implement the yaml struct out despite the work amount of doing so plus it's unseen by client..

I will leave the decision to Ling and Dave.
```

confgenerator/confgenerator.go
```diff
@@ -278,6 +423,59 @@ func extractReceiverFactories(receivers map[string]*receiver) (map[string]*fileR
 	return fileReceiverFactories, syslogReceiverFactories, wineventlogReceiverFactories, nil
 }
 
+func generateOtelReceivers(hostmetricsReceiverFactories map[string]*hostmetricsReceiverFactory, pipelines map[string]*otelPipeline) ([]*otel.HostMetrics, map[string]string, error) {
+	hostMetricsList := []*otel.HostMetrics{}
+	receiveNameMap := make(map[string]string)
```
davidbtucker, line 428: receiverNameMap

otel/conf.go
davidbtucker: Also, we can explore in a follow-up PR, don't need to block this one.

confgenerator/confgenerator.go
```diff
@@ -278,6 +423,59 @@ func extractReceiverFactories(receivers map[string]*receiver) (map[string]*fileR
...
+				stackdriver := otel.Stackdriver{
+					StackdriverID: eID,
+					UserAgent:     "$USERAGENT",
+					Prefix:        "agent.googleapis.com/",
+				}
+				stackdriverList = append(stackdriverList, &stackdriver)
+				exportNameMap[eID] = "stackdriver/" + eID
+				continue
+			}
+			return nil, nil, fmt.Errorf(`exporter %q of pipeline %q is not defined`, eID, pID)
```
davidbtucker, line 473: Should this check be outside the inner for loop? (Same above.)

davidbtucker (changes requested):

confgenerator/confgenerator.go
```diff
@@ -118,6 +162,88 @@ func unifiedConfigReader(input []byte) (unifiedConfig, error) {
...
+func generateOtelServices(receiverNameMap map[string]string, exporterNameMap map[string]string, pipelines map[string]*otelPipeline) ([]*otel.Service, error) {
+	serviceList := []*otel.Service{}
+	var pipelineIDs []string
+	for p := range pipelines {
+		pipelineIDs = append(pipelineIDs, p)
+	}
+	sort.Strings(pipelineIDs)
+	for _, pID := range pipelineIDs {
+		p := pipelines[pID]
+		var pRecevieIDs []string
```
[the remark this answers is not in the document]
sophieyfang: i know i have typo somewhere. it showed on intellij, then it slipped away. i globally seached for "reci" and "Reci". Didn' expect it could be more wrong than that..

otel/conf.go
sophieyfang: SGTM.  left a todo in b/176757501#comment3

confgenerator/confgenerator.go
```diff
@@ -118,6 +162,88 @@ func unifiedConfigReader(input []byte) (unifiedConfig, error) {
...
+		if err != nil {
+			return "", err
+		}
+		serviceList, err = generateOtelServices(receiverNameMap, exporterNameMap, metrics.Service.Pipelines)
+	}
+	otelConfig, err := otel.GenerateOtelConfig(hostMetricsList, stackdriverList, serviceList)
+	if err != nil {
+		return "", err
+	}
+	fmt.Print(otelConfig)
```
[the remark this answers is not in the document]
sophieyfang: Yes! I will remove it.

confgenerator/confgenerator.go
sophieyfang: i think this is right place. The logic resembles here in logging side: https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/confgenerator.go#L325

First, looping a list of pipeLines then for each pipeline,  for looping this pipeline's exporter IDs, trying to find this ID inside the exporters. If it's not found, return the error.

davidbtucker: Oh sorry, I misread this code, thanks for the clarification.

I think it might be clearer to invert this condition:

    if _, ok := exporters[eID]; !ok {
      return nil, nil, ...
    }
    stackdriver := otel.Stackdriver{...}
    ...

davidbtucker (approved):
LGTM with one suggestion.

confgenerator/confgenerator.go
sophieyfang: Yes, this way is much clearer! Done

otel/conf.go
```diff
@@ -0,0 +1,584 @@
...
+      exporters:
+        - stackdriver/agent`
+
+	serviceConf = `metrics/{{.ID}}:
+      receivers:  {{.Receivers}}
+      processors: {{.Processors}}
+      exporters: {{.Exporters}}`
+
+	defaultProcessorConf = `resourcedetection:
+    detectors: [gce, ec2]
```
qingling128, line 95: Consider removing `ec2`?

otel/conf.go
```diff
@@ -0,0 +1,584 @@
...
+    metrics:
+      include:
+        match_type: strict
+        metric_names:
+          - otelcol_process_uptime
+          - otelcol_process_memory_rss
+          - otelcol_grpc_io_client_completed_rpcs
+          - otelcol_googlecloudmonitoring_point_count
+
+  # convert from windows perf counter formats to cloud monitoring formats
```
qingling128, line 386: Is my understanding correct that the `metricstransform/iis` and `metricstransform/mssql` here define the conversion. But because we have not yet enabled the receivers for `iis` and `mssql` on purpose, these are effectively no-op?

qingling128 (approved):
LGTM with a minor suggestion and a question for clarification

otel/conf.go
sophieyfang: Yes, it will only be effective when it's used in the pipelines.

otel/conf.go
sophieyfang: Done

qingling128 (approved):
LGTM

PR #11, opened 2021-01-10
Add valid tests for upper level config

sophieyfang:
What this PR includes:
- remove logName for otel, as otel sends its own logs as eventlog so it has no log files
- cleanup the ec2 as we don't support it
- tweaked the confgenerator_test.go a little bit so as it can share most of the code but takes windows test a slightly different from LInux's
- add valid test cases for windows

What this PR doesn't include:
- invalid test cases are still unclear to me. That will need some internal discussion and will come in as a separate PR.


confgenerator/testdata/valid/windows_metrics-no_conf/golden_otel.conf
```diff
@@ -0,0 +1,402 @@
+receivers:
+  prometheus/agent:
+    config:
+      scrape_configs:
+      - job_name: 'otel-collector'
+        scrape_interval: 1m
+        static_configs:
+        - targets: ['0.0.0.0:8888']
+  
+processors:
```
sophieyfang, line 10: I basically dump all processors required for the default otel windows to function. As long as it's not used by any pipelines, it's simply no-op. but worth asking @james-bebbington if defining the processor will cost any resources for void.

confgenerator/testdata/valid/windows_default_config/golden_fluent_bit_main.conf
```diff
@@ -0,0 +1,160 @@
...
+    storage.type       filesystem
+
+    # https://docs.fluentbit.io/manual/administration/backpressure#mem_buf_limit
+    # This controls how much data the input plugin can hold in memory once the data is ingested into the core.
+    # This is used to deal with backpressure scenarios (e.g: cannot flush data for some reason).
+    # When the input plugin hits "mem_buf_limit", because we have enabled filesystem storage type, mem_buf_limit acts
+    # as a hint to set "how much data can be up in memory", once the limit is reached it continues writing to disk.
+    Mem_Buf_Limit      10M
+
+[INPUT]
```
sophieyfang, line 60: this blob for collectd is no-op. I have a follow-up PR to clean it up once we introduce a isWindows flag to handle the   difference between Linux and Windows (also Path name)

confgenerator/confgenerator_test.go
```diff
@@ -63,46 +65,67 @@ func TestGenerateConfsWithValidInput(t *testing.T) {
 				unifiedConfigFilePath = "default-config.yaml"
 			}
 
+			if d.Name() == "windows_default_config" {
```
davidbtucker, line 68: Nit: Do `else if` or use a `switch` statement.

confgenerator/confgenerator_test.go
```diff
@@ -63,46 +65,67 @@ func TestGenerateConfsWithValidInput(t *testing.T) {
...
 			// Compare the expected and actual and error out in case of diff.
 			updateOrCompareGolden(testName, expectedMainConfig, mainConf, goldenMainPath, t)
 			updateOrCompareGolden(testName, expectedParserConfig, parserConf, goldenParserPath, t)
-			updateOrCompareGolden(testName, expectedCollectdConfig, collectdConf, goldenCollectdPath, t)
+
+			if isWindows {
+				expectedOtelConfig := expectedConfig(testName, goldenOtelPath, t, true)
+				otelConf, err := GenerateOtelConfig(unifiedConfig)
+				if err != nil {
+					t.Errorf("test %q: expect no error, get error running GenerateOtelConfig : %s", testName, err)
```
davidbtucker, line 97: Instead of `t.Errorf()` and `return`, I think you can just do `t.Fatalf()`.

confgenerator/testdata/valid/windows_logging-multiple_file_sources/input.yaml
```diff
@@ -0,0 +1,69 @@
...
+  exporters:
+    google:
+      type: google_cloud_logging
+  service:
+    pipelines:
+      default_pipeline:
+        receivers: [windows_event_log]
+        exporters: [google]
+      pipeline1:
+        receivers:
```
davidbtucker, line 45: Nit: Use the `[...]` syntax here.

confgenerator/testdata/valid/windows_logging-no_conf/input.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright 2020 Google LLC
```
davidbtucker, line 1: Why are the golden files for this test empty?

sophieyfang: 'cause the logging is not defined thus it's expected being empty. See also here:  https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/testdata/valid/logging-no_conf/golden_fluent_bit_main.conf

confgenerator/testdata/valid/windows_default_config/golden_fluent_bit_main.conf
sophieyfang: This will be a follow up PR adding isWindows flag

confgenerator/testdata/valid/windows_metrics-no_conf/golden_otel.conf
sophieyfang: This is noop and cost no resources if not used in pipelines.

confgenerator/testdata/valid/windows_logging-multiple_file_sources/input.yaml
sophieyfang: Next time.

davidbtucker (approved):

PR #12, opened 2021-01-12
Add invalid test for windows

sophieyfang:
## Add invalid tests for windows

- "lib:" is reserved thus we ban client from using lib: receiver_id, export_id and pipeline_id.
- interval can't be "10" it has to end with meaningful unit type, e.x. "s" or "m".
- interval can't be less than 10 section.


confgenerator/confgenerator_test.go
```diff
@@ -160,6 +160,10 @@ func TestGenerateConfigsWithInvalidInput(t *testing.T) {
 				if _, _, err := GenerateFluentBitConfigs(unifiedConfig, defaultLogsDir, defaultStateDir); err == nil {
 					t.Errorf("test %q: GenerateFluentBitConfigs succeeded, want error. file:\n%s", testName, unifiedConfig)
 				}
+			} else if strings.Contains(testName, "windows") {
+				if _, err := GenerateCollectdConfig(unifiedConfig, defaultLogsDir); err == nil {
```
qingling128, line 164: Should this be `GenerateOtelConfigs`? Looks like it's about Windows.

sophieyfang: Nice catch :)

qingling128 (approved):
LGTM

PR #14, opened 2021-01-13
Set default logging config (when logging isn't specified)

sophieyfang:
default logging shall track agents' logs. Even if logging section is totally missing in the upper level conf, it shall still collect agents' logs.


qingling128 (approved):
LGTM

PR #15, opened 2021-01-14
Support windows mssql, iis metrics

sophieyfang:
This is the conf that we agreed upon:

```
metrics:
  receivers:
    hostmetrics:
      type: hostmetrics
      collection_interval: 60s
    mssql:
      type: mssql
      collection_interval: 60s
    iis:
      type: iis
      collection_interval: 60s
  exporters:
    google:
      type: google_cloud_monitoring
  service:
    pipelines:
      default_pipeline:
        receivers: [hostmetrics,mssql,iis]
        exporters: [google]
```


otel/conf.go
```diff
@@ -492,6 +523,52 @@ func validateCollectionInterval(collectionInterval string, pluginName string) (b
 	return true, nil
 }
 
+type MSSQL struct {
```
qingling128, line 526: The `CollectionInterval ` related logic among hostmetrics, iis and mssql seem identical. Can we combine them to a helper function to reduce code duplication?

confgenerator/confgenerator.go
```diff
@@ -209,35 +211,49 @@ func generateOtelServices(receiverNameMap map[string]string, exporterNameMap map
 		if strings.HasPrefix(pID, "lib:") {
 			return nil, fmt.Errorf(`pipeline id prefix 'lib:' is reserved for pre-defined pipelines. Pipeline ID %q is not allowed.`, pID)
 		}
+
 		p := pipelines[pID]
+		// if the pipeline is default_pipeline, then split the pipeline into different streams.
+		if pID == "default_pipeline" {
```
qingling128, line 217: Do we have to restrict this to pID == "default_pipeline"? If users name the pipeline to some other names, what happened? I wonder if we can remove this if check.

confgenerator/confgenerator.go
```diff
@@ -209,35 +211,49 @@ func generateOtelServices(receiverNameMap map[string]string, exporterNameMap map
...
+			for _, rID := range p.Receivers {
+				var pipelineID string
+				var defaultProcessors []string
+				if rID == "hostmetrics" {
```
qingling128, line 221: The receiver ids are arbitrary strings, right? Should we check against the receiver type instead of just the ID?

confgenerator/confgenerator.go
sophieyfang: because we try to hide the underlying processors for hostmetrics, mssql and iis and somehow squeenze those three into one pipeline. Underneath it, inside the OTEL config, those are three pipeline. That's why this default_pipeline is so unique and it should be reserved for this usage.

otel/conf.go
sophieyfang: yes, exactly. Or @davidbtucker are there any solution in golang that's better than a helper function

confgenerator/confgenerator.go
sophieyfang: Good catch

confgenerator/confgenerator.go
davidbtucker: We should have the same behavior for any pipeline, not just the default pipeline.

otel/conf.go
davidbtucker: We could potentially use embedding, but I don't think it's worth it here yet.  Let's address it if there's more duplication in the future.

qingling128: SG

qingling128 (approved):
Left two comments for minor suggestions. LGTM in advance since the changes should be straight forward.

confgenerator/confgenerator.go
```diff
@@ -438,21 +457,41 @@ func generateOtelReceivers(hostmetricsReceiverFactories map[string]*hostmetricsR
...
+				iisList = append(iisList, &iis)
+				receiverNameMap[rID] = "windowsperfcounters/iis_" + rID
+			} else {
+				return nil, nil, nil, nil, fmt.Errorf(`receiver %q of pipeline %q is not defined`, rID, pID)
 			}
 		}
 	}
-	return hostMetricsList, receiverNameMap, nil
+	if len(hostMetricsList) > 1 || len(mssqlList) > 1 || len(iisList) > 1 {
+		return nil, nil, nil, nil, fmt.Errorf(`You have more than one receiver of the same type in "hostmetrics, mssql, iis". This is to avoid duplicate metrics streams.`)
```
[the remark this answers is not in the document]
sophieyfang: Done. Also modified exporter part to "Only one exporter of the same type in [google_cloud_monitoring] is allowed."

confgenerator/confgenerator_test.go
```diff
@@ -150,6 +150,7 @@ func TestGenerateConfigsWithInvalidInput(t *testing.T) {
 		testName := f.Name()
 		t.Run(testName, func(t *testing.T) {
 			unifiedConfigFilePath := fmt.Sprintf(invalidTestdataFilePathFormat, testName)
+			fmt.Println(unifiedConfigFilePath)
```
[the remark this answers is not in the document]
sophieyfang: Removed :)

PR #16, opened 2021-01-20
Remove grace config value 120 seconds to fall back to the default 5 seconds.

qingling128:
Looking at how Fluentd handles this, it turned out we are not flushing at shutdown at all because we are using file based buffering. We are not setting `flush_at_shutdown` explicitly so it defaults to false: https://docs.fluentd.org/configuration/buffer-section#flushing-parameters. So in theory this setting is not important for Fluent Bit either, because we also use file based buffering there.


martijnvans:
If it's not doing anything, I guess I'd suggest removing that setting completely and just picking up the default value of 5 seconds.

After all, 30 seconds is still a significant amount of time that a user has to wait, or for our tests to be slowed down. Especially if it's not for any real reason.

qingling128:
Updated. Ready for review.

davidbtucker (approved):

PR #18, opened 2021-01-21
Add build constraint

quentinmit:
Add build constraint so that `...` wildcard doesn't try to build `cmd/install_services` on non-Windows


igorpeshansky (approved):
LGTM :shipit:

PR #23, opened 2021-01-28
Delay copying local dir to container so that docker build can cache the previous steps.

qingling128:
…


qingling128:
Still testing this (The build is taking a long time...

igorpeshansky (approved):
LGTM :shipit:

Nice catch. Makes a lot of sense! Feel free to merge once you've verified that the build still works.

qingling128:
Tested the Dockerfile and it worked. Merging

PR #26, opened 2021-02-05
Windows subagent support

quentinmit:
The master binary installs itself as a service that automatically generates configuration files and starts its subagents.


qingling128 (approved):
LGTM. Will interate in follow up PRs.

cmd/generate_config/main.go
```diff
@@ -40,56 +37,5 @@ func main() {
...
-		}
-		path := filepath.Join(*outDir, "otel.conf")
-		if err := ioutil.WriteFile(path, []byte(otelConfig), 0644); err != nil {
-			return fmt.Errorf("can't write %q: %w", path, err)
-		}
-	default:
-		return fmt.Errorf("unknown service %q", *service)
-	}
-	return nil
+	return confgenerator.GenerateFiles(*input, *service, *logsDir, *stateDir, *outDir)
```
sophieyfang, line 40: This block of change possibly cause the current break in b/179710945#comment2:

```
Feb  8 22:33:04 centos8-b generate_config[45885]: 2021/02/08 22:33:04 can't create output directory "": mkdir : no such file or directory
```

sophieyfang: It seems that it was broken long before...not because of this PR

PR #27, opened 2021-02-05
Use filepath 

sophieyfang:
remove the hardcode path string and use filepath instead.

`go test -mod=mod ./...`  all pass

but for better referencing, the test_data/valid/windows_default_config needs to update the paths to `C:\ProgramData\Google\Cloud Operations\Ops Agent\run\buffers` and decouple the current configenerator_test.go to use different path prefix for windows. Will do in a followup PR

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -249,13 +249,13 @@ func defaultTails(logsDir string, stateDir string) (tails []*conf.Tail) {
 	return []*conf.Tail{
 		{
 			Tag:  "ops-agent-fluent-bit",
-			DB:   filepath.Join(stateDir, "/buffers/ops-agent-fluent-bit"),
-			Path: filepath.Join(logsDir, "/logging-module.log"),
+			DB:   filepath.Join(stateDir, "buffers", "ops-agent-fluent-bit"),
+			Path: filepath.Join(logsDir, "logging-module.log"),
 		},
 		{
 			Tag:  "ops-agent-collectd",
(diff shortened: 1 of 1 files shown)


confgenerator/confgenerator.go
```diff
@@ -249,13 +249,13 @@ func defaultTails(logsDir string, stateDir string) (tails []*conf.Tail) {
 	return []*conf.Tail{
 		{
 			Tag:  "ops-agent-fluent-bit",
-			DB:   fmt.Sprintf("%s/buffers/ops-agent-fluent-bit", stateDir),
-			Path: fmt.Sprintf("%s/logging-module.log", logsDir),
+			DB:   filepath.Join(stateDir, "buffers", "ops-agent-fluent-bit"),
```
quentinmit, line 252: It's fine to continue having "buffers/ops-agent-fluent-bit" here.

But, why is this needed at all if we're going to pass it in on the command line when starting fluentbit?

sophieyfang: it will have it in a not pretty syntax.
C:\ProgramData\.../buffers/ops-agent-fluent-bit

sophieyfang: Also for the question why it's needed here, because the configuration of fluent-bit needs to know where to tail the agent's own logs.

confgenerator/confgenerator.go
```diff
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
-
+        "path/filepath"
```
qingling128, line 22: `goimports -w .`?

qingling128 (approved):
LGTM

PR #28, opened 2021-02-05
Make log and state directories match the directories baked into the generated config files

quentinmit:
This also reverts to fluentbit v1.7.0-rc3 because -rc4 is broken.


qingling128 (approved):
LGTM

PR #29, opened 2021-02-06
Add Googet packaging for Windows.

igorpeshansky:


cmd/ops_agent_windows/main_windows.go
```diff
@@ -21,10 +27,24 @@ func main() {
...
+			}
+			log.Printf("installed services")
+		} else if *uninstallServices {
+			if err := uninstall(); err != nil {
+				log.Fatal(err)
+			}
+			log.Printf("uninstalled services")
+		} else {
+			// TODO: add an interactive GUI box with the Install, Uninstall, and Cancel buttons.
+			fmt.Println("Invoked as a standalone program with no flags. Nothing to do.")
```
qingling128, line 46: Does it make sense to mention that either `--install` or `--uninstall` is required in the message?

Taking a step back, there was some discussion offline around this that we decided to postpone till this PR. Quoting:

```
Sounds like what we agreed are:
* The binary supports both install and uninstall.
* User specifies a flag --uninstall to uninstall.

The questionable choice is between:
* User specifies no flag to indicate they want to install (this includes starting the services)
* User has to explicit specify --install to install

And we want to match common Windows practice.
```

Dockerfile.windows
```diff
@@ -19,7 +19,8 @@ SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPref
 RUN Write-Host ('Creating folders'); `
     New-Item -Type Directory -Path /local; `
     New-Item -Type Directory -Path /out/bin; `
-    New-Item -Type Directory -Path /out/config;
+    New-Item -Type Directory -Path /out/config; `
+    New-Item -Type Directory -Path /work/bin;
```
quentinmit, line 23: I thought you said you were going to be able to build the package from /out?

cmd/ops_agent_windows/install_windows.go
```diff
@@ -60,8 +64,60 @@ func install() error {
...
+
+func uninstall() error {
+	m, err := mgr.Connect()
+	if err != nil {
+		return err
+	}
+	defer m.Disconnect()
+	errs := []string{}
+	last := len(services) - 1
+	for i := range services {
```
quentinmit, line 81: i := len(services) - 1; i >= 0; i--

so you don't have to do index math below

cmd/ops_agent_windows/install_windows.go
```diff
@@ -60,8 +64,60 @@ func install() error {
...
+		s := services[last-i]
+		serviceHandle, err := m.OpenService(s.name)
+		if err != nil {
+			// Service does not exist, so nothing to delete.
+			continue
+		}
+		defer serviceHandle.Close()
+		if err := stopService(serviceHandle, 30 * time.Second); err != nil {
+			// Don't return until all services have been processed.
+			errs = append(errs, err.Error())
```
quentinmit, line 92: Nit: Use multierror.Group to handle the error collection for you?

cmd/ops_agent_windows/install_windows.go
```diff
@@ -60,8 +64,60 @@ func install() error {
...
+		}
+	}
+	if len(errs) != 0 {
+		return fmt.Errorf(strings.Join(errs, "\n"))
+	}
+	return nil
+}
+
+func stopService(serviceHandle *mgr.Service, timeout time.Duration) error {
+	// TODO: is it possible to use channel-based timeout handling here?
```
quentinmit, line 106: I don't think there's much point unless you have the option to use a blocking or channel-based API for listening for the stop event.

cmd/ops_agent_windows/main_windows.go
quentinmit: I think we were all happy with having a MsgBox come up. In the meantime I'd like to preserve the existing behavior, so if you don't want to do this in the PR, please change the default for install to true.

pkg/goo/install.ps1
```diff
@@ -0,0 +1,28 @@
...
+)
+
+$ErrorActionPreference = 'Stop'
+
+$envFromMatch = {
+  Param($match)
+  (Get-ChildItem -Path Env: | `
+     Where-Object -Property Name -eq $match.Groups[1].Value).Value
+}
+$InstallDir = [regex]::Replace($InstallDir,'^<([^>]+)>',$envFromMatch)
```
quentinmit, line 26: Do we *really* need this complicated machinery? Can you not get Googet to pass in an already-expanded path?

pkg/goo/uninstall.ps1
```diff
@@ -0,0 +1,28 @@
...
+$ErrorActionPreference = 'Stop'
+
+$envFromMatch = {
+  Param($match)
+  (Get-ChildItem -Path Env: | `
+     Where-Object -Property Name -eq $match.Groups[1].Value).Value
+}
+$InstallDir = [regex]::Replace($InstallDir,'^<([^>]+)>',$envFromMatch)
+
+& "$InstallDir\bin\google-cloud-ops-agent.exe" --uninstall
```
quentinmit, line 28: Make --uninstall a param so there's only one script?

cmd/ops_agent_windows/main_windows.go
qingling128: Is there any way to prevent introducing a breaking change when we add a MsgBox?

qingling128: To clarify it a bit, the reason I asked the question above is because we probably need a way for users to invoke the installation via PowerShell etc. without interactive confirmation so that users can script it. When we add a MsgBox, it should not break this use case.

qingling128: Oh I see. I thought the flag `--install` and `--uninstall` are directly used by users. If that's the case, it's fine either way.

qingling128 (approved):
LGTM

cmd/ops_agent_windows/main_windows.go
quentinmit: Printing a message to stdout/stderr does not work in "the GUI case" because double-clicking on the exe will not open a terminal.

pkg/goo/install.ps1
quentinmit: If Googet supports using templates for this in the package definition, I think that's less distasteful. But maybe I don't understand what you mean.

cmd/ops_agent_windows/main_windows.go
```diff
@@ -64,17 +87,20 @@ func initServices() error {
...
 	}{
 		{
 			serviceName,
+			serviceDisplayName,
 			self,
 			[]string{"-in", filepath.Join(base, "../config/config.yaml"), "-out", configOutDir},
 		},
 		{
-			fmt.Sprintf("%s - Metrics Agent", serviceName),
+			fmt.Sprintf("%s-otel", serviceName),
```
quentinmit, line 102: -metrics-agent? I don't think "-otel" is something users are going to understand.

Dockerfile.windows
quentinmit: In that case, maybe just put the output files at /work/out? Then you can have googet pull things from underneath /work, without having two copies of everything and without polluting /work with random build products.

pkg/goo/install.ps1
quentinmit: Looking at the Googet config again, I would guess that we can't do the variable substitution with Go templates because that would be baking the value of `PROGRAMDATA` from the build machine in, and not evaluating it at install time.

What if you just pass in a path relative to PROGRAMDATA and hardcode that environment variable as a prefix? That would significantly simplify the install script.

cmd/ops_agent_windows/main_windows.go
qingling128: Sync'd offline. We'll go with `-opentelemetry-collector` here.

igorpeshansky: Done.

cmd/ops_agent_windows/main_windows.go
qingling128: Discussed offline. Whether to support a GUI / using the exe directly is something we need to decide. If we do decide to support, the ideal UX can be controversial and we need more discussion. Let's punt on supporting those for now. Erroring out and noop when clicking on the exe is better for now, so that if we do decide to support such use cases later, we have the freedom to design the UX without worrying about introducing a breaking change.

quentinmit (approved):

PR #31, opened 2021-02-08
Build fluent-bit with optimizations

hsmatulisgoogle:
Set cmake flag for building fluent-bit with optimizations, but keeping debug symbols.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -72,7 +72,7 @@ function build_fluentbit() {
   cd submodules/fluent-bit
   mkdir -p build
   cd build
-  cmake .. -DCMAKE_INSTALL_PREFIX=$subagentdir/fluent-bit -DFLB_HTTP_SERVER=On
+  cmake .. -DCMAKE_INSTALL_PREFIX=$subagentdir/fluent-bit -DFLB_HTTP_SERVER=On -DCMAKE_BUILD_TYPE=RELWITHDEBINFO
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -72,7 +72,7 @@ function build_fluentbit() {
   cd submodules/fluent-bit
   mkdir -p build
   cd build
-  cmake .. -DCMAKE_INSTALL_PREFIX=$subagentdir/fluent-bit -DFLB_HTTP_SERVER=On
+  cmake .. -DCMAKE_INSTALL_PREFIX=$subagentdir/fluent-bit -DFLB_HTTP_SERVER=On -DCMAKE_BUILD_TYPE=RELWITHDEBINFO
```
quentinmit, line 75: Do you need to do the same thing in Dockerfile.windows?

quentinmit (approved):

build.sh
hsmatulisgoogle: I imagine so, as it looks like the windows build doesn't have debug symbols. I am not familiar with the windows build, do you want me to add that change? (RELEASE -> RELWITHDEBINFO)

quentinmit: Feel free to make that change... It still uses cmake so hopefully the syntax is the same?

hsmatulisgoogle: Done!

PR #32, opened 2021-02-09
Fail Ops Agent startup when standalone agents are already running.

davidbtucker:


cmd/ops_agent_windows/main_windows.go
```diff
@@ -8,12 +8,42 @@ import (
...
 
+// Standalone agents to check for before starting.
+var conflictingAgents = map[string]bool{
+	"StackdriverLogging":    true,
+	"StackdriverMonitoring": true,
+}
+
 func main() {
+	// Fail if any standalone agents are running.
+	mgr, err := mgr.Connect()
```
quentinmit, line 25: You need to call mgr.Disconnect() to close the connection.

I suggest moving all of this into a separate function, then you can use defer mgr.Disconnect()

cmd/ops_agent_windows/main_windows.go
```diff
@@ -8,12 +8,42 @@ import (
...
 const serviceName = "Google Cloud Ops Agent"
 
+// Standalone agents to check for before starting.
+var conflictingAgents = map[string]bool{
+	"StackdriverLogging":    true,
+	"StackdriverMonitoring": true,
+}
+
 func main() {
+	// Fail if any standalone agents are running.
```
quentinmit, line 24: Is it intentional that you are doing this both for installing the agents *and* running the agents? The former will block Googet package installation (and possibly leave the package in a half-installed state?), both will prevent users from intentionally using both with non-conflicting config. What is the desired UX here?

cmd/ops_agent_windows/main_windows.go
davidbtucker: Done.

cmd/ops_agent_windows/main_windows.go
davidbtucker: Moved this logic to under the Execute() workflow.

davidbtucker:
Ok, I addressed Quentin's comments, added logic to condition the check on the Ops Agent configuration, and rebased on master.

Note that I duplicated some code for reading the files.  Open to ideas for de-duping that code, although I think it will be easier once we remove the per-platform config definitions.

confgenerator/files.go
```diff
@@ -61,3 +61,42 @@ func GenerateFiles(input, service, logsDir, stateDir, outDir string) error {
 	}
 	return nil
 }
+
+// TODO(davidbtucker): Refacator the functions in this file.  We should have a single unified config (not per-platform).
+func GenerateFilesFromUnifiedConfigWindows(unified *UnifiedConfigWindows, service, logsDir, stateDir, outDir string) error {
```
qingling128, line 66: Is this part of the code somehow related to https://github.com/GoogleCloudPlatform/ops-agent/blob/01bcc3eb52e74eadca105313c29541b20e3c6d5a/confgenerator/files.go#L10? Seems like similar logic is there as well. Any reason we need to modify config generation when detecting service conflicts? Or is this just some refactoring?

davidbtucker: Right, this is just some refactoring.  At a high level, we want:

1. Read the Ops Agent config
2. Do any checks on the Ops Agent config (e.g. whether it conflicts with standalone agents)
3. From the Ops Agent config, generate the sub-agent configs.

The existing function does all steps 1-3.  The new function just does step 3.

We should refactor these functions to eliminate the duplicate logic, but that will be easier once we remove the per-platform Ops Agent config definitions.

qingling128: Got it. Can we create a bug for this with this context and add a TODO at https://github.com/GoogleCloudPlatform/ops-agent/blob/01bcc3eb52e74eadca105313c29541b20e3c6d5a/confgenerator/files.go#L10 as well?

qingling128 (approved):
LGTM

confgenerator/files.go
davidbtucker: Yep good idea, done.

quentinmit: The TODO is fixed in https://github.com/GoogleCloudPlatform/ops-agent/pull/42

davidbtucker: Thanks Quentin! That simplified this PR a lot.

confgenerator/files.go
```diff
@@ -16,6 +16,10 @@ func GenerateFiles(input, service, logsDir, stateDir, outDir string) error {
 	if err != nil {
 		return err
 	}
+	return GenerateFilesFromUnifiedConfig(uc, service, logsDir, stateDir, outDir)
+}
+
+func GenerateFilesFromUnifiedConfig(uc *UnifiedConfig, service, logsDir, stateDir, outDir string) error {
```
quentinmit, line 22: This should be a method (and it can keep the name `GenerateFiles` in that case)

confgenerator/confgenerator.go
```diff
@@ -100,13 +108,13 @@ func (uc *UnifiedConfig) GenerateFluentBitConfigs(logsDir string, stateDir strin
 	return generateFluentBitConfigs(uc.Logging, logsDir, stateDir)
 }
 
-func ParseUnifiedConfig(input []byte) (UnifiedConfig, error) {
+func ParseUnifiedConfig(input []byte) (*UnifiedConfig, error) {
```
quentinmit, line 111: Why did you change this to a pointer? Since it doesn't have any non-copyable persistent state it should be fine as-is.

cmd/ops_agent_windows/run_windows.go
```diff
@@ -67,14 +77,55 @@ func (s *service) parseFlags(args []string) error {
...
+	}
+	defer mgr.Disconnect()
+	services, err := mgr.ListServices()
+	if err != nil {
+		return fmt.Errorf("failed to list services: %s", err)
+	}
+
+	var errors string
+	if unified.HasLogging() && containsString(services, "StackdriverLogging") {
+		errors += "We detected an existing Windows service for the StackdriverLogging agent, " +
```
quentinmit, line 93: I realize I'm entering crazy town with this suggestion, but, if only one of the agents conflicts should we still start the other agent?

davidbtucker: My current thinking is to keep it as is.

This way will force the user to decide between disabling the standalone agent or disabling the config section.  Either action is easy to do.

If we were to start the non-conflicting sub-agent, the user might be confused that 1/2 of the config is being ignored.

confgenerator/files.go
davidbtucker: Done, thanks.

confgenerator/confgenerator.go
davidbtucker: Reverted.

quentinmit (approved):
Thanks, I think this is much cleaner now.

PR #37, opened 2021-02-18
Update fluent-bit to v1.7.0.

qingling128:
Tested:

logs showed up
image
metrics showed up
image


qingling128:
Seems like the auto-assigner did not kick in. Assigning manually.

davidbtucker (approved):

PR #38, opened 2021-02-20
Use systemd_postun_with_restart for RPM so that agent service enters running state post upgrades or re-installations

sophieyfang:
We shall use `systemd_postun_with_restart` here: https://github.com/GoogleCloudPlatform/ops-agent/blame/master/pkg/rpm/google-cloud-ops-agent.spec#L59
``` 
%postun
%systemd_postun_with_restart apache-httpd.service
``` 

This is based on readings from: https://screenshot.googleplex.com/AizcYk5n4mW4LPo

### DIFF
--- a/pkg/rpm/google-cloud-ops-agent.spec
+++ b/pkg/rpm/google-cloud-ops-agent.spec
@@ -56,6 +56,6 @@ fi
 %systemd_preun google-cloud-ops-agent.target
 
 %postun
-%systemd_postun google-cloud-ops-agent.target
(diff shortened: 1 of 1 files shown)


quentinmit (approved):
This should restart on upgrade, yeah.

qingling128 (approved):
LGTM

qingling128:
Can we change the title to something more user facing? e.g. `Use systemd_postun_with_restart for RPM so that agent service enters running state post upgrades or re-installations`.

PR #40, opened 2021-02-23
Use separate build stages to exploit caching

quentinmit:


Dockerfile.windows
```diff
@@ -70,16 +70,27 @@ RUN Start-Process msiexec.exe `
...
+# Run these in a directory with a go.mod file so that "go get" doesn't need
+# a working Git installation.
+WORKDIR /goget
+RUN go mod init unused; `
+    go get github.com/golangci/golangci-lint/cmd/golangci-lint; `
+    go get github.com/google/addlicense; `
+    go get github.com/pavius/impi/cmd/impi; `
+    go get github.com/google/googet/v2/goopack;
+
+FROM base as fluentbit
```
igorpeshansky, line 86: This `FROM` should probably move after the "Build fluent-bit" comment…

Dockerfile.windows
```diff
@@ -94,11 +105,11 @@ RUN cmake --build . --config Release; `
 # Build OT collector
 ###############################################################################
 
-WORKDIR /work/submodules/opentelemetry-operations-collector
+FROM base as gobuilder
```
igorpeshansky, line 108: I was initially confused why you named this `gobuilder` and not, e.g., `otcollector`, but then realized that you're reusing the same container to build all of the Go stuff. I wonder, though — why not have two builder containers, one for the OT collector, and one for the ops agent code? That way we can rebuild the ops agent without rebuilding the OT collector if needed.

Dockerfile.windows
```diff
@@ -119,9 +130,14 @@ RUN go build -o bin/google-cloud-ops-agent.exe ./cmd/ops_agent_windows; `
...
 ###############################################################################
 
+FROM base as packaging
+
 WORKDIR /work
 
-RUN go get github.com/google/googet/v2/goopack;
+COPY . /work
+
+COPY --from=fluentbit /work/out/bin /work/out/bin
```
igorpeshansky, line 139: [Optional] We should be able to copy the whole `/work/out` from the `fluentbit` container, right? For symmetry, if nothing else?..

igorpeshansky (changes requested):

Dockerfile.windows
quentinmit: Building the OT collector is fairly fast (O(1m)) so I structured it this way so they could benefit from sharing a common Go build cache. We could certainly split the collector out.

igorpeshansky: If it's cached at the container level, that's another minute saved for the build, and the OT collector changes very infrequently, so it's very likely to be cached… In any case, it would be nice to have a comment along the lines of "build all Go stuff in the same container", else others might be confused as I was.

igorpeshansky (commented):
One remaining conversation — once we resolve it, I'm happy to approve.

davidbtucker (approved):

Dockerfile.windows
```diff
@@ -102,11 +102,15 @@ RUN cmake --build . --config Release; `
...
+# Build Go code in one container to exploit Go build caching
 ###############################################################################
 
 FROM base as gobuilder
 
+###############################################################################
+# Build OT collector
+###############################################################################
+
 COPY . /work
```
igorpeshansky, line 114: This one is shared between the OT collector and the Ops agent, so I'd put this above the comment. Feel free to submit once done.

igorpeshansky (approved):
LGTM :shipit:

Dockerfile.windows
quentinmit: I fixed it the other direction by adding a separate COPY for each clause. PTAL.

Dockerfile.windows
```diff
@@ -90,27 +101,31 @@ RUN cmake --build . --config Release; `
...
 
-# TODO: Do something with /work/out/bin/google-cloud-metrics-agent_windows_amd64.exe
-
 ###############################################################################
 # Config generator/service installer
 ###############################################################################
 
 WORKDIR /work
 
+COPY . /work
```
igorpeshansky, line 127: Might make sense to keep the WORKDIR close to the RUN commands, so I'd move the COPY above the WORKDIR. Ditto below. As I've already approved, please merge after addressing.

igorpeshansky (approved):
LGTM :shipit:

Dockerfile.windows
quentinmit: I prefer the current order because the COPY is also a command that mutates the /work directory.

PR #41, opened 2021-02-23
Restart already-running services on reinstall

quentinmit:
Otherwise the old binaries could keep running until the next system reboot.


igorpeshansky (approved):
LGTM :shipit:

PR #42, opened 2021-02-25
Eliminate windows-specific config structs

quentinmit:
I also exported UnifiedConfig so you have a struct you can hang your new methods on.


davidbtucker (approved):
Thanks Quentin!

PR #49, opened 2021-03-22
Add isWindows check

sophieyfang:
- defaulttails() function tails collectd and fluentbit log files while for windows only fluentbit log file is available and otel sends logs directly to the event logs so there is no log file.  (Having it include the collectd tailing plugin doesn’t affect current usage in Windows, it’s just a useless blob of collectd tailing file)

- Windows prefer paths in style `C:\var\log\` rather than ‘/var/log’. For buffer and log path, now we use ‘/var/log’ which does work in Windows but not preferred. 

- config/log file path is different in Windows V.S. Linux. We need to reflect in the [testing](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/confgenerator_test.go) and default config file


sophieyfang:
Tests passed  on both linux and windows:
Screen Shot 2021-03-22 at 6 04 59 PM
Screen Shot 2021-03-22 at 6 05 10 PM

confgenerator/confgenerator_test.go
```diff
@@ -148,15 +166,19 @@ func updateOrCompareGolden(t *testing.T, testName string, expected string, actua
 }
 
 func TestGenerateConfigsWithInvalidInput(t *testing.T) {
-	filePath := invalidTestdataDir
+	filePath := invalidTestdataDir +  "/linux"
+	hostInfo, _ := host.Info()
```
qingling128, line 170: It looks like the `init` method at line 52 determines the value for `isWindows`. Could we reuse that here?

qingling128 (approved):
LGTM

PR #50, opened 2021-03-25
Configure recovery actions and add an rpcss dependency to improve reliability

quentinmit:
Fixes b/180408252
Might fix b/180641561


igorpeshansky (approved):
LGTM :shipit:

PR #51, opened 2021-04-06
rename generate_config to google_cloud_ops_agent_engine for clarity

ekund:
Error messages will then reference the more aptly-named "google_cloud_ops_agent_engine" binary.


davidbtucker (approved):

igorpeshansky:
[Optional] Why not `google_cloud_ops_agent_config_generator`? Also, the bike shed should be blue. 😁

davidbtucker:
> [Optional] Why not `google_cloud_ops_agent_config_generator`? Also, the bike shed should be blue. 😁

The "Ops Agent Engine" terminology came from Joe, so you could discuss with him.

quentinmit:
I'm definitely more confused by this new name. "engine" implies a daemon and this process is not. We certainly might have something in the future that deserves this name, but the config generator ain't it...

quentinmit:
Hi, I'm not going to forget about this. This PR still makes no sense. :)

PR #53, opened 2021-04-07
Force a zypper upgrade before installing dependencies.

qingling128:
The zypper package in the upstream Docker container segfaults on some RPMs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -131,6 +131,7 @@ RUN zypper -n install git systemd autoconf automake flex libtool libcurl-devel l
 && zypper addrepo https://download.opensuse.org/repositories/home:Ledest:devel/openSUSE_Leap_42.3/home:Ledest:devel.repo \
 && zypper addrepo https://download.opensuse.org/repositories/devel:languages:go/openSUSE_Leap_42.3/devel:languages:go.repo \
 && zypper -n --gpg-auto-import-keys refresh \
+&& zypper upgrade \
 && zypper -n install bison>3.4 go \
 # Allow fluent-bit to find systemd
 && ln -fs /usr/lib/systemd /lib/systemd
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -131,7 +131,8 @@ RUN zypper -n install git systemd autoconf automake flex libtool libcurl-devel l
 && zypper addrepo https://download.opensuse.org/repositories/home:Ledest:devel/openSUSE_Leap_42.3/home:Ledest:devel.repo \
 && zypper addrepo https://download.opensuse.org/repositories/devel:languages:go/openSUSE_Leap_42.3/devel:languages:go.repo \
 && zypper -n --gpg-auto-import-keys refresh \
-&& zypper -n install bison>3.4 go \
+# Pin to Golang 1.14 to bypass b/182517088#comment8. Long-term improvement is tracked by b/184743026.
```
davidbtucker, line 134: I think it would be best to use the same Go version across all distros.

Shall we do that in this PR? Or are you leaving that as part of b/184743026?

quentinmit (changes requested):
If the problem is that zypper is segfaulting, downgrading the package doesn't seem like the right fix. Try adding an explicit zypper upgrade before attempting to install go - I bet there's an upstream fix for this in zypper.

qingling128:
> If the problem is that zypper is segfaulting, downgrading the package doesn't seem like the right fix. Try adding an explicit zypper upgrade before attempting to install go - I bet there's an upstream fix for this in zypper.

Initially I thought it's zypper, but it turned out that the Golang 1.14 package does not have this issue. And the last updated time of Golang 1.16 package in that repo matches the date when our nightly job started to fail (See details in b/182517088#comment11). So it's probably somehow related to that package. I did not dig further yet though.

BTW, I'm giving explicit zypper upgrade a try as well now.

qingling128:
> BTW, I'm giving explicit zypper upgrade a try as well now.

An explicit zypper upgrade worked. I pushed a new commit.

quentinmit (approved):

davidbtucker (approved):

PR #54, opened 2021-04-07
make confgenerator generated stanzas in alphabetical order

ekund:
Because generator can be called on a map of values that are used to fill the template, the generator can generate the stanzas in different orders.  This causes unit tests, which test against a specific golden file, to fail sometimes.

To see the failures, rerun the unit tests 10 times in a row:

```
for i in 1 2 3 4 5 6 7 8 9 10; do go clean --testcache; go test -mod=mod ./...; done
```

With this change, there is a consistent order for the output, as well as an updated set of golden files, so that unit tests will pass all the time.

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -20,10 +20,11 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
-	"github.com/shirou/gopsutil/host"
+
 	"github.com/GoogleCloudPlatform/ops-agent/collectd"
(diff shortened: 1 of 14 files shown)


igorpeshansky:
A lot of the changes seem to be from `gofmt`. Would it make sense to split this into two commits — one for the pure formatting, and one for the actual added logic?

ekund:
I split into the the formatting changes, and the rest into the logic.
The logic change includes updating all the golden files to reorder the stanzas in them to be in alphabetical order, so it touches a lot of files.

davidbtucker (approved):

PR #56, opened 2021-04-09
Temporarily pin Golang at 1.14 for sles 12.

qingling128:
Temporary workaround for b/182517088#comment14.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -132,7 +132,8 @@ RUN zypper -n install git systemd autoconf automake flex libtool libcurl-devel l
 && zypper addrepo https://download.opensuse.org/repositories/devel:languages:go/openSUSE_Leap_42.3/devel:languages:go.repo \
 && zypper -n --gpg-auto-import-keys refresh \
 && zypper update \
-&& zypper -n install bison>3.4 go \
+# Pin Golang version to 1.14 to bypass b/182517088#comment14. The long-term solution is tracked by b/184743026.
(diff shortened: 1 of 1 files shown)


qingling128:
Looks like Xenial is pinned to Golang 1.11. I thought about unifying everything to Golang 1.14, then I saw that and figured there was probably some reason.

quentinmit:
The commit message is misleading - as far as we know there is no upstream repo problem, only a zypper bug :)

qingling128:
> The commit message is misleading - as far as we know there is no upstream repo problem, only a zypper bug :)

I wonder why it is only related to Golang and the problem started when the new Golang 1.16 package was released. I'm not sure whether it's the Go package issue or Zypper issue. But I rephrased the title just in case.

quentinmit:
> > The commit message is misleading - as far as we know there is no upstream repo problem, only a zypper bug :)
> 
> I wonder why it is only related to Golang and the problem started when the new Golang 1.16 package was released. I'm not sure whether it's the Go package issue or Zypper issue. But I rephrased the title just in case.

No matter how broken or not the package or repo are, Zypper segfaulting is entirely on Zypper. :)

qingling128:
> No matter how broken or not the package or repo are, Zypper segfaulting is entirely on Zypper. :)

Ah I see. Good to know.

davidbtucker (approved):

PR #57, opened 2021-04-09
More permanent workaround for zypper segfaults

quentinmit:
zypper has a bug that causes random segfaults when downloading packages:
https://bugzilla.opensuse.org/show_bug.cgi?id=1127849

We get unlucky and trigger this when we try to install our
packages. Downloading go separately seems to avoid a segfault with our
exact package set.

(This could happen with any package, so there's no guarantee this won't
break again in the future if the packages change again in the upstream
repos. Unless we can get opensuse to release a new container or zypper,
we'll just have to experiment until we find another reliable sequence of
package installs.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ FROM debian:buster AS buster
 
 # TODO: Factor out the common code without rerunning apt-get on every build.
 
-RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


qingling128 (approved):
LGTM

davidbtucker (approved):

PR #59, opened 2021-04-20
Stop ignoring errors in sub-agent startup

martijnvans:
Previously, when a subagent couldn't be started, the install process would claim that the ops agent was installed successfully and would simply skip the failing agent. This is confusing and hides problems. The way I ran into this was in the case of a flaky install error, which is supposed to be handled by OS Config via a retry loop. However, the retry loop doesn't kick in because it thinks the install succeeded. After this PR, it should at least notice the error and try again.

Additional context is in b/182388102 and b/182387975.


martijnvans:
hi, FYI I've never made a PR before, so I dunno how this works.

qingling128 (approved):
LGTM

PR #60, opened 2021-04-21
Run Travis CI on Windows as well for unit tests and fix broken tests.

qingling128:


sophieyfang:
The test might fail again. I think it's because we use dict thus the order is not fixed.

sophieyfang (approved):

qingling128:
> The test might fail again. I think it's because we use dict thus the order is not fixed.

Good to know. In the few times I tried, the order seems to be the same. But we could force a `sort` if it ends up being flaky.

PR #64, opened 2021-04-21
Enhance tests to ensure the expected error message is thrown by invalid agent config user input.

qingling128:
Note: Some of the error messages that are thrown do not seem to meet the expectation. Will fix them in followup PRs. This PR aims at enhancing the unit test infrastructure itself.


confgenerator/confgenerator_test.go
```diff
@@ -125,24 +127,24 @@ func TestGenerateConfsWithValidInput(t *testing.T) {
...
-	return string(rawExpectedConfig)
+	return rawExpectedConfig
 }
 
-func updateOrCompareGolden(t *testing.T, testName string, expected string, actual string, path string) {
+func updateOrCompareGolden(t *testing.T, testName string, expectedBytes []byte, actual string, path string) {
 	t.Helper()
-	expected = strings.ReplaceAll(expected, "\r\n", "\n")
-	actual = strings.ReplaceAll(actual, "\r\n", "\n")
+	expected := strings.TrimSuffix(strings.TrimSuffix(strings.ReplaceAll(string(expectedBytes), "\r\n", "\n"), "\n"), "\r")
```
sophieyfang, line 146: Not direct relate to this PR but previous changes. The code here seems to deal with the different line ending of Linux and Windows? I think it's no longer needed because now we use lib to detect run-time OS.

sophieyfang: https://github.com/GoogleCloudPlatform/ops-agent/pull/42 is where it got introduced.
https://github.com/GoogleCloudPlatform/ops-agent/pull/49/files is when we should've removed it

qingling128: Just removed them.

qingling128: Looks like Windows freaked out after the two lines are removed:
* https://github.com/GoogleCloudPlatform/ops-agent/runs/2406160651
* https://github.com/GoogleCloudPlatform/ops-agent/runs/2406167768

On a second thought, whether we use lib to detect run-time OS or not does not affect how line ending is handled, right? Assuming we tend to develop on Linux workstations, it would always trigger the discrepancy of line breakers for `testdata/.../windows` when the tests run on Windows (in this case a Travis Windows container).

confgenerator/confgenerator_test.go
```diff
@@ -125,23 +127,23 @@ func TestGenerateConfsWithValidInput(t *testing.T) {
...
 	}
-	return string(rawExpectedConfig)
+	return rawExpectedConfig
 }
 
-func updateOrCompareGolden(t *testing.T, testName string, expected string, actual string, path string) {
+func updateOrCompareGolden(t *testing.T, testName string, expectedBytes []byte, actual string, path string) {
 	t.Helper()
-	expected = strings.ReplaceAll(expected, "\r\n", "\n")
+	expected := strings.ReplaceAll(string(expectedBytes), "\r\n", "\n")
```
sophieyfang, line 146: The reason https://travis-ci.com/github/GoogleCloudPlatform/ops-agent/jobs/500325327 failed is because those windows conf file, we never have the lines ending with `\r\n`. It's still `\n` as Unix..

We can use `unix2dos` to convert it to DOS
https://stackoverflow.com/questions/3569997/how-to-find-out-line-endings-in-a-text-file

confgenerator/confgenerator_test.go
sophieyfang: Test failed because all our windows samples are stilling ending with \n... After PR#49, windows tests is only being trigger if the runtime is windows..

sophieyfang: we will fix it in followups

sophieyfang (approved):

confgenerator/confgenerator_test.go
qingling128: Got it. I'll leave it as is for this PR then. It does have the benefit that if someone (I meant me) generates the Windows side of golden files on a Linux machine, the line breaker discrepancy do not break tests.

PR #68, opened 2021-04-23
Fix systemd behavior on upgrades

igorpeshansky:


pkg/deb/debian/rules
```diff
@@ -11,7 +11,10 @@ override_dh_auto_install:
 	cd ../.. && DESTDIR=$$PWD/pkg/deb/debian/google-cloud-ops-agent ./build.sh
 
 override_dh_installsystemd:
-	dh_installsystemd google-cloud-ops-agent.target
+	dh_installsystemd google-cloud-ops-agent.target \
```
quentinmit, line 14: Nack, I believe I specifically had to override this so that it wouldn't try (and fail) to start the `.service` files.

quentinmit (changes requested):

pkg/deb/debian/rules
igorpeshansky: This is overriding `dh_installsystemd`, not `dh_systemd_start`, so it should only affect how units are installed, right?

PR #69, opened 2021-04-23
Add useragent for fluent-bit

sophieyfang:
1. Add `version.go` file for keeping the `Version` and `BuildDistro` field. It has default values to be used by testing. As those values in testdata folder shall keep static

2. Use `-ldflags` to override `Version` and `BuildDistro` fields during link invocation.

3. Create shared `getUserAgent` function which will later be used by `otel` too 

### Linux

Run below to install agent built from this branch and check the generated conf if it has the right UserAgent
```
sudo REPO_SUFFIX=20210423-4 bash add-google-cloud-ops-agent-repo.sh --also-install
sudo systemctl restart google-cloud-ops-agent.target
cat /var/run/google-cloud-ops-agent-fluent-bit/fluent_bit_main.conf
```
Snippet of conf
```
[OUTPUT]
    # https://docs.fluentbit.io/manual/pipeline/outputs/stackdriver
    Name              stackdriver
    stackdriver_agent Google-Cloud-Ops-Agent-Logging/1.0.6-debian9.13 (Platform=linux;ShortName=debian;ShortVersion=9.13,gzip(gfe))
    Match_Regex       ^(ops-agent-fluent-bit|ops-agent-collectd)$

    # https://docs.fluentbit.io/manual/administration/scheduling-and-retries
    # After 3 retries, a given chunk will be discarded. So bad entries don't accidentally stay around forever.
    Retry_Limit  3

    # https://docs.fluentbit.io/manual/administration/security
    # Enable TLS support.
    tls         On
    # Do not force certificate validation.
    tls.verify  Off

```

### Windows
Screen Shot 2021-04-26 at 12 50 15 PM


sophieyfang:
Hold on reviewing. As i will need to add `ldlflags` for windows side too.: https://github.com/GoogleCloudPlatform/ops-agent/blob/master/Dockerfile.windows#L129

qingling128:
Sure thing!

sophieyfang:
Letting the Dockerfile.windows to embed the distro and version thru passing them via --ldflags when building the ops-agent.
It's quite a challenge when the container is windows.
 
Below are the failure tries
Screen Shot 2021-04-26 at 12 04 20 PM
Screen Shot 2021-04-26 at 12 05 11 PM

Screen Shot 2021-04-26 at 12 06 29 PM
Screen Shot 2021-04-26 at 12 12 49 PM

Screen Shot 2021-04-26 at 12 14 33 PM
Screen Shot 2021-04-26 at 12 27 33 PM

Screen Shot 2021-04-26 at 12 29 12 PM
Screen Shot 2021-04-26 at 12 42 05 PM

Final successful try
Screen Shot 2021-04-26 at 12 41 43 PM
Screen Shot 2021-04-26 at 12 50 15 PM

internal/version/version.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package version
+
+// Version variable will be replaced at link time after `make` has been run.
```
quentinmit, line 17: Why are these two separate strings and not just one?

confgenerator/confgenerator_test.go
```diff
@@ -49,13 +49,20 @@ var (
 )
 
 var platform string
+var hostInfo *host.InfoStat
 
 func init() {
-	hostInfo, _ := host.Info()
+	hostInfo, _ = host.Info()
 	if hostInfo.OS == "windows" {
```
quentinmit, line 56: The test shouldn't be relying on the OS - just hardcode one string and use that in the tests regardless of which platform you're on.

confgenerator/confgenerator.go
```diff
@@ -225,8 +225,27 @@ func defaultStackdriverOutputs() (stackdrivers []*conf.Stackdriver) {
...
+func getUserAgent(prefix string, hostInfo *host.InfoStat) (string, error) {
+	userAgent := map[string]string{
+	    "Prefix":     prefix,
+	    "AgentVersion": version.Version,
+	    "BuildDistro":   version.BuildDistro,
+	    "Platform": hostInfo.OS,
+	    "ShortName": hostInfo.Platform,
+	    "ShortVersion": hostInfo.PlatformVersion,
+	}
+	var userAgentTemplate = template.Must(template.New("mssql").Parse(`{{.Prefix}}/{{.AgentVersion}}-{{.BuildDistro}} (Platform={{.Platform}};ShortName={{.ShortName}};ShortVersion={{.ShortVersion}},gzip(gfe))`))
```
quentinmit, line 237: Please move this template out to a separate constant outside the function (which you'll have to declare with `var` since it's an object).

Also, why is the template called "mssql"?

quentinmit (changes requested):

confgenerator/confgenerator_test.go
sophieyfang: It has to be relying on OS. See this PR: https://github.com/GoogleCloudPlatform/ops-agent/pull/49
Reasons are also listed below
- file path is different on Linux v.s. Windows
- collectd sends its own logs to a file, thus fluent-bit tails that file which  is unique on Linux v.s. windows otel will send its own logs to eventlog
- the ending \r\n is windows v.s. \n is linux

confgenerator/confgenerator.go
sophieyfang: done

confgenerator/confgenerator_test.go
quentinmit: I understand the output is different, but the OS should be an input to the confgenerator and the tests should hardcode one (or both) choice for OS.

internal/version/version.go
quentinmit: Well, because right now you're constructing a version that doesn't actually match the package version. e.g. our Debian package right now is `1.0.5~debian10` but you're going to make `1.0.5-debian10`.

build.sh
```diff
@@ -24,6 +24,10 @@ systemdsystempresetdir=$(pkg-config systemd --variable=systemdsystempresetdir)
 subagentdir=$prefix/subagents
 
 . VERSION
+BUILD_INFO_IMPORT_PATH="github.com/GoogleCloudPlatform/ops-agent/version"
+BUILD_X1="-X ${BUILD_INFO_IMPORT_PATH}.BuildDistro=$(lsb_release -is | tr A-Z a-z)$(lsb_release -rs)"
+BUILD_X2="-X ${BUILD_INFO_IMPORT_PATH}.Version=${PKG_VERSION}"
```
quentinmit, line 29: This should be available as one string in the `${DEB_VERSION}` variable in `debian/rules` and `%{version}-%{release}` in the `.spec` file.

quentinmit (changes requested):

internal/version/version.go
sophieyfang: The syntax of useragent is defined here: b170441793#comment15. We are not going for `1.0.5~debian10` as the package version..

confgenerator/confgenerator_test.go
sophieyfang: Now travis runs this confgenerator.test on two different platforms(Linux and Windows): https://github.com/GoogleCloudPlatform/ops-agent/blob/master/.travis.yml. And it's using the real hostinfo of the running platform

build.sh
sophieyfang: See my other comment. UserAgent format has been defined after rounds of discussions: b/170441793#comment15. I understand that you think for debian `version~build_diistro` shall be the final look as a whole and for rpm packages `version-release` shall be it.

@qingling128  on deciding if we want to have ~ for APT package and - for rpm pacakge

confgenerator/confgenerator_test.go
qingling128: I've been struggling a bit with triggering Windows unit tests from a Linux dev machine (it's a bit painful to always have to start a Windows VM to run Windows unit tests). Quentin's proposal is similar to what I have in mind to improve that experience, and I was planning to send out a PR. It's probably better to address this outside this current PR though as it's not directly related, right?

build.sh
quentinmit: I took that comment as a general example, not a literal version number. But if you want to be pedantic, that comment says `1.0.3-stretch` but your code here is going to generate `1.0.3-debian9`, so you're already not following the format in the comment.

internal/version/version.go
quentinmit: We should not be inventing a new version number format just for this use case. We should either be reporting the code version (`1.0.5`) or the package version (`1.0.5~debian10` or `1.0.5-1`). I think I would prefer the package version (because it contains other info not reported elsewhere - e.g. the package revision), but code vesion would also be defensible. If you want the distro the package was built on separately, you should report that as a separate field, not as part of a bespoke version number format.

build.sh
sophieyfang: My neglect, that can just be changed to:
```sophieyfang_google_com@debian9-ansible-test:~$ lsb_release -c
Codename:	stretch
```
Besides that, what else you think the tradeoff is?

sophieyfang: [sophieyfang_google_com@centos8-5 ~]$ lsb_release -a
LSB Version:	:core-4.1-amd64:core-4.1-noarch
Distributor ID:	CentOS
Description:	CentOS Linux release 8.3.2011
Release:	8.3.2011
Codename:	n/a

sophieyfang_google_com@debian9-ansible-test:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 9.13 (stretch)
Release:	9.13
Codename:	stretch

ummmmm... @qingling128 , for those distros that have codename shall we use code name, for those that doesn't have it shall be use Release just as it is or transformed it to el8?

confgenerator/confgenerator_test.go
sophieyfang: Okay then, we shall address it in a separate PR once we decide on how we want to do the test *across platform*

internal/version/version.go
sophieyfang: I don't think the intention was not to invent a new version number format.  Beloow format has already been decided after rounds of serious discussion i suppose.

```
Google-Cloud-Ops-Agent-Collector/<OPS_AGENT_VERSION>-<BUILD_DISTRO> (OPS_AGENT_COMMENT),gzip(gfe)
```
The useragent will later be used for internal analysis. For example, by looking at the above format, i suppose there wlll analysis of obtaining the build_distro. Thus it makes sense to use separator in this case it's `hyphen` to distinguish it from the version

quentinmit: `1.0.5-stretch` is a new version number format. That version number is not used anywhere else.

sophieyfang: If `-` makes it look like a new version number format, shall we re-discussion this part `<OPS_AGENT_VERSION>-<BUILD_DISTRO>`.  We would be very likely to have an analysis that's to only grab `1.0.5` and see how many VMs uses that version.. So it makes sense to have those two in separate fields

build.sh
sophieyfang: Google-Cloud-Ops-Agent-Logging/1.0.5 (BuildDistro=stretch;Platform=Linux;ShortName=Debian;ShortVersion=9 is the new format

Google-Cloud-Ops-Agent-Logging/1.0.5 (BuildDistro=9;Platform=Linux;ShortName=Debian;ShortVersion=9 is also acceptable

internal/version/version.go
sophieyfang: Google-Cloud-Ops-Agent-Logging/1.0.5 (BuildDistro=9;Platform=Linux;ShortName=Debian;ShortVersion=9 is acceptable accordng to ling

qingling128 (approved):
LGTM

PR #72, opened 2021-05-05
Run confgenerator tests cross platform, where possible

djaglowski:
This PR allows `confgenerator` tests to run cross-platform. The tests also now run in parallel.


djaglowski:
//cc @quentinmit

qingling128:
The latest commit https://github.com/GoogleCloudPlatform/ops-agent/runs/2514781382 from the `master` branch seems to be pass the unit tests for Windows: https://travis-ci.com/github/GoogleCloudPlatform/ops-agent/jobs/503300262. Perhaps `quentin-otlinux` needs another rebase.

BTW, we are running into the same issue that this PR is trying to solve. Some related conversation about running unit tests cross platforms: https://github.com/GoogleCloudPlatform/ops-agent/pull/69#discussion_r620706042 in order to easily run both Linux side and Windows side of the config generation tests from a Linux workstation. Currently the Windows side of the config generation tests need to be run from a Windows VM.

qingling128:
FYI that https://github.com/GoogleCloudPlatform/ops-agent/pull/85 (aka `quentin-otlinux`) has been merged to master now. To get this PR ready for review, could you rebase off the latest master?

confgenerator/testdata/valid/windows/default_config/golden_otel.conf
```diff
@@ -17,6 +17,35 @@ receivers:
...
+          - Total Bytes Received
+          - Total Bytes Sent
+          - Total Connection Attempts (all instances)
+          - Total Delete Requests
+          - Total Get Requests
+          - Total Head Requests
+          - Total Options Requests
+          - Total Post Requests
+          - Total Put Requests
+          - Total Trace Requests
```
djaglowski, line 48: This test case appears to be out of sync with `confgenerator/windows-default-config.yaml`. I added this to match what the test case expects.

confgenerator/testdata/valid/windows/default_config/golden_fluent_bit_main.conf
```diff
@@ -27,34 +27,6 @@
     # we don't want them to all be loaded into the memory.
     storage.max_chunks_up      128
 
-[INPUT]
```
djaglowski, line 30: This test failed. I noticed that is uses the same logging config as the `valid_config` case, so I copied the `golden_fluent_bit_main.conf` from `valid_config` to `default_config`, and this appears to have fixed the problem.

Still, would appreciate careful review on whether this is correct.

djaglowski:
@qingling128 I've rebased onto `master` and believe this is ready for review.

igorpeshansky (changes requested):

.gitignore
```diff
@@ -0,0 +1 @@
+.vscode
```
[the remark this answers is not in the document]
djaglowski: Removed.

confgenerator/confgenerator_test.go
```diff
@@ -48,48 +44,79 @@ var (
...
+	defaultStateDir     string
+	goldenDefaultConfig string
+	*host.InfoStat
+}
+
+var platforms = []platformConfig{
+	platformConfig{
+		defaultLogsDir:      "/var/log/google-cloud-ops-agent/subagents",
+		defaultStateDir:     "/var/lib/google-cloud-ops-agent/fluent-bit",
+		goldenDefaultConfig: "default-config.yaml",
```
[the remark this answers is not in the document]
djaglowski: Fixed

confgenerator/confgenerator_test.go
```diff
@@ -48,48 +44,79 @@ var (
...
+		InfoStat: func() *host.InfoStat {
+			info, _ := host.Info()
+			info.OS = "linux"
+			info.Platform = "linux_platform"
+			info.PlatformVersion = "linux_platform_version"
+			return info
+		}(),
+	},
+	platformConfig{
+		defaultLogsDir:      "C:\\ProgramData\\Google\\Cloud Operations\\Ops Agent\\log",
```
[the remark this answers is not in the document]
djaglowski: Fixed

confgenerator/confgenerator_test.go
```diff
@@ -167,42 +197,52 @@ func updateOrCompareGolden(t *testing.T, testName string, expectedBytes []byte,
...
+			invalidInput := readFileContent(t, platform.OS, testName, invalidInputPath, false)
+			expectedError := readFileContent(t, platform.OS, testName, goldenErrorPath, true)
+			actualError := generateConfigs(invalidInput, platform)
+
 			if actualError == nil {
 				t.Errorf("test %q: generateConfigs succeeded, want error:\n%s\ninvalid input:\n%s", testName, expectedError, invalidInput)
-			} else {
-				updateOrCompareGolden(t, testName, expectedError, actualError.Error(), goldenErrorPath)
 			}
+			updateOrCompareGolden(t, testName, platform.OS, expectedError, actualError.Error(), goldenErrorPath)
```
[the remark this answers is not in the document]
djaglowski: Good catch - changed to `Fatalf`.

confgenerator/confgenerator_test.go
```diff
@@ -48,48 +44,79 @@ var (
...
+	stateDir := platform.defaultStateDir
 	dirs, err := ioutil.ReadDir(dirPath)
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	for _, d := range dirs {
 		testName := d.Name()
-		t.Run(testName, func(t *testing.T) {
+		t.Run(platform.OS+"/"+testName, func(t *testing.T) {
```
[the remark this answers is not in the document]
djaglowski: I split this up as suggested.

It seemed natural to then pull the outer `t.Run` up to the top-level `Test...` func. This led me to thinking that it made sense to merge the helper `test...` func into the top-level `Test...` func, which results in perhaps a questionable amount of nesting. It seems ok to me, but of course let me know if this isn't what you want to see.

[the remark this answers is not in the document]
igorpeshansky: FWIW, I actually liked the fact that your PR renamed the original functions into helpers and avoided the extra level of indentation… Glad you reverted to that.

confgenerator/confgenerator_test.go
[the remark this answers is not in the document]
igorpeshansky: It's my understanding that `Fatalf` is generally reserved for things that go wrong during test execution (e.g., not able to open expected inputs), rather than actual testing assertions. In this case, you could keep the `Errorf` and move `updateOrCompareGolden` back into the `else` branch…

confgenerator/confgenerator_test.go
```diff
@@ -101,41 +128,44 @@ func TestGenerateConfsWithValidInput(t *testing.T) {
...
 			}
 
-			// Retrieve the expected golden conf files.
-			expectedMainConfig := readFileContent(testName, goldenMainPath, t, true)
-			expectedParserConfig := readFileContent(testName, goldenParserPath, t, true)
-			// Generate the actual conf files.
-			mainConf, parserConf, err := uc.GenerateFluentBitConfigs(logsDir, stateDir, hostInfo)
-			if err != nil {
-				t.Fatalf("GenerateFluentBitConfigs got %v", err)
+			// GenerateFluentBitConfigs depends on filepath.Join, so actual runtime.GOOS is consequential
```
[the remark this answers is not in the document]
djaglowski: Thanks, that diff was helpful. 

I implemented a version of this.

confgenerator/confgenerator_test.go
```diff
@@ -48,94 +46,119 @@ var (
...
+					uc, err := ParseUnifiedConfig(data)
+					if err != nil {
+						t.Fatalf("ParseUnifiedConfig got %v", err)
+					}
+					
+					// Retrieve the expected golden conf files.
+					expectedMainConfig := readFileContent(t, platform.OS, testName, goldenMainPath, true)
+					expectedParserConfig := readFileContent(t, platform.OS, testName, goldenParserPath, true)
+					// Generate the actual conf files.
+					setFilepathJoin(platform.OS)
```
igorpeshansky: This kind of global usage prevents us from ever running these tests in parallel, doesn't it? Why not pass the OS into `filepathJoin` instead?

confgenerator/confgenerator_test.go
```diff
@@ -45,78 +47,74 @@ var (
...
-	if info.OS == "windows" {
-		//In order to make test data static, we put static value for platform-wise fields.
-		info.Platform = "win_platform"
-		info.PlatformVersion = "win_platform_version"
-	} else {
-		//In order to make test data static, we put static value for platform-wise fields.
-		info.OS = "linux"
-		info.Platform = "linux_platform"
-		info.PlatformVersion = "linux_platform_version"
+	//In order to make test data static, we put static value for platform-wise fields.
```
igorpeshansky, line 50: [Optional] Let's fix that missing space after the comment marker?

confgenerator/confgenerator_test.go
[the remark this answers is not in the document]
djaglowski: Done. It seemed natural to pull the outer `t.Run` up into the top-level `Test...` func, but of course let me know if that's not what you want to see.

confgenerator/confgenerator_test.go
```diff
@@ -148,11 +175,11 @@ func readFileContent(testName string, filePathFormat string, t *testing.T, respe
 	return rawExpectedConfig
 }
 
-func updateOrCompareGolden(t *testing.T, testName string, expectedBytes []byte, actual string, path string) {
+func updateOrCompareGolden(t *testing.T, testName, goos string, expectedBytes []byte, actual string, path string) {
```
igorpeshansky, line 178: The convention here seems to be to specify the type for each parameter, even if they repeat. Can we please make this consistent?

igorpeshansky (changes requested):

confgenerator/confgenerator_test.go
```diff
@@ -48,94 +46,119 @@ var (
...
+	},
+}
 
-			// Retrieve the expected golden conf files.
-			expectedMainConfig := readFileContent(testName, goldenMainPath, t, true)
-			expectedParserConfig := readFileContent(testName, goldenParserPath, t, true)
-			// Generate the actual conf files.
-			mainConf, parserConf, err := uc.GenerateFluentBitConfigs(logsDir, stateDir, hostInfo)
+func TestGenerateConfsWithValidInput(t *testing.T) {
+	for _, platform := range platforms {
```
igorpeshansky, line 96: Another thing to be careful of if we ever decide to run these tests in parallel: https://golang.org/doc/faq#closures_and_goroutines.

igorpeshansky (commented):
Sorry, one more thing…

djaglowski:
@igorpeshansky, I appreciate the feedback. I've learned a few nuances to golang.

confgenerator/confgenerator_test.go
djaglowski: The reason I went in that direction, was to ensure that `confgenerator` continues to use the real `filepath.Join` in real world situations. 

That said, your point is well taken. This was not a good implementation. I've reworked it to allow parallelism, but also to only override `filepath.Join` in tests.

djaglowski:
Re: `Fatalf` vs `Errorf` - I took a pass through the entire file and switched them as necessary to support the notion you described. I think it got it right, but please let me know if you disagree with my interpretation.

confgenerator/confgenerator_test.go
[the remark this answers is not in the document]
djaglowski: Re: `Fatalf` vs `Errorf` - I took a pass through the entire file and switched them as necessary to support the notion you described. I think it got it right, but please let me know if you disagree with my interpretation.

confgenerator/confgenerator_test.go
djaglowski: I've hardened this by removing any pointers from the `platformConfig` struct. I believe this means that each test will run with a copy of the values we care about. Additionally, the inner tests were dependent on a `testName` variable which I believe was exhibiting the behavior, so I replaced that.

I understand the point you're making, and acknowledge that this could be reworked further to protect against future changes. However, I suspect this is sufficient for now. Let me know if you disagree.

In any case, I've marked all `confgenerator` tests as parallel, and there doesn't appear to be an immediate problem, at least in my local environment:
```
> go test -count=10000 -timeout=5m -mod=mod github.com/GoogleCloudPlatform/ops-agent/confgenerator
ok      github.com/GoogleCloudPlatform/ops-agent/confgenerator  78.127s
```

djaglowski: Well, I see that code coverage has dropped substantially, so I think I understood the implications less well than I expected. Looking into it...

djaglowski: Confirmed that this was doing exactly what you warned me about. 

However, I actually get it now. Should be good to go.

confgenerator/confgenerator_test.go
```diff
@@ -129,39 +125,41 @@ func testGenerateConfsWithValidInput(t *testing.T, platform platformConfig) {
 			// Retrieve the expected golden conf files.
 			expectedMainConfig := readFileContent(t, platform.OS, testName, goldenMainPath, true)
 			expectedParserConfig := readFileContent(t, platform.OS, testName, goldenParserPath, true)
+
 			// Generate the actual conf files.
-			setFilepathJoin(platform.OS)
 			mainConf, parserConf, err := uc.GenerateFluentBitConfigs(platform.defaultLogsDir, platform.defaultStateDir, platform.InfoStat)
-			setFilepathJoin(runtime.GOOS)
 			if err != nil {
-				t.Fatalf("GenerateFluentBitConfigs got %v", err)
```
igorpeshansky, line 110: In this function, it should remain `Fatalf`, because this test cannot proceed when the config generation fails. The test assertion is about the content of the generated config.

confgenerator/confgenerator_test.go
```diff
@@ -129,39 +125,41 @@ func testGenerateConfsWithValidInput(t *testing.T, platform platformConfig) {
...
 				}
-				// Compare the expected and actual and error out in case of diff.
-				updateOrCompareGolden(t, testName, platform.OS, expectedCollectdConfig, collectdConf, goldenCollectdPath)
 			}
 		})
 	}
 }
 
-func readFileContent(t *testing.T, goos, testName, filePathFormat string, respectGolden bool) []byte {
+func readFileContent(t *testing.T, goos string, testName string, filePathFormat string, respectGolden bool) []byte {
```
igorpeshansky, line 179: Just curious, was there a reason to order them [`goos`, `testName`] here, but [`testName`, `goos`] in `updateOrCompareGolden`? FWIW, it's probably better to keep `t` and `testName` next to each other…

confgenerator/confgenerator_test.go
```diff
@@ -89,6 +89,7 @@ func TestGenerateConfsWithValidInput(t *testing.T) {
 	t.Parallel()
 	filepathJoin = testJoin
 	for _, platform := range platforms {
+		platform := platform // create a new 'platform'
```
igorpeshansky, line 107: This comment isn't particularly helpful. You could use the FAQ URL, or just omit the comment, similarly to all the other places this pattern occurs.

confgenerator/confgenerator_test.go
```diff
@@ -85,9 +86,23 @@ func testJoin(goos string, elem ...string) string {
 	return strings.Join(elem, separator)
 }
 
+func overrideFilepathJoin(t *testing.T) {
+	if filepath.Join("a", "b", "c") != filepathJoin(runtime.GOOS, "a", "b", "c") {
+		t.Fatalf("default filepathJoin behaves unexpectedly")
+	}
+	filepathJoin = testJoin
```
igorpeshansky, line 93: You're still updating a global. The test framework is free to schedule all of the goroutines you spawn within these tests first, and then execute all of them at once. This may be ok for the tests spawned in this function, because they all expect to run with the above implementation, but suppose another test is added in the future in a function that forgets to assign this. Not only would its behavior be platform-dependent, but it would also be randomly platform dependent (e.g., if this test function ran in the middle of executing the goroutines from the other function). Why not just use the above implementation in the actual config generator?

confgenerator/confgenerator_test.go
igorpeshansky: Heh, I wasn't necessarily suggesting to make the tests parallel, just avoid painting ourselves into the corner so we couldn't make them parallel in the future. But this is fine too.

igorpeshansky (changes requested):

confgenerator/confgenerator_test.go
djaglowski: No, I just didn't compare the signatures. Will update.

confgenerator/confgenerator_test.go
djaglowski: My motivation was to make use of the real `filepath.Join` "in the wild", since it's more robust than `strings.Join(elem, separator)`. 

In any case, that was my rationale, but this is your project. I'll happily make the change as you've suggested.

confgenerator/confgenerator_test.go
```diff
@@ -161,48 +177,65 @@ func updateOrCompareGolden(t *testing.T, testName string, expectedBytes []byte,
...
-			invalidInput := readFileContent(testName, invalidInputPath, t, false)
-			expectedError := readFileContent(testName, goldenErrorPath, t, true)
-			// The expected error could be triggered by:
-			// 1. Parsing phase of the agent config when the config is not YAML.
-			// 2. Config generation phase when the config is invalid.
-			uc, actualError := ParseUnifiedConfig(invalidInput)
-			if actualError == nil {
-				actualError = generateConfigs(uc, defaultLogsDir, defaultStateDir)
-			}
+			t.Parallel()
```
igorpeshansky, line 205: Mind mentioning in the PR description that tests are now parallel?

igorpeshansky (changes requested):

confgenerator/confgenerator_test.go
djaglowski: No worries. There's an argument to be made either way.

I switched it back and performed the override in `init`. 

One very minor repercussion: In order to pass the code coverage requirement, I need to test the default `filepathJoin` (or test something entirely unrelated). Since `init` overrides the default before there's a chance to test the code, I named it and called it directly in a test. It almost feels silly, but I tried to make the test at least somewhat useful in that it _may_ detect an unintended change to the default.

confgenerator/confgenerator_test.go
```diff
@@ -102,40 +139,43 @@ func TestGenerateConfsWithValidInput(t *testing.T) {
...
+			updateOrCompareGolden(t, testName, platform.OS, expectedParserConfig, parserConf, goldenParserPath)
 
-			if platform == "windows" {
-				expectedOtelConfig := readFileContent(testName, goldenOtelPath, t, true)
-				otelConf, err := uc.GenerateOtelConfig(hostInfo)
+			if platform.OS == "windows" {
+				expectedOtelConfig := readFileContent(t, testName, platform.OS, goldenOtelPath, true)
+				otelConf, err := uc.GenerateOtelConfig(platform.InfoStat)
 				if err != nil {
-					t.Fatalf("GenerateOtelConfig got %v", err)
```
igorpeshansky, line 120: The reasoning for keeping `Fatalf` applies here as well, doesn't it? Also in the non-Windows branch. You could then move the `updateOrCompareGolden` calls out of the `else`…

confgenerator/confgenerator_test.go
```diff
@@ -48,48 +45,88 @@ var (
...
+func TestDefaultFilepathJoin(t *testing.T) {
+	t.Parallel()
+
+	// Test that the default filepathJoin function does not
+	// generate paths that are dependent on the specified OS.
+	abc := filepath.Join("a", "b", "c")
+	linuxAbc := defaultFilepathJoin("linux", "a", "b", "c")
+	windowsAbc := defaultFilepathJoin("windows", "a", "b", "c")
+	if abc != linuxAbc || abc != windowsAbc {
+		t.Fatal("The default filepathJoin function does not match filepath.Join")
```
igorpeshansky, line 99: This is just a test assertion, right? So it can be `Error`.

confgenerator/confgenerator_test.go
```diff
@@ -48,48 +45,88 @@ var (
...
+
+func TestDefaultFilepathJoin(t *testing.T) {
+	t.Parallel()
+
+	// Test that the default filepathJoin function does not
+	// generate paths that are dependent on the specified OS.
+	abc := filepath.Join("a", "b", "c")
+	linuxAbc := defaultFilepathJoin("linux", "a", "b", "c")
+	windowsAbc := defaultFilepathJoin("windows", "a", "b", "c")
+	if abc != linuxAbc || abc != windowsAbc {
```
igorpeshansky, line 98: Nice!

[Optional] The error could indicate which of the two doesn't match the behavior.
Or, if you switch to `Error`, you could split it into two checks, e.g.:
```
	if abc != linuxAbc {
		t.Errorf(`defaultFilepathJoin("linux")` does not match filepath.Join: %q`, linuxAbc)
	}
	if abc != windowsAbc {
		t.Errorf(`defaultFilepathJoin("windows")` does not match filepath.Join: %q`, windowsAbc)
	}
```

igorpeshansky (changes requested):
Thanks for bearing with me. Some minor remaining tweaks.

PR #74, opened 2021-05-06
Remove requirement to uninstall standalone agents

ekund:
In order to ease migration, remove the uninstallation of the standalone agents so that a customer can install the opsagent to run only a monitoring agent, while using that standalone logging agent, or vice versa.

This PR removes the uninstall for Linux and Windows.

### DIFF
--- a/cmd/ops_agent_windows/run_windows.go
+++ b/cmd/ops_agent_windows/run_windows.go
@@ -77,36 +77,6 @@ func (s *service) parseFlags(args []string) error {
 	return fs.Parse(allArgs)
 }
 
-func (s *service) checkForStandaloneAgents(unified *confgenerator.UnifiedConfig) error {
-	mgr, err := mgr.Connect()
-	if err != nil {
-		return fmt.Errorf("failed to connect to service manager: %s", err)
-	}
-	defer mgr.Disconnect()
-	services, err := mgr.ListServices()
-	if err != nil {
-		return fmt.Errorf("failed to list services: %s", err)
-	}
-
-	var errors string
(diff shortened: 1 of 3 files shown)


qingling128:
One concern that drove the original implementation was to prevent users from:
* getting double billed for logging
* getting requests rejected due to conflicts for metrics.

This might happen if:
* Users have a standalone Fluentd agent and an Ops agent that collects system logs as well
or
* Users have a standalone Collectd agent and an Ops agent that collects host metrics as well

Any chance we could check whether the Ops agent config has enabled system log / host metrics, and use that to decide whether to error out or let it through? For Linux, this might require implementing something similar to the Windows side instead of relying on package conflicts settings. How much effort is that?

qingling128:
@davidbtucker for visibility

ekund:
I am concerned about implementing that, since we have the default config baked in to the package.  There is no mechanism right now to install the config with the yaml file edited to remove the config, so I am not clear on how we would even implement that.

qingling128:
One alternative is to give a warning instead of a hard error. Let's discuss a bit more in tomorrow's meeting.

njculver:
Are there any updates on this? With the capability gap that exists between ops agent and the legacy logging agent, it would be very helpful to be able to run the ops agent for monitoring side by side with the legacy logging agent.

qingling128:
Hi, this effort is now being reconsidered in prioritization discussions. Previously it was de-prioritized for competing priorities.

BTW, if you are interested in providing some insights of the remaining gaps between the legacy agent and Ops Agent that are affecting your workloads, it would be very helpful for our prioritization.

PR #78, opened 2021-05-13
Add more test and a check to verify the actual error message in case of bad config.

qingling128:
GitHub shows the diff in a quite hard-to-read way. The actual Go code changes are in these 2 files only:

```
 collectd/conf.go
 confgenerator/confgenerator.go
```

And the testdata structure is like below:

```
ops-agent/confgenerator/testdata/invalid$ tree | grep -v "golden\|input"
.
├── linux
│   ├── all-not_in_yaml_format
│   ├── logging-exporter_reserved_id_prefix
│   ├── logging_exporter_undefined_id
│   ├── logging_exporter_unsupported_type
│   ├── logging-pipeline_reserved_id_prefix
│   ├── logging-processor_parse_regex_type_missing_required_parameter_regex
│   ├── logging-processor_reserved_id_prefix
│   ├── logging_processor_undefined_id
│   ├── logging_processor_unsupported_type
│   ├── logging-receiver_files_type_invalid_parameter_include_paths_is_empty
│   ├── logging-receiver_files_type_invalid_parameter_include_paths_is_empty_list
│   ├── logging-receiver_files_type_invalid_parameter_include_paths_is_empty_string
│   ├── logging-receiver_files_type_missing_required_parameter_include_paths
│   ├── logging-receiver_files_type_unsupported_parameter_channels
│   ├── logging-receiver_files_type_unsupported_parameter_listen_host
│   ├── logging-receiver_files_type_unsupported_parameter_listen_port
│   ├── logging-receiver_files_type_unsupported_parameter_random
│   ├── logging-receiver_files_type_unsupported_parameter_transport_protocol
│   ├── logging-receiver_reserved_id_prefix
│   ├── logging-receiver_syslog_type_invalid_paramater_listen_host_is_empty_string
│   ├── logging-receiver_syslog_type_invalid_paramater_listen_port_is_not_an_int
│   ├── logging-receiver_syslog_type_invalid_paramater_listen_port_is_zero
│   ├── logging-receiver_syslog_type_invalid_paramater_unknown_transport_protocol
│   ├── logging-receiver_syslog_type_unsupported_parameter_channels
│   ├── logging-receiver_syslog_type_unsupported_parameter_exclude_paths
│   ├── logging-receiver_syslog_type_unsupported_parameter_include_paths
│   ├── logging-receiver_syslog_type_unsupported_parameter_random
│   ├── logging_receiver_undefined_id
│   ├── logging_receiver_unsupported_type
│   ├── metrics-exporter_reserved_id_prefix
│   ├── metrics_exporter_undefined_id
│   ├── metrics_exporter_unsupported_type
│   ├── metrics-pipeline_reserved_id_prefix
│   ├── metrics_processor_unsupported_type
│   ├── metrics-receiver_invalid_parameter_collection_interval_below_minimum
│   ├── metrics-receiver_invalid_parameter_malformed_collection_interval
│   ├── metrics-receiver_more_than_one_with_type_hostmetrics
│   ├── metrics-receiver_reserved_id_prefix
│   ├── metrics_receiver_undefined_id
│   └── metrics_receiver_unsupported_type
└── windows
    ├── all-not_in_yaml_format
    ├── logging-receiver_windows_event_log_type_missing_required_parameter_channels
    ├── logging-receiver_windows_event_log_type_unsupported_parameter_exclude_paths
    ├── logging-receiver_windows_event_log_type_unsupported_parameter_include_paths
    ├── logging-receiver_windows_event_log_type_unsupported_parameter_listen_host
    ├── logging-receiver_windows_event_log_type_unsupported_parameter_listen_port
    ├── logging-receiver_windows_event_log_type_unsupported_parameter_random
    ├── logging-receiver_windows_event_log_type_unsupported_parameter_transport_protocol
    ├── metrics-exporter_more_than_one_with_the_same_type
    ├── metrics-exporter_undefined_id
    ├── metrics-exporter_unsupported_type
    ├── metrics-receiver_invalid_parameter_collection_interval_below_minimum
    ├── metrics-receiver_invalid_parameter_malformed_collection_interval
    ├── metrics-receiver_more_than_one_with_type_hostmetrics
    ├── metrics-receiver_more_than_one_with_type_iis
    ├── metrics-receiver_more_than_one_with_type_mssql
    ├── metrics-receiver_undefined_id
    ├── metrics-receiver_unsupported_type
    ├── metrics-reserved_exporter_id_prefix
    ├── metrics-reserved_pipeline_id_prefix
    └── metrics-reserved_receiver_id_prefix
```


qingling128:
It would be great if @davidbtucker could review the golang readability for those two files, and @ekund to review the overall testing strategy to see if it matches what you have in mind to facilitate the `Ops Agent Type Validation / Conversion Refactor` effort.

collectd/conf.go
```diff
@@ -229,22 +238,22 @@ func validatedCollectdConfig(metrics *Metrics) (*collectdConf, error) {
...
-			return nil, fmt.Errorf(`pipeline id prefix 'lib:' is reserved for pre-defined pipelines. Pipeline ID %q is not allowed.`, pipelineID)
+			return nil, reservedIdPrefixError("pipeline", pipelineID)
 		}
 		if len(pipeline.ReceiverIDs) != 1 {
 			return nil, errors.New("exactly one receiver id is required in the metrics service pipeline receiver id list.")
 		}
 		invalidReceiverIDs := findInvalid(definedReceiverIDs, pipeline.ReceiverIDs)
 		if len(invalidReceiverIDs) > 0 {
-			return nil, fmt.Errorf("metrics receivers not defined: %v", invalidReceiverIDs)
+			return nil, fmt.Errorf("metrics receiver %q from pipeline %q is not defined.", invalidReceiverIDs[0], pipelineID)
```
ekund, line 248: In this error message, you switched from printing all of the invalid receiver ids, to just printing the first one.  Shouldn't we print out all of them?  What was the reason for the switch?

collectd/conf.go
```diff
@@ -229,22 +238,22 @@ func validatedCollectdConfig(metrics *Metrics) (*collectdConf, error) {
...
 		}
 
 		if len(pipeline.ExporterIDs) != 1 {
 			return nil, errors.New("exactly one exporter id is required in the metrics service pipeline exporter id list.")
 		}
 		invalidExporterIDs := findInvalid(definedExporterIDs, pipeline.ExporterIDs)
 		if len(invalidExporterIDs) > 0 {
-			return nil, fmt.Errorf("metrics exporters not defined: %v", invalidExporterIDs)
+			return nil, fmt.Errorf("metrics exporter %q from pipeline %q is not defined.", invalidExporterIDs, pipelineID)
```
ekund, line 256: Here, the messages says "exporter" and not "exporters" but we are printing the whole list of them.  I think this should look similar to the one above, on line 248, and they should do something consistent. WDYT?

qingling128:
The `coveralls` coverage decrease warnings can be ignored. The actual coverage increased from what I can see:
* https://coveralls.io/builds/39661104/source?filename=collectd%2Fconf.go
* https://coveralls.io/builds/39661104/source?filename=confgenerator%2Fconfgenerator.go\

The tool might be counting the `input.yaml` and `golden_error` files as well I guess?

ekund:
I am fine with this approach.  I think that we should refactor next, and move all the validation out to a separate module (or modules), but adding all these tests is a good first step, before the refactor.

ekund (changes requested):

collectd/conf.go
qingling128: This logic is nice in theory. In practice it always prints 1 message only, because only one receiver of `hostmetrics` is allowed. I changed the error message to align with the other failure cases.

How to provide a list of error messages instead of one is a more generic problem / improvement we need. That is better to be handled by another PR though. Ideally the validator should return as many errors in a list format to the users so that they can fix all the issues in one shot.

collectd/conf.go
qingling128: Oops. Nice catch! Yeah, keeping them consistent was the intention. Just fixed.

collectd/conf.go
```diff
@@ -144,6 +144,15 @@ LoadPlugin swap
 	"process":    ``,
 }
 
+func reservedIdPrefixError(
```
davidbtucker, line 147: Fyi, Go style would be more like:

    // reservedIdPrefixError returns an error message for when the user
    // tries to define a reserved ID.  The component should be "receiver",
    // "processor", or "exporter".
    func reservedIDPrefixError(component, id string) error {
      ...
    }

collectd/conf.go
```diff
@@ -183,15 +192,15 @@ func validatedCollectdConfig(metrics *Metrics) (*collectdConf, error) {
 	}
 
 	// Validate Metrics.Receivers.
-	if len(metrics.Receivers) != 1 {
-		return nil, errors.New("exactly one metrics receiver with type 'hostmetrics' is required.")
+	if len(metrics.Receivers) > 1 {
+		return nil, errors.New(`At most one metrics receiver with type "hostmetrics" is allowed.`)
```
davidbtucker, line 196: In Go, errors start with lowercase letters.

confgenerator/confgenerator.go
```diff
@@ -116,7 +140,7 @@ func ParseUnifiedConfig(input []byte) (UnifiedConfig, error) {
 	config := UnifiedConfig{}
 	err := yaml.UnmarshalStrict(input, &config)
 	if err != nil {
-		return UnifiedConfig{}, err
+		return UnifiedConfig{}, fmt.Errorf("the agent config file is not valid YAML. detailed error: %s", err.Error())
```
davidbtucker, line 143: Fyi, you can omit the `.Error()` I believe.

confgenerator/confgenerator.go
```diff
@@ -343,77 +369,130 @@ func extractOtelReceiverFactories(receivers map[string]collectd.Receiver) (map[s
...
 			}
 		default:
-			return nil, nil, nil, fmt.Errorf(`receiver %q should have type as one of the "hostmetrics, mssql, iis"`, n)
+			return nil, nil, nil, unsupportedComponentTypeError("windows", "metrics", "receiver", r.Type, n)
 		}
 	}
 	return hostmetricsReceiverFactories, mssqlReceiverFactories, iisReceiverFactories, nil
 }
 
+func unsupportedComponentTypeError(
```
davidbtucker, line 378: Same feedback as earlier function: Move the comments to before the function.

    // unsupportedComponentTypeError returns ...
    func unsupportedComponentTypeError(platform, subagent, component, t, id string) error {
      ...
    }

davidbtucker (commented):
Left a few Go style comments.

ekund (approved):

collectd/conf.go
qingling128: Good to know. Changed.

collectd/conf.go
qingling128: Fixed.

confgenerator/confgenerator.go
qingling128: That seems to work. Removed.

confgenerator/confgenerator.go
qingling128: Good to know. Changed.

collectd/conf.go
davidbtucker: Also, put all the parameters on one line.

davidbtucker (approved):

PR #88, opened 2021-06-07
Hardcode the 'process' label to 'all' to match collectd behavior

sophieyfang:


otel/conf.go
```diff
@@ -402,6 +402,10 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
         operations:
           # change data type from double -> int64
           - action: toggle_scalar_data_type
+          # add process label default to 'all'
```
igorpeshansky, line 405: Why `default to 'all'`? This implies that there could be other values. The value below is constant, so this phrase is redundant, IMO.

igorpeshansky (changes requested):
One minor comment about the comment.

otel/conf.go
sophieyfang: how about `static to all 'all'`. Do you mean remove the comment as the code implies the same already?

igorpeshansky: Yes, I was suggesting to remove that part of the comment.

qingling128: Is this statement correct and appropriate as the comment instead?

```
          # Hard code "process" to "all" because the way this label is
          # populated is unpredictable and could cause potential cardinality
          # issues.
```

quentinmit: I would just say
```
# Hardcode the "process" label to "all" to match collectd behavior.
```
I don't think this is the right place to explain why collectd hardcodes it.

sophieyfang: Have changed it to `# hardcode the 'process' label to 'all' to match collectd behavior`

lowercase h as the rest comments are lowercase started. single quote as otherplaces are single quote. no ending period as other places has no period ending

igorpeshansky: Honestly, I don't think this is the right place to mention collectd either. All of the labels added/changed in this config are added/changed to match collectd behavior. We don't add explanations on any others, so why bother here? Is it not enough to just say `# add process label`?

igorpeshansky (changes requested):

qingling128 (approved):
LGTM except let's just remove the comment. The conversation in this PR should be sufficient for anyone to figure out why we do `all`.

otel/conf.go
sophieyfang: Removed comment as requested

qingling128 (approved):
LGTM

PR #90, opened 2021-06-08
Update Fluent Bit default worker number to 8 to enable multithreading.

qingling128:
The actual change is only in `fluentbit/conf/conf.go`.

Tested by b/182365367


igorpeshansky (approved):
LGTM :shipit:

PR #92, opened 2021-06-08
no fluentbit include/ dir

hopkiw:
The include dir should not be included in the main ops-agent package. If we needed this, the norm would be to ship it in a -devel package. I don't think we need it at all, though.

FLB_DEBUG defaults to true, and if set it overrides CMAKE_BUILD_TYPE to 'Debug'. Since I infer from the line that RelWithDebInfo is the desired target, this PR also sets FLB_DEBUG to false.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -52,13 +52,12 @@ function build_fluentbit() {
   cd submodules/fluent-bit
   mkdir -p build
   cd build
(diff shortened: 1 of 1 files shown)


quentinmit (approved):
Since removing `FLB_DEBUG` has the potential to make changes to throughput or stability, so we need to make sure this is tested after it goes in.

PR #95, opened 2021-06-10
Move fluent-bit binary 

hopkiw:
move the location of fluent-bit binary to /usr/bin


build.sh
```diff
@@ -52,17 +52,22 @@ function build_fluentbit() {
...
+    -DCMAKE_INSTALL_BINDIR=. -DFLB_HTTP_SERVER=ON -DFLB_DEBUG=OFF \
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWITHOUT_HEADERS=ON -DFLB_SHARED_LIB=OFF
   make -j8
   make DESTDIR="$DESTDIR" install
-  # We don't want fluent-bit's service
-  rm "$DESTDIR/lib/systemd/system/fluent-bit.service"
-  rm -r "${DESTDIR}${subagentdir}/fluent-bit/etc"
+  # We don't want fluent-bit's service or configuration, but there are no cmake
+  # flags to disable them. Prune after build.
+  rm -r "${DESTDIR}/usr/lib/google-cloud-ops-agent/etc"
```
qingling128, line 62: Does it make sense to force `fluent-bit` into the path and only remove those specific paths? e.g.

```
rm -r "${DESTDIR}/usr/lib/google-cloud-ops-agent/fluent-bit/etc"
rm -r "${DESTDIR}/lib/fluent-bit"
```

The potential benefit is to avoid accidentally removing other stuff that happens to be in the same paths.

e.g. Can we do `-DCMAKE_INSTALL_PREFIX=/usr/lib/google-cloud-ops-agent/fluent-bit` above?

qingling128 (commented):
Overall looks good to me except a few minor comments.

BTW I also created an internal tracker and asked an additional internal question at b/191185305#comment3.

systemd/google-cloud-ops-agent-fluent-bit.service
```diff
@@ -27,7 +27,7 @@ StateDirectory=google-cloud-ops-agent/fluent-bit
 LogsDirectory=google-cloud-ops-agent/subagents
 Type=simple
 ExecStartPre=@PREFIX@/libexec/google_cloud_ops_agent_engine -service=fluentbit -in @SYSCONFDIR@/google-cloud-ops-agent/config.yaml -logs ${LOGS_DIRECTORY} -state ${STATE_DIRECTORY}
-ExecStart=@PREFIX@/subagents/fluent-bit/bin/fluent-bit --config ${RUNTIME_DIRECTORY}/fluent_bit_main.conf --parser ${RUNTIME_DIRECTORY}/fluent_bit_parser.conf --log_file ${LOGS_DIRECTORY}/logging-module.log --storage_path ${STATE_DIRECTORY}/buffers
+ExecStart=/usr/bin/google-fluent-bit --config ${RUNTIME_DIRECTORY}/fluent_bit_main.conf --parser ${RUNTIME_DIRECTORY}/fluent_bit_parser.conf --log_file ${LOGS_DIRECTORY}/logging-module.log --storage_path ${STATE_DIRECTORY}/buffers
```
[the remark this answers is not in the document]
hopkiw: actually this PR is inconsistent, this should be /usr/lib/google-cloud-ops-agent/fluent-bit, will update

build.sh
hopkiw: if we set `-DCMAKE_INSTALL_PREFIX=/usr/lib/google-cloud-ops-agent/fluent-bit` the final paths will be e.g. /usr/lib/google-cloud-ops-agent/fluent-bit/fluent-bit, /usr/lib/google-cloud-ops-agent/fluent-bit/etc/fluent-bit, and /lib (the systemd install paths are not affected by the prefix variable). so in this case we'd need to do more rm's and also a 'mv'. i set the prefix to /usr/lib/google-cloud-ops-agent because that's where we want the one thing we are keeping, so i optimized for that file. let me know what you think

qingling128: Oh I see. Can we keep the path unchanged, but delete by `/usr/lib/google-cloud-ops-agent/fluent-bit`, `/usr/lib/google-cloud-ops-agent/etc/fluent-bit`? Is there anything else in `/lib` that is not Fluent Bit specific, and might get deleted by accident?

systemd/google-cloud-ops-agent-fluent-bit.service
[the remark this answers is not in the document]
qingling128: Thanks for fixing this!

build.sh
hopkiw: As build.sh is written, the fluent-bit build happens first, so only fluent-bit files are in this directory. I don't follow your alternative suggestion here, could you highlight and make a suggested change in the PR?

qingling128: I meant something like changing

```
rm -r "${DESTDIR}/usr/lib/google-cloud-ops-agent/etc"
rm -r "${DESTDIR}/lib/"
```

to

```
rm -r "${DESTDIR}/usr/lib/google-cloud-ops-agent/fluent-bit"
rm -r "${DESTDIR}/usr/lib/google-cloud-ops-agent/etc/fluent-bit"
rm -r "${DESTDIR}/lib/systemd/system/fluent-bit.service" // Not sure exactly what the path is here
```

I couldn't find a way to add `suggested changes`. I guess Github does not support it yet.

hopkiw: if we change to  `-DCMAKE_INSTALL_PREFIX=/usr/lib/google-cloud-ops-agent/fluent-bit`
then we would run the commands
```
mv "${DESTDIR}/usr/lib/google-cloud-ops-agent/fluent-bit/fluent-bit" "${DESTDIR}/usr/lib/google-cloud-ops-agent/fluent-bit"
rm -r "${DESTDIR}/usr/lib/google-cloud-ops-agent/fluent-bit"
rm -r "${DESTDIR}/lib"
```

this seemed messier but if you want i will update to this

qingling128: Yeah, that's a bit messier. Sounds like there is no way to avoid `rm -r "${DESTDIR}/lib"` any way. Might as well leave as is then.

qingling128 (approved):
LGTM.

Could we run the integration tests from this branch before merging?

Also @quentinmit - Any concern from your side before we merge? Vaguely recall you were looking at this as well at some point.

build.sh
```diff
@@ -52,17 +52,22 @@ function build_fluentbit() {
   cd submodules/fluent-bit
   mkdir -p build
   cd build
-  cmake .. -DCMAKE_INSTALL_PREFIX=$subagentdir/fluent-bit -DFLB_HTTP_SERVER=ON -DFLB_DEBUG=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWITHOUT_HEADERS=ON
+  cmake .. -DCMAKE_INSTALL_PREFIX=/usr/lib/google-cloud-ops-agent \
```
quentinmit, line 55: Took me a while to figure this out. `CMAKE_INSTALL_PREFIX` and `CMAKE_INSTALL_BINDIR` will cause the binary to be put at `/usr/lib/google-cloud-ops-agent/./fluent-bit`. `FLB_SHARED_LIB=OFF` causes it to not generate `libfluent-bit.so` at all. Can you add this in a comment for the next person?

build.sh
quentinmit, line 62: (nit) can you swap these so they're in the same order as before?

quentinmit (approved):
We clarified in chat that `/usr` will be replaced with `$prefix`/`@PREFIX@` again after the other binaries are moved in a separate PR.

qingling128 (approved):
LGTM. Ready for a merge?

sophieyfang:
This failed centos8. After installation, it doesn't start. See b/191865816#comment2

PR #104, opened 2021-06-16
Completely remove the workers config on Windows.

qingling128:
Fixes b/190806301#comment5.

Integration tests are done in b/190806301#comment6.

### DIFF
--- a/fluentbit/conf/conf.go
+++ b/fluentbit/conf/conf.go
@@ -255,8 +255,10 @@ const (
     # https://docs.fluentbit.io/manual/pipeline/outputs/stackdriver
     Name              stackdriver
     resource          gce_instance
-    stackdriver_agent {{.UserAgent}}{{if .Workers}}
-    workers           {{.Workers}}{{end}}
+    stackdriver_agent {{.UserAgent}}
+    {{- if .Workers}}
+    workers           {{.Workers}}
+    {{- end}}
     Match_Regex       ^({{.Match}})$
 
     # https://docs.fluentbit.io/manual/administration/scheduling-and-retries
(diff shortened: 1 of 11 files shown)


fluentbit/conf/conf.go
```diff
@@ -255,8 +255,7 @@ const (
     # https://docs.fluentbit.io/manual/pipeline/outputs/stackdriver
     Name              stackdriver
     resource          gce_instance
-    stackdriver_agent {{.UserAgent}}
-    workers           {{.Workers}}
+    stackdriver_agent {{.UserAgent}}{{.Workers}}
```
davidbtucker, line 258: Could you just do this instead?

    {{if .Workers}}
    workers {{.Workers}}
    {{end}}

quentinmit: Yeah, that would be much better than the code change in confgenerator.go.

You can even get slightly shorter:
```
{{with .Workers}}
workers {{.}}
{{end}}
```

quentinmit (changes requested):

fluentbit/conf/conf.go
qingling128: Changed to 

```
    stackdriver_agent {{.UserAgent}}{{if .Workers}}
    workers           {{.Workers}}{{end}}
```

Instead. It looks a bit weird because I had to make it on the same line to avoid it generating an additional new line.

fluentbit/conf/conf.go
```diff
@@ -255,8 +255,8 @@ const (
     # https://docs.fluentbit.io/manual/pipeline/outputs/stackdriver
     Name              stackdriver
     resource          gce_instance
-    stackdriver_agent {{.UserAgent}}
-    workers           {{.Workers}}
+    stackdriver_agent {{.UserAgent}}{{if .Workers}}
```
davidbtucker, line 258: (Optional) You could put it on a new line and do something like

    {{- if .Workers}}

to clean up the whitespace.

davidbtucker (approved):

fluentbit/conf/conf.go
qingling128: Ah good idea. Changed.

PR #109, opened 2021-06-22
Fix processes/fork_count metric

jsuereth:
- Update host metrics receiver to also pull in `processes` sub-receiver.
- Add rename rule for `processes` fork_count metric.
- Update unit tests, ran manual test

### DIFF
--- a/confgenerator/testdata/valid/linux/default_config/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/default_config/golden_otel.conf
@@ -17,6 +17,7 @@ receivers:
       network:
       paging:
       process:
+      processes:
 processors:
   resourcedetection:
     detectors: [gce]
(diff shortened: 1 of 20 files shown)


qingling128 (approved):
LGTM. Integration tests passed: b/185811300#comment15

sophieyfang:
This PR introduced a no-harm polluting messages in otel logs because `processes` introduces two metrics one we use it for `fork_count` the other for `count_by_state` which we don't support in GA thus shall exclude it:

```
2021-06-24T22:24:54.501Z	info	exporterhelper/queued_retry.go:316	Exporting failed. Will retry the request after interval.	{"kind": "exporter", "name": "googlecloud/google", "error": "rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Unknown metric: agent.googleapis.com/system.processes.count: timeSeries[44,45]", "interval": "6.071613379s"}
```

jsuereth:
@sophieyfang Yeah, sorry bout that, was following up with a PR on a follow-on bug to clean that up.

PR #110, opened 2021-06-23
Redirect info log messages to stdout.

igorpeshansky:
Fixes b/191809260.

### DIFF
--- a/cmd/ops_agent_windows/main_windows.go
+++ b/cmd/ops_agent_windows/main_windows.go
@@ -21,6 +21,7 @@ var (
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


punya:
The code looks good, but why do we want to do this?

qingling128:
> The code looks good, but why do we want to do this?

There was a customer issue reported via an email. (b/191809260 is the tracking bug.) Not a GA blocker, but looks like a low hanging fruit.

igorpeshansky:
We should do it because PowerShell presents any messages sent to stderr as errors.

qingling128 (approved):
LGTM

PR #113, opened 2021-06-25
Remove support for collectd config generation and fold config structs back into the confgenerator package.

igorpeshansky:
A follow-on to #108.


quentinmit (approved):
Yay, definite improvement. There's still lots more simplification that can be done in this code. There's a whole layer of translation that's completely unnecessary.

Either a) the otel config template can just render directly from the UnifiedConfig struct, or b) the functions in confgenerator.go can generate the otel structs directly and then serialize as yaml

But, one thing at a time and that's not a GA blocker. This PR looks ready to merge as-is!

PR #114, opened 2021-06-25
fix rpm spec for fluent-bit path

hopkiw:
missed in previous PR


qingling128 (approved):
LGTM. This fixed all the builds: b/191865816#comment16. The `centos 8` test failures still exist, which can be addressed in a followup PR.

qingling128:
Sync'd with Liam offline that we could merge this.

PR #115, opened 2021-06-25
Support metrics exclusion by prefixes

sophieyfang:
Test evidence:

Error msg when prefixes not ending with "/"
Screen Shot 2021-06-25 at 11 47 55 AM

Before exclusion and After exclusion:
image
image


confgenerator/config/config.go
```diff
@@ -521,10 +510,19 @@ var (
...
-		"mssql":           collectionIntervalValidation,
-		"exclude_metrics": metricPrefixesValidation,
+		"hostmetrics": collectionIntervalValidation,
+		"iis":         collectionIntervalValidation,
+		"mssql":       collectionIntervalValidation,
+		"exclude_metrics": map[string]func(interface{}) error{
+			"metric_prefixes": func(v interface{}) error {
+				for _, prefix := range v.([]string) {
+					if !strings.HasSuffix(prefix, "/") {
+						return fmt.Errorf(`must end with "/"`)
```
quentinmit, line 520: Put the prefix in the error message, such as

```
fmt.Errorf(`prefix %q must end with "/"`, prefix)
```

confgenerator/confgenerator.go
```diff
@@ -377,6 +408,37 @@ func generateOtelExporters(exporters map[string]*config.MetricsExporter, pipelin
...
+		return nil, nil, err
+	}
+	for _, pID := range config.SortedKeys(pipelines) {
+		p := pipelines[pID]
+		for _, processorID := range p.ProcessorIDs {
+			if _, ok := processorNameMap[processorID]; ok {
+				continue
+			}
+			if p, ok := excludemetricsProcessorFactories[processorID]; ok {
+				var metricPrefixes []string
```
quentinmit, line 425: This kind of string manipulation should be done in the template. Just put the raw metric prefix slice into the config struct.

(This is especially important when we have workload metrics to drop, and it should make things shorter to boot.)

quentinmit (changes requested):

confgenerator/config/config.go
sophieyfang: See error msgs in:
```
sophieyfang@sophieyfang:~/tmp/ops-agent$ cat confgenerator/testdata/invalid/linux/metrics-processor_unsupported_type/golden_error
metrics processor "processor_1" with type "unsupported_type" is not supported. Supported metrics processor types: [exclude_metrics].
```

confgenerator/confgenerator.go
sophieyfang: Can you clarify this a bit more?

sophieyfang: Are you implying me to do something similar to this: 
image

also, do we want to do this now. this is not the only place we massage the data in confgenerator. Removing the massage data into struct can be done in a separate PR.

confgenerator/config/config.go
quentinmit: Okay, that doesn't say which list item but it's not terrible.

confgenerator/confgenerator.go
quentinmit: Recreating this comment thread after rebase.

> also, do we want to do this now. this is not the only place we massage the data in confgenerator. Removing the massage data into struct can be done in a separate PR.

Just because existing code is bad is not a good reason to write more bad code. :)

I think there's three different string manipulations that need to happen between the OA config and the OT config:

1. `agent.googleapis.com/` needs to be stripped as a prefix (with a giant TODO to fix this when workload metrics are supported; actually you should probably make sure that it validates that this starts with `agent.googleapis.com/` until workload metrics are supported)
2. Globs need to be turned into regexes
3. Lists of strings needs to be formatted as valid YAML

The third one should definitely happen in the template. Something like

```
metric_names:
{{range .MetricNames}}
  - {{printf "%q" .}}
{{end}}
```

That properly handles the escaping you need, which your current code doesn't even attempt.

The other two are a bit more ambiguous about where they should be; it might be reasonable to keep them here for now but 1 eventually goes away with workload metrics and 2 would be best as a filter function in the template. Your `strings.TrimPrefix` call is fine to remove `agent.googleapis.com`, but you need to actually handle the glob and properly escape the special characters in the string. That would be something like:

```
var literals []string
for _, p := range strings.Split(glob, "*") {
  literals = append(literals, regexp.QuoteMeta(p))
}
metricNames = append(metricNames, fmt.Sprintf(`^%s$`, strings.Join(literals, `.*`)))
```

to replace lines 429-430 in this file.

Escaping is really important, parsing globs is important, and generating YAML in one place instead of multiple is important.

sophieyfang: `for _, p := range strings.Split(glob, "*")` 
This line seems predicting the prefixes could be something like `agent.googleapis.com\c*u\*` which i think a good idea implementation-wise. As we might support more generic field other than `metric_prefixes` to something like `metric_globs` then the same code can be used for both cases?
Thus fixes applied

confgenerator/config/config.go
sophieyfang: Changed to
```
sophieyfang@sophieyfang:~/tmp/ops-agent$ cat confgenerator/testdata/invalid/linux/metrics-processor_exclude_metrics_invalid_parameter_metric_prefixes/golden_error
parameter "metric_prefixes" in "exclude_metrics" type metrics processor "metrics_filter" has invalid value ["agent.googleapis.com/processes/not_end_with_forward_slash_with_star" "not.agent.googleapis.com/processes/*"]: agent.googleapis.com/processes/not_end_with_forward_slash_with_star must end with "/*" | not.agent.googleapis.com/processes/* must start with "agent.googleapis.com/"
```

confgenerator/confgenerator.go
quentinmit: I suggested that because Joe specifically Decided that these needed to be globs and not just prefixes, yeah.

confgenerator/config.go
```diff
@@ -460,6 +498,24 @@ var (
 		"hostmetrics": collectionIntervalValidation,
 		"iis":         collectionIntervalValidation,
 		"mssql":       collectionIntervalValidation,
+		"exclude_metrics": map[string]func(interface{}) error{
+			"metric_prefixes": func(v interface{}) error {
+				var errors []string
+				for _, prefix := range v.([]string) {
+					if !strings.HasSuffix(prefix, "/*") {
+						errors = append(errors, fmt.Sprintf(`%s must end with "/*"`, prefix))
```
quentinmit, line 506: Nit: %q looks better than %s

confgenerator/config.go
```diff
@@ -460,6 +498,24 @@ var (
...
+					}
+					// TODO: Relax prefix must be `agent.googleapis.com/` check, when it starts supporting metrics with other prefixes
+					if !strings.HasPrefix(prefix, "agent.googleapis.com/") {
+						errors = append(errors, fmt.Sprintf(`%s must start with "agent.googleapis.com/"`, prefix))
```
quentinmit, line 510: Ditto

confgenerator/testdata/valid/linux/metrics-exclude_metrics_by_prefixes/golden_collectd.conf
```diff
@@ -0,0 +1,110 @@
+Interval 60
```
quentinmit, line 1: I thought we removed collectd, why is this golden file here?

tmp-config.yaml
```diff
@@ -0,0 +1,50 @@
+# Copyright 2020 Google LLC
```
quentinmit, line 1: Accidentally included this file?

confgenerator/testdata/valid/linux/metrics-exclude_metrics_by_prefixes/input.yaml
```diff
@@ -0,0 +1,50 @@
...
+metrics:
+  receivers:
+    hostmetrics:
+      type: hostmetrics
+      collection_interval: 60s
+  processors:
+    metrics_filter:
+      type: exclude_metrics
+      metric_prefixes:
+      - agent.googleapis.com/processes/*
```
quentinmit, line 40: You should include one with a special character to ensure it gets handled correctly.

e.g.

- agent.googleapis.com/my.metric.name/*

confgenerator/config.go
sophieyfang: Sure~

confgenerator/testdata/valid/linux/metrics-exclude_metrics_by_prefixes/input.yaml
sophieyfang: Will do

confgenerator/confgenerator.go
```diff
@@ -378,6 +410,43 @@ func generateOtelExporters(exporters map[string]*MetricsExporter, pipelines map[
...
+		for _, processorID := range p.ProcessorIDs {
+			if _, ok := processorNameMap[processorID]; ok {
+				continue
+			}
+			if p, ok := excludemetricsProcessorFactories[processorID]; ok {
+				var metricNames []string
+				for _, glob := range p.MetricPrefixes {
+					// TODO: Relax trim prefix `agent.googleapis.com/`, when it starts supporting metrics with other prefixes
+					glob = strings.TrimPrefix(glob, "agent.googleapis.com/")
+					// TODO: Move this glob to regexp into a filter function inside otel/conf.go as template function.
```
igorpeshansky, line 431: `// TODO: Move this glob to regexp into a template function inside otel/conf.go.`

confgenerator/confgenerator.go
```diff
@@ -378,6 +410,43 @@ func generateOtelExporters(exporters map[string]*MetricsExporter, pipelines map[
...
+	for _, pID := range sortedKeys(pipelines) {
+		p := pipelines[pID]
+		for _, processorID := range p.ProcessorIDs {
+			if _, ok := processorNameMap[processorID]; ok {
+				continue
+			}
+			if p, ok := excludemetricsProcessorFactories[processorID]; ok {
+				var metricNames []string
+				for _, glob := range p.MetricPrefixes {
+					// TODO: Relax trim prefix `agent.googleapis.com/`, when it starts supporting metrics with other prefixes
```
igorpeshansky, line 429: What's "it"? Also, a nit: missing trailing `.`. Maybe something like this?

``// TODO: Remove TrimPrefix when we support metrics with other prefixes.``

confgenerator/config.go
```diff
@@ -130,6 +131,16 @@ type MetricsReceiver struct {
 	CollectionInterval string `yaml:"collection_interval"` // time.Duration format
 }
 
+type MetricsProcessorExcludeMetrics struct {
+	MetricPrefixes []string `yaml:"metric_prefixes"`
```
igorpeshansky, line 135: If these have to be globs, are they prefixes anymore?

confgenerator/config.go
```diff
@@ -460,6 +498,24 @@ var (
...
 		"iis":         collectionIntervalValidation,
 		"mssql":       collectionIntervalValidation,
+		"exclude_metrics": map[string]func(interface{}) error{
+			"metric_prefixes": func(v interface{}) error {
+				var errors []string
+				for _, prefix := range v.([]string) {
+					if !strings.HasSuffix(prefix, "/*") {
+						errors = append(errors, fmt.Sprintf(`%q must end with "/*"`, prefix))
+					}
+					// TODO: Relax prefix must be `agent.googleapis.com/` check, when it starts supporting metrics with other prefixes
```
igorpeshansky, line 508: Same question: what's "it"? And same nit: missing trailing `.`. So maybe something like this?

``// TODO: Relax the prefix check when we support metrics with other prefixes.``

confgenerator/config.go
```diff
@@ -460,6 +498,24 @@ var (
...
+					if !strings.HasPrefix(prefix, "agent.googleapis.com/") {
+						errors = append(errors, fmt.Sprintf(`%q must start with "agent.googleapis.com/"`, prefix))
+					}
+				}
+				if len(errors) > 0 {
+					return fmt.Errorf(strings.Join(errors, " | "))
```
igorpeshansky, line 514: In other validation functions, we only return the first error we encounter (this pattern occurs elsewhere in the code, too). We can make a decision to aggregate all errors, but we could also just replicate existing behavior.

igorpeshansky (changes requested):

confgenerator/config.go
sophieyfang: At least for this exclude-metrics filter, I am really not a fan of only reproducing the first encountered error. 
you can see that error msg screenshot in PR script. In the msg it prints the list, but it doesn’t mention which one is the problematic one. And the second point is: there is another possible invalid cause which is not start with our agent.googleapis….I actually really prefer the current code. 
At least for each processor, receiver, exporter, it shall accumulate errors and print it out  at once. 
As for, do we want to process the whole config and accumulate all errors across receivers, processors, exporters, that’s good to have but not as important as the one above.

confgenerator/config.go
sophieyfang: The question will be for @quentinmit. I have this same question too. As Ling posted the final final syntax decision in the bug, it’s metric_prefixes, but this part of code is really doing partial globs. Partial meaning the prefix has to be agent.google…../ and the suffix has to be /*. Then things in between is glob.. is that really a glob?

related comment: https://github.com/GoogleCloudPlatform/ops-agent/pull/115#discussion_r659949800

confgenerator/config.go
igorpeshansky: Honestly, I prefer error accumulation anyway — just pointing out the inconsistency with the existing code.

confgenerator/config.go
quentinmit: I thought we had settled on a different name, yeah. Not sure what exactly it was... `exclude_metrics`? @qingling128 ?

confgenerator/confgenerator.go
sophieyfang: it's now metrics_pattern

confgenerator/config.go
qingling128: `metrics_pattern` is the one we agreed on offline. The change has been made to this PR already.

confgenerator/config.go
qingling128: As discussed offline, having the errors accumulated for this one parameter is fine. So we'll keep it as is.

igorpeshansky (approved):
LGTM :shipit:

PR #116, opened 2021-06-25
Average out cpu utilization over CPUs

sophieyfang:
Test evidence:
image

### DIFF
--- a/otel/conf.go
+++ b/otel/conf.go
@@ -133,7 +133,7 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
         - prometheus/agent
       processors:
         - filter/agent
-        - metricstransform/agent
+        - googlemetricstransform/agent
         - resourcedetection
       exporters:
         - googlecloud/agent
@@ -150,7 +150,7 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
   resourcedetection:
     detectors: [gce]
 
-  # perform custom transformations that aren't supported by the metricstransform processor
+  # perform custom transformations that aren't supported by the googlemetricstransform processor
   agentmetrics/system:
     # 1. converts up down sum types to gauges
     # 2. combines resource process metrics into metrics with processes as labels
@@ -168,7 +168,7 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
           - system.paging.faults
 
   # convert from opentelemetry metric formats to cloud monitoring formats
-  metricstransform/system:
+  googlemetricstransform/system:
     transforms:
       # system.cpu.time -> cpu/usage_time
       - metric_name: system.cpu.time
@@ -505,7 +505,7 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
           - otelcol_googlecloudmonitoring_point_count
 
   # convert from windows perf counter formats to cloud monitoring formats
-  metricstransform/iis:
+  googlemetricstransform/iis:
     transforms:
       - include: \Web Service(_Total)\Current Connections
         action: update
@@ -525,7 +525,7 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
         submatch_case: lower
 
   # convert from windows perf counter formats to cloud monitoring formats
-  metricstransform/mssql:
+  googlemetricstransform/mssql:
     transforms:
       - include: \SQLServer:General Statistics(_Total)\User Connections
         action: update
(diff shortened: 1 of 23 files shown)


otel/conf.go
```diff
@@ -184,10 +184,14 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
         action: update
         new_name: cpu/utilization
         operations:
-          # change label cpu -> cpu_number
-          - action: update_label
-            label: cpu
+          # take avg over cpu dimension, retaining only label: state
```
igorpeshansky, line 187: The convention in the rest of the comments seems to be "_label_name_ label", so that would make it `retaining only state label`.

otel/conf.go
```diff
@@ -184,10 +184,14 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
...
+          - action: aggregate_labels
+            label_set: [state]
+            aggregation_type: mean
+          # create new label cpu_number
+          - action: add_label
             new_label: cpu_number
+            new_value: " "
```
igorpeshansky, line 194: This should be an empty string, not a space.

otel/conf.go
```diff
@@ -184,10 +184,14 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
...
         new_name: cpu/utilization
         operations:
-          # change label cpu -> cpu_number
-          - action: update_label
-            label: cpu
+          # take avg over cpu dimension, retaining only label: state
+          - action: aggregate_labels
+            label_set: [state]
+            aggregation_type: mean
+          # create new label cpu_number
```
igorpeshansky, line 191: `add blank cpu_number label`

igorpeshansky (changes requested):

otel/conf.go
sophieyfang: `new_value: ""` or `new_value: ` will fail this.

2021/06/25 19:28:43 application run finished with error: cannot build pipelines: cannot build pipelines: error creating processor "metricstransform/system" in pipeline "metrics/logging": operation 2: missing required field "new_value" while "action" is add_label

otel/conf.go
sophieyfang: done

otel/conf.go
sophieyfang: done

otel/conf.go
igorpeshansky: Crap. This is a Go problem. The value should really be a blank string — having a space there is incorrect. We should probably pull in the OT folks for advice here. @punya, any thoughts on this?

punya: Hmm, this is enforced by https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.28.0/processor/metricstransformprocessor/factory.go#L128-L130 . I think we should be able to convince people to merge an upstream PR to remove that restriction, but it won't land in time for GA.

If we have to ship with a known bug, which one would be better?
Option 1 -- space instead of blank value
Option 2 -- remove the label

sophieyfang: @qingling128 to decide

igorpeshansky: Both are bad and incompatible with the current solution. Another option is to add a patch into our otel submodule for GA and remove it once the change gets merged upstream. @qingling128 thoughts?

qingling128: Does `a patch into our otel submodule` mean we will need to:
1. Fork https://github.com/open-telemetry/opentelemetry-collector-contrib (The `GoogleCloudPlatform` org takes a long time. So we have to fork to `Stackdriver` org or a personal org for GA timeline, right?) and put a patch there.
2. Update https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector to point to a hash in that fork.
3. Update https://github.com/GoogleCloudPlatform/ops-agent to point to that version of `opentelemetry-operations-collector`.

Or are we talking about something more light weighted?

We have been trying to avoid a fork. But in the long-term, having a fork for urgent / critical bug fixes like this sounds inevitable to me. We'll need to be disciplined to upstream fixes within certain time frame to avoid maintaining a fork that diverges from upstream too far for too long though.

@jsuereth @punya @quentinmit What do you think from the otel perspective?

punya: Yes, that's pretty much what we'd have to do. There might be a way to do steps 1+2 staying within just this repo (I'll make a PR to illustrate today). Step 3 is unavoidable, of course.

qingling128: Discussed offline for this. Decided to do a custom processor (copy, paste, edit it to remove that validation). Details in this PR: https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/50

CC @jsuereth for visibility.

qingling128 (approved):
LGTM

sophieyfang:
Forgot to make it "" instead of " " in cpu_number --Followed up in: https://github.com/GoogleCloudPlatform/ops-agent/pull/121

PR #118, opened 2021-06-28
Extract package version from the VERSION file when building on Windows.

igorpeshansky:
Also allow overriding the build distro.

This fixes the version label for the uptime metric and the user agent string.
The version for the metrics agent is now sent as `google-cloud-ops-agent-metrics/2.0.0-windows-ltsc2019`.


qingling128 (approved):
LGTM

PR #122, opened 2021-06-29
Drop system.cpu.time (cpu/usage_time).

igorpeshansky:
Fixes b/192293922

### DIFF
--- a/confgenerator/testdata/valid/linux/default_config/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/default_config/golden_otel.conf
@@ -35,6 +35,8 @@ processors:
       exclude:
         match_type: strict
         metric_names:
+          # Temporarily exclude system.cpu.time (cpu/usage_time)
+          - system.cpu.time
           - system.network.dropped
           - system.filesystem.inodes.usage
           - system.paging.faults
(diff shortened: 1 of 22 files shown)


confgenerator/testdata/valid/linux/default_config/golden_otel.conf
```diff
@@ -35,6 +35,8 @@ processors:
       exclude:
         match_type: strict
         metric_names:
+          # Temporarily exclude system.cpu.time (cpu/usage_time)
+          - system.cpu.time
```
sophieyfang, line 39: Have you built the agent and test out this config?
As in line 47, system.cpu.time is converted in to cpu/usage_time. I am concerned if it will produce error msgs, as it will be converting something that doesn't exist

sophieyfang: Just tested, it won't cause any warning/error msgs.

sophieyfang (approved):

qingling128 (approved):
LGTM

qingling128:
Igor is out till later tonight, and this seems to be a straight forward enough change. And Sophie tested it out. So I'm gonna merge it.

PR #123, opened 2021-06-30
Fix disk/operation_time to report average operation time

quentinmit:
collectd reported average operation time; OT reports total operation time, so we need to introduce a new metric for average operation time and divide the existing metrics.

Before this can be merged we need to merge https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/49 and update the submodule in this PR.

### DIFF
--- a/otel/conf.go
+++ b/otel/conf.go
@@ -167,6 +167,7 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
           - system.filesystem.inodes.usage
           - system.paging.faults
           - system.disk.operation_time
+          - system.processes.count
 
   # convert from opentelemetry metric formats to cloud monitoring formats
   googlemetricstransform/system:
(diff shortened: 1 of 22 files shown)


otel/conf.go
```diff
@@ -166,6 +166,8 @@ windowsperfcounters/mssql_{{.MSSQLID}}:
           - system.network.dropped
           - system.filesystem.inodes.usage
           - system.paging.faults
+          - system.disk.operation_time
+          - system.processes.count
```
qingling128, line 170: Should this line be reflected in the other golden files as well?

quentinmit: Yes, it should. Forgot to regenerate the golden files again.

qingling128 (approved):
LGTM

otel/conf.go
qingling128: The good thing is that our tests caught it as well, which is what they are designed for :D

PR #124, opened 2021-06-30
Include processors in OTel config only as needed

djaglowski:
Preconfigured OTel processors are typically associated with one or more specific receiver types. For example, a processor that renames MSSQL metrics is only used when a MSSQL receiver exists.

Currently, all predefined processors are included in the OTel config files, regardless of which receivers are used.

This changes introduce an association between processors and receivers, such that processors will be included only when their associated receiver is part of the configuration. This pattern is particularly helpful when considering the impact of integrating additional receivers. Without correlating receivers and processors, each additional receiver type will alter all of the golden configs files. With this change, those files will be unaffected.

Summary of changes:
- `conf.go` contains all the actual code changes
- `valid/linux/*/golden.conf` no longer contain processors for `mssql` or `iis`
- `valid/windows/*golden.conf` contain processors for `mssql` and `iss` only if the corresponding receivers are included.
- A few empty lines were deleted. This is an artifact of the templating layout.


otel/conf.go
```diff
@@ -33,7 +33,14 @@ var confTemplate = template.Must(template.New("conf").Parse(
...
 processors:
-  {{template "defaultprocessor" .}}
+  {{template "systemprocessor" .}}
+{{- if .IIS}}
+  {{template "iisprocessor" .}}
+{{- end}}
+{{- if .MSSQL}}
+  {{template "mssqlprocessor" .}}
+{{- end}}
+  {{template "agentprocessor" .}}
```
djaglowski, line 43: It would make sense to move this line, so that it immediately follows `{{template "systemprocessor" .}}`. 

I opted not to make this change yet, so as to minimize changes to the golden configs.

PR #125, opened 2021-06-30
Set Read_from_Head to True

sophieyfang:


qingling128 (changes requested):
Looks like `submodules/opentelemetry-operations-collector` was updated by accident. Can we revert that? (Doing `git submodule update` should revert it)

qingling128 (approved):
LGTM

PR #126, opened 2021-06-30
Update Fluent Bit to v1.7.9 and enable multithreading on Windows.

jkschulz:
Default to 8 workers (same as Linux). I left it in its own block since it seems likely we'd change Windows or Linux independently in the future, but I can consolidate if we want to reinforce the "consistency" aspect. 

http://b/191158630


confgenerator/confgenerator.go
```diff
@@ -193,8 +193,8 @@ func getUserAgent(prefix string, hostInfo *host.InfoStat) (string, error) {
 func getWorkers(hostInfo *host.InfoStat) int {
```
qingling128, line 193: I'd lean towards removing the conditions here for now, as we do not know when there will be a plan to diverge them. And if there is, it's simple to add it back then.

qingling128 (approved):
LGTM with one suggestion.

Rob-S:
Is there an expected timeframe for when this FluentBit 1.7.9 (or any release 1.7.8 or later, referring to Google Cloud Support Case #28072890) is going to be incorprated into the Ops Agent? We keep having issues with the FluentBit log file filling up the C: drive on various servers.

jkschulz:
> Is there an expected timeframe for when this FluentBit 1.7.9 (or any release 1.7.8 or later, referring to Google Cloud Support Case #28072890) is going to be incorprated into the Ops Agent? We keep having issues with the FluentBit log file filling up the C: drive on various servers.

@Rob-S There ended up being an issue with 1.7.9 (see https://github.com/fluent/fluent-bit/issues/3646#issuecomment-883663469). It's already been [fixed](https://github.com/fluent/fluent-bit/pull/3836) and will be released in 1.8.3. Once that's out (hopefully this week), we'll test and get this in ASAP.

jkschulz:
Replacing this PR with https://github.com/GoogleCloudPlatform/ops-agent/pull/156 &mdash; follow updates there.

PR #129, opened 2021-07-01
ops-agent.service after subagents service

sophieyfang:
### Problem

By 50% chance(because of the inconsistency), on restarting the agent on an invalid conf, client now finds out that agent not running only after missing metrics or logs for a while. I think this might cause some support pain on us considering customizing conf is allowed and sometimes it can go wrong by client.

### Background

While working with jimmy park on re-confirming the behavior of Troubleshooting invalid conf in GA, we realized sometimes `restart` on invalid conf doesn't emit any error msgs and echo $? also says 0. As the document we wrote for beta said it would have error msg if the agent didn't restart right. But we smartly use *might*. 

### `target` behaves consistently with invalid conf
 
- To confirm if `target` behaves differently with `service`. I installed `1.*.* the latest` where we still use `sudo systemctl restart google-cloud-ops-agent.target`. If the conf is invalid on restart, `echo $?` constantly show `1`. See: image

### `service` behave inconsistently on invalid conf
image

## Solution

After reading the code and some doc, I think the inconsistent behavior of `restart service` on invalid conf is because ops-agent.service doesn't wait for `fluentbit.service` and `otel.service`. The conversion code is triggered in `ExecStartPre=` in those two subagents `service` file. When the conversion fails, it happens in those two subagents service. If we let `ops-agent.service` to wait for those two subagents service, it might work. Therefor, I tried putting back this line `After=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opentelemetry-collector.service` inside `ops-agent.service` file. And restarting service on invalid configuration now consistently emit error.

### Test evidence:
image

### DIFF
--- a/systemd/google-cloud-ops-agent.service
+++ b/systemd/google-cloud-ops-agent.service
@@ -15,6 +15,7 @@
 [Unit]
 Description=Google Cloud Ops Agent
 Requires=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opentelemetry-collector.service
+After=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opentelemetry-collector.service
 
 [Service]
 Type=oneshot


quentinmit (dismissed):
Omitting this was a conscious choice.

See https://systemd.network/systemd.unit.html

> If one of the other units fails to activate, and an ordering dependency After= on the failing unit is set, this unit will not be started.

This will cause one broken subagent to take down the others. I specifically did not want that to happen.

quentinmit:
Also, this change will only propagate some failures, not all.

qingling128:
What's the behavior with and without this `After=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opentelemetry-collector.service` setting for the cases below?

* The installation script installs the package and try to start the root service, but the user config is invalid
* The installation script installs the package and try to start the root service, but one of the services for subagents could not start because it has some issues
* The agent was working previously with the 3 services healthy, suddenly one of the services for the subagents failed

sophieyfang:
> Omitting this was a conscious choice.
> 
> See https://systemd.network/systemd.unit.html
> 
> > If one of the other units fails to activate, and an ordering dependency After= on the failing unit is set, this unit will not be started.
> 
> This will cause one broken subagent to take down the others. I specifically did not want that to happen.

Oh, i don't know this.... By broken agent, are you saying:
- one agent binary is broken
- our conversion code has bug, and the agent conf it generates is not understandable by that agent binary
and somehow above cases passes our across distro e2e tests and it gets released publicly

sophieyfang:
> What's the behavior with and without this `After=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opentelemetry-collector.service` setting for the cases below?
> 
> 1 The installation script installs the package and try to start the root service, but the user config is invalid
> 2 The installation script installs the package and try to start the root service, but one of the services for subagents could not start because it has some issues
> 3 The agent was working previously with the 3 services healthy, suddenly one of the services for the subagents failed

## Regarding to point 1: The installation script installs the package and try to start the root service, but the user config is invalid

### TL,DR: if the config.yaml doesn't pass any our validation, both agents will be in failed state in agents. It's because the `ExecStartPre` failed in both agents. We somehow reinforce the behavior. In 1.*, if only logging side fails, only logging agent is in fail state. 

Below statements hold true for repo that `REPO_SUFFIX=20210629-4` still uses target and  `2.0.0` which we released out yesterday
- If the config is very broken in yaml format, both agents will fail on start. 
- If the config is only broken in e.g. logging side -- an invalid receiver type, both agents will fail on start.
Evidence: 
`REPO_SUFFIX=20210629-4` uses still target
image
`2.0.0` which we released yesterday 
image

## Regarding to point2: The installation script installs the package and try to start the root service, but one of the services for subagents could not start because it has some issues

It's quite hard to produce this scenario, if it's the binary itself has problem. Mostly, it will be caught by our e2e test and it won't be released.

## Regarding to point 3: The agent was working previously with the 3 services healthy, suddenly one of the services for the subagents failed

i think `After=xxxx` in `.service` only affects when starting the service not after. but i might be wrong.

sophieyfang:
> Omitting this was a conscious choice.
> 
> See https://systemd.network/systemd.unit.html
> 
> > If one of the other units fails to activate, and an ordering dependency After= on the failing unit is set, this unit will not be started.
> 
> This will cause one broken subagent to take down the others. I specifically did not want that to happen.

I tried it on 1.* where we didn't reinforce conversion check that strict. That one has the after in the `ops-agent.target`, and let config.yaml file only has invalid type in logging section. Restarting the agent, only fluentbit failed on start, the other agent is still running happily. If the ordering within `After=A B` affects one another, If A failed on start, B doesn't get affected. But i might be wrong. Need someone who knows the systemd really well..

image

sophieyfang:
Update it here:
Discussed with @quentinmit @igorpeshansky. Ideally, the config validation check shall be done in top unit file. That way it will also print out the error msg and exit-code correctly.

Orthogonal to above point, do we want to `After=fluentbit.service otel.service` to ensure that the listed unit is fully started up before the configured unit is started?

sophieyfang:
Rephrase `The ordering isn't within the list itself.` to `There is no dependency of win together and lose together rule within the elements inside list of `After=list`

quentinmit (changes requested):
Let's talk about this in a GVC to nail down the exact semantics we actually want for this.

quentinmit:
Covered in #132 and #133, I believe

PR #130, opened 2021-07-01
Remove exporters from default config and ignore in user config

quentinmit:
Since there is currently nothing users can usefully do with exporters, removing the fields from the default configs will reduce user confusion.

The fields are still parsed (but ignored), which means we're backwards compatible, and we can always reintroduce them in the future if needed.


confgenerator/config.go
```diff
@@ -208,8 +209,8 @@ func (l *Logging) Validate(platform string) error {
 	if err := validateComponentIds(l.Processors, subagent, "processor"); err != nil {
 		return err
 	}
-	if err := validateComponentIds(l.Exporters, subagent, "exporter"); err != nil {
-		return err
+	if len(l.Exporters) > 0 {
+		log.Print(`The "metrics.exporters" field is deprecated and will be ignored. Please remove it from your configuration.`)
```
qingling128, line 213: `The "metrics.exporters" field is deprecated as it's no longer needed, and will be ignored. This does not change any functionality. Please remove it from your configuration.`? Anything that can help assure users `nothing is wrong, please ignore and chill. No need to send a customer support case` is good with me.

confgenerator/confmerger.go
```diff
@@ -216,13 +192,12 @@ func mergeConfigs(original, overrides *UnifiedConfig) {
...
-		// Overrides logging.exporters.
-		for k, v := range overrides.Logging.Exporters {
-			original.Logging.Exporters[k] = v
-		}
+		// Skip deprecated logging.exporters.
+		// Override logging.service.pipelines
 		if overrides.Logging.Service != nil {
 			for name, pipeline := range overrides.Logging.Service.Pipelines {
+				// skips logging.service.pipelines.*.exporters
+				pipeline.ExporterIDs = nil
```
qingling128, line 200: Do we need this line after the exporter overrides is removed below?

qingling128: Ah, answered myself: we need it for non-default pipelines. Ignore the comment above

confgenerator/config.go
quentinmit: Updated. Tweaked the grammar slightly on your suggestion.

qingling128 (approved):
LGTM

PR #132, opened 2021-07-02
Fix "inactive (dead)" main service

quentinmit:
Setting `RemainAfterExit` will cause systemd to show "active (exited)" as the state of the parent service, instead of "inactive (dead)".

NOTE that this appears to also cause systemd to propagate failure/restart of one subagent onto the parent agent and then onto the other subagent. This isn't necessarily a bad thing but it's a change. AFAICT it's completely undocumented that failure of a dependency will propagate to the parent - the docs say

> Besides, with or without specifying After=, this unit will be stopped if one of the other units is explicitly stopped.

but it seems to apply whether the subagent is explicitly stopped or it crashes. 

[See [comment](https://github.com/GoogleCloudPlatform/ops-agent/pull/132#issuecomment-873155707), it doesn't seem to consistently pass down error message to the parent]

We could switch to a `Wants` dependency, but then the parent service would also not fail to initially start if one of the children fails to start. I guess we could combine that with config validation in the parent service to get the best of both worlds.

/cc @sophieyfang

### DIFF
--- a/systemd/google-cloud-ops-agent.service
+++ b/systemd/google-cloud-ops-agent.service
@@ -14,7 +14,7 @@
 
 [Unit]
 Description=Google Cloud Ops Agent
-Requires=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opentelemetry-collector.service
+Wants=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opentelemetry-collector.service
 
 [Service]
 Type=oneshot
--- a/systemd/google-cloud-ops-agent.service
+++ b/systemd/google-cloud-ops-agent.service
@@ -21,6 +21,7 @@ Type=oneshot
 # Validate the config.
 ExecStartPre=@PREFIX@/libexec/google_cloud_ops_agent_engine -in @SYSCONFDIR@/google-cloud-ops-agent/config.yaml
 ExecStart=/bin/true
+RemainAfterExit=yes
 
 [Install]
 WantedBy=multi-user.target


qingling128 (approved):
LGTM

quentinmit:
@qingling128 Do you have any thoughts on the discussion above?

sophieyfang:
Since this PR appears to also fix the problem that https://github.com/GoogleCloudPlatform/ops-agent/pull/129 tends to do. I will verify the case tmr, after confirming I will close https://github.com/GoogleCloudPlatform/ops-agent/pull/129. We will still need to discuss about doing the validation in the right place.

qingling128:
> @qingling128 Do you have any thoughts on the discussion above?

Ah I thought we will GVC for the other notes up there. To summarize my thoughts so far:

+1 for making it `show "active (exited)" as the state of the parent service, instead of "inactive (dead)"`. When a user customizes the config, the first thing they will do is to restart the agent. Naturally they will check the service status. `inactive (dead)` might lead them to believe something is wrong.

My vote for the user expectation for the broader question is ideally:

1. When the user first installs the agent, or explicitly restart the agent, in case there is any user config error, the agent should use all the channels we have to tell users that something is wrong (aka it should scream). e.g.
* Service status shows failure (all 3 should fail if possible), and hopefully some messages are in the status output (aka the tail of journal logs) that directly points out the failure is due to user config validation errors.
* Does not ingest metrics. Does not ingest Logs. Partially having metrics (especially uptime metrics) / logs might lead users to believe everything is fine. I believe for the current implementation, the config validation always validates the full config, so this is already what's gonna happen.
* Any the other ways we could think about.

2. If initially the agent is up and healthy, but later on one of the subagent crashes, this is probably not a user config error (in rare cases, a wrong username to mysql db might only causes failures later in the process). For this case, I'd lean towards not having the error on one subagent's side bring down the other side. This way user has more telemetry to debug the crash when they realize something is wrong.

igorpeshansky (approved):
LGTM :shipit:

I like the idea of config validation in the parent service. It may even be possible to generate configs as part of the parent service binary, but I'm on the fence about this, as that would mean the parent service would now need to enumerate the subagent services, and be cognizant of subagent-specific constants (like run directories, etc).

Regarding having the parent service bring down both subagents on error, I'd rather keep the agents running unless something's globally wrong (e.g., the config is invalid).

sophieyfang:
Just checked that using `RemainAfterExit=yes` doesn't consistently pass the error msgs from subagents. The change it didn't is 1  in 10.
See attached screenshot:
image

I think `RemainAfterExit=yes` just bought us a slight more time waiting(not really waiting) on the subagents. Thus the chance of  seeing the expected errors is increasing. But it's not sealed deal

sophieyfang:
> LGTM :shipit:
> 
> I like the idea of config validation in the parent service. It may even be possible to generate configs as part of the parent service binary, but I'm on the fence about this, as that would mean the parent service would now need to enumerate the subagent services, and be cognizant of subagent-specific constants (like run directories, etc).
> 
> Regarding having the parent service bring down both subagents on error, I'd rather keep the agents running unless something's globally wrong (e.g., the config is invalid).

Having the parent do the validation or even the conversion will have to let the two subagents wait for the parent to do the check or even conversion first. And subagents, e.g. fluentbit needs ` -logs ${LOGS_DIRECTORY} -state ${STATE_DIRECTORY}` and that will need parent share or pass down this info to the subagents, e.g. fluentbit. All those might be possible ( i am not saying it out confidently, not super in systemd).

As of now, the simple way to let the `sudo service google-cloud-ops-agent restart` consistently prints out error msgs and echo $? non-zero, we just need to have the parent wait for the subagents: https://github.com/GoogleCloudPlatform/ops-agent/pull/129.

qingling128:
Sounds like the remaining question is about whether we still need #129, but this PR is good to go in, right?

quentinmit:
Merging this PR as-is will cause it to start restarting the parent service and other subagents if any of the subagents die. If we think that's a net improvement we can merge.

qingling128 (approved):
LGTM

quentinmit:
I switched from Requires to Wants and tested. That means that subagent restarts do not propagate to other services. I manually tested start/stop/restart on Debian 10 and the commands do still correctly propagate to the child services.

PR #111, opened 2021-06-23
Split conf to built-in conf and user conf and merge them.

qingling128:
Implements b/191888008.

PR #133, opened 2021-07-02
Explicitly validate config as part of the main service startup.

igorpeshansky:
This also re-enables config validation on Linux, accidentally removed by #111.

### DIFF
--- a/confgenerator/files.go
+++ b/confgenerator/files.go
@@ -53,6 +53,8 @@ func ReadUnifiedConfigFromFile(path string) (UnifiedConfig, error) {
 func GenerateFilesFromConfig(uc *UnifiedConfig, service, logsDir, stateDir, outDir string) error {
 	hostInfo, _ := host.Info()
 	switch service {
+	case "": // Validate-only.
+		return nil
 	case "fluentbit":
 		mainConfig, parserConfig, err := uc.GenerateFluentBitConfigs(logsDir, stateDir, hostInfo)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


qingling128 (approved):
LGTM

confgenerator/files.go
```diff
@@ -67,6 +72,8 @@ func GenerateFilesFromConfig(uc *UnifiedConfig, service, logsDir, stateDir, outD
 		if err = writeConfigFile([]byte(otelConfig), filepath.Join(outDir, "otel.yaml")); err != nil {
 			return err
 		}
+	case "validate-only":
```
quentinmit, line 75: Nit: This isn't a service. I would use an empty string.

systemd/google-cloud-ops-agent.service
```diff
@@ -18,6 +18,8 @@ Requires=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opente
 
 [Service]
 Type=oneshot
+# Validate the config.
+ExecStartPre=@PREFIX@/libexec/google_cloud_ops_agent_engine -service=validate-only -in @SYSCONFDIR@/google-cloud-ops-agent/config.yaml
```
quentinmit, line 22: With an empty string I believe the `-service` argument can go away entirely since empty string is the default value.

igorpeshansky: Yes, my initial implementation did exactly that, but then we'd still need an explicit case for handling it in `GenerateFilesFromConfig`, and I thought it'd be confusing. But, given that you're thinking along the same lines, I can restore that.

quentinmit (approved):

qingling128 (approved):
LGTM

PR #137, opened 2021-07-26
Use libexec instead of opt as base path

hopkiw:
This PR updates file locations of binary components from /opt/ to /usr/libexec/
This is in keeping with FHS and distribution file path requirements. Tested working with EL SElinux rules.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -16,11 +16,11 @@
 
 # Package build scripts can use this by setting $DESTDIR before launching the
 # script.
-prefix=/opt/google-cloud-ops-agent
+prefix=/usr/libexec/google-cloud-ops-agent
 sysconfdir=/etc
 systemdsystemunitdir=$(pkg-config systemd --variable=systemdsystemunitdir)
 systemdsystempresetdir=$(pkg-config systemd --variable=systemdsystempresetdir)
-subagentdir=$prefix/subagents
+# subagentdir=$prefix/subagents
 
 . VERSION
 if [ -z "$BUILD_DISTRO" ]; then
@@ -28,7 +28,7 @@ if [ -z "$BUILD_DISTRO" ]; then
   BUILD_DISTRO=$(lsb_release -is | tr A-Z a-z)${release_version%%.}
 fi
 if [ -z "$CODE_VERSION" ]; then
-  CODE_VERSION=${PKG_VERSION}
+  CODE_VERSION="$PKG_VERSION"
 fi
 BUILD_INFO_IMPORT_PATH="github.com/GoogleCloudPlatform/ops-agent/internal/version"
 BUILD_X1="-X ${BUILD_INFO_IMPORT_PATH}.BuildDistro=${BUILD_DISTRO}"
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -16,83 +16,82 @@
...
+# subagentdir=$prefix/subagents
 
 . VERSION
 if [ -z "$BUILD_DISTRO" ]; then
   release_version="$(lsb_release -rs)" #e.g. 9.13 for debian, 8.3.2011 for centos
   BUILD_DISTRO=$(lsb_release -is | tr A-Z a-z)${release_version%%.}
 fi
 if [ -z "$CODE_VERSION" ]; then
-  CODE_VERSION=${PKG_VERSION}
+  CODE_VERSION="$PKG_VERSION"
```
igorpeshansky, line 31: We probably want to keep the curly braces as well…

build.sh
```diff
@@ -16,83 +16,82 @@
 
 # Package build scripts can use this by setting $DESTDIR before launching the
 # script.
-prefix=/opt/google-cloud-ops-agent
+prefix=/usr/libexec/google-cloud-ops-agent
```
igorpeshansky, line 19: I don't think we want to do that. The prefix controls the locations of all files, including both the binaries and the non-binary data. It's better to set a separate `bindir` and use that for binaries only.

igorpeshansky (changes requested):
I'd like @quentinmit to also take a look at this…

build.sh
hopkiw: this PR adopts google bash style guide, which says no curly braces if the variable is by itself, always curly braces otherwise.

build.sh
hopkiw: i don't know what non-binary data you mean, but this only affects binaries. here is the contents of the el8 RPM produced by this PR:
```
/etc/google-cloud-ops-agent/config.yaml
/usr/lib/.build-id
/usr/lib/.build-id/82
/usr/lib/.build-id/82/acab8ee480462b5e2460f45fbb892ada1bef3e
/usr/lib/systemd/system-preset/90-google-cloud-ops-agent.preset
/usr/lib/systemd/system/google-cloud-ops-agent-fluent-bit.service
/usr/lib/systemd/system/google-cloud-ops-agent-fluent-bit.service.d
/usr/lib/systemd/system/google-cloud-ops-agent-fluent-bit.service.d/directories.conf
/usr/lib/systemd/system/google-cloud-ops-agent-opentelemetry-collector.service
/usr/lib/systemd/system/google-cloud-ops-agent-opentelemetry-collector.service.d
/usr/lib/systemd/system/google-cloud-ops-agent-opentelemetry-collector.service.d/directories.conf
/usr/lib/systemd/system/google-cloud-ops-agent.service
/usr/libexec/google-cloud-ops-agent/fluent-bit
/usr/libexec/google-cloud-ops-agent/google_cloud_ops_agent_engine
/usr/libexec/google-cloud-ops-agent/otelopscol
```

qingling128:
Closing this one for now, as the `burn ops agent to image` effort was aborted.

PR #138, opened 2021-07-28
Simplify special-casing of the default config in tests.

igorpeshansky:
Create golden symlinks to built-in-config-*.yaml, so they can be handled by -update_golden.
Add symlinks to the default config in the all-built_in_config testdata directories, to remove the special input handling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+git:
+  symlinks: true
+
 os:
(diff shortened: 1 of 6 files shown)


qingling128 (approved):
LGTM

qingling128:
Related to b/193544238

PR #139, opened 2021-08-03
Keep config.yaml file if existing

sophieyfang:
## Test when there is no config file

Screen Shot 2021-08-03 at 4 43 48 PM

## Test when there is config file

Screen Shot 2021-08-03 at 4 45 01 PM

## ToDOO

- work with jimmy on documenting the caveat for previous versions as a known issue SGTM. e.g. a note in https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/installation#upgrade and https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/transition#migrate_preview

- later after we have at least two versions of ops-agent that contains this change, we can add a testcase in ops-agent-test.go for checking if existing config file  is kept untouched.


sophieyfang:
Sync'ed with @qingling128. We need to make consistency among Linux and Windows. Thus default-config.yaml isn't necessary to have in this case. I will remove it in next commit

qingling128 (approved):
LGTM

igorpeshansky (commented):
Had some unsent comments on older commits… If you find you want to address them, we can do that in a follow-up PR.

PR #140, opened 2021-08-04
Fix bug where the generated Fluent Bit filters were incorrectly ordered.

davidbtucker:
The new test config is a good example.  It has:

    pipeline2:
      processors: [beta, alpha]
    pipeline1:
      processors: [alpha, beta]

The code previously created a flat list of 4 Fluent Bit filters:

    [pipeline2-beta, pipeline2-alpha, pipeline1-alpha, pipeline1-beta]

which was then sorted, putting the pipeline2 filters out-of-order:

    [pipeline1-alpha, pipeline1-beta, pipeline2-alpha, pipeline2-beta]

This PR keeps the pipelines in groups:

    [[pipeline2-beta, pipeline2-alpha], [pipeline1-alpha, pipeline1-beta]]

and then does not sort within each group:

    [[pipeline1-alpha, pipeline1-beta], [pipeline2-beta, pipeline2-alpha]]

thus maintaining the correct order of filters in the generated Fluent
Bit config.

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -218,7 +218,7 @@ func (uc *UnifiedConfig) GenerateFluentBitConfigs(logsDir string, stateDir strin
 	fbStackdrivers := defaultStackdriverOutputs(hostInfo)
 	fbSyslogs := []*conf.Syslog{}
 	fbWinEventlogs := []*conf.WindowsEventlog{}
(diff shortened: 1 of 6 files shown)


sophieyfang (approved):
Generally looks good to me.

As igor and quentin are working on refactoring. I feel like it will be on their path. Although it doesn't seem like a big conflict as the change isn't big.

PR #141, opened 2021-08-04
WIP: configuration cleanup.

igorpeshansky:
We can pick-and-choose individual commits from this draft PR into separate PRs, as needed. I've tried to keep the commits small, and it may make sense to review them one at a time.

PR #143, opened 2021-08-06
Cleanup configuration representation.

igorpeshansky:
Extracted from #141 and rebased on top of the merged #142. Can split up further if needed.

PR #145, opened 2021-08-06
Clean up configuration generation for otel configs.

igorpeshansky:
Extracted from #141.


quentinmit:
I rebased this on `master` now that #143 is merged. Still quite a bit of cleanup I'd like to do, but I guess we could split that into a separate PR if you'd like to review this so far.

quentinmit:
I finished the work from Friday to make otel config generation completely modular. The diff is now `12 files changed, 874 insertions(+), 1820 deletions(-)`, so a net decrease of almost 1,000 LOC.

Please review, @davidbtucker

confgenerator/otel/modular.go
```diff
@@ -0,0 +1,118 @@
...
+		receiverName := pipeline.Receiver.name(prefix)
+		receivers[receiverName] = pipeline.Receiver.Config
+		var processorNames []string
+		for i, processor := range pipeline.Processors {
+			name := processor.name(fmt.Sprintf("%s_%d", prefix, i))
+			processorNames = append(processorNames, name)
+			processors[name] = processor.Config
+		}
+		processorNames = append(processorNames, globalProcessorNames...)
+		pipelines["metrics/"+prefix] = map[string]interface{}{
```
davidbtucker, line 105: Document why we prefix with `metrics/` here.

confgenerator/otel/modular.go
```diff
@@ -0,0 +1,118 @@
...
+	Exporter         Component
+}
+
+func (c ModularConfig) Generate() (string, error) {
+	receivers := map[string]interface{}{}
+	processors := map[string]interface{}{}
+	exporters := map[string]interface{}{}
+	pipelines := map[string]interface{}{}
+
+	configMap := map[string]interface{}{
```
davidbtucker, line 77: A sample output would make this function more readable.  Something like:

    // metrics/mypipe:
    //   receivers: [hostmetrics/mypipe]
    //   processors: [filter/mypipe_1, metrics_filter/mypipe_2, resourcedetection/_global_0]
    //   exporters: [googlecloud]

confgenerator/confgenerator.go
```diff
@@ -37,262 +36,70 @@ func defaultFilepathJoin(_ string, elem ...string) string {
...
-	stackdriverList := []*otel.Stackdriver{}
-	exportNameMap := make(map[string]string)
-	for _, pID := range sortedKeys(pipelines) {
-		p := pipelines[pID]
-		for _, eID := range p.ExporterIDs {
-			exporter, ok := exporters[eID]
+			receiver, ok := m.Receivers[rID]
 			if !ok {
-				continue
+				return nil, fmt.Errorf("receiver %q not found", rID)
```
davidbtucker, line 84: It would be nice if we could report this error using that line number + column format.

davidbtucker (commented):
First round of comments.

confgenerator/otel/modular.go
```diff
@@ -0,0 +1,118 @@
...
+
+func configToYaml(config interface{}) ([]byte, error) {
+	outMap := make(map[string]interface{})
+	if err := mapstructure.Decode(config, &outMap); err != nil {
+		return nil, err
+	}
+	return yaml.Marshal(outMap)
+}
+
+type ModularReceiver interface {
```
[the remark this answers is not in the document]
quentinmit: Actually this interface is now unused, so I just deleted it. :)

confgenerator/otel/modular.go
```diff
@@ -0,0 +1,118 @@
...
+	Pipelines() []Pipeline
+}
+
+type ModularProcessor interface {
```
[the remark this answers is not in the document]
quentinmit: Ditto.

confgenerator/metric_processors.go
```diff
@@ -14,16 +14,182 @@
...
+		metricNames = append(metricNames, fmt.Sprintf(`^%s$`, strings.Join(literals, `.*`)))
+	}
+	return []otel.Component{metricsFilter(
+		"exclude",
+		"regexp",
+		metricNames...,
+	)}
+}
+
+// metricsFilter returns an otel.Component that filters metrics.
```
qingling128, line 54: Can we split the util stuff below to a separate file? Potentially into the `otel` package?

quentinmit: Done, though I did intentionally put it here until the interface was stable and I felt comfortable with exporting it. Oh well, we're still probably going to be the only user of the code.

confgenerator/otel/modular.go
```diff
@@ -0,0 +1,118 @@
...
+// limitations under the License.
+
+// Package otel provides data structures to represent and generate otel configuration.
+package otel
+
+import (
+	"fmt"
+
+	yaml "github.com/goccy/go-yaml"
+	"github.com/mitchellh/mapstructure"
```
qingling128, line 22: Does Otel also use `mapstructure`?

Another dumb question: What are the reasons we need `otel/modular.go` instead of using something Otel has? I guess

1. Otel does not already have something like this module?
2. Otel has something like this module, but it's not public?
2. We want to enforce rules that are specific to us (e.g. only one exporter. Always prefix with `metrics/` for pipelines.)?

qingling128: Got it

quentinmit:
I just pushed a couple commits to make the golden file diffs smaller. Here's a command you can use to diff them:
```
pip install yq
sudo apt install jq
for i in testdata/valid/*/*/golden_otel.conf; do echo "--- $i"; diff -U500 <(git cat-file blob origin/master:./$i | ~/.local/bin/yq -y -S) <(git cat-file blob HEAD:./$i | ~/.local/bin/yq -y -S); done
```

confgenerator/config.go
```diff
@@ -356,23 +357,31 @@ type LoggingPipeline struct {
...
 type metricsReceiverMap map[string]MetricsReceiver
 type metricsProcessorMap map[string]MetricsProcessor
-type metricsExporterMap map[string]MetricsExporter
 type Metrics struct {
 	Receivers  metricsReceiverMap  `yaml:"receivers" validate:"dive,keys,startsnotwith=lib:"`
 	Processors metricsProcessorMap `yaml:"processors" validate:"dive,keys,startsnotwith=lib:"`
 	// Exporters are deprecated and ignored, so do not have any validation.
-	Exporters metricsExporterMap `yaml:"exporters,omitempty"`
-	Service   *MetricsService    `yaml:"service"`
+	Exporters map[string]interface{} `yaml:"exporters,omitempty"`
```
igorpeshansky, line 364: We will now allow arbitrary YAML in the `exporters` field, and it's completely unvalidated. Is this where we want to be with respect to customer configs? Or do we want to restrict what they put in there before we ignore it?

quentinmit: We already dropped validation on this field, both partially before and now even more. It's a lot of work to maintain validation on a field that's already ignored and deprecated. Anyone who is specifying this field has been doing it since pre-2.0, so they already had a chance to get the field validated. I don't think it's worth the effort to maintain unused validation.

igorpeshansky (commented):
LGTM :shipit:

confgenerator/otel/modular.go
```diff
@@ -44,7 +44,7 @@ func (c Component) name(suffix string) string {
 	return c.Type
 }
 
-// configToYaml converts a tree of structss into a YAML file.
+// configToYaml converts a tree of structs into a YAML file.
```
qingling128, line 47: LOL

qingling128 (approved):
LGTM. Integration tests passed: http://b/195762251#comment4

confgenerator/confgenerator.go
```diff
@@ -37,262 +36,70 @@ func defaultFilepathJoin(_ string, elem ...string) string {
...
-			case "google_cloud_monitoring":
-				if _, ok := exportNameMap[eID]; !ok {
-					stackdriver := otel.Stackdriver{
-						StackdriverID: eID,
-						Prefix:        "agent.googleapis.com/",
-					}
-					stackdriverList = append(stackdriverList, &stackdriver)
-					exportNameMap[eID] = "googlecloud/" + eID
+			for i, receiverPipeline := range receiver.Pipelines() {
+				prefix := fmt.Sprintf("%s_%s", strings.ReplaceAll(pID, "_", "__"), strings.ReplaceAll(rID, "_", "__"))
```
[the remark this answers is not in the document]
quentinmit: Replacing `_` with `__` prevents collisions between different pipelines. If we went the other direction, there would be no way to distinguish between (pipeline `default__pipeline` receiver `hostmetrics`) and (pipeline `default` receiver `pipeline__hostmetrics`).

We could certainly introduce smarter anti-collision logic though (e.g. default to the bare pipeline name and add suffixes only when there would be a collision). But I'd like to save that for later PRs.

confgenerator/otel/modular.go
quentinmit: My MBP keyboard introduces spurious `s` keypresses all over the place :(

PR #146, opened 2021-08-06
Clean up configuration generation for fluentbit configs.

igorpeshansky:
Extracted from #141.

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -106,16 +106,35 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
 	return out, nil
 }
 
+type sortKey struct {
+	n   int
+	tag string
+}
+
+func sortKeyLess(a, b sortKey) bool {
+	return a.n < b.n || (a.n == b.n && a.tag < b.tag)
+}
+
+func inputSortKey(i fluentbit.Input) sortKey {
+	switch r := i.(type) {
+	case *fluentbit.Tail:
+		return sortKey{n: 1, tag: r.Tag}
+	case *fluentbit.Syslog:
+		return sortKey{n: 2, tag: r.Tag}
+	case *fluentbit.WindowsEventlog:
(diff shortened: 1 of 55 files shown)


igorpeshansky (commented):
This is now ready for review. @quentinmit, the last commit will have minor conflicts with your branch. It does introduce useful functionality, IMO, but we can discuss reverting it as well.

confgenerator/confgenerator.go
```diff
@@ -102,20 +103,58 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
...
+	tag string
+}
+
+func sortKeyLess(a, b sortKey) bool {
+	return a.n < b.n || (a.n == b.n && a.tag < b.tag)
+}
+
+func inputSortKey(i fluentbit.Input) sortKey {
+	switch r := i.(type) {
+	case *fluentbit.Tail:
```
qingling128, line 117: Curious: Do we need the `1`, `2`, `3` based ordering here because otherwise the result is non-deterministic or because we have to enforce the exact orders of these configs? For some reason, we can not just rely on `tail`, `syslog` and `windows_event_log` to be sorted alphabetically?

Can we add a comment for the rationale?

qingling128: I'm leaning towards the receivers / processors being sorted alphabetically in the long run, so that we do not need to manually maintain this sorting order.

qingling128: Sync'd offline, the sorting here is just to minimize the diff in the golden files. The actual sorting will not be depending on the types and will be handled in a follow up PR.

qingling128 (approved):
LGTM

confgenerator/confgenerator.go
quentinmit: I believe the final sort order will be that pipelines will be generated alphabetically and then within each pipeline the filters will be generated in the user-specified order. I don't particularly care about this sorting code because it's just going to be removed soon anyway.

quentinmit (approved):

PR #151, opened 2021-08-18
Remove the assumption of the agent.googleapis.com domain

quentinmit:


confgenerator/confgenerator.go
```diff
@@ -64,7 +64,9 @@ func (uc *UnifiedConfig) GenerateOtelConfig(hostInfo *host.InfoStat) (string, er
 			Config: map[string]interface{}{
 				"user_agent": userAgent,
 				"metric": map[string]interface{}{
-					"prefix": "agent.googleapis.com/",
+					// Receivers are responsible for sending fully-qualified metric names.
+					// NB: If a receiver fails to send a full URL, OT will add the prefix `custom.googleapis.com/opencensus/`."
```
igorpeshansky, line 68: Is there a stray `"` at the end of this comment?

confgenerator/mssql.go
```diff
@@ -65,6 +65,7 @@ func (m MetricsReceiverMssql) Pipelines() []otel.Pipeline {
 					`\SQLServer:Databases(_Total)\Write Transactions/sec`,
 					"mssql/write_transaction_rate",
 				),
+				otel.AddPrefix("agent.googleapis.com"),
```
jsuereth, line 68: This is WAY easier to review!  Great work on the refactor.

confgenerator/confgenerator.go
```diff
@@ -64,7 +64,9 @@ func (uc *UnifiedConfig) GenerateOtelConfig(hostInfo *host.InfoStat) (string, er
...
+					"prefix": "",
```
igorpeshansky, line 69: Just curious, would omitting the prefix setting have equivalent behavior? Specifying it explicitly makes sense anyway, if only as a place to hang a comment off of…

confgenerator/confgenerator.go
```diff
@@ -64,7 +64,9 @@ func (uc *UnifiedConfig) GenerateOtelConfig(hostInfo *host.InfoStat) (string, er
 			Config: map[string]interface{}{
 				"user_agent": userAgent,
 				"metric": map[string]interface{}{
-					"prefix": "agent.googleapis.com/",
+					// Receivers are responsible for sending fully-qualified metric names.
```
jsuereth, line 67: Can you open a ticket (P3) about writing a test to verify `custom.googleapis.com` metrics aren't slipping through?

jsuereth (approved):

confgenerator/confgenerator.go
qingling128, line 69: Does it make sense to default to `agent.googleapis.com/` if not set? That way it errors out and developer who does not intend to send metrics to `agent.googleapis.com/` could easily tell something is not set up correctly.

confgenerator/otel/processors.go
```diff
@@ -47,6 +48,18 @@ func MetricsTransform(metrics ...map[string]interface{}) Component {
...
 
+// AddPrefix returns a config snippet that adds a prefix to all metrics.
+func AddPrefix(prefix string) map[string]interface{} {
+	// $ needs to be escaped because reasons.
+	// https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricstransformprocessor#rename-multiple-metrics-using-substitution
+	return map[string]interface{}{
+		"include":    `^(.*)$$`,
+		"match_type": "regexp",
+		"action":     "update",
+		"new_name":   path.Join(prefix, `$${1}`),
```
igorpeshansky, line 59: Erm, wouldn't this use the `\\` separator when run on Windows? Why not just use ``prefix+`/$${1}` ``? Or, better yet, just set `prefix` to `"agent.googleapis.com/"` (with the trailing slash), to match the semantics of the exporter…

confgenerator/confgenerator.go
qingling128: Sync'd offline. OT does not give a way to customize the default. If we override the default then we lose the behavior of only adding the prefix when the metric name isn't a URL. So NVM

qingling128 (approved):
LGTM

igorpeshansky (changes requested):

confgenerator/agentmetrics.go
```diff
@@ -72,6 +72,7 @@ func (r MetricsReceiverAgent) Pipeline() otel.Pipeline {
 					// change data type from double -> int64
 					otel.ToggleScalarDataType,
 				),
+				otel.AddPrefix("agent.googleapis.com"),
```
igorpeshansky, line 75: Add a TODO to fold this back into the `RenameMetric` calls? Also in other files…

igorpeshansky (changes requested):
Forgot one.

confgenerator/confgenerator.go
quentinmit: Done.

confgenerator/confgenerator.go
quentinmit: Yeah, we don't have that choice right now. Sadly.

confgenerator/confgenerator.go
quentinmit: Indeed...

confgenerator/agentmetrics.go
jsuereth: Don't we plan to use Rename for non-agent metrics?

confgenerator/confgenerator.go
quentinmit: I think it would be equivalent but I decided to go for less magic.

confgenerator/agentmetrics.go
quentinmit: I'm not 100% convinced that's the right choice. It might be, but not sure.

confgenerator/confgenerator.go
jsuereth: We are doing a rewrite of this, so we should capture this use case down in a bug/FR

confgenerator/otel/processors.go
jsuereth: These are URLs, does windows really use \\ for urls?

jsuereth (approved):

confgenerator/otel/processors.go
quentinmit: No, `path` always uses forward slash. It's specifically designed for the URL usecase. (Also it's what the OT code itself uses.)

`filepath` is the package that would use `\` on Windows.

You are incorrect about the semantics of the exporter; the trailing slash is not necessary because the exporter also uses `path.Join`.

PR #153, opened 2021-08-19
Add nginxreceiver and normalizesumsprocessor

djaglowski:
Additions:
- A new receiver of type `nginx`, which is supported on both windows and linux
- A new validation tag, `url`, which performs some minimal validation on nginx's `endpoint` field
- A new processor type, `normalizesums`, which is useful for any receiver that submits counter metrics to google cloud monitoring (include `nginx`)
- Several golden config test cases for nginx, both valid and invalid


.gitmodules
```diff
@@ -6,4 +6,4 @@
 	url = ../../fluent/fluent-bit.git
 [submodule "submodules/opentelemetry-operations-collector"]
 	path = submodules/opentelemetry-operations-collector
-	url = ../../GoogleCloudPlatform/opentelemetry-operations-collector.git
+	url = git@github.com:observIQ/opentelemetry-operations-collector.git
```
djaglowski, line 9: This is temporary, and will be reverted when the underlying PR is merged.

confgenerator/config.go
```diff
@@ -127,6 +128,17 @@ func newValidator() *validator.Validate {
 		}
 		return t >= tmin
 	})
+	// url validates that the value is a valid url
+	v.RegisterValidation("url", func(fl validator.FieldLevel) bool {
```
djaglowski, line 132: We could consider having `http_url`, `jmx_url`, etc, so that some more specific validation could be done.

.gitmodules
quentinmit: You can use an `insteadOf` in your Git config to override this without changing the `.gitmodules` file.

jsuereth: underlying PR is merged :)

jsuereth:
QQ: Is this PR meant to have a documentation markdown file on how to set up and configure the receiver, or is that a different PR?

qingling128:
The code itself LGTM except we could revert the temporary `.gitmodules` change now.

Besides the documentation need Josh mentioned, we'd also like to get a successful integration test run. This requires some collaboration with @martijnvans . The initial integration test PR is at https://github.com/GoogleCloudPlatform/ops-agent/pull/154.

martijnvans:
I am sending https://github.com/GoogleCloudPlatform/ops-agent/pull/157 with  clarified instructions for how to test a new application.

.gitmodules
djaglowski: This change has been reverted, and the submodule has been updated accordingly.

djaglowski:
> QQ: Is this PR meant to have a documentation markdown file on how to set up and configure the receiver, or is that a different PR?

I will add documentation before marking this PR ready for review.

djaglowski:
> The code itself LGTM except we could revert the temporary `.gitmodules` change now.
> 
> Besides the documentation need Josh mentioned, we'd also like to get a successful integration test run. This requires some collaboration with @martijnvans . The initial integration test PR is at #154.

> I am sending #157 with clarified instructions for how to test a new application.

I'm following these threads, and will add an integration test when they are merged.

confgenerator/nginx.go
```diff
@@ -0,0 +1,51 @@
...
+package confgenerator
+
+import "github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
+
+type MetricsReceiverNginx struct {
+	ConfigComponent `yaml:",inline"`
+
+	MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"required,url"`
```
quentinmit, line 24: @jsuereth As I mentioned to you I don't love the name `endpoint` because it doesn't help the user figure out what it should be set to. (e.g. something like `status_url` would be clearer)

@djaglowski Is there not a sane default for this? `http://localhost/status` perhaps?

jsuereth: Regarding the name endpoint:
1. `endpoint` is what was chosen by OTel prior to this project.  
2. [Prometheus exporter](https://github.com/nginxinc/nginx-prometheus-exporter#running-the-exporter-binary) uses a vey non-ops-agent, non-otel looking endpoint.   They do recommend the user use a default exposed path of "stub_status".
3. `endpoint` is canonical in most receivers in otel.

If you think Ops Agent should deviate, that's worth a discussion.  I do not think we should look significantly different from otel without a good reason.



Regarding you question on sane default, the answer is not really.  users *have* to configure the plugin.  The [default documentation](https://nginx.org/en/docs/http/ngx_http_stub_status_module.html#stub_status) recommends (`/status`) which likely conflicts with a lot of hosted web applications.  

We have two options:

1. We can default to `/status` because this is what the default set up documentation suggests.
2. We can default to `/sub_status` because this is the default recommended by prometheus.
3. We can not have a default since a user *has* to configure this URL.

I'd vote for option #1 (preffered) with #2 as my second.

confgenerator/config.go
djaglowski: This is my first encounter with the `validator` library, so I just wasn't aware that it contained built in validators. It looks to do just what we want, so I'll switch to that.

confgenerator/config.go
```diff
@@ -127,6 +128,17 @@ func newValidator() *validator.Validate {
...
+		u, err := url.Parse(fl.Field().String())
+		if err != nil {
+			return false
+		}
+		if u.Port() == "" {
```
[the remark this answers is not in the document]
djaglowski: Switching to the builtin validator, but point taken.

confgenerator/nginx.go
djaglowski: I agree we should stick with `endpoint`. This will be the closest thing to a "universal" field across 100+ receivers, and I think that providing users with a familiar anchor point for each config is more useful than hinting at detailed meaning. 

Also would second that #1 is the best default because it aligns with the set up docs.

quentinmit: I don't see the value in a "universal" field - especially one that might contain an HTTP URL for one receiver, a Unix socket path for another, and a database connection string for a third. Unless you're saying `endpoint` always means a URL? I'm happy to punt this discussion to after we have the first working demo.

jsuereth: endpoint always means a URL in opentelemetry collector.  I think the discussion on URL vs. Endpoint should be amongst @qingling128 and team, outside the scope of this PR.   Can you schedule/make sure that happens Quentin?

Also, found another possible default for status, the current cloud docs recommend [this config](https://raw.githubusercontent.com/Stackdriver/stackdriver-agent-service-configs/master/etc/nginx/conf.d/status.conf) for nginx, which means we probably should default to `/nginx_status` to ease migration.   We do *not* have a discovery mechanism to *look* for 443/80 port openings.

Given all that, (1) or (2) as default seems fine for now. I think we can (in a non-breaking way) improve the default overtime as long as we don't regress the behavior.

djaglowski: @quentinmit, can you double-check / correct me on this?

---

I've removed the `required` tag and added the default as follows:
```go
type MetricsReceiverNginx struct {
	ConfigComponent `yaml:",inline"`
	MetricsReceiverShared `yaml:",inline"`
	Endpoint string `yaml:"endpoint" validate:"url=http://localhost/status"`
}
```

This should invalidate several of the golden error cases that I added (they should not generate any error), but I'm getting the following error instead: `the agent config file is not valid. detailed error: [19:16] Key: 'MetricsReceiverNginx.endpoint' Error:Field validation for 'endpoint' failed on the 'url' tag`

I believe the validator is disregarding the absence of the `required` tag, and enforcing that `""` must be a valid `url`, which of course it is not.

---

The same behavior can be replicated on `master` with `collection_interval`, by removing the `required` tag from `MetricsReceiverShared.CollectionInterval`, and correspondingly the `collection_interval` from a test config, such as `testdata/valid/*/metrics-receiver_custom_collection_interval/input.yaml`:
```
> go test -mod=mod github.com/GoogleCloudPlatform/ops-agent/confgenerator
--- FAIL: TestGenerateConfsWithValidInput (0.00s)
    --- FAIL: TestGenerateConfsWithValidInput/linux (0.00s)
        --- FAIL: TestGenerateConfsWithValidInput/linux/metrics-receiver_custom_collection_interval (0.00s)
            confgenerator_test.go:134: MergeConfFiles("testdata/valid/linux/metrics-receiver_custom_collection_interval/built-in-config.yaml", "testdata/valid/linux/metrics-receiver_custom_collection_interval/input.yaml", "testdata/valid/linux/metrics-receiver_custom_collection_interval/merged-config.yaml") got: the agent config file is not valid. detailed error: Key: 'MetricsReceiverShared.collection_interval' Error:Field validation for 'collection_interval' failed on the 'duration' tag
FAIL
FAIL    github.com/GoogleCloudPlatform/ops-agent/confgenerator  0.216s
FAIL
```

---

If I'm correct in my assessment here, then we either need to fix this quickly, or accept that we'll not have a default in this first release.

quentinmit: I believe you're looking for the `omitempty` tag:

```
	Endpoint string `yaml:"endpoint" validate:"omitempty,url"`
```

and then you should replace `""` with the default value at config generation time. I'm not sure if there's a way to include the default value itself directly in the struct tags, but I'm pretty sure it's not with `url=http://foo`.

See https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Omit_Empty

djaglowski: Thanks Quentin. Now using `omitempty` in place of `required`, rather than just removing `required`. Default is set if necessary in the `Pipelines()` method.

djaglowski:
As discussed today, there will be two follow up PRs, one for docs, and the other for the integration test.

qingling128 (commented):
The code change in this PR LGTM. If we merge it to `master` branch NOW, it could be released any day (we do weekly releases plus on-demand ones). Since the integration test is punted to a separate PR, have we done any manual testing for this PR?

Alternatively, we can LGTM and push this one to a feature branch (e.g. `nginx`), and push the following doc and integration test PRs to that branch as well, and then merge them together to `master` when they are ready to be released.

qingling128 (approved):
LGTM

qingling128:
Per discussion offline:
- We are merging this PR to an `nginx` branch and push follow up doc and integration PRs there first. Once the launch is approved internally, we will get this `nginx` branch merged to the `master` branch. This way other bug fixes and features on the `master` branch can be released any time.

qingling128:
@jsuereth @quentinmit - Any additional comments?

jsuereth (approved):

confgenerator/nginx.go
```diff
@@ -0,0 +1,56 @@
...
+			Type: "nginx",
+			Config: map[string]interface{}{
+				"collection_interval": r.CollectionIntervalString(),
+				"endpoint":            r.Endpoint,
+			},
+		},
+		Processors: []otel.Component{
+			otel.NormalizeSums(),
+			otel.MetricsTransform(
+				otel.AddPrefix("custom.googleapis.com"),
```
jsuereth, line 48: I think this is supposed to be `workload.googleapis.com`

jsuereth (changes requested):
Sorry I missed this earlier.

confgenerator/nginx.go
qingling128: Nice catch!

quentinmit (approved):

confgenerator/nginx.go
djaglowski: I've been told that this namespace will eventually be available outside of google, and I've periodically attempted to write to it, but it still appears to be restricted:
```
2021-08-23T19:27:41.949-0400    info    exporterhelper/queued_retry.go:316      Exporting failed. Will retry the request after interval.        {"kind": "exporter", "name": "googlecloud", "error": "rpc error: code = PermissionDenied desc = User dan.jaglowski@bluemedora.com does not have permission to write to metric workloads.googleapis.com/nginx.requests.", "interval": "10.402171842s"}
```

qingling128: Should it be `workloads.googleapis.com` instead of `workloads.googleapis.com`?

jsuereth: @djaglowski Did you try just "workload.googleapis.com"  It's not plural.

djaglowski: For some reason I was sure I'd been told `workloads`, but checking `workload`, it works. Pushed the update.

jsuereth (approved):

confgenerator/nginx.go
qingling128: Summarizing the discussion offline.

**TL;DR** - We'll change the `endpoint` parameter name to `stub_status_url`.

**Longer story**
Nginx has various different plugins that users can use to expose metrics (today we support the `stub_status_module` only). e.g.
* http://nginx.org/en/docs/http/ngx_http_stub_status_module.html
* http://nginx.org/en/docs/http/ngx_http_api_module.html
* http://nginx.org/en/docs/http/ngx_http_status_module.html (commercial only)
* https://github.com/nginxinc/nginx-prometheus-exporter
* https://github.com/hnlq715/nginx-vts-exporter
* https://github.com/knyar/nginx-lua-prometheus

To ensure the config model is forward compatible, the options in the future in case we have to support some different flavors are:

Option 1 - Different Nginx related receiver for each of the plugins/modules above
Option 2 - Same Nginx receiver with different parameters that indicates the usage of different plugins/modules

We're leaning towards 2, while it should always be possible to do 1 if absolutely necessary. In order to reduce potential conflicts of what `endpoint` means for different Nginx modules, we'd like to make the parameter name specific to the metrics plugins/modules for forward compatibility.

qingling128: Follow up PR to address this is https://github.com/GoogleCloudPlatform/ops-agent/pull/166.

PR #159, opened 2021-08-23
Add nginx docs

djaglowski:
This is superset of #153, intended specifically to add documentation for the nginx receiver.


docs/nginx.md
```diff
@@ -0,0 +1,42 @@
+# Nginx Receiver
+
+The `nginx` receiver can fetch stats from a NGINX instance using the mod_status endpoint.
+
+
+## Prerequisites
+
+[ngx_http_stub_status_module](http://nginx.org/en/docs/http/ngx_http_stub_status_module.html) must be enabled on the NGINX instance. This may require editing the NGINX configuration.
```
qingling128, line 8: Can we expand a bit on the details for what exactly users need to do. Some reference from the legacy monitoring agent doc: https://cloud.google.com/monitoring/agent/plugins/nginx#prerequisites

docs/nginx.md
```diff
@@ -0,0 +1,42 @@
...
+| `type`                | required                  | Must be `nginx`. |
+| `endpoint`            | `http://localhost/status` | The url exposed by the NGINX stats module. |
+| `collection_interval` | `60s`                     | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
+
+Example Configuration:
+
+```yaml
+metrics:
+  receivers:
+    mynginxmetrics:
```
qingling128, line 24: The convention for the receiver ID is to split by `underscore`, so probably `my_nginx_metrics`.

docs/nginx.md
```diff
@@ -0,0 +1,42 @@
...
+      type: nginx
+      endpoint: http://localhost:80/status
+      collection_interval: 30s
+  service:
+    pipelines:
+      nginxpipeline:
+        receivers:
+          - nginxmetrics
```
qingling128, line 32: The receiver ID (`nginxmetrics`) does not seem to match the ID (`mynginxmetrics`) defined above. This would likely throw a validation error.

docs/nginx.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+## Metrics
+
+| Metric                     | Data Type | Unit        | Labels | Description |
+| ---                        | ---       | ---         | ---    | ---         | 
+| nginx.requests             | sum       | requests    |        | Total number of requests made to the server. |
```
qingling128, line 39: At Ops Agent level, the metrics documentation should be about the Google Monitoring format instead of OTEL, so that users can easily locate it in Metrics Explorer in the UI. Should this be something like `XXX.googleapis.com/nginx...`? @jsuereth might know more details of the expectation.

docs/nginx.md
djaglowski: I've expanded quite a bit, borrowing from the legacy docs.

@jsirianni, would you mind double checking my instructions?

docs/nginx.md
djaglowski: Fixed

docs/nginx.md
djaglowski: Fixed

docs/nginx.md
djaglowski: I've prepended each with `custom.googleapis.com`. I think this makes sense, but depends on resolution to [this comment](https://github.com/GoogleCloudPlatform/ops-agent/pull/153#discussion_r694355052).

djaglowski: These are now documented correctly as `workload.googleapis.com/nginx.foo`.

qingling128 (approved):
LGTM

PR #163, opened 2021-08-23
Preparatory work for Fluentbit marshaling replacement

quentinmit:
This PR refactors a bunch of code and makes the generated configs much closer to what they will be when they are marshaled without templates.

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -165,7 +165,7 @@ func (uc *UnifiedConfig) GenerateFluentBitConfigs(logsDir string, stateDir strin
 		}
 
 		var err error
-		extractedInputs, err := generateFluentBitInputs(logging.Receivers, logging.Service.Pipelines, hostInfo)
+		extractedInputs, err := generateFluentBitInputs(logging.Receivers, logging.Service.Pipelines)
 		if err != nil {
 			return "", "", err
 		}
@@ -283,7 +283,7 @@ func getUserAgent(prefix string, hostInfo *host.InfoStat) (string, error) {
 	return expandTemplate(userAgentTemplate, prefix, extraParams)
 }
 
-func generateFluentBitInputs(receivers map[string]LoggingReceiver, pipelines map[string]*LoggingPipeline, hostInfo *host.InfoStat) ([]fluentbit.Input, error) {
+func generateFluentBitInputs(receivers map[string]LoggingReceiver, pipelines map[string]*LoggingPipeline) ([]fluentbit.Input, error) {
 	inputs := []fluentbit.Input{}
 	for _, pID := range sortedKeys(pipelines) {
 		p := pipelines[pID]
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -215,10 +215,11 @@ func (uc *UnifiedConfig) GenerateFluentBitConfigs(logsDir string, stateDir strin
 	}
 
 	mainConfig, parserConfig, err := fluentbit.Config{
-		Inputs:  inputs,
-		Outputs: outputs,
-		Filters: filters,
-		Parsers: parsers,
+		StateDir: stateDir,
+		Inputs:   inputs,
+		Outputs:  outputs,
+		Filters:  filters,
+		Parsers:  parsers,
 
 		UserAgent: userAgent,
 	}.Generate()
@@ -233,12 +234,10 @@ func defaultTails(logsDir string, stateDir string, hostInfo *host.InfoStat) (tai
 	tails = []fluentbit.Input{}
 	tailFluentbit := fluentbit.Tail{
 		Tag:          "ops-agent-fluent-bit",
-		DB:           filepathJoin(hostInfo.OS, stateDir, "buffers", "ops-agent-fluent-bit"),
 		IncludePaths: []string{filepathJoin(hostInfo.OS, logsDir, "logging-module.log")},
 	}
 	tailCollectd := fluentbit.Tail{
 		Tag:          "ops-agent-collectd",
-		DB:           filepathJoin(hostInfo.OS, stateDir, "buffers", "ops-agent-collectd"),
 		IncludePaths: []string{filepathJoin(hostInfo.OS, logsDir, "metrics-module.log")},
 	}
 	tails = append(tails, &tailFluentbit)
@@ -302,7 +301,6 @@ func generateFluentBitInputs(receivers map[string]LoggingReceiver, pipelines map
 				case *LoggingReceiverFiles:
 					fbTail := fluentbit.Tail{
 						Tag:          fmt.Sprintf("%s.%s", pID, rID),
-						DB:           filepathJoin(hostInfo.OS, stateDir, "buffers", pID+"_"+rID),
 						IncludePaths: r.IncludePaths,
 					}
 					if len(r.ExcludePaths) != 0 {
@@ -321,7 +319,6 @@ func generateFluentBitInputs(receivers map[string]LoggingReceiver, pipelines map
 						Tag:          fmt.Sprintf("%s.%s", pID, rID),
 						Channels:     strings.Join(r.Channels, ","),
 						Interval_Sec: "1",
-						DB:           filepathJoin(hostInfo.OS, stateDir, "buffers", pID+"_"+rID),
 					})
 				}
 			}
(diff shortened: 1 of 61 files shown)


confgenerator/fluentbit/outputs.go
```diff
@@ -0,0 +1,44 @@
...
+
+package fluentbit
+
+import "fmt"
+
+func (o Stackdriver) Component() Component {
+	config := map[string]string{
+		// https://docs.fluentbit.io/manual/pipeline/outputs/stackdriver
+		"Name":              "stackdriver",
+		"Match_Regex":       fmt.Sprintf("^(%s)$", o.Match), // FIXME: properly escape, join from list
```
davidbtucker, line 23: FIXME

confgenerator/fluentbit/inputs.go
```diff
@@ -0,0 +1,110 @@
...
+	return path.Join("${buffers_dir}", dir)
+}
+
+func (i Tail) Component() Component {
+	config := map[string]string{
+		// https://docs.fluentbit.io/manual/pipeline/inputs/tail#config
+		"Name": "tail",
+		"Tag":  i.Tag,
+		// TODO: Escaping?
+		"Path":           strings.Join(i.IncludePaths, ","),
```
davidbtucker, line 38: gofmt

confgenerator/confgenerator.go
```diff
@@ -300,12 +292,11 @@ func generateFluentBitInputs(receivers map[string]LoggingReceiver, pipelines map
 				switch r := r.(type) {
 				case *LoggingReceiverFiles:
 					fbTail := fluentbit.Tail{
-						Tag:  fmt.Sprintf("%s.%s", pID, rID),
-						DB:   filepathJoin(hostInfo.OS, stateDir, "buffers", pID+"_"+rID),
-						Path: strings.Join(r.IncludePaths, ","),
+						Tag:          fmt.Sprintf("%s.%s", pID, rID),
+						IncludePaths: r.IncludePaths,
 					}
 					if len(r.ExcludePaths) != 0 {
```
davidbtucker, line 298: Can remove this if?

confgenerator/fluentbit/conf.go
```diff
@@ -63,113 +56,80 @@ var mainConfTemplate = template.Must(template.New("fluentBitMainConf").Parse(`[S
...
-    Retry_Limit  3
-
-    # https://docs.fluentbit.io/manual/administration/security
-    # Enable TLS support.
-    tls         On
-    # Do not force certificate validation.
-    tls.verify  Off
+    tls               On
+    tls.verify        Off
+    {{- if .Workers}}
```
davidbtucker, line 130: Use `with` here too?

confgenerator/fluentbit/outputs.go
quentinmit: This isn't a regression and I'm planning to fix in the next PR when I refactor this struct and how it's used.

confgenerator/fluentbit/inputs.go
quentinmit: This is already gofmt'd (the value aligns with the next line's value)

confgenerator/confgenerator.go
quentinmit: This is going away in the next PR...

confgenerator/fluentbit/conf.go
quentinmit: As is this...

davidbtucker (approved):

confgenerator/fluentbit/conf.go
```diff
@@ -440,9 +401,8 @@ func (s Syslog) Generate() (string, error) {
 // A WindowsEventlog represents the configuration data for fluentbit's winlog input plugin
 type WindowsEventlog struct {
 	Tag          string
-	Channels     string
-	Interval_Sec string
-	DB           string
+	Channels     string // XXX: proper type
```
igorpeshansky, line 404: Should this be `[]string` as well?

confgenerator/fluentbit/service.go
```diff
@@ -0,0 +1,53 @@
...
+// limitations under the License.
+
+package fluentbit
+
+type Service struct {
+}
+
+func (s Service) Component() Component {
+	return Component{
+		Kind: "SYSTEM",
```
igorpeshansky, line 22: `SERVICE`

confgenerator/confgenerator.go
```diff
@@ -291,7 +283,7 @@ func getUserAgent(prefix string, hostInfo *host.InfoStat) (string, error) {
 	return expandTemplate(userAgentTemplate, prefix, extraParams)
 }
 
-func generateFluentBitInputs(receivers map[string]LoggingReceiver, pipelines map[string]*LoggingPipeline, stateDir string, hostInfo *host.InfoStat) ([]fluentbit.Input, error) {
+func generateFluentBitInputs(receivers map[string]LoggingReceiver, pipelines map[string]*LoggingPipeline, hostInfo *host.InfoStat) ([]fluentbit.Input, error) {
```
igorpeshansky, line 286: `hostInfo` is unused here as well, though I imagine this is also going away.

confgenerator/fluentbit/modular.go
```diff
@@ -0,0 +1,56 @@
...
+func (c Component) generateSection() string {
+	var lines []string
+	var maxLen int
+	for k := range c.Config {
+		if len(k) > maxLen {
+			maxLen = len(k)
+		}
+	}
+	for k, v := range c.Config {
+		lines = append(lines, fmt.Sprintf("    %-*s %s", maxLen, k, v))
```
igorpeshansky, line 40: If you guard this with `v != ""`, you could create keys with empty values, making the code simpler.
Otherwise you might accidentally get a blank key with no value, which would be a syntax error.

confgenerator/fluentbit/inputs.go
```diff
@@ -0,0 +1,110 @@
...
+	dir := strings.ReplaceAll(strings.ReplaceAll(tag, ".", "_"), "/", "_")
+	return path.Join("${buffers_dir}", dir)
+}
+
+func (i Tail) Component() Component {
+	config := map[string]string{
+		// https://docs.fluentbit.io/manual/pipeline/inputs/tail#config
+		"Name": "tail",
+		"Tag":  i.Tag,
+		// TODO: Escaping?
```
igorpeshansky, line 37: [Optional] You could put this `TODO` at the end of the relevant line, to ensure they don't get accidentally separated. Ditto in line 64 below.

confgenerator/fluentbit/inputs.go
```diff
@@ -0,0 +1,110 @@
...
+	"strings"
+)
+
+// TODO: Move structs out of conf.go
+
+// DBPath returns the database path for the given log tag
+func DBPath(tag string) string {
+	// TODO: More sanitization?
+	dir := strings.ReplaceAll(strings.ReplaceAll(tag, ".", "_"), "/", "_")
+	return path.Join("${buffers_dir}", dir)
```
igorpeshansky, line 29: This will have different results when run on Linux and Windows. The whole reason we introduced `filepathJoin` was to be able to run both Linux and Windows tests on either system without worrying about path separator discrepancies. In this case, the full semantics of `path.Join` don't apply anyway — this is just a string concatenation. Since we've determined that `/` works on Windows, can we just replace this with `return "${buffers_dir}/" + dir` instead?

igorpeshansky (changes requested):
A couple of bugs to fix; the rest are suggestions…

confgenerator/fluentbit/inputs.go
igorpeshansky: Never mind, I missed the fact that this was `path.Join` and not `filepath.Join`. Go is confusing.

confgenerator/fluentbit/conf.go
quentinmit: In theory... hence the comment. I wasn't addressing it because this type is going to be merged, but w/e, I improved it.

confgenerator/fluentbit/service.go
quentinmit: Already fixed in 164, but backported here.

confgenerator/confgenerator.go
quentinmit: Yes, but w/e, fixed.

confgenerator/fluentbit/modular.go
quentinmit: I don't like the idea of magically omitting empty values. We might actually need that to work at some point.

confgenerator/fluentbit/inputs.go
quentinmit: I think I'm going to be pushing these into the receiver structs, so I'll deal with that then.

igorpeshansky (approved):
LGTM :shipit:

PR #164, opened 2021-08-24
Replace Fluentbit config generation with modular implementation

quentinmit:


confgenerator/confgenerator.go
```diff
@@ -97,6 +99,144 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
...
+				receiver, ok := l.Receivers[rID]
+				if !ok {
+					return nil, fmt.Errorf("receiver %q not found", rID)
+				}
+				tag := fmt.Sprintf("%s.%s", pID, rID)
+				components := receiver.Components(tag)
+				for i, pID := range p.ProcessorIDs {
+					processor, ok := l.Processors[pID]
+					if !ok {
+						return nil, fmt.Errorf("processor %q not found", pID)
```
igorpeshansky, line 142: Wouldn't this mean one of the default processors (e.g., `lib:apache`)? We still have to generate a filter component, but use the default parser instead…

confgenerator/confgenerator.go
```diff
@@ -97,6 +99,144 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
...
+					}
+					components = append(components, processor.Components(tag, i)...)
+				}
+				components = append(components, setLogNameComponents(tag, rID)...)
+				logNames = append(logNames, regexp.QuoteMeta(rID))
+				sources = append(sources, fbSource{tag, components})
+			}
+		}
+		sort.Slice(sources, func(i, j int) bool { return sources[i].tag < sources[j].tag })
```
igorpeshansky, line 151: [Optional] If you sorted the pipeline keys and the receiver ids, you wouldn't need the additional `tag` field (or the whole `fbSource` struct), right?

confgenerator/confgenerator.go
```diff
@@ -97,6 +99,144 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
...
+		var logNames []string
+		for pID, p := range l.Service.Pipelines {
+			for _, rID := range p.ReceiverIDs {
+				receiver, ok := l.Receivers[rID]
+				if !ok {
+					return nil, fmt.Errorf("receiver %q not found", rID)
+				}
+				tag := fmt.Sprintf("%s.%s", pID, rID)
+				components := receiver.Components(tag)
+				for i, pID := range p.ProcessorIDs {
```
igorpeshansky, line 139: `pID` already represents the pipeline id. I know you don't really need it anymore, but having the same loop variable name in a nested loop is confusing.

confgenerator/confgenerator.go
```diff
@@ -97,6 +99,144 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
...
+					processor, ok := l.Processors[pID]
+					if !ok {
+						return nil, fmt.Errorf("processor %q not found", pID)
+					}
+					components = append(components, processor.Components(tag, i)...)
```
igorpeshansky, line 144: Just curious, why wouldn't you use the processor id in the parser name? It might be easier than counting to track where the parser config came from. Unless the same processor id can appear more than once in a pipeline, but I suspect that would be an invalid config…

confgenerator/fluentbit/modular.go
```diff
@@ -43,14 +43,29 @@ func (c Component) generateSection() string {
...
 	Components []Component
-	LogName    string
-	Internal   bool
 }
 
-type ModularConfig struct {
-	Pipelines      []Pipeline
-	InternalOutput Component
-	Output         Component
+func (c ModularConfig) Generate() (string, string, error) {
```
igorpeshansky, line 51: Would help to name the output parameters…

confgenerator/fluentbit/modular.go
```diff
@@ -43,14 +43,29 @@ func (c Component) generateSection() string {
...
+	parts = append(parts, "")
+
+	var parserParts []string
+
+	for _, c := range c.Components {
+		out := c.generateSection()
+		if c.Kind == "PARSER" {
+			parserParts = append(parserParts, out)
+		} else {
+			parts = append(parts, out)
```
igorpeshansky, line 67: [Optional] To minimize the diffs, it might help to group these by `c.Kind` and emit them in the order `SERVICE`,`INPUT`,`FILTER`,`OUTPUT`.

confgenerator/logging_processors.go
```diff
@@ -23,6 +25,15 @@ func (r LoggingProcessorParseJson) Type() string {
 	return "parse_json"
 }
 
+func (p LoggingProcessorParseJson) Components(tag string, i int) []fluentbit.Component {
```
igorpeshansky, line 28: [Optional] Might be worth renaming the parameter of `Type()` above to `p` as well… Ditto for `LoggingProcessorParseRegex`.

confgenerator/confgenerator.go
```diff
@@ -97,6 +99,144 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
...
+	filter := fluentbit.Component{
+		Kind: "FILTER",
+		Config: map[string]string{
+			"Match":    tag,
+			"Name":     "parser",
+			"Key_Name": p.Field,
+			"Parser":   parserName,
+		},
+	}
+	parser := fluentbit.Component{
```
igorpeshansky, line 225: Might be worth commenting that the parser component returned from here is incomplete.

confgenerator/logging_receivers.go
```diff
@@ -80,6 +82,15 @@ func (r LoggingReceiverWinevtlog) Type() string {
 	return "windows_event_log"
 }
 
+func (r LoggingReceiverWinevtlog) Components(tag string) []fluentbit.Component {
+	return []fluentbit.Component{
+		fluentbit.WindowsEventlog{
```
igorpeshansky, line 87: Just curious. You've replaced a bunch of filters with direct inlined component construction (presumably because they only appear once). This struct also only appears once in the new modular code — was there a particular reason you've kept it? Same question about `Syslog`, actually…

igorpeshansky (changes requested):
Some preliminary comments (only on the commits that are on top of #163).

confgenerator/confgenerator.go
quentinmit: This was still TODO but it's addressed now. There are no default parsers.

confgenerator/confgenerator.go
quentinmit: This setup feels cleaner and I believe is fewer lines of code overall.

confgenerator/confgenerator.go
quentinmit: I believe that's a technically valid config.

confgenerator/fluentbit/modular.go
quentinmit: That's not what named output parameters are for. Go guidance is that you should not name output parameters unless you specifically need them.

confgenerator/fluentbit/modular.go
quentinmit: This makes the diff slightly smaller, and I am grumpy about the complexity, but I did it. We can revert in a subsequent cleanup CL.

confgenerator/logging_receivers.go
quentinmit: That's because structs like `Syslog` and `WindowsEventlog` are shared across the old and new code. Yes, they will both go away as part of removing all the unused code.

confgenerator/confgenerator.go
```diff
@@ -97,6 +99,152 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
...
+type fbSource struct {
+	tag        string
+	components []fluentbit.Component
+}
+
+func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.InfoStat) ([]fluentbit.Component, error) {
+	var out []fluentbit.Component
+	out = append(out, fluentbit.Service{}.Component())
+
+	if l != nil && l.Service != nil {
```
igorpeshansky, line 128: In the case of `Metrics`, we do this check in `GenerateOtelConfig`, while here you put everything in `generateFluentbitComponents`… Should this be consistent?

confgenerator/confgenerator.go
igorpeshansky: A little too meta, but fine.

confgenerator/confgenerator.go
```diff
@@ -97,6 +99,152 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
...
+				sources = append(sources, fbSource{tag, components})
+			}
+		}
+		sort.Slice(sources, func(i, j int) bool { return sources[i].tag < sources[j].tag })
+		sort.Strings(logNames)
+
+		for _, s := range sources {
+			out = append(out, s.components...)
+		}
+		if len(logNames) > 0 {
```
igorpeshansky, line 160: Is this possible? Would this mean a `Logging` section with no pipelines, or a pipeline with no receivers? Do we want to support such a config? Would it make sense to instead add a validator to prevent this?

quentinmit:
I pushed the golden diffs, so I would appreciate a review of those looking for any bugs. Then I can rip out all the now-unused code.

confgenerator/confgenerator.go
```diff
@@ -97,6 +99,152 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
 	return out, nil
 }
 
+func (uc *UnifiedConfig) GenerateFluentBitConfigsModular(logsDir string, stateDir string, hostInfo *host.InfoStat) (string, string, error) {
```
igorpeshansky, line 102: Just wondering, would it be better to rename the old `GenerateFluentbitConfigs` to `GenerateFluentbitConfigsLegacy`, and name this one `GenerateFluentbitConfigs`?..

confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_fluent_bit_parser.conf
```diff
@@ -1,52 +0,0 @@
-[PARSER]
```
igorpeshansky, line 1: I can't tell if this file is deleted or just empty…

igorpeshansky (changes requested):
The golden diffs look fine to me. Have some questions about the implementation.

confgenerator/confgenerator.go
quentinmit: I'm going to delete the legacy one in a further commit on this branch and update the name of this function at the same time.

confgenerator/confgenerator.go
quentinmit: This was needed to handle one of the test inputs. Therefore I believe for backwards compatibility reasons we can't reject it.

confgenerator/confgenerator.go
quentinmit: I did it that way because I needed to sort the components that exist even without a logging section.

qingling128: Is there a similar need on the metrics side? (e.g. need to sort the components that exist even without a metrics section)

confgenerator/confgenerator.go
```diff
@@ -97,6 +99,152 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
...
+				Match:     strings.Join(logNames, "|"),
+				Workers:   8,
+				UserAgent: userAgent,
+			}.Component())
+		}
+	}
+	// TODO: Use receivers instead of generating Tail objects directly.
+	if hostInfo.OS != "windows" {
+		out = append(out, fluentbit.Tail{
+			Tag:          "ops-agent-collectd",
```
qingling128, line 171: [Optional because it's not introduced by this PR] Should this and line 180 be updated to `ops-agent-open-telemetry-collector` instead now? In theory it's no-op now because otel writes to syslog instead of `metrics-module.log`.

confgenerator/fluentbit/inputs.go
```diff
@@ -92,7 +92,15 @@ func (i Syslog) Component() Component {
 			// as a hint to set "how much data can be up in memory", once the limit is reached it continues writing to disk.
 			"Mem_Buf_Limit": "10M",
 		},
-	}
+	}, {
+		// FIXME: This is not new, but we shouldn't be disabling syslog protocol parsing by passing a custom Parser - Fluentbit includes builtin syslog protocol support, and we should enable/expose that.
```
qingling128, line 96: Agree that we can add a config parameter to control the syslog format. The default would have to stay raw though to avoid breaking changes.

confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_fluent_bit_parser.conf
qingling128: Looks like it's empty: https://github.com/GoogleCloudPlatform/ops-agent/blob/fcad3090ec18d3ece2c7467099ab83302fc07b9d/confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_fluent_bit_parser.conf. Was it removed on purpose?

confgenerator/confgenerator.go
```diff
@@ -97,6 +99,152 @@ func (m *Metrics) generateOtelPipelines() (map[string]otel.Pipeline, error) {
...
+		Match:     "ops-agent-fluent-bit|ops-agent-collectd",
+		Workers:   8,
+		UserAgent: userAgent,
+	}.Component())
+
+	return out, nil
+}
+
+func setLogNameComponents(tag, logName string) []fluentbit.Component {
+	// TODO: Can we just set log_name_key in the output plugin and avoid this mess?
```
qingling128, line 189: If we consistently append the pipeline ID in the tag, we can strip off anything before the first `.` in the output plugin. That should work as well.

Alternatively, something like below might also work (not verified though):

Current config
```
[FILTER]
    Add   logName log_source_id1
    Match pipeline1.log_source_id1
    Name  modify

[FILTER]
    Emitter_Mem_Buf_Limit 10M
    Emitter_Storage.type  filesystem
    Match                 pipeline1.log_source_id1
    Name                  rewrite_tag
    Rule                  $logName .* $logName false

[FILTER]
    Match  log_source_id1
    Name   modify
    Remove logName
```

Alternative config
https://docs.fluentbit.io/manual/pipeline/filters/rewrite-tag#regex
// Match pipeline_id as group 1 and receiver_id as group 2 and rewrite the tag to group 2.
```
[FILTER]
    Emitter_Mem_Buf_Limit 10M
    Emitter_Storage.type  filesystem
    Match                 pipeline1.log_source_id1
    Name                  rewrite_tag
    Rule                  $logName ^([^.]*)\.(.*)$ $2 false
```

I'm leaning towards the former though for simplicity.

confgenerator/testdata/valid/linux/logging-processor_order/golden_fluent_bit_parser.conf
```diff
@@ -1,60 +1,15 @@
 [PARSER]
-    Format      regex
-    Name        lib:default_message_parser
-    Regex       ^(?<message>.*)$
+    Format json
+    Name   pipeline1.sample_logs.0
```
qingling128, line 3: Since the processor names (`alpha` and `beta`) are not preserved in the `parser.conf`, and their signature is identical, it feels like even if the processor order is wrong, this test will no longer detect the problem. Do we need to change it to `pipeline1.sample_logs.alpha.0`, or change the test case to make `alpha` and `beta` slightly different?

confgenerator/confgenerator.go
igorpeshansky: The OT collector does not write its logs to disk. We could technically delete this — the only reason I left it around is as a reminder to get the OT collector to write the logs here.

confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_fluent_bit_parser.conf
igorpeshansky: The new code only generates the parsers that are used, and most of our test inputs don't use the default parsers. Perhaps we should add one catch-all test that uses an instance of each, just to make sure all of them can be generated…

confgenerator/testdata/valid/linux/logging-processor_order/golden_fluent_bit_parser.conf
igorpeshansky: I've thought about it, but the parsers are identical, and the filters do have distinct parts, so we can distinguish them that way. It is a loss of readability to lose the parser name, but it doesn't hurt test coverage as far as I can see.

confgenerator/fluentbit/modular.go
igorpeshansky: I didn't find any explicit guidance in either the Go style or the Go best practices on the named parameters. However, Go is supposed to be all about not surprising readers, and not specifying which of the strings is which opens up a possibility of mixing them up. Would you mind pointing me at the guidance about output parameters that you're referencing?

confgenerator/confgenerator.go
igorpeshansky: Just a reminder that this would still be worth doing, IMO.

igorpeshansky (changes requested):

confgenerator/confgenerator.go
qingling128: Deleting it sounds good to me too.

confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_fluent_bit_parser.conf
qingling128: Ah I see. That makes sense. Resolving this one for now.

confgenerator/testdata/valid/linux/logging-processor_order/golden_fluent_bit_parser.conf
qingling128: I meant something like below:

Right now if we flip the order of two processors, the tests still pass (which is as expected because the two processors are both `parse_json`, but it feels weird as the goal of this unit test is to verify the order):

```
$ git status
On branch quentin-fluentbit-modular
Your branch is up to date with 'origin/quentin-fluentbit-modular'.

$ git diff
diff --git a/confgenerator/testdata/valid/linux/logging-processor_order/input.yaml b/confgenerator/testdata/valid/linux/logging-processor_order/input.yaml
index 61d53c4..fca7d35 100644
--- a/confgenerator/testdata/valid/linux/logging-processor_order/input.yaml
+++ b/confgenerator/testdata/valid/linux/logging-processor_order/input.yaml
@@ -36,5 +36,5 @@ logging:
         exporters: [google]
       pipeline1:
         receivers: [sample_logs]
-        processors: [alpha, beta]
+        processors: [beta, alpha]
         exporters: [google]

$ testall
?   	github.com/GoogleCloudPlatform/ops-agent/cmd/google_cloud_ops_agent_engine	[no test files]
ok  	github.com/GoogleCloudPlatform/ops-agent/confgenerator	(cached)
ok  	github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit	(cached)
?   	github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel	[no test files]
?   	github.com/GoogleCloudPlatform/ops-agent/internal/version	[no test files]
```

For comparison, the test fails as expected in `master` branch:

```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

$ git diff
diff --git a/confgenerator/testdata/valid/linux/logging-processor_order/input.yaml b/confgenerator/testdata/valid/linux/logging-processor_order/input.yaml
index 61d53c4..fca7d35 100644
--- a/confgenerator/testdata/valid/linux/logging-processor_order/input.yaml
+++ b/confgenerator/testdata/valid/linux/logging-processor_order/input.yaml
@@ -36,5 +36,5 @@ logging:
         exporters: [google]
       pipeline1:
         receivers: [sample_logs]
-        processors: [alpha, beta]
+        processors: [beta, alpha]
         exporters: [google]

$ testall
?   	github.com/GoogleCloudPlatform/ops-agent/cmd/google_cloud_ops_agent_engine	[no test files]
--- FAIL: TestGenerateConfsWithValidInput (0.00s)
    --- FAIL: TestGenerateConfsWithValidInput/linux (0.00s)
        --- FAIL: TestGenerateConfsWithValidInput/linux/logging-processor_order (0.04s)
            confgenerator_test.go:112: merged config:
...
                  	"ssage\n    Match    pipeline1.*\n    Name     parser\n    Parser   ",
                - 	"alph",
                + 	"bet",
                  	"a\n\n[FILTER]\n    Key_Name message\n    Match    pipeline1.*\n    Na",
                  	"me     parser\n    Parser   ",
                - 	"bet",
                + 	"alph",
                  	"a\n\n[FILTER]\n    Key_Name message\n    Match    pipeline2.*\n    Na",
                  	"me     parser\n    Parser   beta\n\n[FILTER]\n    Key_Name message\n ",
                  	... // 1986 identical bytes
                  }, "")
FAIL
FAIL	github.com/GoogleCloudPlatform/ops-agent/confgenerator	0.129s
ok  	github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit	0.021s
?   	github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel	[no test files]
?   	github.com/GoogleCloudPlatform/ops-agent/internal/version	[no test files]
FAIL
```

confgenerator/confgenerator.go
igorpeshansky: It was indeed a pipeline with no receivers (`logging-default_overrides_disable_all/input.yaml`). SIgh.

confgenerator/confgenerator.go
quentinmit: I've been preserving it at Igor's requests, but I'm just as happy to delete it.

confgenerator/confgenerator.go
quentinmit: I'm pretty sure fluentbit would let us set `log_name_key logName` in the output plugin and it would just consume the `logName` label directly. I prefer that over anything involving regexes.

confgenerator/confgenerator.go
quentinmit: This comment thread is still open. I plan to go back and comment every exported function after I delete all the unused code.

confgenerator/fluentbit/modular.go
quentinmit: The relevant style entry is https://golang.org/doc/effective_go#named-results (plus more internal guidance that I can't link to in a public PR). I agree that's not a strong recommendation; I have personally been given the explicit feedback in readability reviews that named result parameters should not be used unless those names are needed inside the function.

This function's return parameters will obviously be documented with a comment.

confgenerator/logging_processors.go
quentinmit: I am trying to minimize the diff. I agree it's silly the other one is r, but... cleanup can be separate.

confgenerator/testdata/valid/linux/logging-processor_order/golden_fluent_bit_parser.conf
quentinmit: This is a great observation. I've updated the test to make the processors different.

confgenerator/confgenerator.go
igorpeshansky: If we can indeed do this, it would be a very nice cleanup, but it's probably best to defer it to a follow-up PR. Also, we've been lucky so far that nobody complained about us tromping all over the `logName` field — there's a reason the default `log_name_key` value is `logging.googleapis.com/logName` and not just `logName`.

confgenerator/confgenerator.go
```diff
@@ -424,126 +271,3 @@ func getUserAgent(prefix string, hostInfo *host.InfoStat) (string, error) {
 	}
 	return expandTemplate(userAgentTemplate, prefix, extraParams)
 }
-
-func generateFluentBitInputs(receivers map[string]LoggingReceiver, pipelines map[string]*LoggingPipeline) ([]fluentbit.Input, error) {
-	inputs := []fluentbit.Input{}
-	for _, pID := range sortedKeys(pipelines) {
```
igorpeshansky, line 288: [Optional] There is now more unused code in `confgenerator/config.go`, e.g., type switch cases in `mapKeys`. Some structs in `confgenerator/fluentbit/conf.go` are probably unused as well…

quentinmit: As I said to you in chat, "I want to start ripping out all the unused code, but I want to make sure the diffs are okay first before I lose the ability to switch back."

You've done your review, now give me a chance to actually delete things.

quentinmit:
I've toggled the draft bit off now, PTAL.

confgenerator/fluentbit/modular.go
quentinmit: My understanding from going through the readability process is that naming result parameters makes it possible to use a bare `return`, and if you're not actually using that in the function body, you should not use named result parameters to prevent an accidental bare return.

I agree this seems like a fairly weak argument. I think I'm fine going against that guidance if you really want. (Note that this function signature is not new and it didn't have named result parameters before...)

confgenerator/confgenerator.go
quentinmit: Removed.

confgenerator/confgenerator.go
quentinmit: Hence why I left it as a `TODO`.

confgenerator/confgenerator.go
quentinmit: This mirrors the OT side, I agree but I'd rather clean that up later.

PR #166, opened 2021-08-24
Change nginx 'endpoint' to 'status_url'

djaglowski:


djaglowski:
@qingling128, I think this was what we agreed on earlier. Let me know if you'd like to see a different name.

quentinmit:
I think you have some merge errors - I don't believe there should be any changes to the golden files in this PR.

djaglowski:
@quentinmit Good catch. I've tracked this down, and opened #167 to address the unexpected changes.

Some of these changes are expected though, where the renamed parameter is involved in golden test cases.

quentinmit:
> @quentinmit Good catch. I've tracked this down, and opened #167 to address the unexpected changes.
> 
> Some of these changes are expected though, where the renamed parameter is involved in golden test cases.

I expect the changes to `input.yaml` but not changes to any of the `golden_*` files. The renamed parameter should still generate exactly the same OT config.

djaglowski:
> I expect the changes to input.yaml but not changes to any of the golden_* files. The renamed parameter should still generate exactly the same OT config.

I've merged in changed from #167, so the changes should be much clearer now. 

Changes to nginx tests:
- `input.yaml`s that contained `endpoint` have been updated to use `status_url`
- Golden error files that called out `endpoint`, now call out `status_url`
- Test cases that contained the work `endpoint` renamed to `status_url`
- No changes to otel configs

qingling128 (commented):
Summarizing the discussion offline.

**TL;DR** - We'll change the `endpoint` parameter name to `stub_status_url`.

**Longer story**
Nginx has various different plugins that users can use to expose metrics (today we support the `stub_status_module` only). e.g.
* http://nginx.org/en/docs/http/ngx_http_stub_status_module.html
* http://nginx.org/en/docs/http/ngx_http_api_module.html
* http://nginx.org/en/docs/http/ngx_http_status_module.html (commercial only)
* https://github.com/nginxinc/nginx-prometheus-exporter
* https://github.com/hnlq715/nginx-vts-exporter
* https://github.com/knyar/nginx-lua-prometheus

To ensure the config model is forward compatible, the options in the future in case we have to support some different flavors are:

Option 1 - Different Nginx related receiver for each of the plugins/modules above
Option 2 - Same Nginx receiver with different parameters that indicates the usage of different plugins/modules

We're leaning towards 2, while it should always be possible to do 1 if absolutely necessary. In order to reduce potential conflicts of what `endpoint` means for different Nginx modules, we'd like to make the parameter name specific to the metrics plugins/modules for forward compatibility.

PR #174, opened 2021-08-26
Add JVM receiver

djaglowski:
**Note**: This PR is in a preliminary state and has a few issues to work through. It is being opened now in order to allow for code review and kokoro validation.

Adds a JVM receiver, based on [opentelemetry-collector-contrib/jmxreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/jmxreceiver).

This receiver depends on an additional Java executable, which has been added to the build process and distribution packages.

Known issues & TODOs:
- Intermittent linux build failures have occurred. These appear to stem from one issue, related to jmx metric gatherer's gradle build process. This is being evaluated.
- Manual testing has focused heavily on centos8. Other OS's need additional validation. This is in progress.
- The readme is incomplete, particularly the prerequisites section. In progress.


Dockerfile.windows
```diff
@@ -115,12 +115,11 @@ ADD https://javadl.oracle.com/webapps/download/AutoDL?BundleId=242044_3d5a2bb8f8
 RUN powershell start-process -filepath C:\local\jdk-8u251-windows-x64.exe -passthru -wait -argumentlist '/s,INSTALLDIR=C:\Java,/L,install64.log'
 
 COPY submodules/opentelemetry-java-contrib /work/submodules/opentelemetry-java-contrib
-COPY .git/modules/submodules/opentelemetry-java-contrib /work/submodules/opentelemetry-java-contrib/gitroot
 
 WORKDIR /work/submodules/opentelemetry-java-contrib
 
-RUN ./gradlew --no-daemon :jmx-metrics:build '-Pgit.root="./gitroot"'
-RUN Copy-Item -Path jmx-metrics/build/libs/opentelemetry-java-contrib-jmx-metrics-1.0.0-alpha.jar -Destination /work/out/bin/;
+RUN ./gradlew --no-daemon :jmx-metrics:build
```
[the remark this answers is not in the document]
jsuereth: Yeah, that's only possible w/ nebula.  We'll revisit after bumping this to 1.5.x when it's updated.

confgenerator/jvm.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (r MetricsReceiverJVM) Pipelines() []otel.Pipeline {
+	if r.Endpoint == "" {
+		r.Endpoint = defaultJVMEndpoint
+	}
+
+	jarPath, err := findJarPath()
+	if err != nil {
+		_ = err // TODO(djaglowski) return nil, fmt.Errorf("create jvmreceiver: %v, err)
```
qingling128, line 48: Is `return nil, fmt.Errorf("create jvmreceiver: %v, err)` commented out here by accident or on purpose?

confgenerator/testdata/valid/linux/metrics-receiver_jvm/golden_fluent_bit_main.conf
```diff
@@ -0,0 +1,94 @@
+@SET buffers_dir=/var/lib/google-cloud-ops-agent/fluent-bit/buffers
```
qingling128, line 1: Some of the Fluent Bit golden files changed as part of PR 164. We'll need to rebase / merge `master` HEAD and rerun `-update_golden` in this PR in order to get unit tests to pass:

```
go test -mod=mod github.com/GoogleCloudPlatform/ops-agent/confgenerator -update_golden -v
```

docs/jvm.md
```diff
@@ -0,0 +1,46 @@
+# JVM Receiver
+
+The `jvm` receiver can fetch stats from a Jave Virtual Machine via [JMX](https://www.oracle.com/java/technologies/javase/javamanagement.html).
+
+
+## Prerequisites
+
+The receiver requires that you enable the [status module](http://jvm.org/en/docs/http/ngx_http_stub_status_module.html) on your JVM instance.
+
+In order to expose a JMX endpoint, you must set the `com.sun.management.jmxremote.port system property`. It is recommended to also set the `com.sun.management.jmxremote.rmi.port` system property to the same port. To expose JMX endpoint remotely, you must also set the `java.rmi.server.hostname` system property. Java system properties can be set via command line args by prepending the property name with `-D` . For example: `-Dcom.sun.management.jmxremote.port`
```
qingling128, line 10:
```suggestion
In order to expose a JMX endpoint, you must set the `com.sun.management.jmxremote.port` system property. It is recommended to also set the `com.sun.management.jmxremote.rmi.port` system property to the same port. To expose JMX endpoint remotely, you must also set the `java.rmi.server.hostname` system property. Java system properties can be set via command line args by prepending the property name with `-D` . For example: `-Dcom.sun.management.jmxremote.port`
```

confgenerator/jvm.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return []otel.Pipeline{{
+		Receiver: otel.Component{
+			Type: "jmx",
+			Config: map[string]interface{}{
```
jsuereth, line 54: Mentioned this to Craig in standup today, we need this to include credentials (username/password) for JMX.

jsuereth:
Current integration test failure:

```
Aug 27 21:20:23 github-test-20210827-0b874-1559107e-6335-4496-80c0-7680dfb91665 otelopscol[6196]: 2021-08-27T21:20:23.842Z#011debug#011subprocess/subprocess.go:287#011Error: Unable to access jarfile opentelemetry-java-contrib-jmx-metrics-1.0.0-alpha.jar#011{"kind": "receiver", "name": "jmx/jvm__pipeline_jvm__metrics"}
```

dehaansa:
> Current integration test failure:
> 
> ```
> Aug 27 21:20:23 github-test-20210827-0b874-1559107e-6335-4496-80c0-7680dfb91665 otelopscol[6196]: 2021-08-27T21:20:23.842Z#011debug#011subprocess/subprocess.go:287#011Error: Unable to access jarfile opentelemetry-java-contrib-jmx-metrics-1.0.0-alpha.jar#011{"kind": "receiver", "name": "jmx/jvm__pipeline_jvm__metrics"}
> ```

@jsuereth It looks like however it's being run isn't using the system.d service, as it's missing the BIN_DIRECTORY variable set there. Can you clarify how this is being run, in order to try to understand/resolve the issue?

jsuereth:
@dehaansa Sorry, I've been dealing some critical issues, haven't had time to look.  I think @martijnvans should be able to speak to how these are run in the integration test.  I'll try to run locally myself to see if I can help your local dev flow.

confgenerator/jvm.go
```diff
@@ -48,15 +50,25 @@ func (r MetricsReceiverJVM) Pipelines() []otel.Pipeline {
...
+		"endpoint":            r.Endpoint,
+		"jar_path":            jarPath,
+	}
+
+	// Only set the username & password fields if provided
+	if r.Username != "" {
+		config["username"] = r.Username
+		if r.Password != "" {
+			config["password"] = r.Password
+		}
```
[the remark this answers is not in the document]
dehaansa: Great questions, I'll see what I can figure out about user/password options and revisit this.

confgenerator/jvm.go
djaglowski: This is on purpose, but we should discuss.

How would you like errors to be handled in this method? 

Should `Pipelines()` return `([]otel.Pipeline, error)`?

dehaansa:
Looks like we're all passing now.

```
2021/09/02 08:01:33 Attempt 1 of debian-10 test of nginx finished with err=<nil>, retryable=false
2021/09/02 08:01:46 Attempt 1 of debian-10 test of jvm finished with err=<nil>, retryable=false
--- PASS: TestThirdPartyApps (0.00s)
    --- PASS: TestThirdPartyApps/ops-agent (0.00s)
        --- PASS: TestThirdPartyApps/ops-agent/debian-10 (0.00s)
            --- PASS: TestThirdPartyApps/ops-agent/debian-10/nginx (156.41s)
            --- PASS: TestThirdPartyApps/ops-agent/debian-10/jvm (167.21s)
PASS
```

confgenerator/jvm.go
```diff
@@ -0,0 +1,98 @@
...
+func findJarPath() (string, error) {
+	jarName := "opentelemetry-java-contrib-jmx-metrics-1.0.0-alpha.jar"
+
+	executableDir, err := osext.ExecutableFolder()
+	if err != nil {
+		return jarName, fmt.Errorf("could not determine binary path: %w", err)
+	}
+
+	// TODO(djaglowski) differentiate behavior via build tags
+	if runtime.GOOS != "windows" {
```
jsuereth, line 90: have you been able to test that this works on windows?

Dockerfile.windows
```diff
@@ -101,6 +104,24 @@ RUN cmake --build . --config Release; `
...
+ADD https://javadl.oracle.com/webapps/download/AutoDL?BundleId=242044_3d5a2bb8f8d4428bbe94aed7ec7ae784 /local/jdk-8u251-windows-x64.exe
+RUN powershell start-process -filepath C:\local\jdk-8u251-windows-x64.exe -passthru -wait -argumentlist '/s,INSTALLDIR=C:\Java,/L,install64.log'
+
+COPY submodules/opentelemetry-java-contrib /work/submodules/opentelemetry-java-contrib
+
+WORKDIR /work/submodules/opentelemetry-java-contrib
+
+RUN ./gradlew --no-daemon :jmx-metrics:build
+# TODO: Parameterize this jar name once we can control the release artifact
+RUN Copy-Item -Path contrib/jmx-metrics/build/libs/opentelemetry-java-contrib-jmx-metrics-1.0.0-alpha.jar -Destination /work/out/bin/;
```
jsuereth, line 123: Would you mind renaming this JAR file to not include the version?

- The version should be encoded inside the JAR
- Ideally we don't have to copy-paste the JAR name all over the build every time we bump its version.
- If we upgrade the ops-agent in-place on a machine, we should be replacing the JAR file instead of writing a new one.

build.sh
```diff
@@ -48,6 +48,14 @@ function build_otel() {
   go build -o "$DESTDIR$subagentdir/opentelemetry-collector/otelopscol" ./cmd/otelopscol
 }
 
+function build_otel_jmx() {
+  cd submodules/opentelemetry-java-contrib
+  mkdir -p "$DESTDIR$subagentdir/opentelemetry-collector/"
+  ./gradlew --no-daemon :jmx-metrics:build
+  # TODO: Parameterize this jar name once we can control the release artifact
+  cp "contrib/jmx-metrics/build/libs/opentelemetry-java-contrib-jmx-metrics-1.0.0-alpha.jar" "$DESTDIR$subagentdir/opentelemetry-collector/"
```
jsuereth, line 56: You can also erase the version-number in the jar when you copy here.

confgenerator/jvm.go
```diff
@@ -0,0 +1,98 @@
...
+			otel.NormalizeSums(),
+			otel.MetricsTransform(
+				otel.AddPrefix("workload.googleapis.com"),
+			),
+		},
+	}}
+}
+
+func findJarPath() (string, error) {
+	jarName := "opentelemetry-java-contrib-jmx-metrics-1.0.0-alpha.jar"
```
jsuereth, line 82: Ideally we aren't encoding the version number of the JAR chosen here, that should be an integration-time decision.

pkg/goo/google-cloud-ops-agent.goospec
```diff
@@ -17,6 +17,7 @@
     "out\\bin\\vcruntime140.dll": "{{$INSTALL_DIR}}/bin/vcruntime140.dll",
     "out\\bin\\fluent-bit.exe": "{{$INSTALL_DIR}}/bin/fluent-bit.exe",
     "out\\bin\\fluent-bit.dll": "{{$INSTALL_DIR}}/bin/fluent-bit.dll",
+    "out\\bin\\opentelemetry-java-contrib-jmx-metrics-1.0.0-alpha.jar": "{{$INSTALL_DIR}}/bin/opentelemetry-java-contrib-jmx-metrics-1.0.0-alpha.jar",
```
jsuereth, line 20: Similar comment on removing version number here.

pkg/goo/google-cloud-ops-agent.goospec
```diff
@@ -36,6 +37,7 @@
         "out/bin/fluent-bit.exe",
         "out/bin/fluent-bit.dll",
         "out/bin/google-cloud-metrics-agent_{{$GOOS}}_{{$GOARCH}}.exe",
+        "out/bin/opentelemetry-java-contrib-jmx-metrics-1.0.0-alpha.jar",
```
jsuereth, line 40: Same comment here.

confgenerator/jvm.go
djaglowski: We tested this manually on an earlier commit, and will do so again with the latest code.

jsuereth (approved):
Approved for java-integration related components.

confgenerator/jvm.go
dehaansa: Confirming here that the latest windows build & manual test was successful.

confgenerator/jvm.go
qingling128: Sync'd with Quentin offline.

**TL;DR**

The suggestion is to embed the jar in the binary for now (https://pkg.go.dev/embed) to bypass the need to throw this error and get unblocked quickly. This buys us some time to come up with a cleaner solution for similar needs (e.g. JAVA_HOME) as elaborated below.

**The long story**

In general, user config errors should fail config generation entirely instead of leaving the agent partially working, so that users can quickly detect that something is wrong and fix them. This helps limit the number of customer cases when the root cause is user config error.

However, the error here is about a case when we can't find the jar or Java etc., which is a mix of user error and missing an external dependency.

If in the future we have better diagnose signals (e.g. uptime metrics for each receivers), it's more reasonable to let things fail partially. Today if we let the agent work partially, the service status will still show it as normal. Unless the user looks at the agent logs, or try to query JVM metrics, they won't notice that something is broken, which could be a confusing UX. So for now we'd like to stick with failing the whole config generation.

That means we need a way to return errors from receivers. For configurable paths (e.g. the java home), it could be done through field validation. For things that are not user-configurable, we could allow receivers to define a Validate method to perform extra validation, and if it exists, we'll arrange for that to be called as part of overall validation.

Another idea is to build a mechanism to compute field defaults, and then rely on the existing validation logic to fail if those fields are blank. For example, `JAVA_HOME` should have a computed default, that the user can override, and then only if both of those fail and the field is still empty, we return a config error.

confgenerator/confgenerator_test.go
```diff
@@ -89,6 +90,14 @@ func TestGenerateConfsWithValidInput(t *testing.T) {
 }
 
 func testGenerateConfsWithValidInput(t *testing.T, platform platformConfig) {
+	getExecutableFolder = func() (string, error) {
+		if runtime.GOOS != "windows" {
```
quentinmit, line 94: We generate both windows and linux configs on both platforms, so you can't look at `runtime.GOOS` here :(

Dockerfile.windows
```diff
@@ -101,6 +104,24 @@ RUN cmake --build . --config Release; `
...
 # TODO: Do something with /work/out/bin/fluent-bit.{exe,dll}
 
+
+###############################################################################
+# Build JMX Project
+###############################################################################
+
+FROM base as javacontrib
+
+ADD https://javadl.oracle.com/webapps/download/AutoDL?BundleId=242044_3d5a2bb8f8d4428bbe94aed7ec7ae784 /local/jdk-8u251-windows-x64.exe
```
quentinmit, line 114: Is this the right JDK?

dehaansa: Manually validated on windows with correct JDK.

jsuereth:
Note:   Looked at coveralls drop.  TL;DR; It's dinging you for not having tests that trigger the "not finding jmx-receiver jar" and related code.   I think that's fine for now, but if someone has an idea on a useful coverage test, that'd be ideal to add.

confgenerator/confgenerator_test.go
qingling128: Looks like this has been addressed in https://github.com/GoogleCloudPlatform/ops-agent/commit/91068764f0d1fb3c7b70bf7dfc3a2134c1f7e719#diff-85d8cd0bc6302678b49b04ef15857774aaab001501d09e2df6424d9e20f6ad3d.

qingling128 (approved):
LGTM

qingling128:
Per sync in the standup, this is ready to be merged.

docs/jvm.md
```diff
@@ -0,0 +1,46 @@
...
+## Metrics
+
+| Metric                                             | Data Type | Unit        | Labels | Description |
+| ---                                                | ---       | ---         | ---    | ---         | 
+| workload.googleapis.com/jvm.classes.loade          | gauge     | 1           |        | Current number of loaded classes |
+| workload.googleapis.com/jvm.gc.collections.count   | sum       | 1           | name   | Total number of garbage collections |
+| workload.googleapis.com/jvm.gc.collections.elapsed | sum       | ms          | name   | Time spent garbage collecting |
+| workload.googleapis.com/jvm.memory.heap            | gauge     | by          |        | Current heap usage |
+| workload.googleapis.com/jvm.memory.nonheap         | gauge     | by          |        | Current non-heap usage |
+| workload.googleapis.com/jvm.memory.jvm.memory.pool | gauge     | by          | name   | Current memory pool usage |
```
igorpeshansky, line 45: > `jvm.memory.jvm.memory.pool`

Is this a typo?

docs/jvm.md
```diff
@@ -0,0 +1,46 @@
...
+      jvm_pipeline:
+        receivers:
+          - jvm_metrics
+```
+
+## Metrics
+
+| Metric                                             | Data Type | Unit        | Labels | Description |
+| ---                                                | ---       | ---         | ---    | ---         | 
+| workload.googleapis.com/jvm.classes.loade          | gauge     | 1           |        | Current number of loaded classes |
```
igorpeshansky, line 40: > `jvm.classes.loade`

Is this a typo?

docs/jvm.md
```diff
@@ -0,0 +1,46 @@
...
+| workload.googleapis.com/jvm.gc.collections.count   | sum       | 1           | name   | Total number of garbage collections |
```
igorpeshansky, line 41: These names seem to contradict the [API metric type naming guidelines](https://cloud.google.com/monitoring/api/v3/naming-conventions#naming-types-and-labels). @quentinmit, would this be a problem long-term?

docs/jvm.md
dehaansa: Yes, this is a typo. The correct name is 'jvm.memory.pool'.

docs/jvm.md
dehaansa: Yes, this is also a typo.. The correct name is 'jvm.classes.loaded'

remoe:
This should be optional? When one has a system with small amount of memory and not using java, then this should not be installed ?

PR #181, opened 2021-08-27
Add build image for Ubuntu 21.04 (Hirsute)

jefferbrecht:
Duplicating the commands for an existing Ubuntu target worked fine in testing (with the change for hirsute-backports).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,7 +55,6 @@ RUN ./pkg/deb/build.sh
 FROM ubuntu:hirsute AS hirsute
 
 RUN set -x; apt-get update && \
-    DEBIAN_FRONTEND=noninteractive apt-get -y -t hirsute-backports install debhelper && \
     DEBIAN_FRONTEND=noninteractive apt-get -y install git systemd \
     autoconf libtool libcurl4-openssl-dev libltdl-dev libssl-dev libyajl-dev \
     build-essential cmake bison flex file libsystemd-dev \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -52,6 +52,23 @@ COPY . /work
 WORKDIR /work
 RUN ./pkg/deb/build.sh
 
+FROM ubuntu:hirsute AS hirsute
+
+RUN set -x; apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get -y -t hirsute-backports install debhelper && \
```
igorpeshansky, line 58: You actually shouldn't need to explicitly install `debhelper` either — it's going to be pulled in via `cdbs` [1]. We only had to do this on Xenial because the automatically installed version of `debhelper` was too old. Notice that we don't do this on either Bionic or Focal.

[1] [cdbs](https://packages.ubuntu.com/hirsute/cdbs) -> [dh-translations](https://packages.ubuntu.com/hirsute/dh-translations) -> [debhelper](https://packages.ubuntu.com/hirsute/debhelper)

igorpeshansky (changes requested):

Dockerfile
```diff
@@ -99,7 +99,23 @@ RUN set -xe; \
...
+    DEBIAN_FRONTEND=noninteractive apt-get -y install git systemd \
+    autoconf libtool libcurl4-openssl-dev libltdl-dev libssl-dev libyajl-dev \
+    build-essential cmake bison flex file libsystemd-dev \
+    devscripts cdbs pkg-config
 
+ADD https://golang.org/dl/go1.16.3.linux-amd64.tar.gz /tmp/go1.16.3.linux-amd64.tar.gz
```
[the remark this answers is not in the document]
jefferbrecht: Test failures are unrelated to my PR and will be fixed by #176, so I will be making this change.

Dockerfile
jefferbrecht: Updated.

igorpeshansky (approved):
LGTM :shipit:

PR #189, opened 2021-09-01
Switch to GitHub Actions.

igorpeshansky:


.github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
...
+        go get golang.org/x/tools/cmd/goimports
+        go install github.com/mattn/goveralls@latest
+    - run: |
+        git config --global core.symlinks true
+    - uses: actions/checkout@v2
+    - if: ${{ runner.os == 'Linux' }}
+      run: if [ $(goimports -l .) ]; then goimports -d .; echo 'Failed the goimports format check. Please format the code using "goimports -w ."'; exit 1; fi
+    - run: go test -mod=mod -coverprofile=covprofile ./...
+    - env:
+        COVERALLS_TOKEN: ${{ secrets.github_token }}
```
qingling128, line 24: Just curious: Where is this token set up? It would be great if we could document it somewhere (Either here or in the internal team site works)

.github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
+on: [push, pull_request]
+name: Test
+jobs:
+  test:
+    strategy:
+      fail-fast: false
```
qingling128, line 6: Why `false`? Sounds like failing fast shortens the wait.

qingling128 (approved):
LGTM

.github/workflows/main.yml
igorpeshansky: It's provided by GitHub itself: https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

.github/workflows/main.yml
igorpeshansky: This is fail-fast for jobs, not workflows. Failing fast stops any in-progress parallel jobs. Since we want to parallelize running tests on Linux and Windows, it's [recommended](https://ncorti.com/blog/howto-github-actions-build-matrix) to turn that off.

confgenerator/testdata/.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Golden error messages expect inputs to have Unix line endings.
+input.yaml text eol=lf
```
igorpeshansky, line 2: Our tests depended on the default Travis behavior of using LF line endings even on Windows. This makes the dependency explicit.

igorpeshansky (commented):
Thanks, PTAL.

qingling128:
That answered all my questions. Feel free to submit.

qingling128 (approved):
The new name LGTM

.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+    - uses: actions/setup-go@v2
+      with:
+        go-version: 1.x
+    - run: |
+        go get golang.org/x/tools/cmd/goimports
+        go install github.com/mattn/goveralls@latest
+    - run: |
+        git config --global core.symlinks true
+    - uses: actions/checkout@v2
```
igorpeshansky, line 19: Travis [clones submodules by default](https://docs.travis-ci.com/user/customizing-the-build/#git-submodules). Fixed in #196.

PR #190, opened 2021-09-01
Minor re-arrangement of the Nginx integration tests.

qingling128:
The goal is to align the test structure with https://github.com/GoogleCloudPlatform/ops-agent/blob/master/dev-docs/new-receivers.md#test-an-application-manually, so that the doc team can easily verify things when working on the documentation.

For the reference: public Nginx doc is at https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/nginx.


djaglowski (approved):

jsuereth (approved):

martijnvans (approved):
Looks great, makes me wonder why  we were doing that stuff in the install file in the first place. I think it used to be necessary to avoid a 2 minute  sleep inside collectd, but it's obsolete now. (for both collectd and the ops agent)

PR #191, opened 2021-09-03
Update centos and SLES scripts for nginx

martijnvans:
This PR is just like https://github.com/GoogleCloudPlatform/ops-agent/pull/190, but it changes the scripts in the centos_rhel and sles directories as well.

I'm doing this because I'm seeing some odd behavior when testing other distros, and I think this PR will help with that. It is a bit mysterious though, because what I'm seeing is some passes and some failures, while I would expect to see 100% failures since the endpoint where the metrics are being exposed doesn't match. I'm not sure what's going on with that.

After this nginx tests are passing on all Linux distros I've tried.


qingling128 (approved):
LGTM.

Potential improvement in the future is to combine the linux distros into one when there is no difference among the 3 distro categories (debian/ubuntu, centos/rhel, sles).

PR #196, opened 2021-09-08
Check out submodules when running GitHub actions.

igorpeshansky:
Travis used to do that by default. This fixes a migration regression from #189.


qingling128 (approved):
LGTM

PR #197, opened 2021-09-08
Document JVM authentication parameters

dehaansa:
Document the parameters required to configure the JVM receiver if a JVM's JMX configuration requires authentication.


qingling128 (approved):
LGTM

PR #202, opened 2021-09-13
cmd: log built-in-config.yaml and merged-config.yaml

ridwanmsharif:
This change logs the config files on both windows and linux. It does so
by logging to STDOUT on linux and to the event log on windows.

This is what it looks like in cloud logging:
image

### DIFF
--- a/cmd/google_cloud_ops_agent_engine/main.go
+++ b/cmd/google_cloud_ops_agent_engine/main.go
@@ -51,7 +51,10 @@ func run() error {
 	log.Printf("Built-in config:\n%s", builtInConfig)
 	log.Printf("Merged config:\n%s", mergedConfig)
 
-	hostInfo, _ := host.Info()
+	hostInfo, err := host.Info()
+	if err != nil {
+		return err
+	}
 	uc, err := confgenerator.ParseUnifiedConfigAndValidate(mergedConfig, hostInfo.OS)
 	if err != nil {
 		return err
--- a/cmd/google_cloud_ops_agent_engine/main.go
+++ b/cmd/google_cloud_ops_agent_engine/main.go
@@ -18,10 +18,10 @@ import (
 	"flag"
 	"log"
 	"os"
-	"path/filepath"
 
 	"github.com/GoogleCloudPlatform/ops-agent/apps"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	"github.com/shirou/gopsutil/host"
 )
 
 var (
@@ -40,9 +40,21 @@ func main() {
 }
 func run() error {
 	// TODO(lingshi) Move this to a shared place across Linux and Windows.
-	confDebugFolder := filepath.Join(os.Getenv("RUNTIME_DIRECTORY"), "conf", "debug")
-	if err := confgenerator.MergeConfFiles(*input, confDebugFolder, "linux", apps.BuiltInConfStructs); err != nil {
+	builtInConfig, mergedConfig, err := confgenerator.MergeConfFiles(*input, "linux", apps.BuiltInConfStructs)
+	if err != nil {
 		return err
 	}
-	return confgenerator.GenerateFiles(filepath.Join(confDebugFolder, "merged-config.yaml"), *service, *logsDir, *stateDir, *outDir)
+
+	// Log the built-in and merged config files to STDOUT. These are then written
+	// by journald to var/log/syslog and so to Cloud Logging once the ops-agent is
+	// running.
+	log.Printf("Built-in config:\n%s", builtInConfig)
+	log.Printf("Merged config:\n%s", mergedConfig)
+
+	hostInfo, _ := host.Info()
+	uc, err := confgenerator.ParseUnifiedConfigAndValidate(mergedConfig, hostInfo.OS)
+	if err != nil {
+		return err
+	}
+	return confgenerator.GenerateFilesFromConfig(&uc, *service, *logsDir, *stateDir, *outDir)
 }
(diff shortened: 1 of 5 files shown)


ridwanmsharif:
Alternatively, we could add separate fluent bit [`head`](https://docs.fluentbit.io/manual/pipeline/inputs/head) input plugin to read from the config files directly. And for those receivers have a multi line processor that tries to combine everything into one LogEntry. Its a bit tricker.

The [rule definition](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/multiline-parsing#rules-definition) would be something like:
start_state: "# Generated by Google Cloud Ops Agent" -> A string we would write as the first line of the config files
cont: Negative lookahead of the start_state

With this alternative approach, we would get a single LogEntry with the entire config file (though I don't know what it would do if the input ends up in separate chunks, if the files are too large for example).

I went with the simpler approach for the PR -> just logging to std out. 

cc'ing @davidbtucker and @quentinmit to get some thoughts around how we should self-log something like this.

davidbtucker:
Hey. I personally don't mind the one-entry-per-line for now.

One thing though: Could we get the indentation to show up?
https://user-images.githubusercontent.com/18472685/151616324-8afc4731-b186-46ba-aaf8-a54b0bd53d1f.png
shows everything as left-aligned.

Thanks.

Dave

cmd/google_cloud_ops_agent_engine/main.go
```diff
@@ -44,5 +45,27 @@ func run() error {
...
 	}
+	if err := logConfigFiles(filepath.Join(confDebugFolder, "built-in-config.yaml"), filepath.Join(confDebugFolder, "merged-config.yaml")); err != nil {
+		return err
+	}
 	return confgenerator.GenerateFiles(filepath.Join(confDebugFolder, "merged-config.yaml"), *service, *logsDir, *stateDir, *outDir)
 }
+
+// logConfigFiles logs the built-in and merged config files to STDOUT. These are then written by journald to var/log/syslog and so to
+// Cloud Logging once the ops-agent is running.
+func logConfigFiles(builtInConfigFile, mergedConfigFile string) error {
```
davidbtucker, line 56: I wonder if it would make more sense to do this within `mergeConfFiles`. Maybe see what @quentinmit thinks.

ridwanmsharif: I thought about that but since this is logged differently for Linux (stdout -> journald -> logging agent) and Windows (event log -> logging agent), I thought it probably should live outside the common code

quentinmit: My main comment here is that we shouldn't be reading the files to log them; we should be serializing the in-memory config directly out to the log. I believe we have plans to get rid of the on-disk merged config file at some point.

I don't particularly care about formatting; we can clean that up as part of the improved/structured logging work that @davidbtucker is embarking on.

ridwanmsharif: Restructured some of the changes so we don't rely on reading from files and instead use the in memory structures. Additionally, I moved some code around so we don't rely on reading from these files unnecessarily.

These files are now written to just for debugging purposes.

ridwanmsharif:
> Could we get the indentation to show up? 

The indentation is actually preserved, I see the indentation in jsonPayload but looks like Cloud Logging's UI removes leading whitespace in the quickview

cmd/google_cloud_ops_agent_engine/main.go
```diff
@@ -41,8 +44,34 @@ func main() {
 func run() error {
 	// TODO(lingshi) Move this to a shared place across Linux and Windows.
 	confDebugFolder := filepath.Join(os.Getenv("RUNTIME_DIRECTORY"), "conf", "debug")
-	if err := confgenerator.MergeConfFiles(*input, confDebugFolder, "linux", apps.BuiltInConfStructs); err != nil {
+	mergedConfig, err := confgenerator.MergeConfFiles(*input, confDebugFolder, "linux", apps.BuiltInConfStructs)
```
davidbtucker, line 47: In the interest of minimizing this PR's diff, could we just have `MergeConfFiles` return the built-in config bytes as well?

I also think we should refactor this code (outside of this PR) to simplify the flow. Maybe something like:

1. Read input files
2. Generate outputs
3. Log inputs + outputs
4. Write output files

@quentinmit might have other ideas too.

quentinmit: 4 is unnecessary, the only reason we write the intermediate files now is because of the current structure of the code. 1-3 is sufficient for this feature of logging the config.

ridwanmsharif: Have we decided that we don't want to keep those files at all? In the event that Fluent Bit fails for whatever reason, we won't be able to see the logs in Cloud Logging and having these files locally for customers to send over and/or look at themselves could be pretty useful. I kept it here just for debuggability reasons but happy to remove it if we decide.

quentinmit: The configs will still end up in the logs on the local disk without 4, so it doesn't affect the "Fluent Bit fails" case.

ridwanmsharif:
Had some free cycles so fixed up this PR. The latest commit makes it so we no longer write the debug files to disk as we talked about.

cmd/google_cloud_ops_agent_engine/main.go
ridwanmsharif: Refactored to not write these output files and merge in-memory.

quentinmit (approved):
+100, thanks for the cleanup. Still more work to do but this is already a nice improvement.

PR #203, opened 2021-09-13
NGINX logs severity by response class and latency + docs

draffensperger:


confgenerator/nginx.go
```diff
@@ -70,17 +70,19 @@ func (LoggingProcessorNginxAccess) Type() string {
 
 func (p LoggingProcessorNginxAccess) Components(tag string, uid string) []fluentbit.Component {
 	c := LoggingProcessorParseRegex{
-		// Sample line: ::1 - - [26/Aug/2021:16:49:43 +0000] "GET / HTTP/1.1" 200 10701 "-" "curl/7.64.0"
+		// Sample line (default config): ::1 - - [26/Aug/2021:16:49:43 +0000] "GET / HTTP/1.1" 200 10701 "-" "curl/7.64.0"
+		// Sample line (with latency in seconds): ::1 - - [26/Aug/2021:16:49:43 +0000] "GET / HTTP/1.1" 200 10701 "-" "curl/7.64.0" 1.002s
```
quentinmit, line 74: The only documented log format is "combined" at http://nginx.org/en/docs/http/ngx_http_log_module.html and it does not contain a latency.

We shouldn't be encouraging users to define custom log formats. If users want a latency or response code metric, we should just expose those metrics in the metrics receiver.

confgenerator/nginx.go
```diff
@@ -70,17 +70,19 @@ func (LoggingProcessorNginxAccess) Type() string {
...
 			TimeFormat: "%d/%b/%Y:%H:%M:%S %z",
-			Types: map[string]string{
-				"http_request_status": "integer",
-				// N.B. "http_request_responseSize" is a string containing an integer.
-				// https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest.FIELDS.response_size
-			},
+			// The `http_request_status` field is an integer, however the FluentBit
+			// "Key_Value_Matches" condition only works on strings, so it is
+			// intentionally left as a string type, though it is correctly
+      // interpreted as an integer when it is exported to Cloud Logging.
```
quentinmit, line 83: How does this get turned into an integer? I'm not aware of any code that will do that (and we don't want it to happen for `responseSize` / it doesn't seem to be happening for `responseSize`)

confgenerator/nginx.go
```diff
@@ -99,6 +101,25 @@ func (p LoggingProcessorNginxAccess) Components(tag string, uid string) []fluent
 			},
 		})
 	}
+	// Assign severity based on HTTP status code class (enables use of system
```
quentinmit, line 104: This is very, very weird to me. "5xx" responses are arguably errors, but the other levels are all normal responses.

qingling128:
Linked this to b/207227303. Closing for now. Will revisit when that work gets prioritized.

PR #205, opened 2021-09-16
Require TCP logging receivers to always specify "format" until we converge on a default behavior

quentinmit:


quentinmit:
Closing as dupe :(

PR #206, opened 2021-09-16
Add required 'format' parameter to TCP input plugin.

jkschulz:
http://b/195571656

### DIFF
--- a/confgenerator/logging_receivers.go
+++ b/confgenerator/logging_receivers.go
@@ -156,6 +156,7 @@ func init() {
 type LoggingReceiverTCP struct {
 	ConfigComponent `yaml:",inline"`
 
+	Format     string `yaml:"format,omitempty" validate:"required,oneof=json"`
 	ListenHost string `yaml:"listen_host,omitempty" validate:"omitempty,ip"`
 	ListenPort uint16 `yaml:"listen_port,omitempty"`
 }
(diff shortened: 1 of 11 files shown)


confgenerator/logging_receivers.go
```diff
@@ -180,7 +181,7 @@ func (r LoggingReceiverTCP) Components(tag string) []fluentbit.Component {
 			"Tag":    tag,
 			"Listen": r.ListenHost,
 			"Port":   fmt.Sprintf("%d", r.ListenPort),
-			"Format": "json",
```
quentinmit, line 183: I would keep this hardcoded as it was before, because we aren't necessarily using the same strings/config format as fluentbit. The rest of this config is all still hardcoded for streaming JSON.

confgenerator/testdata/invalid/linux/logging-receiver_tcp_type_missing_required_parameter_format/golden_error
```diff
@@ -0,0 +1,8 @@
+[3:13] "format" is a required field
```
quentinmit, line 1: I prefer the error you get if you don't mark this `required` (`"format" must be one of [json]`) but I don't feel strongly about it

quentinmit (approved):

confgenerator/testdata/invalid/linux/logging-receiver_tcp_type_missing_required_parameter_format/golden_error
jkschulz: Custom error messages from the validator can be a future PR. Leaving as is for consistency.

confgenerator/logging_receivers.go
quentinmit: (My fear is that someone will see your change and think "great, we already planned ahead for more formats, I don't need to change anything to support the next format")

qingling128 (approved):
LGTM

PR #212, opened 2021-09-22
Make "docker build --target=..." do something sensible.

igorpeshansky:
Previously, this would copy out the whole contents of the build container. This change renames the original targets to `...-build` and reuses the original target names to represent the artifacts from the corresponding build container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 # Build as DOCKER_BUILDKIT=1 docker build -o /tmp/out .
+# or DOCKER_BUILDKIT=1 docker build -o /tmp/out . --target=buster
 # Generated tarball(s) will end up in /tmp/out
 
-FROM debian:buster AS buster
+FROM debian:buster AS buster-build
 
 # TODO: Factor out the common code without rerunning apt-get on every build.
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -193,33 +194,54 @@ COPY . /work
 WORKDIR /work
 RUN ./pkg/rpm/build.sh
 
-FROM scratch
-COPY --from=buster /tmp/google-cloud-ops-agent.tgz /google-cloud-ops-agent-debian-buster.tgz
-COPY --from=buster /google-cloud-ops-agent*.deb /
+FROM scratch AS buster
```
davidbtucker, line 197: Personal nit (feel free to ignore):

I would find `debian10-buster` clearer than just `buster`.  It would benefit those of us who don't keep a (nickname -> distro) map in main memory, and also make the list at 238-247 clearer.

davidbtucker (approved):

Dockerfile
igorpeshansky: This is a backward-compatible change — the target was already called `buster` (now `buster-build`), I'm just reusing the name for another target. We can discuss renaming the targets in a follow-up PR.

ridwanmsharif (approved):
LGTM

PR #213, opened 2021-09-23
Implement Apache Httpd log receivers

dehaansa:
Add `apache_access` and `apache_error` log receivers, following the patterns established in the similar [nginx log receivers](https://github.com/GoogleCloudPlatform/ops-agent/pull/176).

Sample parsed access log:
image

Sample parsed error log:
image

### DIFF
--- a/confgenerator/apache.go
+++ b/confgenerator/apache.go
@@ -1,0 +1,167 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package confgenerator
+
+import (
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
+)
+
+type LoggingProcessorApacheAccess struct {
+	ConfigComponent `yaml:",inline"`
+}
+
+func (LoggingProcessorApacheAccess) Type() string {
+	return "apache_access"
+}
+
+func (p LoggingProcessorApacheAccess) Components(tag string, uid string) []fluentbit.Component {
+	c := LoggingProcessorParseRegex{
+		// Documentation: https://httpd.apache.org/docs/current/logs.html#accesslog
+		// Sample "common" line: 127.0.0.1 - frank [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326
+		// Sample "combined" line: ::1 - - [26/Aug/2021:16:49:43 +0000] "GET / HTTP/1.1" 200 10701 "-" "curl/7.64.0"
+		Regex: `^(?<http_request_remoteIp>[^ ]*) (?<host>[^ ]*) (?<user>[^ ]*) \[(?<time>[^\]]*)\] "(?<http_request_requestMethod>\S+)(?: +(?<http_request_requestUrl>[^\"]*?)(?: +(?<http_request_protocol>\S+))?)?" (?<http_request_status>[^ ]*) (?<http_request_responseSize>[^ ]*)(?: "(?<http_request_referer>[^\"]*)" "(?<http_request_userAgent>[^\"]*)")?$`,
+		LoggingProcessorParseShared: LoggingProcessorParseShared{
+			TimeKey:    "time",
+			TimeFormat: "%d/%b/%Y:%H:%M:%S %z",
+			Types: map[string]string{
+				"http_request_status": "integer",
+				// N.B. "http_request_responseSize" is a string containing an integer.
+				// https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest.FIELDS.response_size
+			},
+		},
+	}.Components(tag, uid)
+	// apache logs "-" when a field does not have a value. Remove the field entirely when this happens.
+	for _, field := range []string{
+		"host",
+		"user",
+		"http_request_referer",
+	} {
+		c = append(c, fluentbit.Component{
+			Kind: "FILTER",
+			Config: map[string]string{
+				"Name":      "modify",
+				"Match":     tag,
+				"Condition": fmt.Sprintf("Key_Value_Equals %s -", field),
+				"Remove":    field,
+			},
+		})
+	}
+	// Generate the httpRequest structure.
+	c = append(c, fluentbit.Component{
+		Kind: "FILTER",
+		Config: map[string]string{
+			"Name":          "nest",
+			"Match":         tag,
+			"Operation":     "nest",
+			"Wildcard":      "http_request_*",
+			"Nest_under":    "logging.googleapis.com/http_request",
+			"Remove_prefix": "http_request_",
+		},
+	})
+	return c
+}
+
+type LoggingProcessorApacheError struct {
+	ConfigComponent `yaml:",inline"`
+}
+
+func (LoggingProcessorApacheError) Type() string {
+	return "apache_error"
+}
+
+func (p LoggingProcessorApacheError) Components(tag string, uid string) []fluentbit.Component {
+	c := LoggingProcessorParseRegex{
+		// Documentation: https://httpd.apache.org/docs/current/logs.html#errorlog
+		// Sample line 2.4: [Fri Sep 09 10:42:29.902022 2011] [core:error] [pid 35708:tid 4328636416] (13)Permission denied [client 72.15.99.187] File does not exist: /usr/local/apache2/htdocs/favicon.ico
+		// Sample line 2.2: [Fri Sep 09 10:42:29.902022 2011] [error] [pid 35708:tid 4328636416] [client 72.15.99.187] File does not exist: /usr/local/apache2/htdocs/favicon.ico
+		// TODO - Support time parsing for version 2.0 where smallest resolution is seconds
+		// Sample line 2.0: [Wed Oct 11 14:32:52 2000] [error] [client 127.0.0.1] client denied by server configuration: /export/home/live/ap/htdocs/test
+		Regex: `^\[(?<time>[^\]]+)\] \[(?:(?<module>\w+):)?(?<level>[\w\d]+)\](?: \[pid (?<pid>\d+)(?::tid (?<tid>[0-9]+))?\])?(?: (?<error_code>[^\[]*))?(?: \[client (?<client>[^\]]*)\])? (?<message>.*)$`,
+		LoggingProcessorParseShared: LoggingProcessorParseShared{
+			TimeKey:    "time",
+			TimeFormat: "%a %b %d %H:%M:%S.%L %Y",
+			Types: map[string]string{
+				"pid": "integer",
+				"tid": "integer",
+			},
+		},
+	}.Components(tag, uid)
+	for _, l := range []struct{ level, severity string }{
+		{"emerg", "EMERGENCY"},
+		{"alert", "ALERT"},
+		{"crit", "CRITICAL"},
+		{"error", "ERROR"},
+		{"warn", "WARNING"},
+		{"notice", "NOTICE"},
+		{"info", "INFO"},
+		{"debug", "DEBUG"},
+		{"trace1", "DEBUG"},
+		{"trace2", "DEBUG"},
+		{"trace3", "DEBUG"},
+		{"trace4", "DEBUG"},
+		{"trace5", "DEBUG"},
+		{"trace6", "DEBUG"},
+		{"trace7", "DEBUG"},
+		{"trace8", "DEBUG"},
+	} {
+		c = append(c, fluentbit.Component{
+			Kind: "FILTER",
+			Config: map[string]string{
+				"Name":      "modify",
+				"Match":     tag,
+				"Condition": fmt.Sprintf("Key_Value_Equals level %s", l.level),
+				"Add":       fmt.Sprintf("logging.googleapis.com/severity %s", l.severity),
+			},
+		})
+	}
+	return c
+}
+
+type LoggingReceiverApacheAccess struct {
+	LoggingProcessorApacheAccess `yaml:",inline"`
+	LoggingReceiverFilesMixin    `yaml:",inline" validate:"structonly"`
+}
+
+func (r LoggingReceiverApacheAccess) Components(tag string) []fluentbit.Component {
+	if len(r.IncludePaths) == 0 {
+		r.IncludePaths = []string{"/var/log/apache2/access.log"}
+	}
+	c := r.LoggingReceiverFilesMixin.Components(tag)
+	c = append(c, r.LoggingProcessorApacheAccess.Components(tag, "apache_access")...)
+	return c
+}
+
+type LoggingReceiverApacheError struct {
+	LoggingProcessorApacheError `yaml:",inline"`
+	LoggingReceiverFilesMixin   `yaml:",inline" validate:"structonly"`
+}
+
+func (r LoggingReceiverApacheError) Components(tag string) []fluentbit.Component {
+	if len(r.IncludePaths) == 0 {
+		r.IncludePaths = []string{"/var/log/apache2/error.log"}
+	}
+	c := r.LoggingReceiverFilesMixin.Components(tag)
+	c = append(c, r.LoggingProcessorApacheError.Components(tag, "apache_error")...)
+	return c
+}
+
+func init() {
+	loggingProcessorTypes.registerType(func() component { return &LoggingProcessorApacheAccess{} })
+	loggingProcessorTypes.registerType(func() component { return &LoggingProcessorApacheError{} })
+	loggingReceiverTypes.registerType(func() component { return &LoggingReceiverApacheAccess{} })
+	loggingReceiverTypes.registerType(func() component { return &LoggingReceiverApacheError{} })
+}
--- a/integration_test/third_party_apps_data/applications/apache/expected_logs.yaml
+++ b/integration_test/third_party_apps_data/applications/apache/expected_logs.yaml
@@ -1,0 +1,12 @@
+log_entries:
+- log_name: "apache_access"
+  field_matchers:
+    httpRequest.remoteIp: "::1"
+    httpRequest.requestUrl: "/forbidden.html"
+- log_name: "apache_error"
+  field_matchers:
+    jsonPayload.client: "::1"
+    jsonPayload.level: "error"
+    jsonPayload.module: "core"
+    jsonPayload.message: '.*file permissions deny server access.*'
+    jsonPayload.error_code: '.*Permission denied.*'
--- a/integration_test/third_party_apps_data/agent/ops-agent/linux/enable_apache
+++ b/integration_test/third_party_apps_data/agent/ops-agent/linux/enable_apache
@@ -1,0 +1,34 @@
+# Configures Ops Agent to collect telemetry from the app and restart Ops Agent.
+
+set -e
+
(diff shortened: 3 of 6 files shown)


dehaansa:
From what I can tell kokoro *would* pass except it is expecting a metric_name.txt file and the metrics for apache are coming in another PR. Could kokoro become more flexible, or is there another way I can work around this?

confgenerator/apache.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+type LoggingReceiverApacheAccess struct {
+	LoggingProcessorApacheAccess `yaml:",inline"`
+	LoggingReceiverFilesMixin    `yaml:",inline" validate:"structonly"`
+}
+
+func (r LoggingReceiverApacheAccess) Components(tag string) []fluentbit.Component {
+	if len(r.IncludePaths) == 0 {
+		r.IncludePaths = []string{"/var/log/apache2/access.log"}
```
qingling128, line 141: I recall the default paths were different based on VM distros (some examples in https://github.com/GoogleCloudPlatform/ops-agent/blob/master/dev-docs/new-receivers.md#1-research-what-telemetry-the-application-generates)

Is that no longer the case with newer versions, or is it because our integration tests failed to catch it?

qingling128: Ah, looking at the [latest build](https://pantheon.corp.google.com/storage/browser/ops-agents-public-buckets-test-logs/prod/stackdriver_agents/testing/consumer/third_party_apps/presubmit_github/264/20210928-123920/logs), the 3rd party tests only ran on Debian 10. That's probably why the tests did not catch this.

qingling128: I just sent a change internally to extend it to:

```
      'debian-10',
      'centos-8',
      'rhel-7',
      'sles-15',
      'ubuntu-2004-lts',
```

Waiting for review / approval there.

dehaansa: This was the main location we found in our research (also in that document it calls out "Apache 2 seems to be dominating. We support that version only"). If desired, we can add the httpd path as well to the default include paths array.

confgenerator/apache.go
```diff
@@ -0,0 +1,167 @@
...
+	for _, l := range []struct{ level, severity string }{
+		{"emerg", "EMERGENCY"},
+		{"alert", "ALERT"},
+		{"crit", "CRITICAL"},
+		{"error", "ERROR"},
+		{"warn", "WARNING"},
+		{"notice", "NOTICE"},
+		{"info", "INFO"},
+		{"debug", "DEBUG"},
+		{"trace1", "DEBUG"},
```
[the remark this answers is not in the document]
dehaansa: Trace8 is the maximum. Simply for simplicity of the code I opted to not make a separate key_value_matches block, but we could add it in instead of the current pattern.

integration_test/third_party_apps_data/applications/apache/expected_logs.yaml
```diff
@@ -0,0 +1,12 @@
+log_entries:
```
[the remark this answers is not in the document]
dehaansa: The other available fields in httpRequest aren't in the default Apache Httpd log formats, so they aren't available. When we enhance these processors to support custom log formats they may be available, but currently they don't exist in the log formats.

docs/apache.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Configuration
+
+To configure a receiver for your apache access logs, specify the following fields:
+
+| Field                 | Default                       | Description |
+| ---                   | ---                           | ---         |
+| `type`                | required                      | Must be `apache_access`. |
+| `include_paths`       | `[/var/log/apache2/access.log]` | The log files to read. |
+| `exclude_paths`       | `[]`                          | Log files to exclude (if `include_paths` contains a glob or directory). |
```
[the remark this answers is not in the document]
dehaansa: We could link to `https://docs.fluentbit.io/manual/pipeline/inputs/tail#config` where this parameter is directly applied, but even that documentation only states "common wildcards". I'm not sure what the best documentation to provide would be here.

integration_test/third_party_apps_data/applications/apache/expected_logs.yaml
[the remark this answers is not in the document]
qingling128: Weird. I thought the sample log above is pulled from log entries generated by the Kokoro tests that are triggered by this PR based on the `/forbidden.html` URL. Maybe I missed something. Checking.

confgenerator/apache.go
qingling128: +1 that we do not need to support Apache 1.

The issue here is slightly different though. As mentioned in https://github.com/GoogleCloudPlatform/ops-agent/blob/master/dev-docs/new-receivers.md#1-research-what-telemetry-the-application-generates, the access / error log paths differ by distro, even for Apache 2 alone.

```
Default log file paths (RHEL / Red Hat / CentOS / Fedora) | /var/log/httpd/access_log and /var/log/httpd/error_log
Default log file paths (Debian / Ubuntu) | /var/log/apache2/access.log and /var/log/apache2/error.log
```

Spot checked centos 8:
```
[lingshi@centos8 ~]$ sudo service httpd status
Redirecting to /bin/systemctl status httpd.service
● httpd.service - The Apache HTTP Server
   Loaded: loaded (/usr/lib/systemd/system/httpd.service; disabled; vendor preset: disabled)
   Active: active (running) since Tue 2021-09-28 20:27:26 UTC; 29s ago
     Docs: man:httpd.service(8)
 Main PID: 49498 (httpd)
   Status: "Running, listening on: port 80"
    Tasks: 213 (limit: 22244)
   Memory: 24.5M
   CGroup: /system.slice/httpd.service
           ├─49498 /usr/sbin/httpd -DFOREGROUND
           ├─49499 /usr/sbin/httpd -DFOREGROUND
           ├─49500 /usr/sbin/httpd -DFOREGROUND
           ├─49501 /usr/sbin/httpd -DFOREGROUND
           └─49502 /usr/sbin/httpd -DFOREGROUND

Sep 28 20:27:26 centos8 systemd[1]: Starting The Apache HTTP Server...
Sep 28 20:27:26 centos8 systemd[1]: Started The Apache HTTP Server.
Sep 28 20:27:26 centos8 httpd[49498]: Server configured, listening on: port 80
[lingshi@centos8 ~]$ sudo cat /var/log/apache2/access.log
cat: /var/log/apache2/access.log: No such file or directory
[lingshi@centos8 ~]$ sudo cat /var/log/httpd/access_log
[lingshi@centos8 ~]$ httpd -v
Server version: Apache/2.4.37 (centos)
Server built:   May 20 2021 04:33:06
```

confgenerator/apache.go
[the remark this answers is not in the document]
qingling128: Got it. Can we add a comment to mention the rationale? Optionally also add a link to the Apache documentation reference that Trace8 is the maximum (if there is a link)?

confgenerator/apache.go
dehaansa: Okay, I've updated the default include paths to support both common default paths.

confgenerator/apache.go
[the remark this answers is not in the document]
dehaansa: Added a comment & doc link, let me know if you have any further thoughts on it.

integration_test/third_party_apps_data/applications/apache/expected_logs.yaml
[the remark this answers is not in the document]
dehaansa: I might also be misunderstanding what you're looking for - it appears to me that all of the relevant httpRequest components that exist in the logs are present in the httpRequest object, from my understanding it sounded like you were saying some were not being parsed but were present in the logs.

confgenerator/apache.go
qingling128: Looks great. Thanks!

confgenerator/apache.go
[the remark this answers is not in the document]
qingling128: Perfect. Thanks!

integration_test/third_party_apps_data/applications/apache/expected_logs.yaml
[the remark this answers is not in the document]
qingling128: Oops, let me clarify: I meant changing the `expected_logs.yaml` file content here to something that verifies the other fields as well. e.g.:

```
- log_name: "apache_access"
  field_matchers:
    httpRequest.remoteIp: "::1"
    httpRequest.requestUrl: "/forbidden.html"
    httpRequest.protocol: "HTTP/1.1"
    httpRequest.remoteIp: "::1"
    httpRequest.requestMethod: "GET"
    httpRequest.requestUrl: "/forbidden.html"
    httpRequest.responseSize: "435"
    httpRequest.status: 403
    httpRequest.userAgent: "curl/7.64.0"
```
unless there are special reasons like some of the fields are non-deterministic.

Right now it only has checks for two fields:

```
- log_name: "apache_access"
  field_matchers:
    httpRequest.remoteIp: "::1"
    httpRequest.requestUrl: "/forbidden.html"
```

qingling128 (approved):
LGTM

integration_test/third_party_apps_data/applications/apache/expected_logs.yaml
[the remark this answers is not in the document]
dehaansa: Makes sense! The only one I'm unsure of is if responseSize could be different on different distributions for some reason. I think we should put it in for now and if it turns out to be an issue in the future we'll remove.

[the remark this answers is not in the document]
dehaansa: responseSize caused the following so I've just removed it for now:
```
2021/09/28 16:45:31 Attempt 1 of debian-10 test of apache finished with err=QueryLog() failed: rpc error: code = InvalidArgument desc = Field 'response_size' is of incorrect type. Regular expression matches can only be performed against string fields., retryable=false
```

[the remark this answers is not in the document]
qingling128: Sounds good. Thanks for taking care of this!

qingling128 (approved):
LGTM

qingling128:
Note: We are planning to release Ops Agent `2.3.1` today (Sept 29). Please do not merge this PR before https://github.com/GoogleCloudPlatform/ops-agent/releases/tag/2.3.1 is out.

apps/apache.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (p LoggingProcessorApacheError) Components(tag string, uid string) []fluentbit.Component {
+	c := confgenerator.LoggingProcessorParseRegex{
+		// Documentation: https://httpd.apache.org/docs/current/logs.html#errorlog
+		// Sample line 2.4: [Fri Sep 09 10:42:29.902022 2011] [core:error] [pid 35708:tid 4328636416] (13)Permission denied [client 72.15.99.187] File does not exist: /usr/local/apache2/htdocs/favicon.ico
+		// Sample line 2.2: [Fri Sep 09 10:42:29.902022 2011] [error] [pid 35708:tid 4328636416] [client 72.15.99.187] File does not exist: /usr/local/apache2/htdocs/favicon.ico
+		// TODO - Support time parsing for version 2.0 where smallest resolution is seconds
+		// Sample line 2.0: [Wed Oct 11 14:32:52 2000] [error] [client 127.0.0.1] client denied by server configuration: /export/home/live/ap/htdocs/test
+		Regex: `^\[(?<time>[^\]]+)\] \[(?:(?<module>\w+):)?(?<level>[\w\d]+)\](?: \[pid (?<pid>\d+)(?::tid (?<tid>[0-9]+))?\])?(?: (?<errorCode>[^\[]*))?(?: \[client (?<client>[^\]]*)\])? (?<message>.*)$`,
```
quentinmit, line 94: Will this pattern cause the message to end up in the errorCode field when the error log does not contain a `[client` field? e.g. here's a real error log entry from 2.4:

```
[Thu Sep 30 03:18:29.239182 2021] [ssl:error] [pid 2451:tid 140169666050176] AH02217: ssl_stapling_init_cert: Can't retrieve issuer certificate!
```

apps/apache.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+type LoggingReceiverApacheAccess struct {
+	LoggingProcessorApacheAccess            `yaml:",inline"`
+	confgenerator.LoggingReceiverFilesMixin `yaml:",inline" validate:"structonly"`
+}
+
+func (r LoggingReceiverApacheAccess) Components(tag string) []fluentbit.Component {
+	if len(r.IncludePaths) == 0 {
+		r.IncludePaths = []string{"/var/log/apache2/access.log", "/var/log/httpd/access_log"}
```
quentinmit, line 145: Did we confirm these paths cover all of the supported Ops Agent distros? (The former is definitely current Debian/Ubuntu, the latter I believe is current Fedora, what about SLES etc.?)

quentinmit (approved):

apps/apache.go
qingling128: Filed a bug yesterday to extend the 3rd party app automated tests to more distros, then found there is already a WIP ticket for it: b/198830714. It requires some additional work. For now, we might need to manually spot check SLES.

apps/apache.go
dehaansa: Yeah, this fails to parse that message appropriately. The below would capture the message you sent properly, but if we're missing a semicolon at the end of the errorCode it would fail. Do you think we should try to capture error code even if it could be fragile, or just capture errorCode only if it precedes a client and is clear like that.
```
^\[(?<time>[^\]]+)\] \[(?:(?<module>\w+):)?(?<level>[\w\d]+)\](?: \[pid (?<pid>\d+)(?::tid (?<tid>[0-9]+))?\])?(?: (?<errorCode>[^\[:]*):?)?(?: \[client (?<client>[^\]]*)\])? (?<message>.*)$
```

apps/apache.go
dehaansa: According to a bit of searching it looks like the SLES paths are similar to Debain/Ubuntu but with and underscore rather than a .log ("/var/log/apache2/error_log" for example). My initial searching I misread them to be the same as D/U. Added the paths to the defaults.

docs/apache.md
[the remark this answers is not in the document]
qingling128: For consistency, let's follow the `files` receiver: 
image

apps/apache.go
qingling128: I'd lean towards something more restricted and we can loosen it if needed because it's forward compatible. Because the other way around could be considered a breaking change.

docs/apache.md
```diff
@@ -0,0 +1,69 @@
+# `apache_access` and `apache_error` Logging Receivers
+
+## Configuration
+
+To configure a receiver for your apache access logs, specify the following fields:
+
+| Field                 | Default                       | Description |
+| ---                   | ---                           | ---         |
+| `type`                | required                      | Must be `apache_access`. |
+| `include_paths`       | `[/var/log/apache2/access.log]` | The log files to read. |
```
qingling128, line 10:
```suggestion
| `include_paths`       | `[/var/log/apache2/access.log]` | A list of filesystem paths to read by tailing each file. A wild card (`*`) can be used in the paths; for example, `/var/log/apache*/*.log`.
```

docs/apache.md
qingling128, line 11:
```suggestion
| `exclude_paths`       | `[]`                          | A list of filesystem path patterns to exclude from the set matched by `include_paths`.
```

apps/apache.go
qingling128: Looks like this has been updated. Resolving this thread.

apps/apache.go
dehaansa: @qingling128 my understanding is that you mean to use the above new regex, in order to capture errorCode when available, even thought it could be fragile. Is that correct?

qingling128:
Integration tests failed. But when I manually checked the ingested logs, it was okay:

image

image

Re-run the tests.

dehaansa:
@qingling128 from the logs in kokoro, it looks like integration tests failed from a panic in the test running code, perhaps something in the apparent update to run against more distributions.

```runtime: running GoogleExitFunction

goroutine 89 [running]:
testing.tRunner.func1.2({0x55fffca8ef40, 0x55fffd6bce10})
	third_party/go/gc/src/testing/testing.go:1209 +0x24e
testing.tRunner.func1()
	third_party/go/gc/src/testing/testing.go:1212 +0x225
panic(--- FAIL: TestProcessorOrder (637.54s)
{0x55fffca8ef40, 0x55fffd6bce10})
	third_party/go/gc/src/runtime/panic.go    --- FAIL: TestProcessorOrder/debian-10 (637.54s)
:814 +0x207
ops_agent_test.TestProcessorOrder.func1(0xc000603ba0, {0xc0000480b2, 0x9})
	cloud/monitoring/agents/kokoro/tests/ops_agent_test.go:364 +0x5b3
gce.RunForEachPlatform.func1(0x0)
	cloud/monitoring/agents/kokoro/tests/gce_testing.go:1383 +0x25
testing.tRunner(0xc000603ba0, 0xc000576220)
	third_party/go/gc/src/testing/testing.go:1259 +0x102
created by testing.(*T).Run
	third_party/go/gc/src/testing/testing.go:1306 +0x35f

panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x55fffc032c93]

goroutine 89 [running]:
testing.tRunner.func1.2({0x55fffca8ef40, 0x55fffd6bce10})
	third_party/go/gc/src/testing/testing.go:1209 +0x24e
testing.tRunner.func1()
	third_party/go/gc/src/testing/testing.go:1212 +0x225
panic({0x55fffca8ef40, 0x55fffd6bce10})
	third_party/go/gc/src/runtime/panic.go:814 +0x207
ops_agent_test.TestProcessorOrder.func1(0xc000603ba0, {0xc0000480b2, 0x9})
	cloud/monitoring/agents/kokoro/tests/ops_agent_test.go:364 +0x5b3
gce.RunForEachPlatform.func1(0x0)
	cloud/monitoring/agents/kokoro/tests/gce_testing.go:1383 +0x25
testing.tRunner(0xc000603ba0, 0xc000576220)
	third_party/go/gc/src/testing/testing.go:1259 +0x102
created by testing.(*T).Run
	third_party/go/gc/src/testing/testing.go:1306 +0x35f

exit status 2```

**Comment by qingling128** [comment]
The tests only run for Debian 10 today. The change to extend it to more distros turned out to be a non-trivial effort. That would take some time to land.

File: apps/apache.go (hunk 5 above)
Comment (qingling128) [reply apps/apache.go:94]: It seems like the new regex above is not able to capture the 3 log formats in the comment:

```
		// Sample line 2.4: [Fri Sep 09 10:42:29.902022 2011] [core:error] [pid 35708:tid 4328636416] (13)Permission denied [client 72.15.99.187] File does not exist: /usr/local/apache2/htdocs/favicon.ico
		// Sample line 2.2: [Fri Sep 09 10:42:29.902022 2011] [error] [pid 35708:tid 4328636416] [client 72.15.99.187] File does not exist: /usr/local/apache2/htdocs/favicon.ico
		// Sample line 2.0: [Wed Oct 11 14:32:52 2000] [error] [client 127.0.0.1] client denied by server configuration: /export/home/live/ap/htdocs/test
```

dehaansa: It works properly from my understanding, and I just went and validated with regex101.com as well. What fields are you concerned about?

dehaansa:
> The tests only run for Debian 10 today. The change to extend it to more distros turned out to be a non-trivial effort. That would take some time to land.

Ah, okay, must have just been a strange fluke as they seem to have passed on re-run.

apps/apache.go
qingling128: Maybe I missed something, regex101 did not parse those 3 when I tried. For instance:

image

qingling128: Sync'd offline. It was the newline that accidentally failed the match. The proposed new regex actually work for all 4 cases.

qingling128 (approved):
LGTM

apps/apache.go
qingling128: The latest change LGTM.

PR #214, opened 2021-09-23
Add build image for Debian 11 (Bullseye)

ridwanmsharif:
This change adds the commands for a Debian 11 image. Uses Debian 10 as a
template.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,9 +16,26 @@
 # or DOCKER_BUILDKIT=1 docker build -o /tmp/out . --target=buster
 # Generated tarball(s) will end up in /tmp/out
 
-FROM debian:buster AS buster-build
+FROM debian:bullseye AS bullseye-build
 
-# TODO: Factor out the common code without rerunning apt-get on every build.
+# TODO: Factor out the common code without rerunning apt-get on every debian and
+# ubuntu build.
+
+RUN set -x; apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get -y install git systemd \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -16,6 +16,24 @@
 # or DOCKER_BUILDKIT=1 docker build -o /tmp/out . --target=buster
 # Generated tarball(s) will end up in /tmp/out
 
+FROM debian:bullseye AS bullseye-build
+
+# TODO: Factor out the common code without rerunning apt-get on every build.
```
jefferbrecht, line 21: Minor, if we're keeping this TODO around it's probably fine to just leave it on the first entry and remove the duplicated one from the buster section.

jefferbrecht (approved):

jefferbrecht:
Kokoro should be working now. Not sure if there's a way to re-trigger the actions without committing something.

Dockerfile
ridwanmsharif: Done.

Dockerfile
```diff
@@ -16,9 +16,9 @@
 # or DOCKER_BUILDKIT=1 docker build -o /tmp/out . --target=buster
 # Generated tarball(s) will end up in /tmp/out
 
-FROM debian:buster AS buster-build
+FROM debian:bullseye AS bullseye-build
 
-# TODO: Factor out the common code without rerunning apt-get on every build.
+# TODO: Factor out the common code without rerunning apt-get on every debian build.
```
quentinmit, line 21: I believe this comment applies to both Debian and Ubuntu

quentinmit (approved):

PR #215, opened 2021-09-23
Apache web server metrics

djaglowski:
This PR adds an Apache web server metric receiver, with the following caveats:
1. The corresponding kokoro setup code is included in #213. This PR should be merged after that one.
2. The [underlying receiver](https://github.com/observIQ/opentelemetry-components/tree/master/receiver/httpdreceiver) is not currently imported in the build of the operations-collector submodule.


docs/apache.md
```diff
@@ -0,0 +1,58 @@
...
+## Prerequisites
+
+The receiver requires that you enable the [mod_status module](https://httpd.apache.org/docs/2.4/mod/mod_status.html) on your Apache web server.
+
+ This requires adding the following lines to the server’s `httpd.conf` file:
+
+```
+<Location "/server-status">
+    SetHandler server-status
+    Require host example.com
```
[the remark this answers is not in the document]
djaglowski: Good catch. Updated.

docs/apache.md
```diff
@@ -1,3 +1,62 @@
...
+          - httpd_metrics
+```
+
+## Metrics
+
+The Ops Agent collects the following metrics from your Apache web servers.
+
+| Metric                                            | Data Type | Unit        | Labels              | Description |
+| ---                                               | ---       | ---         | ---                 | ---         | 
+| workload.googleapis.com/httpd.current_connections | gauge     | connections |      server_name        | The number of active connections currently attached to the HTTP server.  |
```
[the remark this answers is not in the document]
quentinmit: Again I'm not sure if we want to diverge from the upstream receiver like this. @jsuereth ?

Other metrics systems seem pretty consistent with `apache` and not `httpd`:
- [Legacy Google Agent](https://cloud.google.com/monitoring/api/metrics_agent#agent-apache)
- [Prometheus](https://github.com/Lusitaniae/apache_exporter#collectors)
- [Telegraf](https://github.com/influxdata/telegraf/blob/release-1.15/plugins/inputs/apache/README.md#measurements--fields)
- [Munin](https://gallery.munin-monitoring.org/plugins/munin/apache/)
- [SignalFX/collectd](https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-apache.html)

In fact I didn't run across any that call it `httpd`. AFAIK that name is only tightly bound to Apache on RedHat-derived distros. I'm curious what research you found that points the other way?

dehaansa:
Adding the agentmetrics processor to the agentmetrics pipeline fixed the service_version label issue, but ops agent tests are still failing due to missing `swap/bytes_used` metric. We believe this is due to the operations-collector pulling in this change https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/4854 in the upstream, and we're looking into if that would cause issues in the processor for the hostmetrics pipeline

docs/apache.md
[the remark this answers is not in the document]
jsuereth: commented elsewhere as well.  I think given prometheus + legacy Google agent usage here, we should likely not diverge unless we felt this caused user friction.

Specficially, we don't want users of Prometheus on GKE (or Legacy Google Agent on GCE) to find completely different metric names vs. what they see from Ops Agent.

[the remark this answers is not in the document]
quentinmit: @jsuereth I think you parsed this backwards; OT's upstream receiver is the only one of that set I listed above that uses `httpd` as a prefix and not `apache`. So making things consistent for Prometheus users *does* mean diverging from the OT receiver.

quentinmit:
> Adding the agentmetrics processor to the agentmetrics pipeline fixed the service_version label issue, but ops agent tests are still failing due to missing `swap/bytes_used` metric. We believe this is due to the operations-collector pulling in this change [open-telemetry/opentelemetry-collector-contrib#4854](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/4854) in the upstream, and we're looking into if that would cause issues in the processor for the hostmetrics pipeline

Please split these changes into a separate PR since they're isolated from the Apache receiver.

docs/apache.md
```diff
@@ -0,0 +1,58 @@
...
+| workload.googleapis.com/httpd.uptime              | sum       | s           |     server_name     | The amount of time that the server has been running in seconds.  |
```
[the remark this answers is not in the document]
qingling128: Addressed by https://github.com/GoogleCloudPlatform/ops-agent/pull/215/files#diff-3934432bf4db161a04479d4460bfbc28eeeab791cbbb49cc166f89e376e0633fR51. Resolving.

docs/apache.md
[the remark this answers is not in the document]
jsuereth: I'm suggesting ops-agent should use `apache` and we should think about pushing that allthe way into Otel at some point.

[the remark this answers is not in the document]
qingling128: The PR has been updated to rename `httpd` to `apache` in the prefix: https://github.com/GoogleCloudPlatform/ops-agent/pull/215/files#diff-3934432bf4db161a04479d4460bfbc28eeeab791cbbb49cc166f89e376e0633fR58. Confirmed offline that `apache` prefix is the way to go for us.

qingling128:
Update: After some investigation, the `ops agent integration test` failure is now being addressed from the internal test suite. Nothing needs to change in this PR for that.

docs/apache.md
```diff
@@ -1,3 +1,61 @@
...
+| `type`                | required                  | Must be `apache`. |
+| `server_status_url`     | `http://localhost:8080/server-status?auto` | The url exposed by the `mod_status` module. |
+| `collection_interval` | `60s`                     | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
+
+Example Configuration:
+
+```yaml
+metrics:
+  receivers:
+    httpd_metrics:
```
qingling128, line 36:
```suggestion
    apache_metrics:
```

docs/apache.md
```diff
@@ -1,3 +1,61 @@
...
+      type: httpd
+      server_status_url: http://localhost:8080/server-status?auto
+      collection_interval: 30s
+  service:
+    pipelines:
+      httpd_pipeline:
```
qingling128, line 42:
```suggestion
      apache_pipeline:
```

docs/apache.md
```diff
@@ -1,3 +1,61 @@
...
+        receivers:
+          - httpd_metrics
```
qingling128, line 44:
```suggestion
          - apache_metrics
```

docs/apache.md
```diff
@@ -1,3 +1,61 @@
...
+| `server_status_url`     | `http://localhost:8080/server-status?auto` | The url exposed by the `mod_status` module. |
+| `collection_interval` | `60s`                     | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
+
+Example Configuration:
+
+```yaml
+metrics:
+  receivers:
+    httpd_metrics:
+      type: httpd
```
qingling128, line 37:
```suggestion
      type: apache
```

qingling128 (approved):
LGTM

jsuereth (approved):

confgenerator/agentmetrics.go
```diff
@@ -41,6 +41,10 @@ func (r MetricsReceiverAgent) Pipeline() otel.Pipeline {
 			},
 		},
 		Processors: []otel.Component{
+			{
```
quentinmit, line 44: Please split this change (I guess including the submodule update and label change) into a separate PR so we can review and test it separately, since it changes so many files.

But also... is there something different that the new code in the agentmetrics processor is doing that's different from the `otel.AggregateLabels` calls you added in `hostmetrics.go`? Can it be handled with a smaller diff by just adding calls to `otel.AggregateLabels` in the `MetricsTransform` below?

dehaansa: Sorry, I missed that this was still in there. The AggregateLabels calls should actually be irrelevant, but I believe the agentmetrics processor will still be necessary for resolving the issue that we attempted to resolve back in https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/59#discussion_r733070787 the 0.37 update PR. 

There were two different issues with ops agent integration tests that were encountered during this PR, the swap/bytes_used that I was attempting to resolve with AggregateLabels, and the service_version label issue that was being resolved with the agentmetrics processor added to the agentmetrics pipeline.

dehaansa: https://github.com/GoogleCloudPlatform/ops-agent/pull/261

jsuereth: #261 is merged

qingling128 (approved):
LGTM

PR #218, opened 2021-09-28
MySQL metrics

Mrod1598:
This PR adds an MySQL metric receiver, with the following caveats:

1. ~~The corresponding kokoro setup code is included in a, soon to come, log receiver PR for MySQL. This PR should be merged after that one.~~
2. ~~The [underlying receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/5446) is not currently imported in the build of the operations-collector submodule.~~


apps/mysql.go
```diff
@@ -15,13 +15,65 @@
...
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"omitempty,url"`
+
+	Password string `yaml:"password" validate:"omitempty"`
+
+	Username string `yaml:"username" validate:"omitempty"`
+}
+
+const defaultEndpoint = "localhost:3306"
```
quentinmit, line 38: This receiver is another example where you're using the `url` validator and the default value is not a URL. We need to be very clear and explicit with how users should configure this - are there valid URLs that can go here? What does a Unix socket path look like? etc.

quentinmit, line 38: I do not think we should default to a TCP connection. Especially on distros that automatically configure Unix socket authentication, we should make sure we are using a Unix socket so the zero-config setup works out of the box.

Mrod1598: So this is going to be `host:port` as well. I've made an upstream PR to update it to add an additional parameter to enable different protocols.

Mrod1598: I'll update this to detect the default socket file from `/var/run/mysqld/mysql.sock` once the upstream PR gets Merged.

qingling128: Is there a link to the upstream PR?

quentinmit: I don't think we should merge this without Unix socket support

Mrod1598: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6138/files

qingling128: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6138 was merged.

qingling128: open-telemetry/opentelemetry-collector-contrib#6138 was merged.

qingling128:
There seem to be some conflict with HEAD. Can we resolve that?

qingling128:
Current status: waiting for https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/63 (Update Collector & Receivers to 0.39) to be merged first.

apps/mysql.go
```diff
@@ -15,13 +15,75 @@
...
 
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
+
+	"fmt"
+	"strings"
 )
 
+type MetricsReceiverMySQL struct {
```
quentinmit, line 28: Nit: Is this what the linter likes? I'm surprised it doesn't prefer `MetricsReceiverMySql`.

apps/mysql.go
```diff
@@ -15,13 +15,75 @@
...
+const defaultMySQLTCPEndpoint = "localhost:3306"
+const defaultMySQLUnixEndpoint = "/var/run/mysqld/mysql.sock"
+
+func (r MetricsReceiverMySQL) Type() string {
+	return "mysql"
+}
+
+func (r MetricsReceiverMySQL) Pipelines() []otel.Pipeline {
+	if r.Endpoint == "" {
+		if _, err := os.Stat(defaultMySQLUnixEndpoint); err != nil {
```
quentinmit, line 49: This means that the config generation is non-hermetic and can't be unit tested. Do we know that there are environments where we can't default to the unix socket? (What are they? Windows?) Even if those environments do require TCP, they would also presumably require a password, so I think we can just unconditionally default to TCP.

Also see above; this can be racy with MySQL startup.

apps/mysql.go
```diff
@@ -15,13 +15,75 @@
...
 
+type MetricsReceiverMySQL struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"omitempty,hostname_port|file"`
+
+	Password  string `yaml:"password" validate:"omitempty"`
+	Username  string `yaml:"username" validate:"omitempty"`
```
quentinmit, line 36: We don't seem to have a default for username; does an empty username get turned into `root` somewhere downstream?

apps/mysql.go
```diff
@@ -15,13 +15,75 @@
...
+			otel.NormalizeSums(),
+			otel.MetricsTransform(
+				otel.AddPrefix("workload.googleapis.com"),
+			),
+		},
+	}}
+}
+
+func init() {
+	confgenerator.MetricsReceiverTypes.RegisterType(func() confgenerator.Component { return &MetricsReceiverMySQL{} })
```
quentinmit, line 78: Style nit: I would just use two init() functions and keep the other one at the end of the file.

apps/mysql.go
```diff
@@ -15,13 +15,75 @@
...
+	"fmt"
+	"strings"
 )
 
+type MetricsReceiverMySQL struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"omitempty,hostname_port|file"`
```
quentinmit, line 33: You can't use the `file` validator because that requires the file to exist at validation time. (Since Ops Agent might start before MySQL, we can't guarantee the MySQL socket exists at config generation time.)

apps/mysql.go
```diff
@@ -15,13 +15,75 @@
...
+
+	Password  string `yaml:"password" validate:"omitempty"`
+	Username  string `yaml:"username" validate:"omitempty"`
+	Transport string `yaml:"transport" validate:"omitempty"`
```
quentinmit, line 37: What are the valid values for `transport`? They should be validated (and documented in the docs).

MySQL usually calls this parameter [`protocol`](http://download.nust.na/pub6/mysql/doc/refman/5.4/en/connecting.html#option_general_protocol) and the valid values are `TCP`, `SOCKET`, `PIPE`, and `MEMORY`, but also client libraries usually infer the correct value so users don't normally need to specify it. (Can we do that?)

docs/mysql.md
```diff
@@ -1,3 +1,75 @@
...
+| `transport`           |                           | Defines the network to use for connecting to the server. |
+
+
+Example Configuration:
+
+```yaml
+metrics:
+  receivers:
+    mysql_metrics:
+      type: mysql
```
quentinmit, line 38: This example configuration implies that the user needs to set all of these fields; can we have two examples, the first with none of the fields set, and the second with a description like "TCP connection with a username and password"?

docs/mysql.md
```diff
@@ -1,3 +1,75 @@
+# `mysql` Metrics Receiver
+
+The mysql receiver can retrieve stats from your mysql instance by connecting as a monitoring user.
+
+
+## Prerequisites
+
+It is recommended that you create a dedicated monitoring user. The monitoring user must be granted the following permissions:
```
quentinmit, line 8: Can you add before this something like:

```md
The `mysql` receiver defaults to connecting to a local MySQL server using a Unix socket and Unix authentication as the `root` user.
```

to make the default behavior clear?

docs/mysql.md
```diff
@@ -1,3 +1,75 @@
...
+## Configuration
+
+Following the guide for [Configuring the Ops Agent](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/configuration#file-location), add the required elements for your mysql configuration.
+
+To configure a receiver for your mysql metrics, specify the following fields:
+
+| Field                 | Default                   | Description |
+| ---                   | ---                       | ---         |
+| `type`                | required                  | Must be `mysql`. |
+| `endpoint`            | `localhost:3306`          | The hostname:port or socket path used by mysql |
```
quentinmit, line 25: This is not the correct default

docs/mysql.md
```diff
@@ -1,3 +1,75 @@
...
+| `collection_interval` | `60s`                     | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
+| `username`            |                           | The username used to connect to the server. |
```
quentinmit, line 27: `root`?

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_mysql
```diff
@@ -16,7 +16,18 @@ sudo tee /etc/google-cloud-ops-agent/config.yaml > /dev/null << EOF
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+metrics:
+  receivers:
+    mysql_metrics:
+      type: mysql
+      collection_interval: 10s
+      username: root
```
quentinmit, line 24: We should be testing the defaults (i.e. the unit test should not specify `username`, `password`, or `collection_interval`).

apps/mysql.go
Mrod1598: bit confused on the desired behavior you're looking for. Are you saying we should just default to Unix sockets without trying to os.Stat it and just not bother with TCP defaults?

quentinmit: I'm asking if that works in all the situations where the current code does. I think it does but I'm not sure about any edge cases.

apps/mysql.go
Mrod1598: Otel is pushing to try and use [`config.NetAddr`](https://pkg.go.dev/go.opentelemetry.io/collector/config/confignet) struct for connections, which uses Transport as the wording here. but I'll change it to protocol for the ops-agent level.

As for defaulting for the protocols and what those defaults should be, I've tested using `unix` and `tcp` for the values of transport since it's based off of [net.Dial](https://pkg.go.dev/net#Dial) for the [client](https://github.com/go-sql-driver/mysql) and those worked. Since the protocols don't seem to line up with the shell of MySql, perhaps the ones outlined by net.Dial are the ones we should be using.

apps/mysql.go
Mrod1598: From what I'm seeing, yes I think that'll work. There seems there was a way to set no password for mysql prior to 5.7 for windows machines but that it was removed.

apps/mysql.go
Mrod1598: it doesn't have a preferences it seems, so I've updated to your suggestion.

apps/mysql.go
Mrod1598: moved to two seperate funcs.

apps/mysql.go
Mrod1598: I'm not seeing a good replacement for it. and if we remove file, we'll have to remove `hostname_port` or it'll always use that. Any suggestions?

apps/mysql.go
Mrod1598: it does now.

docs/mysql.md
Mrod1598: added

docs/mysql.md
Mrod1598: added

docs/mysql.md
Mrod1598: updated.

docs/mysql.md
Mrod1598: updated

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_mysql
Mrod1598: updated

apps/mysql.go
Mrod1598: @quentinmit So, we can split the field into two separate mutually exclusive fields, one for unix paths that we don't validate, and one for tcp that we validate with `hostname_port`. This would also solve the transport issue comment so we can automatically default to `tcp` or `unix` based on what field is present and require that only one of the two fields are present.

quentinmit: How about `hostname_port|startswith=/` and select a transport based on the leading `/`?

apps/mysql.go
quentinmit: If the values are `config.NetAddr` transports, we can keep the field called `transport` - that way people won't expect to be able to use mysql protocol names. (Though if we can infer it from whether the address starts with `/` or not that would be even better. Not sure how we would support the Windows protocols but I'd guess the underlying library doesn't either.)

docs/mysql.md
quentinmit: Thanks. Why do we recommend a dedicated monitoring user? I think I would phrase it as "If you want to create a dedicated monitoring user, the monitoring user must be granted ..."

quentinmit (approved):
LGTM if the comments are all resolved. Thanks!

docs/mysql.md
Mrod1598: The reasoning behind it is so they wouldn't be giving root access to everything with the default config. But I'll update the wording if you think its better?

apps/mysql.go
Mrod1598: that works

apps/mysql.go
Mrod1598: ok then we should be good with the update to the validation I just pushed.

Mrod1598:
@quentinmit do you mind resolving the comments you don't think we need anymore discussion on just so we're on the same page?

apps/mysql.go
qingling128: Looks like username and password are no longe required. https://github.com/GoogleCloudPlatform/ops-agent/pull/218/commits/4a4a8195aa05fd30f5b6f598783022e4307fcaee and https://github.com/GoogleCloudPlatform/ops-agent/pull/218/commits/d9f213f6764b6dca1e2ea9f456c31be1a089657e also removed the related username and password settings. Resolving this.

apps/mysql.go
qingling128: Confirmed that this has been addressed in https://github.com/GoogleCloudPlatform/ops-agent/commit/9858567c7473a8286d3628d9024cb12c5dfe6b26. Resolving this thread.

apps/mysql.go
qingling128: Confirmed that this has been changed in https://github.com/GoogleCloudPlatform/ops-agent/commit/8e3667ab3e1680c58a907b2525bc4df9ba3120b7. Resolving this thread.

docs/mysql.md
qingling128: Resolving this thread as it has been addressed in https://github.com/GoogleCloudPlatform/ops-agent/commit/c55e07c2a4a26fe674f3feb49536b32e7521642f.

docs/mysql.md
qingling128: Resolving this as it has been addressed by https://github.com/GoogleCloudPlatform/ops-agent/commit/c55e07c2a4a26fe674f3feb49536b32e7521642f.

docs/mysql.md
qingling128: Resolving this as it has been addressed by https://github.com/GoogleCloudPlatform/ops-agent/commit/c55e07c2a4a26fe674f3feb49536b32e7521642f.

apps/mysql.go
qingling128: Looks like this has been resolved.

apps/mysql.go
qingling128: Can we also add a `oneof` validation to https://github.com/GoogleCloudPlatform/ops-agent/blame/78981928d79a29f3f33da56b16fd577f8604d5cd/apps/mysql.go#L35? Similar to https://github.com/GoogleCloudPlatform/ops-agent/blob/bd6630930fa4a1d672e522963bc8d9badcb30304/confgenerator/logging_receivers.go#L118. Plus a related unit test for invalid `transport` value.

docs/mysql.md
qingling128: @quentinmit WDYT?

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_mysql
qingling128: Looks like `collection_interval` was removed in https://github.com/GoogleCloudPlatform/ops-agent/pull/218/commits/d9f213f6764b6dca1e2ea9f456c31be1a089657e (which addressed this thread) but added back in https://github.com/GoogleCloudPlatform/ops-agent/commit/ab5db88f88f20c97eeef19093f4ac089793517e1.

Any special reason behind it? @Mrod1598

qingling128:
Remaining open threads. PTAL

@quentinmit 
* https://github.com/GoogleCloudPlatform/ops-agent/pull/218#discussion_r759182997

@Mrod1598 
* https://github.com/GoogleCloudPlatform/ops-agent/pull/218#discussion_r759184208
* https://github.com/GoogleCloudPlatform/ops-agent/pull/218/files/8620d7199eb6a617dc369657eade36956d49c609#r759181253

apps/mysql.go
Mrod1598: Added.

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_mysql
Mrod1598: becauase it was saying it was required without it

apps/mysql.go
qingling128: Thanks!

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_mysql
qingling128: Looks like `collection_interval` is indeed defined as `required`. It should be optional though. Since it's not directly related to this PR, I'll file a separate bug.

docs/mysql.md
quentinmit: Ops agent is currently running as root on the machine, so it already has de facto root access to the local MySQL server. So I don't think there's much risk mitigation from telling users to create another user (and that would mean using passwords, which are less secure than local authentication).

qingling128 (approved):
LGTM module the integration tests pass, and the markdown file is modified as discussed in https://github.com/GoogleCloudPlatform/ops-agent/pull/218#discussion_r759182997.

qingling128:
https://github.com/GoogleCloudPlatform/ops-agent/pull/306 was just merged. This probably needs rebasing to get the tests pass.

dehaansa:
More windows failures & a sles-12-sp4 failure (TestSystemLogByDefault/sles-12-sp4-sap) on ops agent tests, but 3rd party tests looked good.

docs/mysql.md
qingling128: The sentence in question has been removed.

qingling128 (approved):
The new change to remove `transport` config (because it can be inferred by `endpoint`, and this fixes the config UX discrepancy with Memcached) LGTM.

qingling128:
Merging this since the `third party apps integration test` passed before the rebasing (merge conflict in the golden file is pretty trivial and passed unit tests), and the failures in `ops agent integration test` are all related to sles 15 upstream repo reliability issues.

PR #219, opened 2021-09-28
Mysql log parsers

dehaansa:
Add a log receiver for mysql error logs, general & slow query logs. 

Error:
image

Slow:
image

General: (Regex has been updated to hide \s+ before query, will update screenshot.
image


integration_test/third_party_apps_data/applications/mysql/metric_name.txt
```diff
@@ -1 +1 @@
-agent.googleapis.com/mysql/command_count
+agent.googleapis.com/cpu/utilization
```
qingling128, line 1: [Optional] Is the `cpu/utilization` metrics here just as a place holder before we add metrics support? We can skip the metrics checking by leaving this file empty.

apps/mysql.go
```diff
@@ -0,0 +1,87 @@
...
+	return "mysql_error"
+}
+
+func (p LoggingProcessorMysqlError) Components(tag string, uid string) []fluentbit.Component {
+	c := confgenerator.LoggingProcessorParseRegex{
+		// Format documented: https://dev.mysql.com/doc/refman/8.0/en/error-log-format.html
+		// Older versions of mysql should have the same general format, but may not have tid, error_code, subsystem.
+		// Sample Line: 2020-08-06T14:25:02.936146Z 0 [Warning] [MY-010068] [Server] CA certificate /var/mysql/sslinfo/cacert.pem is self signed.
+		// Sample Line: 2020-08-06T14:25:03.109022Z 5 [Note] Event Scheduler: scheduler thread started with id 5
+		Regex: `^(?<time>\d{4}-\d{2}-\d{2}(?:T|\s)\d{2}:\d{2}:\d{2}.\d+(?:Z|[+-]\d{2}:\d{2})?)(?:\s+(?<tid>\d+))?(?:\s+\[(?<level>[^\]]+)])?(?:\s+\[(?<error_code>[^\]]+)])?(?:\s+\[(?<subsystem>[^\]]+)])?\s+(?<message>.*)$`,
```
qingling128, line 38: [Optional] The timestamp regex here seems a bit verbose. Alternatively we could do something simpler (e.g. something like https://github.com/fluent/fluent-bit/blob/0660382260120ae069dcc6dd15b7fd28659d1a23/conf/parsers_extra.conf#L22) since the TimeFormat below already specifies the detailed format.

qingling128 (approved):
LGTM

qingling128:
LGTM. Not a hard requirement, but it would be great to include Mysql logging support (error logs / general logs / slow query logs) in the same Ops Agent release.

integration_test/third_party_apps_data/applications/mysql/metric_name.txt
dehaansa: Yup, I was intending to come back and fix that once I had validated the empty file in Apache PR, and forgot, thanks!

qingling128:
Note: We are planning to release Ops Agent 2.3.1 today (Sept 29). Please do not merge this PR before https://github.com/GoogleCloudPlatform/ops-agent/releases/tag/2.3.1 is out.

apps/mysql.go
quentinmit, line 38: Why do we use `error_code` here but `errorCode` in Apache? Does this match the field name conventions from fluentd? If you're trying to match MySQL's JSON logging format, they use `err_code` and only contain the int, not the whole string with `MY-`.

apps/mysql.go
```diff
@@ -0,0 +1,87 @@
...
+
+package apps
+
+import (
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
+)
+
```
quentinmit, line 23: Nit: Leave a TODO comment listing the other types of logs we plan to add?

apps/mysql.go
```diff
@@ -0,0 +1,87 @@
...
+	confgenerator.ConfigComponent `yaml:",inline"`
+}
+
+func (LoggingProcessorMysqlError) Type() string {
+	return "mysql_error"
+}
+
+func (p LoggingProcessorMysqlError) Components(tag string, uid string) []fluentbit.Component {
+	c := confgenerator.LoggingProcessorParseRegex{
+		// Format documented: https://dev.mysql.com/doc/refman/8.0/en/error-log-format.html
```
quentinmit, line 34: It looks like the error log format for MySQL changed drastically in 8.x. MySQL 5.x and apparently MariaDB use a wildly different format:

https://mariadb.com/kb/en/error-log/#format

I'm not sure how to handle this but in general I think users expect MySQL and MariaDB to be interchangeable.  Maybe it's sufficient to call out in the docs "This receiver only supports Oracle MySQL >= 8.0 and does not support MySQL 5.x or MariaDB"?

apps/mysql.go
jsuereth: We should aim for consistency, barring some extenuating reason to diverge (e.g. user-expectation, new conventions in upstream etc.).   Also curious what the user-expectation here is and if we should unify w/ what was done for Apache / nginx.

apps/mysql.go
jsuereth: Does it make sense to have "supported version" as  flag or create different receivers for the different versions (with latest being the default) ?

I know we had internal discussion on this, @quentinmit @qingling128 what's the recent thinking here?

quentinmit: I believe the recent thinking is "add a version field if needed, but try to avoid it". e.g. if we can write a regex that easily supports both, or two separate regexes that both apply (I think that's possible here?), or declare that we don't support the old version, that's preferable.

Oracle MySQL 8.0 is only 3 years old. RHEL 7 includes MariaDB (and presumably therefore the old log format). RHEL 8 includes both MariaDB and MySQL 8.0.

integration_test/third_party_apps_data/applications/mysql/metric_name.txt
qingling128: Thanks for fixing it!

apps/mysql.go
qingling128: `Nginx` set the error code in the httpRequest.status field: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest.

`Apache`, as Quentin mentioned, sets an `errorCode` field: https://github.com/GoogleCloudPlatform/ops-agent/pull/213/files/1dff1e9ab954dd85135d108912835c6ce8dcc10a#diff-3934432bf4db161a04479d4460bfbc28eeeab791cbbb49cc166f89e376e0633fR94

The special log fields in the other section of logging config documentation uses camel case:
https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/configuration#special-fields

quentinmit: Nginx error logs do not parse the error code out into a separate field. They just treat it as part of the message. (I don't think we should extrapolate that here, since MySQL error codes are more well-defined and expected on every log line.)

Nginx access logs fill in `httpRequest.status`, but that's a completely different kind of code, not an error code, so I don't think that applies either.

dehaansa: This was an oversight on my part, I was intending to be consistent with the other log parsing examples and stick with camel case. Updating.

apps/mysql.go
dehaansa: From what I understand of Fluent Bit we can apply two Parsers to the same Filter and that should handle either log format. It'll require a bit of confgenerator updating but shouldn't be too bad. It will attempt to apply each regex to each line so the sooner they fail (more specific time format) the better for performance.

apps/mysql.go
qingling128: Thanks for adding it!

apps/mysql.go
qingling128: Yes Fluent Bit support multiple parsers for different formats:

image

apps/mysql.go
qingling128: This has been updated. Resolving the thread.

qingling128: Actually, can we update the screenshot in the PR description as well? (Not a blocker for the PR)

apps/mysql.go
dehaansa: Once https://github.com/GoogleCloudPlatform/ops-agent/pull/225 is approved/merged an additional parser can be added to support multiple error file formats.

dehaansa: Yup! Working on that & query log parsers, currently on another branch in case this one was merged ahead, but otherwise intent to bring into this PR.

apps/mysql.go
```diff
@@ -0,0 +1,268 @@
...
...{2}T\d{2}:\d{2}:\d{2}.\d+Z)\s# User@Host:\s+(?<user>[^\[]*)\[(?<database>[^\]]*)\]\s+@\s+((?<host>[^\s]+)\s)?\[(?:(?<ipAddress>[\w\d\.:]+)?)\]\s+Id:\s+(?<tid>\d+)\s+#\s+Query_time:\s+(?<queryTime>[\d\.]+)\s+Lock_time:\s+(?<lockTime>[\d\.]+)\s+Rows_sent:\s+(?<rowsSent>\d+)\s+Rows_examined:\s(?<rowsExamined>\d+)(?:\s+Thread_id:\s+\d+)?(?:\s+Errno:\s(?<errorNumber>\d+))?(?:\s+Killed:\s(?<killed>\d+))?(?:\s+Bytes_received:\s(?<bytesReceived>\d+))?(?:\s+Bytes_sent:\s(?<bytesSent>\d+))?(?:\s+Read_first:\s(?<readFirst>\d+))?(?:\s+Read_last:\s(?<readLast>\d+))?(?:\s+Read_key:\s(?<readKey>\d+))?(?:\s+Read_next:\s(?<readNext>\d+))?(?:\s+Read_prev:\s(?<readPrev>\d+))?(?:\s+Read_rnd:\s(?<readRnd>\d+))?(?:\s+Read_rnd_next:\s(?<readRndNext>\d+))?(?:\s+Sort_merge_passes:\s(?<sortMergePasses>\d+))?(?:\s+Sort_range_count:\s(?<sortRangeCount>\d+))?(?:\s+Sort_rows:\s(?<sortRows>\d+))?(?:\s+Sort_scan_count:\s(?<sortScanCount>\d+))?(?:\s+Created_tmp_disk_tables:\s(?<createdTmpDiskTables>\d+))?(?:\s+Created_tmp_tables:\s(?<createdTmpTables>\d+))?(?:\s+Start:\s(?<startTime>\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d+Z))?(?:\s+End:\s(?<endTime>\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d+Z))?\s+(?<message>[\s\S]+)`,
```
quentinmit, line 157: This regex is bordering on unreadable. Do you think you could build it up with code instead? (e.g. a for loop around the list of fields?)

If you don't think that's worth the effort it's also fine with me if you don't change this. I'm just going to have to have faith that the integration tests work.

dehaansa: I'd certainly say it's well beyond unreadable. I can look at building the majority of it from a set of fields, and see if that ends up looking cleaner or not.

qingling128: Looks like this has been addressed by https://github.com/GoogleCloudPlatform/ops-agent/pull/219/commits/fbba49c764cefb96223a2d4ec97ece59511db3f3

sophieyfang:
When logs are spread across chunks, it's unsafe to use tail + multiline parser as the filter. It needs to use the multiline parser as part of the input parser. I've also tested such logs when across chunks won't be concatenated successfully

image

PR #220, opened 2021-09-29
Cassandra metrics receiver

dehaansa:
Cassandra metrics receiver first draft. As Cassandra is JVM/JMX based, there are no additional open telemetry receivers required for this integration. 

Metrics collected are listed here: https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/docs/target-systems/cassandra.md

One possibly contentious aspect of this PR is that the Cassandra receiver can be configured to also collect JVM metrics via a configuration parameter (collect_jvm_metrics). I chose to implement this rather than always requiring a user to create two receivers as the JMX receiver supports collecting multiple "target" scripts in one JVM, and creating two receivers would create two JVMs and have considerable memory overhead. We could abstract this away from the user and combine JMX receivers when the configurations matched, but I didn't try to implement that kind of logic in this PR.

TODO
- Finish documenting metrics
- Validate CI deployments in kokoro
- Implement any feedback from scope document

### DIFF
--- a/apps/cassandra.go
+++ b/apps/cassandra.go
@@ -15,6 +15,7 @@
 package apps
 
 import (
+	"fmt"
 	"log"
 
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
@@ -29,6 +30,8 @@ type MetricsReceiverCassandra struct {
 	Endpoint string `yaml:"endpoint" validate:"omitempty,url"`
 	Username string `yaml:"username"`
 	Password string `yaml:"password"`
+
+	CollectJVMMetics *bool `yaml:"collect_jvm_metrics"`
 }
 
 const defaultCassandraEndpoint = "localhost:9999"
@@ -47,8 +50,13 @@ func (r MetricsReceiverCassandra) Pipelines() []otel.Pipeline {
 		log.Printf(`Encountered an error discovering the location of the JMX Metrics Exporter, %v`, err)
 	}
 
+	targetSystem := "cassandra"
+	if r.CollectJVMMetics == nil || *r.CollectJVMMetics {
+		targetSystem = fmt.Sprintf("%s,%s", targetSystem, "jvm")
+	}
+
 	config := map[string]interface{}{
-		"target_system":       "cassandra,jvm",
+		"target_system":       targetSystem,
 		"collection_interval": r.CollectionIntervalString(),
 		"endpoint":            r.Endpoint,
 		"jar_path":            jarPath,
--- a/apps/cassandra.go
+++ b/apps/cassandra.go
@@ -1,0 +1,81 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apps
+
+import (
+	"log"
+
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
+)
+
+type MetricsReceiverCassandra struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"omitempty,url"`
+	Username string `yaml:"username"`
+	Password string `yaml:"password"`
+}
+
+const defaultCassandraEndpoint = "localhost:9999"
+
+func (r MetricsReceiverCassandra) Type() string {
+	return "cassandra"
+}
+
+func (r MetricsReceiverCassandra) Pipelines() []otel.Pipeline {
+	if r.Endpoint == "" {
+		r.Endpoint = defaultCassandraEndpoint
+	}
+
+	jarPath, err := FindJarPath()
+	if err != nil {
+		log.Printf(`Encountered an error discovering the location of the JMX Metrics Exporter, %v`, err)
+	}
+
+	config := map[string]interface{}{
+		"target_system":       "cassandra,jvm",
+		"collection_interval": r.CollectionIntervalString(),
+		"endpoint":            r.Endpoint,
+		"jar_path":            jarPath,
+	}
+
+	// Only set the username & password fields if provided
+	if r.Username != "" {
+		config["username"] = r.Username
+	}
+	if r.Password != "" {
+		config["password"] = r.Password
+	}
+
+	return []otel.Pipeline{{
+		Receiver: otel.Component{
+			Type:   "jmx",
+			Config: config,
+		},
+		Processors: []otel.Component{
+			otel.NormalizeSums(),
+			otel.MetricsTransform(
+				otel.AddPrefix("workload.googleapis.com"),
+			),
+		},
+	}}
+}
+
+func init() {
+	confgenerator.MetricsReceiverTypes.RegisterType(func() confgenerator.Component { return &MetricsReceiverCassandra{} })
+}
--- a/confgenerator/testdata/invalid/linux/metrics-receiver_unsupported_type/golden_error
+++ b/confgenerator/testdata/invalid/linux/metrics-receiver_unsupported_type/golden_error
@@ -1,1 +1,1 @@
-metrics receiver with type "unsupported_type" is not supported. Supported metrics receiver types: [hostmetrics, jvm, nginx].
+metrics receiver with type "unsupported_type" is not supported. Supported metrics receiver types: [cassandra, hostmetrics, jvm, nginx].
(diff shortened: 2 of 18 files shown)


docs/cassandra.md
```diff
@@ -0,0 +1,64 @@
...
+| workload.googleapis.com/cassandra.client.request.range_slice.unavailable.count  | sum       | 1           |        | Number of token range read request unavailable exceptions encountered |
+| workload.googleapis.com/cassandra.compaction.tasks.completed                    | sum       | 1           |        | Number of completed compactions since server start |
+| workload.googleapis.com/cassandra.compaction.tasks.pending                      | gauge     | 1           |        | Estimated number of compactions remaining to perform |
+| workload.googleapis.com/cassandra.storage.load.count                            | sum       | bytes       |        | Size of the on disk data size this node manages |
+| workload.googleapis.com/cassandra.storage.total_hints.count                     | sum       | 1           |        | Number of hint messages written to this node since start |
+| workload.googleapis.com/cassandra.storage.total_hints.in_progress.count         | sum       | 1           |        | Number of hints attempting to be sent currently |
```
dehaansa, line 64: This metric is collected as a Sum/Counter (even in latest release of the jmx-metrics java-contrib component) and is documented as such in [cassandra's docs](https://cassandra.apache.org/doc/latest/cassandra/operating/metrics.html) (search for TotalHintsInProgress) but it really seems like it would be a Gauge. I don't have a system with load on it that is causing this metric to be non-zero but I'm curious if it's correct in the scripts & docs.

docs/cassandra.md
```diff
@@ -0,0 +1,64 @@
...
+| workload.googleapis.com/cassandra.client.request.read.latency.count             | sum       | µs          |        | Total standard read request latency |
+| workload.googleapis.com/cassandra.client.request.read.latency.max               | gauge     | µs          |        | Maximum standard read request latency |
+| workload.googleapis.com/cassandra.client.request.read.timeout.count             | sum       | 1           |        | Number of standard read request timeouts encountered |
+| workload.googleapis.com/cassandra.client.request.read.unavailable.count         | sum       | 1           |        | Number of standard read request unavailable exceptions encountered |
+| workload.googleapis.com/cassandra.client.request.write.latency.50p              | gauge     | µs          |        | Standard write request latency - 50th percentile |
+| workload.googleapis.com/cassandra.client.request.write.latency.99p              | gauge     | µs          |        | Standard write request latency - 99th percentile |
+| workload.googleapis.com/cassandra.client.request.write.latency.count            | sum       | µs          |        | Total standard write request latency |
```
quentinmit, line 50: It's really weird for the write request count to have `.latency.` in the name, but I suppose that's an unfortunate consequence of how these are exposed.

apps/cassandra.go
```diff
@@ -0,0 +1,89 @@
...
+type MetricsReceiverCassandra struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"omitempty,url"`
+	Username string `yaml:"username"`
+	Password string `yaml:"password"`
+
+	CollectJVMMetics *bool `yaml:"collect_jvm_metrics"`
```
quentinmit, line 34: I'm not a big fan of collecting metrics that collide with the JVM receiver by default when users aren't expecting them, at least until we resolve the separate issues around how to handle monitoring for multiple processes.

I'd like to suggest a third alternative to consider: Collect the JVM metrics, but consider prefixing them with `cassandra.` I think that's a better UX now, but it may conflict with what we want to do around extra resource attributes in the future.

quentinmit (commented):
LGTM code-wise but I'm not marking approved because of the open business logic questions.

docs/cassandra.md
dehaansa: I believe this metric represents the total amount of latency experienced over all requests, which is different from the count of write requests. Agreed that the naming is not great.

quentinmit: If this is the total latency, which metric contains a request count?

(If there isn't one, I don't know what you can do with total latency if there isn't a count to divide it with.)

apps/cassandra.go
jsuereth: The decision to have cassandra report JVM metrics is inline with how the OTel JMX receiver works and user expectations around cassandra monitoring (where JVM GC is a big portion of it).

When you mix user expectations (monitoring Cassandra needs JVM GC monitoring), concerns around having multiple JVM GC metrics with different names, etc.  I think having cassandra default to collecting JVM metrics is the right thing.   The only concern I have is whether we add a label to them denoting "this was cassandra".  That shouldn't be necessary given we'll (eventually) have a label on all JVM metrics denoting which resource they came from.

TL;DR:   I'm not a fan of splitting JVM from Cassandra.  This matches user expectation and has precedence in other solutions.

docs/cassandra.md
jsuereth: I took a look into the code.  Cassandra is still using codahale's metric library (Is that dropwizard?).    They're using an underlying counter for this metric in their own code.  I'm not sure the behavior of that library, so this is a good open question.   It's entirely possibly they're sending a monotonic sum expecting users to calculate rate afterwards.

confgenerator/testdata/invalid/linux/metrics-receiver_unsupported_type/golden_error
```diff
@@ -1 +1 @@
-metrics receiver with type "unsupported_type" is not supported. Supported metrics receiver types: [hostmetrics, jvm, nginx].
```
qingling128, line 1: Any Cassandra specific invalid config input we want to test against?

One potential item is that the current implementation does not support more than one JVM based receivers as discussed in https://github.com/GoogleCloudPlatform/ops-agent/pull/220/files#r724704460. In other words, at most one metrics receiver with type "cassandra" or "jvm" is allowed. Adding this validation has its pros and cons though:

* Pros: Prevent users from accidentally specifying multiple JVM based receivers in the config without realizing something is wrong. If the errors generated in the logs makes it very clear this is the cause, it might not be that bad. (Is that the case?) Otherwise this would generate unnecessary support cases. Making it clear in the documentation would reduce the chance, but eliminating it.
* Cons: When we do support multiple JVM based receivers, we would need to do work to remove this validation.

Reference for similar validation:
* https://github.com/GoogleCloudPlatform/ops-agent/blob/8bd47ae99fbf38e6b54824cb4dcb0de7cb5961d1/confgenerator/config.go#L613
* https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/testdata/invalid/windows/metrics-receiver_more_than_one_with_type_iis/golden_error

If the effort to add this validation is relatively low, I'd prefer we add a validation to prevent users from trying out something that is not supported and wasting the effort.

WDYT?

apps/cassandra.go
qingling128: I like that idea. Also commented about it in https://github.com/GoogleCloudPlatform/ops-agent/pull/220#discussion_r724711297. Sam seems to be onboard too.

docs/cassandra.md
quentinmit: Okay, that kind of makes sense. Can you fix the description for these metrics in the docs here?

(BTW it might be better to use "cumulative" or "delta" in the data type column here, instead of "sum" which doesn't mean anything in the context of GCM.)

Kind of unfortunately that read and write request counts are two separate metrics; normally I think I would want to see that modeled as a single metric with a label for read/write. But I assume we're also constrained there by the upstream receiver, since we don't want to apply transformations to metrics.

docs/cassandra.md
quentinmit: What do you mean by "calculate rate afterwards"? The name and description of this metric talk about the number of concurrent hints, which is not a rate of anything (time is not a denominator for that metric).

docs/cassandra.md
dehaansa: Yeah, that's the currently behavior of the upstream receiver. We could certainly look at updating that in the future.

Descriptions fixed, sum updated to cumulative.

dehaansa:
Per conversation about metrics, waiting on the following PRs to be ready.
- https://github.com/open-telemetry/opentelemetry-java-contrib/pull/113 
- https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/58

docs/cassandra.md
quentinmit: I noticed that there's actually two different axes here. It feels like the request count should really be one metric with `operation={read,write,range_slice}` and `status={ok,timeout,unavailable}` labels, since right now there's just 3x3 metrics representing the cross product of those labels.

qingling128: +1 for the two proposed changes:

1. Change `50p/99p/max` to histogram with `{operation=read / write}` label. This seems to be a more long-term change since the support for histogram is not there yet.

2. Change
```
workload.googleapis.com/cassandra.client.request.read.latency.count
workload.googleapis.com/cassandra.client.request.read.timeout.count
workload.googleapis.com/cassandra.client.request.read.unavailable.count
workload.googleapis.com/cassandra.client.request.write.latency.count
workload.googleapis.com/cassandra.client.request.write.timeout.count
workload.googleapis.com/cassandra.client.request.write.unavailable.count
```

to something like

```
workload.googleapis.com/cassandra.client.request.count {operation=read / write, status=ok / timeout / unavailable}
```

This one seems like a more straight forward improvement. Related upstream PR: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/43 (Handle list). While it could be done in the future, it does require a breaking change or we have to give customer options to switch between both metrics structures for backward compatibility. It might be worth waiting a bit instead of going though that hassle.

qingling128: Current status: waiting for upstream PR: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/118 (
Update cassandra jmx metrics script to combine similar metrics into labelled metric).

PR #221, opened 2021-09-29
Redis Metrics

Mrod1598:
This PR adds a Redis metric receiver, with the following caveat:

~~The underlying [receiver](https://github.com/observIQ/opentelemetry-collector-contrib/tree/main/receiver/redisreceiver) is not currently imported in the build of the operations-collector submodule.~~

Also updated the operations-collector submodule to bring in the redis receiver.


docs/redis.md
```diff
@@ -1,3 +1,72 @@
...
+Following the guide for [Configuring the Ops Agent](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/configuration#file-location), add the required elements for your redis configuration.
+
+To configure a receiver for your redis metrics, specify the following fields:
+
+| Field                 | Default                   | Description |
+| ---                   | ---                       | ---         |
+| `type`                | required                  | Must be `redis`. |
+| `endpoint`            | `localhost:6379`          | The url exposed by redis |
+| `collection_interval` | `60s`                     | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
+| `password`            |                           | The password used to connect to the server.
```
jsuereth, line 17: Nit:  end table w/ `|`

docs/redis.md
```diff
@@ -1,3 +1,72 @@
...
+The Ops Agent collects the following metrics from your redis servers.
+
+| Metric                                | Data Type | Unit | Labels  | Description    |
+| ---                                   | ---       | ---  | ---     | ---            | 
+| workload.googleapis.com/redis/uptime	                              | sum	  | s		   |         | Number of seconds since Redis server start |
+| workload.googleapis.com/redis/cpu/time	                            | sum	  | s	       | state:  | System CPU consumed by the Redis server in seconds since server start |
+| workload.googleapis.com/redis/clients/connected	                    | sum	  | 1		   |         | Number of client connections (excluding connections from replicas) |
+| workload.googleapis.com/redis/clients/max_input_buffer	            | gauge	| 1		   |         | Biggest input buffer among current client connections |
+| workload.googleapis.com/redis/clients/max_output_buffer	            | gauge	| 1		   |         | Longest output list among current client connections |
+| workload.googleapis.com/redis/clients/blocked	                      | sum	  | 1		   |         | Number of clients pending on a blocking call |
```
jsuereth, line 47: Is this number of active clients blocked, or total since start?

docs/redis.md
```diff
@@ -1,3 +1,72 @@
...
+| workload.googleapis.com/redis/commands	                            | gauge	| {ops}/s  |         | Number of commands processed per second |
+| workload.googleapis.com/redis/commands/processed	                  | sum	  | 1		   |         | Total number of commands processed by the server |
+| workload.googleapis.com/redis/net/input	                            | sum	  | by	   |         | The total number of bytes read from the network |
+| workload.googleapis.com/redis/net/output	                          | sum	  | by	   |         | The total number of bytes written to the network |
+| workload.googleapis.com/redis/keyspace/hits	                        | sum	  | 1		   |         | Number of successful lookup of keys in the main dictionary |
+| workload.googleapis.com/redis/keyspace/misses	                      | sum	  | 1	       |         | Number of failed lookup of keys in the main dictionary |
+| workload.googleapis.com/redis/latest_fork	                          | guage	| us	   |         | Duration of the latest fork operation in microseconds |
+| workload.googleapis.com/redis/slaves/connected	                    | sum	  | 1		   |         | Number of connected replicas |
```
jsuereth, line 65: Does this never drop?  Is it a gauge?

docs/redis.md
Mrod1598: added

docs/redis.md
djaglowski: Should non-monotonic sums be documented as gauges at the op-agent receiver level, given that they end up being gauges as far as the user is concerned?

jsuereth: Yeah, for Ops-Agent we should document it as a gauge.

apps/redis.go
```diff
@@ -17,8 +17,49 @@ package apps
...
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
 )
 
+type MetricsReceiverRedis struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"omitempty,url"`
+	Password string `yaml:"password" validate:"omitempty"`
```
quentinmit, line 29: It doesn't look like you're using the password parameter anywhere?

apps/redis.go
```diff
@@ -17,8 +17,49 @@ package apps
...
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
 )
 
+type MetricsReceiverRedis struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"omitempty,url"`
```
quentinmit, line 28: This is being validated as a URL but the default is `localhost:6379`. Which is it, a URL or a host:port?

quentinmit (dismissed):

apps/redis.go
Mrod1598: Added to otel config and updated the golden confs

docs/redis.md
```diff
@@ -1,3 +1,71 @@
...
+| workload.googleapis.com/redis/memory/used	                          | gauge	 | by	     |         | Total number of bytes allocated by Redis using its allocator |
+| workload.googleapis.com/redis/memory/peak	                          | gauge	 | by	     |         | Peak memory consumed by Redis (in bytes) |
+| workload.googleapis.com/redis/memory/rss	                          | gauge	 | by	     |         | Number of bytes that Redis allocated as seen by the operating system |
+| workload.googleapis.com/redis/memory/lua	                          | gauge	 | by	     |         | Number of bytes used by the Lua engine |
+| workload.googleapis.com/redis/memory/fragmentation_ratio	          | gauge	 | 1		   |         | Ratio between used_memory_rss and used_memory |
+| workload.googleapis.com/redis/rdb/changes_since_last_save	          | gauge  | 1		   |         | Number of changes since the last dump |
+| workload.googleapis.com/redis/commands	                            | gauge	 | {ops}/s |         | Number of commands processed per second |
```
quentinmit, line 58: Why are we exposing this at all? Isn't it just the rate of the following metric? That seems entirely redundant (and metrics containing precalculated rates are hard to effectively use)

docs/redis.md
```diff
@@ -1,3 +1,71 @@
...
+          - redis_metrics
+```
+
+## Metrics
+
+The Ops Agent collects the following metrics from your redis servers.
+
+| Metric                                                              | Data Type | Unit    | Labels  | Description    |
+| ---                                                                 | ---    | ---     | ---     | ---            | 
+| workload.googleapis.com/redis/uptime	                              | sum	   | s		   |         | Number of seconds since Redis server start |
```
quentinmit, line 42: Our normal convention is that uptime metrics should be gauges, since users care about the value and not the rate. I'm not sure how that intersects with OT standards.

quentinmit, line 42: Why are all of these metrics slash-separated and not dot-separated? The scoping doc appears to call this "redis.uptime", though it also uses slashes for the subdirectories below.

apps/redis.go
quentinmit: That makes sense. I don't think it makes sense to document a default that doesn't have a scheme if we're intending this to be interpreted as a URL. I'd suggest that we update the documented default to "redis://localhost:6379" and documenting it as "Redis [URL](https://www.iana.org/assignments/uri-schemes/prov/redis). If the scheme is unspecified, `redis` is assumed." WDYT?

I just took a quick survey of some of the clients (geez, Redis has lot!?) and they seem super inconsistent about how they handle this, BTW:

- Redigo has an `address` [parameter](https://pkg.go.dev/github.com/gomodule/redigo/redis#Dial) that takes `host:port`, and an alternate API for parsing `redis://` URLs
- go-redis has an `Addr` [field](https://pkg.go.dev/github.com/go-redis/redis/v8#Options) that takes `host:port` with no option for URL that I can see
- aioredis has a `from_url` [method](https://aioredis.readthedocs.io/en/latest/api/high-level/#aioredis.client.Redis.from_url) that takes a URL, and no option for `host:port` that I see
- redis-py has `host` and `port` [parameters](https://redis-py.readthedocs.io/en/stable/#redis.Connection), and an alternate `from_url` method that takes a URL
- Walrus only documents `host` and `port` [parameters](https://walrus.readthedocs.io/en/latest/getting-started.html)
- redis-rs takes a `params` [argument](https://docs.rs/redis/0.21.3/redis/#connection-parameters) which can be cast from a URL string or be an instance of a struct with host and port

apps/redis.go
quentinmit: Thanks; I assume that explicitly specifying it in the config file as an empty string still works as a default?

Can you leave a TODO here with the plan to add username for ACL authentication?
(https://redis.io/topics/security and https://redis.io/commands/auth)

Mrod1598: Will do

docs/redis.md
Mrod1598: This was a preexisting receiver and it already had this metric.

docs/redis.md
qingling128: Can we add some translation to keep the separator consistent?

Mrod1598: I'm not sure what the meaning is here. it seems as though, because it's a sum it is just the value of the amount of seconds.  So, it's just the total amount of time it's been up not the rate of time that it's increasing by.

djaglowski: I think this raises a larger question of whether or not metric naming is going to be deferred to the OTel community, and if not, to what extent should naming be "corrected" at the ops agent level?

docs/redis.md
quentinmit: We could exclude it if we choose to, yes.

Mrod1598: would we like to?

docs/redis.md
Mrod1598: > Can we add some translation to keep the separator consistent?

for this one at least, I've opened a PR against the contrib repo to update the separator to be more consistent with the existing metrics.

apps/redis.go
quentinmit: Okay. Can we also name the field `address` to be more consistent with all the client libraries, since it doesn't accept a URL?

What about SSL and Unix sockets, since we don't have the scheme to set those kinds of connections?

apps/redis.go
Mrod1598: Added

docs/redis.md
Mrod1598: do we have an answer to this question yet?

apps/redis.go
qingling128: Looks like the requested changes have been applied. Thanks!

docs/redis.md
qingling128: Current status: This is being changed in upstream PR: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/5837 (Update Redis receiver's metric names)

quentinmit: That's merged now - can you update this PR?

Mrod1598: updated

qingling128 (approved):
LGTM

docs/redis.md
qingling128: Thanks for updating it!

quentinmit: I think this doc still needs to be updated with the new metric names, and to remove the uptime metric.

qingling128: Nice catch. @Mrod1598 - Can you help update the `docs/redis.md` file?

Mrod1598: updated

docs/redis.md
```diff
@@ -12,10 +12,19 @@ To configure a receiver for your redis metrics, specify the following fields:
...
-| `address`            | `localhost:6379`          | The url exposed by redis |
+| `address`             | `localhost:6379`          | The url exposed by redis |
 | `collection_interval` | `60s`                     | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
 | `password`            |                           | The password used to connect to the server. |
 
+optional fields are: 
+| Field                 | Default                   | Description |
+| ---                   | ---                       | ---         |
+| `transport` | `tcp`| Defines the network to use for connecting to the server. |
+| `insecure` | `true` | whether to disable client transport security for the exporter's connection. |
```
qingling128, line 23: Can the value of config parameter insecure be inferred by the presence of the other 3 parameters: CertFilePath, CaFilePath, KeyFilePath? If so, we could cut one redundant parameter.

docs/redis.md
```diff
@@ -26,6 +35,10 @@ metrics:
       address: localhost:6379
       collection_interval: 30s
       password: pwd
+      insecure: false
```
qingling128, line 38: Can we add some unit tests for setting TLS?

* A `valid` test for `secure` path
* A `valid` test for the `insecure` path
* Some `invalid` tests to verify the errors we throw when conflicting parameter values are specified together (e.g. `insecure` is true or omitted, but `cert_file` is set

qingling128 (commented):
Left some comments for the TLS change that was newly added and needs some additional work. I'm fine with splitting the TLS change into a follow up PR, too.

qingling128 (changes requested):
Removing LGTM for now for the newly added TLS change so we don't accidentally merge this one.

docs/redis.md
dehaansa: @qingling128 How would we throw an error on your 3rd request? The only kind of config validation that I'm aware of in the ops agent right now are things like data type validation (is this a valid URL), all the deeper validation is currently in the receivers themselves, which don't show until config has been generated & otel collector attempts to run

qingling128: https://github.com/go-playground/validator#other has a few options that could work for the case here: e.g. `required_without` and `exclude_with`.

Something like:

```
  A string `validate:"required_without=B,excluded_with=B"`
```

docs/redis.md
Mrod1598: good point. set that up.

docs/redis.md
Mrod1598: setup that validation and those test cases

confgenerator/testdata/invalid/linux/metrics-receiver_redis_cert_missing_key/golden_error
```diff
@@ -0,0 +1,9 @@
+[17:18] Key: 'MetricsReceiverRedis.key_file' Error:Field validation for 'key_file' failed on the 'required_with' tag
```
qingling128, line 1: Can we add a rule to https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/config.go#L85 for `required_with` validation too so that the error message is more descriptive and customized to Ops Agent config?

Something like: 'key_file' is a required field when 'cert_file' is present.

docs/redis.md
quentinmit: Without an explicit config option for TLS, there's now no way to enable TLS without setting explicit certificates. I don't know enough about how Redis is typically deployed; is it really the case that Redis is never deployed with normal globally-trusted SSL certificates? (in which case the host defafult SSL

docs/redis.md
qingling128: Do we know whether both of these cases are valid at otel level?

* insecure == true and none of the other 3 parameters are set
* insecure == false and none of the other 3 parameters are set

I was under the impression that `insecure == false and none of the other 3 parameters are set` is not a valid case, thus I recommended omitting `insecure`. But Quentin pointed out a good point that an unspecified CA path might mean to use the system default CAs, so it could potentially be a valid case though.

@Mrod1598 Do you know or can you test it out whether that is something otel supports?

qingling128: In fact, it would be better if we can separate the TLS implementation for Redis to a follow up PR and take the time we need to get it right. There is some complexity to it regarding:

* What config parameters make the most sense for Redis users. e.g. Unlike some other apps, Redis out of the box does not come with TLS enabled and require explicit user actions to enable it.
* An unspecified CA path might mean to use the system default CAs. It's unclear yet whether otel supports this. It could affect config design.
* Some receivers in the future might need TLS support as well, so the TLS related config we invent today would likely be the convention moving forward.
* We might need some end to end tests for TLS (Does otel have anything related already?)

qingling128: CC @quentinmit for visibility as we chatted about this offline.

Mrod1598: Ok removed TLS for the time being. will make that PR when this gets merged.

Mrod1598: > * or Redis users. e.g. Unlike some other apps, Redis out of the box does not come with TLS enabled and require explicit user actions to enable it.
> * An unspecified CA path might mean to use the system default CAs. It's unclear yet whether otel supports this. It could affe

will still look into this and see what I can figure out.

Mrod1598: > I was under the impression that `insecure == false and none of the other 3 parameters are set` is not a valid case, thus I recommended omitting `insecure`. But Quentin pointed out a good point that an unspecified CA path might mean to use the system default CAs, so it could potentially be a valid case though.
> 

just tested it and `insecure == false and none of the other 3 parameters are set`  is a valid config.

confgenerator/testdata/invalid/linux/metrics-receiver_redis_cert_missing_key/golden_error
dehaansa: Will be done in the follow up TLS PR.

qingling128 (approved):
LGTM

PR #222, opened 2021-09-29
Redis log receiver

dehaansa:
Add support for a redis log receiver. Dependent on approval of scope doc.

image

### DIFF
--- a/apps/redis.go
+++ b/apps/redis.go
@@ -1,0 +1,102 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apps
+
+import (
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
+)
+
+type LoggingProcessorRedis struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+}
+
+func (LoggingProcessorRedis) Type() string {
+	return "redis"
+}
+
+func (p LoggingProcessorRedis) Components(tag string, uid string) []fluentbit.Component {
+	c := confgenerator.LoggingProcessorParseRegex{
+		// Documentation: https://github.com/redis/redis/blob/6.2/src/server.c#L1122
+		// Sample line: 534:M 28 Apr 2020 11:30:29.988 * DB loaded from disk: 0.002 seconds
+		// Sample older format line: [4018] 14 Nov 07:01:22.119 * Background saving terminated with success
+		Regex: `^\[?(?<pid>\d+):?(?<roleChar>[A-Z])?\]?\s+(?<time>\d{2}\s+\w+(?:\s+\d{4})?\s+\d{2}:\d{2}:\d{2}.\d{3})\s+(?<level>(\*|#|-|\.))\s+(?<message>.*)$`,
+		LoggingProcessorParseShared: confgenerator.LoggingProcessorParseShared{
+			TimeKey:    "time",
+			TimeFormat: "%d %b %Y %H:%M:%S.%L",
+			Types: map[string]string{
+				"pid": "integer",
+			},
+		},
+	}.Components(tag, uid)
+
+	for _, l := range []struct{ level, severity string }{
+		// Log levels documented: https://github.com/redis/redis/blob/6.2/src/server.c#L1124
+		{".", "DEBUG"},
+		{"-", "INFO"},
+		{"*", "NOTICE"},
+		{"#", "WARNING"},
+	} {
+		c = append(c, fluentbit.Component{
+			Kind: "FILTER",
+			Config: map[string]string{
+				"Name":      "modify",
+				"Match":     tag,
+				"Condition": fmt.Sprintf("Key_Value_Equals level %s", l.level),
+				"Add":       fmt.Sprintf("logging.googleapis.com/severity %s", l.severity),
+			},
+		})
+	}
+
+	for _, l := range []struct{ roleChar, role string }{
+		// Role translation documented: https://github.com/redis/redis/blob/6.2/src/server.c#L1149
+		{"X", "sentinel"},
+		{"C", "RDB/AOF writing child"},
+		{"S", "slave"},
+		{"M", "master"},
+	} {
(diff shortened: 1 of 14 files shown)


apps/redis.go
```diff
@@ -0,0 +1,102 @@
...
+	}.Components(tag, uid)
+
+	for _, l := range []struct{ level, severity string }{
+		// Log levels documented: https://github.com/redis/redis/blob/6.2/src/server.c#L1124
+		{".", "DEBUG"},
+		{"-", "INFO"},
+		{"*", "NOTICE"},
+		{"#", "WARNING"},
+	} {
+		c = append(c, fluentbit.Component{
```
quentinmit, line 54: This pattern has appeared several times now; could you add a helper to generate these from a `map[string]string` and `src` and `dest` field names?

apps/redis.go
```diff
@@ -0,0 +1,102 @@
...
+func (LoggingProcessorRedis) Type() string {
+	return "redis"
+}
+
+func (p LoggingProcessorRedis) Components(tag string, uid string) []fluentbit.Component {
+	c := confgenerator.LoggingProcessorParseRegex{
+		// Documentation: https://github.com/redis/redis/blob/6.2/src/server.c#L1122
+		// Sample line: 534:M 28 Apr 2020 11:30:29.988 * DB loaded from disk: 0.002 seconds
+		// Sample older format line: [4018] 14 Nov 07:01:22.119 * Background saving terminated with success
+		Regex: `^\[?(?<pid>\d+):?(?<roleChar>[A-Z])?\]?\s+(?<time>\d{2}\s+\w+(?:\s+\d{4})?\s+\d{2}:\d{2}:\d{2}.\d{3})\s+(?<level>(\*|#|-|\.))\s+(?<message>.*)$`,
```
quentinmit, line 37: (Like on PR 219) Do we have a convention for camel-case vs. underscores? I see `roleChar` here (contrast with `error_code` on PR 219)

apps/redis.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type LoggingReceiverRedis struct {
+	LoggingProcessorRedis                   `yaml:",inline"`
+	confgenerator.LoggingReceiverFilesMixin `yaml:",inline" validate:"structonly"`
+}
+
+func (r LoggingReceiverRedis) Components(tag string) []fluentbit.Component {
+	if len(r.IncludePaths) == 0 {
+		r.IncludePaths = []string{"/var/log/redis/redis-server.log*", "/var/log/redis_*", "/var/log/redis/redis_*.log*"}
```
quentinmit, line 92: Why does this have `redis-server.log*`? I imagine this will catch rotated log files, and possibly cause them to get processed a second time when they are renamed?

quentinmit, line 92: Do we know these paths are accurate for all the supported distros?

(Maybe we should start leaving a comment on each set of include paths listing which distros go with which paths?)

dehaansa: Yeah, that seems like an oversight, I'll fix it.

apps/redis.go
qingling128: +1 for camelCase

apps/redis.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (LoggingProcessorRedis) Type() string {
+	return "redis"
+}
+
+func (p LoggingProcessorRedis) Components(tag string, uid string) []fluentbit.Component {
+	c := confgenerator.LoggingProcessorParseRegex{
+		// Documentation: https://github.com/redis/redis/blob/6.2/src/server.c#L1122
+		// Sample line: 534:M 28 Apr 2020 11:30:29.988 * DB loaded from disk: 0.002 seconds
+		// Sample older format line: [4018] 14 Nov 07:01:22.119 * Background saving terminated with success
```
quentinmit, line 34: If you can, can you actually list the Redis versions that apply for each of these sample lines?

apps/utils.go
```diff
@@ -0,0 +1,38 @@
...
+
+package apps
+
+import (
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
+)
+
+func TranslationComponents(tag, src, dest string, translations []struct{ srcVal, destVal string }) []fluentbit.Component {
```
quentinmit, line 23: Thanks for writing this helper! Can you please put it in the `fluentbit` package?

integration_test/third_party_apps_data/applications/redis/expected_logs.yaml
```diff
@@ -0,0 +1,7 @@
+log_entries:
+- log_name: "redis"
+  field_matchers:
+    jsonPayload.level: "\\*"
```
quentinmit, line 4: Can you test for `severity` as well?

quentinmit (approved):

apps/redis.go
```diff
@@ -0,0 +1,100 @@
...
+	confgenerator.LoggingReceiverFilesMixin `yaml:",inline" validate:"structonly"`
+}
+
+func (r LoggingReceiverRedis) Components(tag string) []fluentbit.Component {
+	if len(r.IncludePaths) == 0 {
+		r.IncludePaths = []string{
+			// Default log path on Ubuntu / Debian
+			"/var/log/redis/redis-server.log",
+			// Default log path built from src
+			"/var/log/redis_6379.log",
```
jsuereth, line 83: IIUC - 6379 is the default port of redis, and any install automatically select this path based on the port you specify when installing.

Just want to confirm my understanding of where the "magic" number comes from here.  If so, would you mind adding that to the comment?

integration_test/third_party_apps_data/applications/redis/expected_logs.yaml
dehaansa: I tried to test using `severity` and something in the way the tooling is built couldn't handle it. I saw an error like the following:

```
2021/09/29 15:14:57 Attempt 1 of debian-10 test of redis finished with err=QueryLog() failed: rpc error: code = InvalidArgument desc = Field 'severity' is of incorrect type. Regular expression matches can only be performed against string fields., retryable=false
=== CONT  TestThirdPartyApps/ops-agent/debian-10/redis
    third_party_apps_test.go:385: Non-retryable error: QueryLog() failed: rpc error: code = InvalidArgument desc = Field 'severity' is of incorrect type. Regular expression matches can only be performed against string fields.
```

apps/redis.go
dehaansa: That's correct that the magic number is the default redis port. I believe you're correct that it would modify the filename during install if you selected a non-standard port.

qingling128 (approved):
LGTM

integration_test/third_party_apps_data/applications/redis/expected_logs.yaml
qingling128: Probably because `severity` is an enum: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity.

Proposed a potential fix for the internal testing infra and filed a ticket for tracking: b/202287040

qingling128:
@dehaansa - Ah, I missed one thing: Can we send a follow up PR for the markdown documentation? Similar to https://github.com/observIQ/ops-agent/blob/master/docs/apache.md

PR #225, opened 2021-10-02
Generate multiline parsers for fluent bit

dehaansa:
This config used in test is copied from cassandra development, it works properly for cassandra system logs parsing.

### DIFF
--- a/confgenerator/fluentbit/modular.go
+++ b/confgenerator/fluentbit/modular.go
@@ -25,7 +25,8 @@ type Component struct {
 	// Kind is "INPUT", "FILTER", "PARSER", etc.
 	Kind string
 	// Config is a set of key-value configuration pairs
-	Config map[string]string
+	Config           map[string]string
+	AdditionalConfig []string
 }
 
 func (c Component) generateSection() string {
@@ -40,6 +41,13 @@ func (c Component) generateSection() string {
 		lines = append(lines, fmt.Sprintf("    %-*s %s", maxLen, k, v))
 	}
 	sort.Strings(lines)
+
+	// Used for Multiline config where several "rule" lines
+	// must be placed at the end of a parser config
+	for _, line := range c.AdditionalConfig {
+		lines = append(lines, fmt.Sprintf("    %s", line))
+	}
+
 	return fmt.Sprintf("[%s]\n%s\n", c.Kind, strings.Join(lines, "\n"))
 }
 
@@ -63,8 +71,9 @@ func (c ModularConfig) Generate() (string, string, error) {
 		out := c.generateSection()
 		sectionsByKind[c.Kind] = append(sectionsByKind[c.Kind], out)
 	}
-	parserParts := sectionsByKind["PARSER"]
+	parserParts := append(sectionsByKind["PARSER"], sectionsByKind["MULTILINE_PARSER"]...)
 	delete(sectionsByKind, "PARSER")
+	delete(sectionsByKind, "MULTILINE_PARSER")
 	for _, k := range []string{"SERVICE", "INPUT", "FILTER", "OUTPUT"} {
 		parts = append(parts, sectionsByKind[k]...)
 		delete(sectionsByKind, k)
--- a/confgenerator/logging_processors.go
+++ b/confgenerator/logging_processors.go
@@ -14,7 +14,11 @@
 
 package confgenerator
 
-import "github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
+import (
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
+)
 
 // A LoggingProcessorParseJson parses the specified field as JSON.
 type LoggingProcessorParseJson struct {
@@ -59,6 +63,75 @@ func (p LoggingProcessorParseRegex) Components(tag, uid string) []fluentbit.Comp
 		filter,
 		parser,
 	}
+}
+
+type MultilineRule struct {
+	StateName string `yaml:"state_name" validate:"required"`
+	Regex     string `yaml:"regex,omitempty" validate:"required"`
+	NextState string `yaml:"next_state" validate:"required"`
+}
+
+func (r MultilineRule) AsString() string {
+	return fmt.Sprintf(`rule    "%s"    "%s"    "%s"`, r.StateName, r.Regex, r.NextState)
+}
+
+// A LoggingProcessorParseMultiline applies a set of regex rules to the specified lines, storing the named capture groups as keys in the log record.
+//     #
+//     # Regex rules for multiline parsing
+//     # ---------------------------------
+//     #
+//     # configuration hints:
+//     #
+//     #  - first state always has the name: start_state
+//     #  - every field in the rule must be inside double quotes
+//     #
+//     # rules |   state name  | regex pattern                  | next state
+//     # ------|---------------|--------------------------------------------
+//     rule      "start_state"   "/(Dec \d+ \d+\:\d+\:\d+)(.*)/"  "cont"
+//     rule      "cont"          "/^\s+at.*/"                     "cont"
+type LoggingProcessorParseMultiline struct {
+	ConfigComponent             `yaml:",inline"`
+	LoggingProcessorParseShared `yaml:",inline"`
+
+	Rules []MultilineRule `yaml:"rules" validate:"required"`
+	Regex string          `yaml:"regex,omitempty" validate:"required"`
+}
+
+func (r LoggingProcessorParseMultiline) Type() string {
+	return "parse_regex"
+}
+
+func (p LoggingProcessorParseMultiline) Components(tag, uid string) []fluentbit.Component {
+	multilineParserName := fmt.Sprintf("%s.%s.multiline", tag, uid)
+	rulesText := []string{}
+	for _, rule := range p.Rules {
+		rulesText = append(rulesText, rule.AsString())
+	}
+
+	multilineComponents := []fluentbit.Component{
+		fluentbit.Component{
+			Kind: "FILTER",
+			Config: map[string]string{
+				"Name":                  "multiline",
+				"Match":                 tag,
+				"Multiline.Key_Content": "message",
+				"Multiline.Parser":      multilineParserName,
+			},
+		},
+		fluentbit.Component{
+			Kind: "MULTILINE_PARSER",
+			Config: map[string]string{
+				"Name": multilineParserName,
+				"Type": "regex",
+			},
+			AdditionalConfig: rulesText,
+		},
+	}
+
+	filter, parser := p.LoggingProcessorParseShared.Components(tag, uid)
+	parser.Config["Format"] = "regex"
+	parser.Config["Regex"] = p.Regex
+	return append(multilineComponents, filter, parser)
 }
 
 func init() {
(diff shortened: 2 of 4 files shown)


confgenerator/logging_processors.go
```diff
@@ -39,77 +44,184 @@ func init() {
...
+type LoggingProcessorParseRegexSimple struct {
+	ConfigComponent        `yaml:",inline"`
+	fluentbit.ParserShared `yaml:",inline"`
+	Field                  string `yaml:"field,omitempty"`
 
 	Regex string `yaml:"regex,omitempty" validate:"required"`
 }
 
-func (r LoggingProcessorParseRegex) Type() string {
+func (r LoggingProcessorParseRegexSimple) Type() string {
```
dehaansa, line 57: This "simple" regex parser was left in place as I wanted this update to not break previously configured one-off regex parsers. If that's not really a concern it could be removed leaving only the multiple parser version.

qingling128 (commented):
My main concern is that we are introducing a new Ops Agent user visible receiver `parse_regex_complex`. We do want to support multi format parsers. But it likely would also be applicable for json parsers and needs more time to design and integration test it.

Can we leave the multi parsers code at fluent bit subagent level for now in this PR since the goal is to facilitate specific use cases of some 3rd party applications?

confgenerator/logging_processors.go
dehaansa: Resolved.

dehaansa:
> My main concern is that we are introducing a new Ops Agent user visible receiver `parse_regex_complex`. We do want to support multi format parsers. But it likely would also be applicable for json parsers and needs more time to design and integration test it.
> 
> Can we leave the multi parsers code at fluent bit subagent level for now in this PR since the goal is to facilitate specific use cases of some 3rd party applications?

Does this also apply to the parse_regex_multiline receiver? It is currently user visible, but I can not expose it as well and keep it internally useable by fluent bit.

qingling128:
Yes, it applies to `parse_regex_multiline` as well.

dehaansa:
Code coverage reduction is due to the removal of user-facing parse_regex_complex & parse_regex_multiline, which prevents me from writing a confgenerator/testdata unit test for either of them. When we have log receivers using them, that code will get code coverage again. Alternatively, I could write an out of band unit test to cover the code.

confgenerator/logging_processors.go
```diff
@@ -39,77 +44,182 @@ func init() {
 	LoggingProcessorTypes.RegisterType(func() Component { return &LoggingProcessorParseJson{} })
 }
 
-// A LoggingProcessorParseRegex applies a regex to the specified field, storing the named capture groups as keys in the log record.
-type LoggingProcessorParseRegex struct {
-	ConfigComponent             `yaml:",inline"`
-	LoggingProcessorParseShared `yaml:",inline"`
+// A LoggingProcessorParseRegexSimple applies a regex to the specified field, storing the named capture groups as keys in the log record.
+// This was maintained in addition to the parse_regex_complex to ensure backward compatibility with any existing configurations
+type LoggingProcessorParseRegexSimple struct {
```
quentinmit, line 49: Can you please keep type name as `LoggingProcessorParseRegex` since the receiver type is still `parse_regex` (and we can't change that)?

confgenerator/logging_processors.go
```diff
@@ -39,77 +44,182 @@ func init() {
...
+	}
+}
+
+type RegexParser struct {
+	Regex  string                 `yaml:"regex,omitempty" validate:"required"`
+	Parser fluentbit.ParserShared `yaml:",inline"`
+}
+
+// A LoggingProcessorParseRegex applies a set of regexes to the specified field, storing the named capture groups as keys in the log record.
+type LoggingProcessorParseRegex struct {
```
quentinmit, line 78: Please rename this type instead (e.g. `LoggingProcessorParseRegexComplex`)

confgenerator/logging_processors.go
```diff
@@ -39,77 +44,182 @@ func init() {
...
+}
+
+type MultilineRule struct {
+	StateName string `yaml:"state_name" validate:"required"`
+	Regex     string `yaml:"regex,omitempty" validate:"required"`
+	NextState string `yaml:"next_state" validate:"required"`
+}
+
+func (r MultilineRule) AsString() string {
+	return fmt.Sprintf(`"%s"    "%s"    "%s"`, r.StateName, r.Regex, r.NextState)
```
quentinmit, line 112: What kind of escaping do these strings need (especially the regex)?

confgenerator/fluentbit/modular.go
```diff
@@ -26,6 +26,67 @@ type Component struct {
 	Kind string
 	// Config is a set of key-value configuration pairs
 	Config map[string]string
+	// RepeatedConfig is used for configuration pairs where the
+	// key can appear in the output fluent bit config multiple times
+	RepeatedConfig map[string][]string
+}
+
+// ParserShared holds common parameters that are used by all processors that are implemented with fluentbit's "parser" filter.
```
quentinmit, line 34: Can you please put this new type in a separate file since it doesn't interact with the rest of this file? (e.g. `processors.go`?

Out of curiosity, what was the reason this moved into a different package? The structure I was trying to establish was the the `confgenerator/fluentbit` package only contained non-user-visible code and structures, and all of the user-visible types with YAML tags etc. were in `confgenerator` or `apps`.

confgenerator/fluentbit/modular.go
```diff
@@ -36,10 +97,25 @@ func (c Component) generateSection() string {
...
 		lines = append(lines, fmt.Sprintf("    %-*s %s", maxLen, k, v))
 	}
 	sort.Strings(lines)
+
+	// Used for Multiline config where several "rule" lines
+	// must be placed at the end of a parser config, and when multiple "Parser"
+	// are provided to one parser filter
+	for k, list := range c.RepeatedConfig {
+		for _, v := range list {
+			lines = append(lines, fmt.Sprintf("    %-*s %s", maxLen, k, v))
```
quentinmit, line 115: Nit: Refactor this to a common function? e.g.

```
addLine := func(k, v string) { lines = append(lines, fmt.Sprintf("    %-*s %s", maxLen, k, v)) }
```

confgenerator/logging_processors.go
```diff
@@ -39,77 +44,182 @@ func init() {
...
+		fluentbit.ParserFilterComponent(tag, p.Field, []string{parserName}),
+	}
+}
+
+type RegexParser struct {
+	Regex  string                 `yaml:"regex,omitempty" validate:"required"`
+	Parser fluentbit.ParserShared `yaml:",inline"`
+}
+
+// A LoggingProcessorParseRegex applies a set of regexes to the specified field, storing the named capture groups as keys in the log record.
```
quentinmit, line 77: Since we're not exposing any of these processors to users, can we simplify these structs? It seems like there's a lot of redundancy here (e.g. `Type` methods, yaml struct tags, etc.) that isn't being used at all.

confgenerator/logging_processors.go
dehaansa: The only thing I'm aware of is that they shouldn't have unescaped double quotes, otherwise they are just quoted regex literals (https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/multiline-parsing)

confgenerator/fluentbit/modular.go
dehaansa: I moved it intending to slide the functions that felt less user visible (generating the Component objects) into the fluentbit support package, but keep the Component function tightly associated with the shared parser struct. Additionally, the struct is intended to specify common parts that are very specific to Fluent Bit.

I'm happy to move it all back, or to move the struct back and keep the functions in fluentbit, or whatever makes sense.

dehaansa: Moved into processors.go for now, but happy to make another change if you have a different thought @quentinmit

confgenerator/logging_processors.go
quentinmit: I think it's likely that some of these patterns will have double quotes in them, so does that mean we should add a call to ```strings.Replace(r.Regex, `"`, `\"`)``` here (is that the proper escaping syntax?)?

dehaansa: I think that's good enough for now. Some concern that a developer could try to escape it themselves if they are familiar with the format, but should be caught pretty quickly in testing. Could write a more complex replace that would only escape if not already escaped but I think it makes sense to keep it simple for now

confgenerator/logging_processors.go
```diff
@@ -52,13 +60,109 @@ func (r LoggingProcessorParseRegex) Type() string {
...
+}
+
+type RegexParser struct {
+	Regex  string
+	Parser fluentbit.ParserShared
+}
+
+// A LoggingProcessorParseRegexComplex applies a set of regexes to the specified field, storing the named capture groups as keys in the log record.
+type LoggingProcessorParseRegexComplex struct {
+	ConfigComponent
```
quentinmit, line 80: I don't think `ConfigComponent` is used anymore.

All of these new types and methods would be candidates to move into the `fluentbit` package since they're no longer user-facing. Up to you, I think based on how likely it is that we'll want to expose these to user config. (Personally, I'd guess not likely, because they are incredibly fluentbit-specific.)

confgenerator/fluentbit/modular.go
```diff
@@ -26,6 +26,9 @@ type Component struct {
 	Kind string
 	// Config is a set of key-value configuration pairs
 	Config map[string]string
+	// RepeatedConfig is used for configuration pairs where the
+	// key can appear in the output fluent bit config multiple times
+	RepeatedConfig map[string][]string
```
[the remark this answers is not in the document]
quentinmit: +1, this is a great observation. `[][2]string` even has the advantage of being fewer characters :P

Merging them to one field would mean that we lose the current behavior of sorting the keys. That might make some of the current uses harder, because items would have to be added in order (and accessing items in `Config` would be harder if the caller wants to mutate them), but I'm 50/50 on whether it's a net win to merge them into one.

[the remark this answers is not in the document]
dehaansa: @jefferbrecht I was working on the change you recommended, and it seemed cleaner to do it as a separate PR (https://github.com/GoogleCloudPlatform/ops-agent/pull/231). The change is really all in this commit https://github.com/GoogleCloudPlatform/ops-agent/pull/231/commits/4c60f04f6ef40abacd23bb9b1a67ded5c4b142a6 but it changed enough things across the fluent bit code base that it felt cleaner to do it separately.

confgenerator/logging_processors.go
dehaansa: I tested the `\"` escape on a deployment of the ops-agent I had set up for cassandra testing and it worked correctly.

confgenerator/fluentbit/modular.go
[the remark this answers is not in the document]
qingling128: Regarding sorting, can we sort by the key then value? Either way, we could continue this discussion in the follow up PR. I left a comment about it in https://github.com/GoogleCloudPlatform/ops-agent/pull/231#discussion_r723785292.

confgenerator/logging_processors.go
qingling128: Let's make sure we have a test case in the actual application PR to cover this .

quentinmit (approved):
LGTM modulo a resolution for the comment thread below

confgenerator/logging_processors.go
dehaansa: I would probably prefer it here, and also refactoring it out without making a circular dependency looks complex, so I think we'll keep it there for now and possibly tackle it in a refactor PR if it makes sense in the future.

qingling128: Sounds good

PR #226, opened 2021-10-04
Implement Systemd Logging Receiver

hsmatulisgoogle:
Sample log entry:

```
{
  insertId: "1do3uimfsdg4tl"
  jsonPayload: {
    CODE_FILE: "src/core/job.c"
    CODE_FUNCTION: "job_log_status_message"
    CODE_LINE: "781"
    MESSAGE: "Stopped Google Cloud Ops Agent - Logging Agent."
    MESSAGE_ID: "9d1aaa27d60140bd96365438aad20286"
    PRIORITY: "6"
    RESULT: "done"
    SYSLOG_FACILITY: "3"
    SYSLOG_IDENTIFIER: "systemd"
    UNIT: "google-cloud-ops-agent-fluent-bit.service"
    _BOOT_ID: "a71a67f16b4047d28950adf93aad1760"
    _CAP_EFFECTIVE: "1fffffffff"
    _CMDLINE: "/usr/lib/systemd/systemd --switched-root --system --deserialize 22"
    _COMM: "systemd"
    _EXE: "/usr/lib/systemd/systemd"
    _GID: "0"
    _HOSTNAME: "systemd-centos7"
    _MACHINE_ID: "XXXX"
    _PID: "1"
    _SELINUX_CONTEXT: "system_u:system_r:init_t:s0"
    _SOURCE_REALTIME_TIMESTAMP: "1633383692647691"
    _SYSTEMD_CGROUP: "/"
    _TRANSPORT: "journal"
    _UID: "0"
  }
  logName: "projects/XXXXX/logs/systemd_logs"
  receiveTimestamp: "2021-10-04T21:41:33.775290488Z"
  resource: {
    labels: {
      instance_id: "XXXXX"
      project_id: "XXXXX"
      zone: "us-central1-a"
    }
    type: "gce_instance"
  }
  timestamp: "2021-10-04T21:41:32.650252Z"
}
```


confgenerator/logging_receivers.go
```diff
@@ -228,3 +228,28 @@ func (r LoggingReceiverWindowsEventLog) Components(tag string) []fluentbit.Compo
...
+
+func (r LoggingReceiverSystemd) Type() string {
+	return "systemd"
+}
+
+func (r LoggingReceiverSystemd) Components(tag string) []fluentbit.Component {
+	return []fluentbit.Component{{
+		Kind: "INPUT",
+		Config: map[string]string{
+			// https://docs.fluentbit.io/manual/pipeline/inputs/windows-event-log
```
qingling128, line 245:
```suggestion
			// https://docs.fluentbit.io/manual/pipeline/inputs/systemd
```

confgenerator/testdata/invalid/windows/logging-receiver_systemd/golden_error
```diff
@@ -0,0 +1 @@
+logging receiver with type "systemd" is not supported. Supported logging receiver types: [files, nginx_access, nginx_error, syslog, tcp, windows_event_log].
```
qingling128, line 1: For naming consistency, could we make the folder name explicit that it's not supported? e.g.

```
confgenerator/testdata/invalid/windows/logging-receiver_unsupported_type_systemd
```

hsmatulisgoogle: Done

qingling128 (approved):
LGTM, modulo the internal testing passes.

confgenerator/logging_receivers.go
```diff
@@ -228,3 +228,28 @@ func (r LoggingReceiverWindowsEventLog) Components(tag string) []fluentbit.Compo
 func init() {
 	LoggingReceiverTypes.RegisterType(func() Component { return &LoggingReceiverWindowsEventLog{} }, "windows")
 }
+
+// A LoggingReceiverSystemd represents the user configuration for a Systemd/journald receiver.
+type LoggingReceiverSystemd struct {
```
quentinmit, line 233: Why is there no configuration for this receiver? Without the ability to configure journald matches it's going to collect a bunch of redundant log messages.

confgenerator/logging_receivers.go
```diff
@@ -228,3 +228,28 @@ func (r LoggingReceiverWindowsEventLog) Components(tag string) []fluentbit.Compo
...
+	ConfigComponent `yaml:",inline"`
+}
+
+func (r LoggingReceiverSystemd) Type() string {
+	return "systemd"
```
quentinmit, line 238: IMO `systemd` is a bad name for this receiver; systemd supports multiple different logging systems and journald can collect logs from more than just systemd, so I think a bare `systemd` is misleading. I know that's what fluentbit calls it, but I would suggest something like `systemd-journald` or `journald`

quentinmit (changes requested):
Where are the docs for this receiver? I'd like to see what the structure of the generated log entries will be.

confgenerator/logging_receivers.go
qingling128: Detailed reasoning for the decision was recorded in the internal design doc `go/ops-agent-systemd-receiver-design`.

qingling128:
Tested: http://b/195571765#comment3.

confgenerator/logging_receivers.go
quentinmit: `systemd_journald` seems like a fine name. Thanks!

qingling128:
@googlebot I consent

VERSION
```diff
@@ -1,2 +1 @@
-PKG_VERSION="2.4.0"
-
+PKG_VERSION="2.4.0"
```
quentinmit, line 1: Accidental change?

quentinmit (approved):
It sounds like the intent is to match the JSON format from systemd, so bare keys get put directly into the `jsonPayload`: https://www.freedesktop.org/wiki/Software/systemd/json/

Binary values and duplicate keys are not currently handled correctly; currently binary values have non-ASCII bytes replaced with spaces, and only the last value for a duplicate key is kept. Fixing those both seem reasonable as bug fixes after the fact. Ling and I confirmed that neither of these block a log entry from being accepted by the Cloud Logging API.

qingling128:
Rebased to get rid of the cla requirement. Will merge once tests pass.

VERSION
qingling128: Fixed

PR #228, opened 2021-10-04
Remove expired root certificate on sles12.

igorpeshansky:
This is similar to GoogleCloudPlatform/google-fluentd#363.


qingling128 (approved):
LGTM

PR #229, opened 2021-10-05
Cassandra logs

dehaansa:
Cassandra logs receivers for System, Debug, and GC logs. Dependent on multiline parser config https://github.com/GoogleCloudPlatform/ops-agent/pull/225.

TODO
- Find & document default log paths.

GC Logs example:
image

Debug & System are the same content, just logged at different severity. Example:
image


apps/cassandra.go
```diff
@@ -0,0 +1,206 @@
...
+	c := confgenerator.LoggingProcessorParseMultilineRegex{
+		LoggingProcessorParseRegexComplex: confgenerator.LoggingProcessorParseRegexComplex{
+			Parsers: []confgenerator.RegexParser{
+				confgenerator.RegexParser{
+					// Vast majority of lines look like the first, with time stopped & time stopping
+					// Sample line: 2021-10-02T04:18:28.284+0000: 3.315: Total time for which application threads were stopped: 0.0002390 seconds, Stopping threads took: 0.0000281 seconds
+					// Sample line: 2021-10-05T01:20:52.695+0000: 4.434: [GC (CMS Initial Mark) [1 CMS-initial-mark: 0K(3686400K)] 36082K(4055040K), 0.0130057 secs] [Times: user=0.04 sys=0.00, real=0.01 secs]
+					// Sample line: 2021-10-05T01:20:52.741+0000: 4.481: [CMS-concurrent-preclean-start]
+					// Lines may also contain more detailed GC Heap information in the following lines
+					Regex: `^(?<time>\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3,6}(?:Z|[+-]\d{2}:?\d{2})):\s+(?<uptime>\d+\.\d{3,6}):\s+(?<message>(?:.*(?:stopped: (?<timeStopped>\d+\.\d+)).*(?:took: (?<timeStopping>\d+\.\d+)[\s\S]*)|[\s\S]+))`,
```
qingling128, line 115: Can we make the regex for timeStopped and timeStopping more strict to avoid accidentally matching something else and hopefully optimize the regex performance a bit?

e.g.

Match `Total time for which application threads were stopped` instead of `stopped`, and match `Stopping threads took` instead of `took` and match ` seconds $` (end of line). Does that have any down side?

confgenerator/testdata/valid/linux/logging-receiver_cassandra/golden_fluent_bit_parser.conf
```diff
@@ -0,0 +1,224 @@
+[PARSER]
```
qingling128, line 1: Any Cassandra specific invalid user config we need to test against?

docs/cassandra.md
```diff
@@ -0,0 +1,74 @@
...
+
+GC logs contain the following fields in the [`LogEntry`](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry):
+
+| Field | Type | Description |
+| ---   | ---- | ----------- |
+| `jsonPayload.uptime` | number | Seconds the JVM has been active |
+| `jsonPayload.timeStopped` | number | Seconds the JVM was stopped for garbage collection |
+| `jsonPayload.timeStopping` | number | Seconds the JVM took to stop threads before garbage collection |
+| `jsonPayload.message` | string | Log message |
+| `severity` | string ([`LogSeverity`](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity)) | Log entry level (translated) |
```
qingling128, line 73: In the example logs from PR description, it looks like `severity` is missing for GC logs. Do we need to hard code it to `INFO` manually because it's not present in the log itself?

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_cassandra
```diff
@@ -0,0 +1,46 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+logging:
+  receivers:
+    cassandra_system:
+      type: cassandra_system
+      include_paths:
```
qingling128, line 24: Do we have to explicitly configure `include_paths` here? If `/home/test_user/apache-cassandra/logs/XXX` paths are used by default, can we just add them to the default paths? Or is that just something we use for testing (I could not find its reference in the testing bits though)

dehaansa: That's specifically set up for testing SLES where there's no default package install, it's just a manual download, unzip and run. Since kokoro only validates against Debian right now I haven't fully validated that it works in there but it's designed for kokoro testing, not for a default install path.

docs/cassandra.md
dehaansa: This one was a copy paste error, there should be no severity in GC logs.

confgenerator/testdata/valid/linux/logging-receiver_cassandra/golden_fluent_bit_parser.conf
dehaansa: Nothing I could find.

apps/cassandra.go
```diff
@@ -0,0 +1,206 @@
...
+func (LoggingProcessorCassandraDebug) Type() string {
+	return "cassandra_debug"
+}
+
+func (p LoggingProcessorCassandraDebug) Components(tag string, uid string) []fluentbit.Component {
+	return javaLogParsingComponents(tag, uid)
+}
+
+func javaLogParsingComponents(tag string, uid string) []fluentbit.Component {
+	c := confgenerator.LoggingProcessorParseMultilineRegex{
```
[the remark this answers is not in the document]
dehaansa: I don't believe those are other formats, just examples of what the content of the logs could be. Will validate and toss some more examples in the comments.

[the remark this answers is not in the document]
dehaansa: Confirmed that those are just examples of logs in the normal format, documented a few more examples in the comments.

apps/cassandra.go
qingling128: Cool. Resolving this.

apps/cassandra.go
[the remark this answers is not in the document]
qingling128: Great!

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_cassandra
qingling128: Ah I see. Can we add a comment here to clarify why we need to explicitly specify `include_paths`, or customize the Cassandra logs on SLES to write to `/var/log/cassandra/XXX`?

quentinmit: Can you do the SLES case as a separate receiver, so that we can still test the built-in default paths on Debian/Ubuntu/RedHat?

quentinmit (approved):

PR #231, opened 2021-10-06
Refactor component config to be a list of key value pairs

dehaansa:
Requested here https://github.com/GoogleCloudPlatform/ops-agent/pull/225 but it felt like it significantly changed enough code to make it a separate quick PR.

After first implementation & review, new implementation is much simpler and does not affect existing config, just extends ability to generate config with required order or repeated keys.


confgenerator/testdata/valid/linux/metrics-receiver_nginx_missing_status_url/golden_otel.conf
```diff
@@ -356,8 +356,10 @@ receivers:
       process: {}
       processes: {}
   nginx/nginxpipeline_nginxmetrics:
-    collection_interval: 30s
-    endpoint: http://localhost/status
+  - - collection_interval
```
qingling128, line 359: Looks like this part is accidentally broken. Mind fixing it?

confgenerator/fluentbit/modular.go
```diff
@@ -24,22 +24,24 @@ import (
...
+			maxLen = len(line[0])
 		}
 	}
-	for k, v := range c.Config {
-		lines = append(lines, fmt.Sprintf("    %-*s %s", maxLen, k, v))
+
+	for _, line := range c.Config {
+		lines = append(lines, fmt.Sprintf("    %-*s %s", maxLen, line[0], line[1]))
 	}
-	sort.Strings(lines)
```
qingling128, line 42: I vaguely recall we added explicit sorting to ensure the generated config order is deterministic. Is there a reason that prevents us from sorting them in the new structure? If there are repeated keys, we can sort by key first, then value I guess.

Are the field order difference in the `golden_fluent_bit_parser.conf` files below are related to this?

dehaansa: Yes, the field order differences in the golden configs are due to this. The issue with sorting in the new structure is some of the repeated keys require intentional ordering. Rule for multiline might be more flexible, I haven't tested it, but Parser for multi-parser filter definitely uses ordering. We can name them so they always sort correctly, but something mentioned in the other ticket specified "Grep" parsers requiring specifically ordered Exclude and Regex fields, which I'm not as familiar with but might be harder to guarantee ordering correctly.

quentinmit: I think my personal inclination is that it's better overall to just keep two separate fields (a sorted map for almost everything, and RepeatedConfig as `[][2]string` for the very limited cases where we need to control the order of options). I don't think the benefits of the struct simplification outweigh the more cumbersome config construction and the out-of-order generated configs.

jefferbrecht: Sure, I can get on board with that -- it keeps the diff simpler too. One small nit, I'd rename `RepeatedConfig` to `OrderedConfig`.

quentinmit: Oh, gotcha. I didn't see that `RepeatedConfig` was currently still a map.

(BTW, it's definitely a gross hack, but I had originally planned that if we ever needed repeated config in an edge cases it could be "handled" by just putting newlines in the value of a key... I think it's probably far better to build an explicit marshaling interface for it, though.)

jefferbrecht (approved):
LGTM

PR #234, opened 2021-10-12
Jmx jar update

dehaansa:
Update the version of JMX Metrics receiver jar. Currently on latest commit, not on a tag, as this gets us the updated cassandra metrics (https://github.com/open-telemetry/opentelemetry-java-contrib/pull/111).

Also parameterized it for release by placing it in the VERSION file with the PKG_VERSION, if this is a conflict with something that I don't have visibility into or not preferred, let me know and I can create a new BUILD_CONFIG or similar file for this purpose


jsuereth (approved):

VERSION
```diff
@@ -1,2 +1,2 @@
 PKG_VERSION="2.5.0"
-
+JMX_METRICS_JAR_VERSION="1.6.0"
```
igorpeshansky, line 2: This is an internal implementation detail. I think you're right that we should add a `BUILD_CONFIG` file or something. Unless we can extract this info from the `opentelemetry-java-contrib` submodule sources…

dehaansa: I'll see if I can find a good way to dig it out of source, otherwise will refactor to a BUILD_CONFIG file

jsuereth: This version string is meaningless as it stands, you could make it ANY valid number.  This should not be a blocking comment.

jsuereth: Specifically

- The version string is NOT used to generate properties files or any runtime version number (they're still hardcoded)
- The version string is erased on the JAR when we bundle it in ops-agent
- The string just needs to be three valid numbers right now.
- We plan to migrate this build out of the ops-agent buid with an explicit verson choice that's not tied to git submodule.   While there's still an issue with the nebula plugin in gradlew, we're stuck with this kind of hack right now.

dehaansa: No luck on digging out of source, my understanding is that the main point of the nebula plugin is that versioning is all stored in git tags, nothing in source.

jsuereth: Sam - I see you moved it into BUILD_CONFIG.   Unfortuantely that version number isn't actually a valid config parameter.

Can you add a comment to it that it's ONLY used for build wiring but has no impact on the version selected/used?

dehaansa: @jsuereth the way it's used in the Powershell scripting for the Windows build might not play very nice with a comment without some more significant changes, but I can give it a try.

@igorpeshansky we're using this workaround to specify version rather than allowing Nebula plugin to define version because the way CI builds the images doesn't download git config/history for submodules.

PR #238, opened 2021-10-15
Demo for test_config.yaml. Illustrate test issues on other platforms.

martijnvans:
This PR should fail the third_party_apps test. Hopefully the errors are not too hard to disentangle, though this is where it would be really great if the ResultStore team would let us use their UI instead of... I don't know, opening sponge_log.xml in a web browser or something...


qingling128:
Cleaning up obsolete PRs. Feel free to reopen if it's still relevant.

Issue #251, opened 2021-10-26
Include default log rotation rules for subagent logs


yuryu:
Subagent logs can grow quickly and fill up the disk. Currently ops agent doesn't have any logrotate rules for these log files and they continue to accumulate indefinitely.

For example, in one of the VMs I'm running, the logging-module.log file is 2.4 GB after 3 weeks. All installations will eventually run out of disk.

Users could create a rule under `/etc/logrotate.d` but a reasonable default should be provided out of the box.

franciscovalentecastro:
The following PR `Introduce log rotation by default for self log files` https://github.com/GoogleCloudPlatform/ops-agent/pull/1102 released in Ops Agent [2.31.0](https://github.com/GoogleCloudPlatform/ops-agent/releases/tag/2.31.0) provides log rotation by default.

PR #254, opened 2021-10-29
Provide a logrotate conf for subagent logs

yuryu:
The rationale is provided in #251. This PR adds a default logrotate conf for Linux packages.

The config is based on the default debian syslog rule, and does:
- maintain up to four copies (up to a month)
- rotate weekly
- don't fail if log files are missing
- don't rotate if log is empty
- compress logs two generations older

We don't need to restart fluent-bit because currently fluent-bit opens and closes log files every time it writes something.


pkg/rpm/google-cloud-ops-agent.spec
```diff
@@ -46,6 +47,7 @@ CODE_VERSION=%{version} BUILD_DISTRO=${build_distro#.} DESTDIR="%{buildroot}" ./
 %{_prefix}/libexec/google_cloud_ops_agent_engine
 %{_unitdir}/%{name}*
 %{_unitdir}-preset/*-%{name}*
+%{_logrotatedir}/google-cloud-ops-agent
```
igorpeshansky, line 50: Should this be marked as "config"? I assume users will potentially want to edit this…

default-logrotate-conf
```diff
@@ -0,0 +1,9 @@
+/var/log/google-cloud-ops-agent/subagents/logging-module.log
+{
+        rotate 4
```
igorpeshansky, line 3: We probably want `create` here as well…

build.sh
```diff
@@ -116,6 +116,8 @@ function build_systemd() {
 # TODO: Build sample config file
 mkdir -p "$DESTDIR/$sysconfdir/google-cloud-ops-agent/"
 cp "confgenerator/default-config.yaml" "$DESTDIR/$sysconfdir/google-cloud-ops-agent/config.yaml"
+mkdir -p "$DESTDIR/$sysconfdir/logrotate.d"
+cp "default-logrotate-conf" "$DESTDIR/$sysconfdir/logrotate.d/google-cloud-ops-agent"
```
igorpeshansky, line 120: It probably doesn't make sense to pollute the top-level directory with miscellaneous configs. I would suggest creating an `etc/logrotate.d` directory at the top level, but I'll let @hsmatulisgoogle make that call…

igorpeshansky (changes requested):

igorpeshansky (commented):
Forgot to ask — what about Windows? Wouldn't it suffer from the same issues?

yuryu:
I'll address the comments shortly. I haven't used it on Windows and don't know, but it looks like fluent-bit is [configured](https://github.com/GoogleCloudPlatform/ops-agent/blob/6a9855a5647a6bea5214e87ae5f30c5baf618739/cmd/ops_agent_windows/main_windows.go#L123) to produce a log file on Windows as well? In that case I'd assume yes. Windows doesn't have logrotate so we would need to work around it (I think many services use the system event log instead of local files).

build.sh
yuryu: moved to etc/logrotate.d. Let me know if another path is preferrable.

jeremyvisser:
We have a fleet of RHEL and OEL machines affected by this issue, and while @igorpeshansky's suggestion to implement Windows log rotation is made in good faith, it may be the perfect preventing good from happening.

I think that Windows log rotation is a valid thing to ask for, but can we merge the Linux–only log rotation for the time being since the log files in the wild are not getting any smaller?

inglor:
I just investigated an issue with one of the VMs running out of space and the culprit was the logging from the agent resulted in 22G file. Please merge this.

yuryu:
I don't know why the check is failing. I didn't change anything about Windows but somehow it's Windows that's failing? Let me know if there's anything I can do.

PR #260, opened 2021-11-02
docs: add step to update submodules

ridwanmsharif:
This change adds a step that updates a submodule during build
optionally. I've forgotten to do this a few times and this tip should
help avoid this common issue.


igorpeshansky (approved):
LGTM :shipit:

PR #264, opened 2021-11-06
Don't trim the "agent.googleapis.com/" prefix in `exclude_metrics`.

igorpeshansky:
This prevented the generated regexps from matching the new metric names with the prefix added.

Also escape `$` characters in generated Otel collector configs for `exclude_metrics`.

http://b/205077265.

### DIFF
--- a/apps/metric_processors.go
+++ b/apps/metric_processors.go
@@ -36,8 +36,6 @@ func (p MetricsProcessorExcludeMetrics) Type() string {
 func (p MetricsProcessorExcludeMetrics) Processors() []otel.Component {
 	var metricNames []string
 	for _, glob := range p.MetricsPattern {
-		// TODO: Remove TrimPrefix when we support metrics with other prefixes.
-		glob = strings.TrimPrefix(glob, "agent.googleapis.com/")
 		var literals []string
 		for _, g := range strings.Split(glob, "*") {
 			literals = append(literals, regexp.QuoteMeta(g))
--- a/apps/metric_processors.go
+++ b/apps/metric_processors.go
@@ -36,11 +36,13 @@ func (p MetricsProcessorExcludeMetrics) Type() string {
 func (p MetricsProcessorExcludeMetrics) Processors() []otel.Component {
 	var metricNames []string
 	for _, glob := range p.MetricsPattern {
+		// $ needs to be escaped because reasons.
+		// https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricstransformprocessor#rename-multiple-metrics-using-substitution
 		var literals []string
 		for _, g := range strings.Split(glob, "*") {
-			literals = append(literals, regexp.QuoteMeta(g))
+			literals = append(literals, strings.Replace(regexp.QuoteMeta(g), "$", "$$", -1))
 		}
-		metricNames = append(metricNames, fmt.Sprintf(`^%s$`, strings.Join(literals, `.*`)))
+		metricNames = append(metricNames, fmt.Sprintf(`^%s$$`, strings.Join(literals, `.*`)))
 	}
 	return []otel.Component{otel.MetricsFilter(
 		"exclude",
--- a/confgenerator/testdata/valid/linux/metrics-default_overrides_disable_all/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/metrics-default_overrides_disable_all/golden_otel.conf
@@ -32,7 +32,7 @@ processors:
       exclude:
         match_type: regexp
         metric_names:
-        - ^.*$
+        - "^agent\\.googleapis\\.com/.*$"
   metricstransform/agent_1:
     transforms:
     - action: update
--- a/confgenerator/testdata/valid/linux/metrics-default_overrides_disable_all/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/metrics-default_overrides_disable_all/golden_otel.conf
@@ -32,7 +32,7 @@ processors:
       exclude:
         match_type: regexp
         metric_names:
-        - "^agent\\.googleapis\\.com/.*$"
+        - "^agent\\.googleapis\\.com/.*$$"
   metricstransform/agent_1:
     transforms:
     - action: update
--- a/confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_with_special_chars/input.yaml
+++ b/confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_with_special_chars/input.yaml
@@ -1,0 +1,22 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+metrics:
+  processors:
+    metrics_filter:
+      type: exclude_metrics
+      metrics_pattern:
+      # Escaped the comma to work around https://github.com/goccy/go-yaml/issues/269
+      # TODO: remove the escape once the issue is fixed.
+      - agent.googleapis.com/x$y(%28)z\,a':w-+!/*
--- a/confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_by_prefixes/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_by_prefixes/golden_otel.conf
@@ -32,8 +32,8 @@ processors:
       exclude:
         match_type: regexp
         metric_names:
-        - ^processes/.*$
-        - ^cpu/.*$
+        - "^agent\\.googleapis\\.com/processes/.*$"
+        - "^agent\\.googleapis\\.com/cpu/.*$"
   metricstransform/agent_1:
     transforms:
     - action: update
--- a/confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_by_prefixes/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_by_prefixes/golden_otel.conf
@@ -32,8 +32,8 @@ processors:
       exclude:
         match_type: regexp
         metric_names:
-        - "^agent\\.googleapis\\.com/processes/.*$"
-        - "^agent\\.googleapis\\.com/cpu/.*$"
+        - "^agent\\.googleapis\\.com/processes/.*$$"
+        - "^agent\\.googleapis\\.com/cpu/.*$$"
   metricstransform/agent_1:
     transforms:
     - action: update
--- a/confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_with_globs/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_with_globs/golden_otel.conf
@@ -32,8 +32,8 @@ processors:
       exclude:
         match_type: regexp
         metric_names:
-        - "^agent\\.googleapis\\.com/proce.*ses/.*$"
-        - "^agent\\.googleapis\\.com/c.*u/.*$"
+        - "^agent\\.googleapis\\.com/proce.*ses/.*$$"
+        - "^agent\\.googleapis\\.com/c.*u/.*$$"
   metricstransform/agent_1:
     transforms:
     - action: update
--- a/confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_with_globs/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_with_globs/golden_otel.conf
@@ -32,8 +32,8 @@ processors:
       exclude:
         match_type: regexp
         metric_names:
-        - ^proce.*ses/.*$
-        - ^c.*u/.*$
+        - "^agent\\.googleapis\\.com/proce.*ses/.*$"
+        - "^agent\\.googleapis\\.com/c.*u/.*$"
   metricstransform/agent_1:
     transforms:
     - action: update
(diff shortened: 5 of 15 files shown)


qingling128 (approved):
LGTM

apps/metric_processors.go
```diff
@@ -42,7 +41,9 @@ func (p MetricsProcessorExcludeMetrics) Processors() []otel.Component {
 		for _, g := range strings.Split(glob, "*") {
 			literals = append(literals, regexp.QuoteMeta(g))
```
quentinmit, line 42: I don't think the `QuoteMeta` is going to escape `$` properly (it's presumably going to do `\$` instead of `$$`)

quentinmit (changes requested):

apps/metric_processors.go
quentinmit: Actually I think maybe it needs to be `\$$`? So do a `strings.Replace` after `QuoteMeta`?

igorpeshansky: Are you talking about the case where the customer puts a `$` into an `exclude_metrics` glob? Is `$` even a valid character in metric names [1]? Should we just add validation instead?

[1] https://cloud.google.com/monitoring/api/v3/naming-conventions#naming-types-and-labels

igorpeshansky (commented):
FYI, since this was not about quoting `$` signs, I'm about to split the IIS changes into a separate PR and squash the `exclude_metrics` commits together. I'll request a re-review after.

jkschulz (approved):
LGTM pending testing

qingling128:
> LGTM pending testing

Do we have a tracking bug for adding an integration test for `exclude_metrics`?

igorpeshansky (commented):
> LGTM pending testing

@jkschulz This was tested manually; removing the anchors fixes the behavior.

> Do we have a tracking bug for adding an integration test for `exclude_metrics`?

@qingling128 I'll create one.

apps/metric_processors.go
quentinmit: Yeah, I'm talking about the case where `$` is in the glob. Even if it's not allowed by GCM, I think OT still allows it, and either way we shouldn't generate a bad config file. I'm happy with properly escaping it or forbidding it with a validator.

igorpeshansky: Added validation and a test that specifically checks for a `$` in the glob.

confgenerator/testdata/invalid/linux/metrics-processor_exclude_metrics_invalid_parameter_metric_prefixes/golden_error
```diff
@@ -1,9 +1,9 @@
-[6:7] "metrics_pattern[0]" must end with "/*","metrics_pattern[1]" must start with "agent.googleapis.com/"
+[6:7] "metrics_pattern[0]" must end with "/*","metrics_pattern[1]" must start with "agent.googleapis.com/","metrics_pattern[2]" must be a metric type glob pattern
```
sophieyfang, line 1: Shall this be [6:8]? As you added a new metrics pattern in here on line 8 which is wrong in syntax

igorpeshansky: That is an excellent question, but this is, in fact, the exact error message produced by the validation library (generated via `-update_golden`). There may be a bug in our interface to the `yaml` library's validator functionality. Perhaps @quentinmit has some insight here…

apps/metric_processors.go
igorpeshansky: Just realized that this newly-added validation will eventually reject patterns for some of our third-party workload metrics (e.g., `workload.googleapis.com/jvm.memory.heap.*` — this pattern is not possible today because of the `endswith=/*` rule, but will become reasonable once that restriction is relaxed). Should we change the validation now, or wait until it actually becomes an issue (I assume we're not renaming the metrics at this point)?
@qingling128 FYI.

qingling128: Relaxing this validation would make it possible for users to exclude specific metrics. I think that's useful, but let's run it by Harp via internal channel for potential concerns he had expressed before.

confgenerator/testdata/invalid/linux/metrics-processor_exclude_metrics_invalid_parameter_metric_prefixes/golden_error
igorpeshansky: Ohh. Never mind. The "7" in "[6:7]" is the _column_, not the end line of the range ([code](https://github.com/goccy/go-yaml/blob/master/internal/errors/error.go#L186)). Looks like we don't actually track the range of lines for multiple errors.

apps/metric_processors.go
```diff
@@ -42,7 +41,9 @@ func (p MetricsProcessorExcludeMetrics) Processors() []otel.Component {
...
 		}
-		metricNames = append(metricNames, fmt.Sprintf(`^%s$`, strings.Join(literals, `.*`)))
+		// All filter regexes are automatically anchored.
+		// https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/coreinternal/processor/filterset/regexp/regexpfilterset.go#L79
```
sophieyfang, line 45: It's not anchored. Though it did solve the problem.

apps/metric_processors.go
igorpeshansky: Yeah, the naming conventions doc referenced above seems inaccurate. I'm trying to confirm the actual validation, so we can determine whether `$` can appear in metric types. b/207141764 tracks the doc update.

apps/metric_processors.go
igorpeshansky: Discussed offline. @sophieyfang found the actual cause of this, which was #151 not updating [the removal of the prefix](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/apps/metric_processors.go#L40). I'll update this PR and request another review after testing.

apps/metric_processors.go
igorpeshansky: We are currently also enforcing that any metric patterns in `exclude_metrics` filters start with `agent.googleapis.com/`, which will also prevent excluding workload metrics. If we do relax the restrictions, we should relax both of those.

That said, I've removed the validation and added the code to escape the `$` instead.

igorpeshansky:
Looks like this PR needs to be rebased off the latest master.

apps/metric_processors.go
```diff
@@ -36,13 +36,13 @@ func (p MetricsProcessorExcludeMetrics) Type() string {
...
 	var metricNames []string
 	for _, glob := range p.MetricsPattern {
-		// TODO: Remove TrimPrefix when we support metrics with other prefixes.
-		glob = strings.TrimPrefix(glob, "agent.googleapis.com/")
+		// $ needs to be escaped because reasons.
+		// https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricstransformprocessor#rename-multiple-metrics-using-substitution
 		var literals []string
 		for _, g := range strings.Split(glob, "*") {
-			literals = append(literals, regexp.QuoteMeta(g))
+			literals = append(literals, strings.Replace(regexp.QuoteMeta(g), "$", "$$", -1))
```
sophieyfang, line 43: I think filterprocessor doesn't need to escape $

confgenerator/testdata/valid/linux/metrics-default_overrides_disable_all/golden_otel.conf
```diff
@@ -32,7 +32,7 @@ processors:
       exclude:
         match_type: regexp
         metric_names:
-        - ^.*$
+        - "^agent\\.googleapis\\.com/.*$$"
```
sophieyfang, line 35: I don't think that it needs "$$"

sophieyfang (approved):

apps/metric_processors.go
quentinmit: Why did you remove this escaping? `$` is special in the otel configuration; it starts an environment variable expansion.

confgenerator/testdata/valid/linux/metrics-default_overrides_disable_all/golden_otel.conf
quentinmit: I believe the `$$` is needed to properly handle the future case where it doesn't end in `/*`.

apps/metric_processors.go
sophieyfang: when I tested, ^$ works just fine. What’s the context that it needs to be escaped as $$? I saw metricstransformer uses $$ in recent changes. I don’t be think filterprocessor needs it.

sophieyfang: I just confirmed again it works just fine with a single $.

```
curl -sSO https://dl.google.com/cloudagents/add-google-cloud-ops-agent-repo.sh
sudo bash add-google-cloud-ops-agent-repo.sh --also-install
sudo service google-cloud-ops-agent.service stop
sudo vim /run/google-cloud-ops-agent-opentelemetry-collector/otel.yaml <---modify this file to exclude `^processes/.*$` and let it output to logging exporter
sudo /opt/google-cloud-ops-agent/subagents/opentelemetry-collector/otelopscol --add-instance-id=false --config=/run/google-cloud-ops-agent-opentelemetry-collector/otel.yaml
```

confgenerator/testdata/valid/linux/metrics-default_overrides_disable_all/golden_otel.conf
sophieyfang: In regexp mode, the metrics names are just any regular expressions.  Why it will stop working when it not ends with /.*?

igorpeshansky: I've tested with the single `$` — it looks like only metricstransform needed the escaping.

apps/metric_processors.go
igorpeshansky: I've also successfully tested with the single `$` — it looks like only metricstransform needed the escaping.

apps/metric_processors.go
igorpeshansky: A-and it turns out we don't need to escape the `$` in filterprocessor.

igorpeshansky:
This has been tested successfully, and is now ready for review.

apps/metric_processors.go
quentinmit: @sophieyfang I don't think your test is valid because `^agent.*processes/.*$` and `^agent.*processes/.*` have the same behavior. 

@igorpeshansky What exactly did you test? I would expect `^agent.*processes/c` to exclude `agent/processes/count_by_state` but `^agent.*processes/c$` to not. So if it's treating the latter as the former, that would be the behavior change.

Also it's possible a trailing $ is treated specially; the other case to test is `^workload.*foo\$bar`, which needs to match a metric named `foo$bar` (and not just be a failure to match anything). The latter is what the `strings.Replace` is for.

confgenerator/testdata/valid/linux/metrics-default_overrides_disable_all/golden_otel.conf
quentinmit: There's nothing specific to `metricstransform` - https://opentelemetry.io/docs/collector/configuration/#configuration-environment-variables

apps/metric_processors.go
sophieyfang: Tested again with:
“””
filter/default__pipeline_hostmetrics_3:
    metrics:
      include:
        match_type: regexp
        metric_names:
        - ^agent.googleapis.com/processes/disk/read_bytes_count$
 “””
It does only outputread_bytes_count metrics when I tested it. What error/mis-behavior is expected by this expanding of $? I don’t see it emit any error or stop outputting metrics

I agree that .*$ works the same as .*

confgenerator/testdata/valid/linux/metrics-default_overrides_disable_all/golden_otel.conf
igorpeshansky: Ah. That explains why it worked without escaping in this specific case — `os.Expand` ignores trailing `$` signs. Reverted to the escaping commit.

apps/metric_processors.go
igorpeshansky: I've tested the code that generates a single `$` within the current constraints, which means that the regexp ended in a `.*`. The first example you gave isn't valid as of today. The second example (`foo$bar`) is harder to test, since no such metric is generated.

I've just tried something like this: `metrics_pattern: ["agent.googleapis.com/processes$XYZ/*"]`, and there's definitely substitution going on, as the following got logged:
```
Nov 24 19:27:06 b-205077265 otelopscol[16397]: 2021-11-24T19:27:06.139Z        info        filterprocessor@v0.37.1-0.20211028205244-e6fab4102b84/filter_processor.go:73        Metric filter configured        {"kind": "processor", "name": "filter/default__pipeline_hostmetrics_3", "include match_type": "", "include expressions": [], "include metric names": [], "include metrics with resource attributes": null, "exclude match_type": "regexp", "exclude expressions": [], "exclude metric names": ["^agent\\.googleapis\\.com/processes\\/.*$"], "exclude metrics with resource attributes": null}
```

The trailing `$` is indeed treated differently by `os.Expand`, which explains why our tests worked. I've reverted to the commit that does the escaping.

sophieyfang (approved):

apps/metric_processors.go
igorpeshansky: The trailing `$` happens to work when un-escaped, but we do need the escapes. Sigh.

quentinmit (approved):

apps/metric_processors.go
sophieyfang: $$ sounds good to me

sophieyfang (approved):

confgenerator/testdata/valid/linux/metrics-processor_exclude_metrics_type_filter_with_special_chars/input.yaml
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+metrics:
+  processors:
+    metrics_filter:
+      type: exclude_metrics
+      metrics_pattern:
+      # Escaped the comma to work around https://github.com/goccy/go-yaml/issues/269
+      # TODO: remove the escape once the issue is fixed.
+      - agent.googleapis.com/x$y(%28)z\,a':w-+!/*
```
jkschulz, line 22: this is valid? should we be failing config validation if this doesn't match monitoring api [defined format](http://cloud/monitoring/api/v3/naming-conventions) (plus `.*`)? fine for that to be a separate pr doe

igorpeshansky: Strangely (and sadly), this is a valid metric name. Granted, it's not one that would ever be produced with the `agent.googleapis.com/` prefix, but that's not what this test verifies. The page you are referencing is outdated — see b/207141764.

PR #268, opened 2021-11-10
Update Redis Config Options

Mrod1598:
Creating this PR to reopen the discussion about TLS. There doesn't seem to be a reasonable default for where the certs would live. Although, otel does allow that as a valid config option for TLS.


apps/redis.go
```diff
@@ -37,18 +41,31 @@ func (r MetricsReceiverRedis) Type() string {
...
 
 func (r MetricsReceiverRedis) Pipelines() []otel.Pipeline {
-
 	if r.Address == "" {
 		r.Address = defaultRedisEndpoint
 	}
 
+	var insecure string
+	if r.CertFilePath != "" || r.KeyFilePath != "" || r.CaFilePath != "" {
+		insecure = "false"
```
quentinmit, line 50: `CaFilePath == ""` doesn't imply `insecure=="false"` because you can use the system default trust store to validate the server certificate.

confgenerator/config.go
```diff
@@ -94,6 +94,8 @@ func (ve validationError) Error() string {
 		return fmt.Sprintf("%q must be one of [%s]", ve.Field(), ve.Param())
 	case "required":
 		return fmt.Sprintf("%q is a required field", ve.Field())
+	case "required_with":
+		return fmt.Sprintf("%q is a required to be set alongw with the '%s' field", ve.Field(), ve.Param())
```
quentinmit, line 98: `"%q is required if %q is specified"` perhaps?

confgenerator/testdata/valid/linux/metrics-receiver_redis_missing_address/golden_otel.conf
```diff
@@ -386,6 +386,12 @@ receivers:
     collection_interval: 30s
     endpoint: localhost:6379
     password: pwd
+    tls:
```
quentinmit, line 389: Does OT treat all of these being set to "" the same as not specifying TLS at all?

Mrod1598: pretty much, it sets `insecure: true` as the default.

confgenerator/config.go
Mrod1598: Updated

apps/redis.go
Mrod1598: I think what you meant to say is `CaFilePath == ""` doesn't imply `insecure=="true"` because that means don't use tls and is currently how it's set up. Do what we want to do then is just set it to false if nothing is set so TLS IS enabled?

quentinmit: I'm not sure what you mean, so let me try to just enumerate the possible states an HTTP client socket can be in:

- HTTP (no SSL/TLS) - I think this is what `insecure: true` means
- HTTPS with server certificate validated with system trust store - `insecure: false` with no other settings
- HTTPS with server certificate validated with a custom certificate authority - `insecure: false` and `ca_file_path: /foo/bar`
- HTTPS with no server certificate validation - `insecure: false` with `insecure_skip_verify: true`

and then the last three repeated with the addition of a client certificate with `cert_file_path` and `key_file_path` specified.

All 7 of those are different states and it should be possible to configure all of them. And the default behavior should either be the first one or the second one depending on whether the server defaults to using SSL or not using SSL.

Mrod1598: ah ok, appreciate that, and from [what I'm seeing](https://redis.io/topics/encryption#:~:text=By%20default%2C%20Redis%20uses%20mutual,ca%2Dcert%2Ddir%20\).), Redis does not use TLS/SSL by default so the default should be `insecure: true`.

qingling128: Reached agreement offline. See the common util shared by https://github.com/GoogleCloudPlatform/ops-agent/pull/320 for the expected implementation.

qingling128:
TLS related convention has been established offline. https://github.com/GoogleCloudPlatform/ops-agent/pull/320 has implemented common utils for this, and the doc was shared offline.

Next step for this PR:
* Rebase off master to pick up the change in the PR above
* Refactor it to use the common utils

Mrod1598:
The default TLS for Redis is `insecure: true`. https://redis.io/topics/encryption

docs/redis.md
```diff
@@ -78,6 +86,8 @@ To configure a receiver for your redis logs, specify the following fields:
 | `exclude_paths`       | `[]`                          | A list of filesystem path patterns to exclude from the set matched by `include_paths`.
 
 
+
+
 Example Configuration:
 
 ```yaml
```
qingling128, line 93: Looks like there are some `submodules/fluent-bit` changes that accidentally got in. Might need rebasing?

qingling128 (approved):
LGTM modulo the conflicts against master HEAD is resolved and the `submodules/fluent-bit` diff was addressed.

docs/redis.md
```diff
@@ -16,6 +16,14 @@ To configure a receiver for your redis metrics, specify the following fields:
 | `collection_interval` | `60s`                     | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
 | `password`            |                           | The password used to connect to the server. |
 
+optional fields are: 
+| Field                 | Default                   | Description |
```
qingling128, line 20: What about `insecure` and `insecure_skip_verify`?

Mrod1598: So,  since docs are now pointing to documentation outside this repo from this [PR](https://github.com/GoogleCloudPlatform/ops-agent/pull/366), how are updates to the documentation going to be added?

docs/redis.md
Mrod1598: Fixed

docs/redis.md
qingling128: We are in the middle of automating the documentation related process.

For the moment, the general workflow is:
1. Documentation temporarily goes to GitHub markdown
2. Public doc at https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party is derived from the GitHub markdown
3. Once 2 is published, we remove the markdown to avoid having to maintain 2 copies of things.

For this specific case of Redis, options are:
* Leave a section like `Optional fields` in the markdown for now. We'll delete it once the content has been moved to the public side.
* Leave a comment with the documentation change required. We'll take it and add it to the public doc. No need for a follow up cleanup PR.

qingling128 (approved):
LGTM

docs/redis.md
Mrod1598: Added comment in doc

qingling128 (approved):
LGTM

confgenerator/testdata/valid/linux/metrics-receiver_redis_missing_address/golden_otel.conf
```diff
@@ -426,6 +426,8 @@ receivers:
...
+      insecure: true
```
quentinmit, line 430: What integration test should be covering this change? I don't see any changes to integration tests in this PR.

qingling128: Good question.

The current integration test setup only test one case for each app on all representative distros. Might need some work there to support testing multiple scenarios for one application.

A workaround is to do something like https://github.com/GoogleCloudPlatform/ops-agent/tree/master/integration_test/third_party_apps_data/applications/shared_tls

To balance the test flakiness concerns and code coverage need, it's probably ok for the shared TLS util to trigger tests on one distro, assuming the underlying logic is identical across distros.

Does the upstream Open Telemetry collector have any TLS related tests?

Mrod1598: Doesn't look like it.

Mrod1598: Would you like the shared_TLS in this PR or is that a seperate PR to be made?

qingling128: Either way is OK. If this PR has been tested manually, we can get it in first then add integration tests in a follow up one.

Mrod1598: verified that it is working through manual testing again with the updates. 
Screen Shot 2022-02-07 at 8 52 11 PM

Mrod1598:
Failures are related to couchdb and tomcat

confgenerator/testdata/valid/linux/metrics-receiver_redis_missing_address/golden_otel.conf
qingling128: Thanks!

qingling128 (approved):
LGTM

PR #269, opened 2021-11-11
Adjust log name via logging.googleapis.com/logName instead of using a rewrite_tag.

qingling128:
The current implementation has a section that changes log tag from `{PIPELINE_ID}`.`{RECEIVER_ID}` to `{RECEIVER_ID}`, because the output plugin decides the LogEntry.LogName based on the log tag, and we do not want to expose `{PIPELINE_ID}` in log name. This involves a `rewrite_tag` filter that is very heavy performance wise (It throws all log entries back to the pool to be re-processed.

This change refactor the config to make use of the `logging.googleapis.com/logName` field instead, which is a field that the output plugin recognizes already.

## Old Config

```
[FILTER]
    Add   logName syslog
    Match default_pipeline.syslog
    Name  modify

[FILTER]
    Emitter_Mem_Buf_Limit 10M
    Emitter_Storage.type  filesystem
    Match                 default_pipeline.syslog
    Name                  rewrite_tag
    Rule                  $logName .* $logName false

[FILTER]
    Match  syslog
    Name   modify
    Remove logName

[OUTPUT]
    Match_Regex       ^(syslog)$
    Name              stackdriver
    Retry_Limit       3
```

# New config

```
[FILTER]
    Add   logging.googleapis.com/logName syslog
    Match default_pipeline.syslog
    Name  modify

[OUTPUT]
    Match_Regex       ^(default_pipeline\.syslog)$
    Name              stackdriver
    Retry_Limit       3
```

Manually tested: b/206012639#comment2

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -137,7 +137,7 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
 			components []fluentbit.Component
 		}
 		var sources []fbSource
-		var logNames []string
+		var tags []string
 		for pID, p := range l.Service.Pipelines {
 			for _, rID := range p.ReceiverIDs {
 				receiver, ok := l.Receivers[rID]
@@ -157,18 +157,18 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
 					components = append(components, processor.Components(tag, strconv.Itoa(i))...)
 				}
 				components = append(components, setLogNameComponents(tag, rID)...)
-				logNames = append(logNames, regexp.QuoteMeta(tag))
+				tags = append(tags, regexp.QuoteMeta(tag))
 				sources = append(sources, fbSource{tag, components})
 			}
 		}
 		sort.Slice(sources, func(i, j int) bool { return sources[i].tag < sources[j].tag })
-		sort.Strings(logNames)
+		sort.Strings(tags)
 
 		for _, s := range sources {
 			out = append(out, s.components...)
 		}
-		if len(logNames) > 0 {
-			out = append(out, stackdriverOutputComponent(strings.Join(logNames, "|"), userAgent))
+		if len(tags) > 0 {
+			out = append(out, stackdriverOutputComponent(strings.Join(tags, "|"), userAgent))
 		}
 	}
 	out = append(out, LoggingReceiverFilesMixin{
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -157,7 +157,7 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
 					components = append(components, processor.Components(tag, strconv.Itoa(i))...)
 				}
 				components = append(components, setLogNameComponents(tag, rID)...)
-				logNames = append(logNames, regexp.QuoteMeta(rID))
+				logNames = append(logNames, regexp.QuoteMeta(tag))
 				sources = append(sources, fbSource{tag, components})
 			}
 		}
--- a/confgenerator/logging.go
+++ b/confgenerator/logging.go
@@ -22,32 +22,13 @@ import (
 
 // setLogNameComponents generates a series of components that rewrites the tag on log entries tagged `tag` to be `logName`.
 func setLogNameComponents(tag, logName string) []fluentbit.Component {
-	// TODO: Can we just set log_name_key in the output plugin and avoid this mess?
 	return []fluentbit.Component{
 		{
 			Kind: "FILTER",
 			Config: map[string]string{
 				"Match": tag,
-				"Add":   fmt.Sprintf("logName %s", logName),
+				"Add":   fmt.Sprintf("logging.googleapis.com/logName %s", logName),
 				"Name":  "modify",
-			},
-		},
-		{
-			Kind: "FILTER",
-			Config: map[string]string{
-				"Emitter_Mem_Buf_Limit": "10M",
-				"Emitter_Storage.type":  "filesystem",
-				"Match":                 tag,
-				"Name":                  "rewrite_tag",
-				"Rule":                  "$logName .* $logName false",
-			},
-		},
-		{
-			Kind: "FILTER",
-			Config: map[string]string{
-				"Match":  logName,
-				"Name":   "modify",
-				"Remove": "logName",
 			},
 		},
 	}
--- a/confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_fluent_bit_main.conf
+++ b/confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_fluent_bit_main.conf
@@ -43,24 +43,12 @@
     storage.type      filesystem
 
 [FILTER]
-    Add   logName syslog
+    Add   logging.googleapis.com/logName syslog
     Match default_pipeline.syslog
     Name  modify
 
-[FILTER]
-    Emitter_Mem_Buf_Limit 10M
-    Emitter_Storage.type  filesystem
-    Match                 default_pipeline.syslog
-    Name                  rewrite_tag
-    Rule                  $logName .* $logName false
-
-[FILTER]
-    Match  syslog
-    Name   modify
-    Remove logName
-
 [OUTPUT]
-    Match_Regex       ^(syslog)$
+    Match_Regex       ^(default_pipeline\.syslog)$
     Name              stackdriver
     Retry_Limit       3
     resource          gce_instance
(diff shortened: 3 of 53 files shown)


jkschulz (approved):
YAS SO EXCITED FOR THIS

quentinmit:
How much of a performance improvement is this? While it's true that `rewrite_tag` reprocesses the logs, none of the filters should match the new tag so I would expect it to be pretty light-weight.

qingling128:
> How much of a performance improvement is this? While it's true that `rewrite_tag` reprocesses the logs, none of the filters should match the new tag so I would expect it to be pretty light-weight.

`rewrite_tag` also triggers the log entries to be taken out of memory, stored on disk for buffering / queuing again, then loaded back again into memory. Also it has some bugs (some are fixed, but we are still seeing new issues under high load) on its own, causing things to pile up / fail to be registered back to the pool.

qingling128:
Tests are failing for probably unrelated reasons. Will hold on this PR until the tests are recovered.

confgenerator/confgenerator.go
```diff
@@ -157,7 +157,7 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
 					components = append(components, processor.Components(tag, strconv.Itoa(i))...)
 				}
 				components = append(components, setLogNameComponents(tag, rID)...)
-				logNames = append(logNames, regexp.QuoteMeta(rID))
+				logNames = append(logNames, regexp.QuoteMeta(tag))
```
igorpeshansky, line 160: This is no longer `logNames`, is it? Want to call it `tags` or `tagRegexps` instead?

igorpeshansky (changes requested):

confgenerator/confgenerator.go
qingling128: Changed to `tags`.

igorpeshansky (approved):
LGTM :shipit:

sophieyfang:
For the LogsViewer UI, after this change, does it now show as `logging.googleapis.com/syslog` instead of `syslog`?

According to  below configuration, the fluentbit modify filter will add a new key `logging.googleapis.com/syslog` into the record. Do you mean the stackdriver output plugin recognizes pattern of `logging.googleapis.com/.*` and it would prefer (syslog in this case) over tag name `default_pipeline.syslog`?

```
This change refactor the config to make use of the logging.googleapis.com/logName field instead, which is a field that the output plugin recognizes already.
```

```
[FILTER]
    Add   logging.googleapis.com/logName syslog
    Match default_pipeline.syslog
    Name  modify

[OUTPUT]
    Match_Regex       ^(default_pipeline\.syslog)$
    Name              stackdriver
    Retry_Limit       3
```

sophieyfang:
> For the LogsViewer UI, after this change, does it now show as `logging.googleapis.com/syslog` instead of `syslog`?
> 
> According to below configuration, the fluentbit modify filter will add a new key `logging.googleapis.com/syslog` into the record. Do you mean the stackdriver output plugin recognizes pattern of `logging.googleapis.com/.*` and it would prefer (syslog in this case) over tag name `default_pipeline.syslog`?
> 
> ```
> This change refactor the config to make use of the logging.googleapis.com/logName field instead, which is a field that the output plugin recognizes already.
> ```
> 
> ```
> [FILTER]
>     Add   logging.googleapis.com/logName syslog
>     Match default_pipeline.syslog
>     Name  modify
> 
> [OUTPUT]
>     Match_Regex       ^(default_pipeline\.syslog)$
>     Name              stackdriver
>     Retry_Limit       3
>     resource          gce_instance
> ```

Actually a new key value pair as `"logging.googleapis.com/logName": "syslog"` into the record. If stackdriver output plugin recognizes of `logging.googleapis.com/logName` and it would prefer (syslog in this case) over tag name `default_pipeline.syslog`. We shall update the stackdriver output plugin document to reflect the case. Now It feels a secret usecase

qingling128:
The resulting log entry in the UI looks identical as before: b/206012639#comment2. The output plugin recognizes a `logging.googleapis.com/logName` and move its value to `LogEntry.LogName` instead (aka it will remove the `logging.googleapis.com/logName` field after extracting the value.

sophieyfang:
> The resulting log entry in the UI looks identical as before: b/206012639#comment2. The output plugin recognizes a `logging.googleapis.com/logName` and move its value to `LogEntry.LogName` instead (aka it will remove the `logging.googleapis.com/logName` field after extracting the value.

Then below description seems not  right? More info: https://docs.fluentbit.io/manual/pipeline/outputs/stackdriver
We didn't set `labels_key` in the output plugin,  but still it's doing the things. We shall update this [public doc](https://docs.fluentbit.io/manual/pipeline/outputs/stackdriver) i guess.
image

sophieyfang:
Oh.. by default it's `logging.googleapis.com/labels` and `labels` can be any string and for our case in here, it's logName..i feel like it deserves a good example for explanation, i will send a PR in that public doc.

jkschulz:
@sophieyfang `labels` and `logName` are [separate](https://github.com/fluent/fluent-bit/blob/0ba2dc2196e4c2c19a6075f8154afe7be137119d/plugins/out_stackdriver/stackdriver.c#L1820-L1821) things (see http://cloud/logging/docs/reference/v2/rest/v2/LogEntry). 
[`labels_key`](https://github.com/fluent/fluent-bit/blob/0ba2dc2196e4c2c19a6075f8154afe7be137119d/plugins/out_stackdriver/stackdriver_conf.c#L428-L434) is defined in the documentation, but [`log_name_key`](https://github.com/fluent/fluent-bit/blob/0ba2dc2196e4c2c19a6075f8154afe7be137119d/plugins/out_stackdriver/stackdriver_conf.c#L317-L323) is not yet (though it should be). 

`log_name_key` works the same way, but for `logName`, using a default value of `logging.googleapis.com/logName`.

So
- if you set `log_name_key` to `nameA` in the output plugin, Fluent Bit will 
  - look for a [key named `nameA`](https://github.com/fluent/fluent-bit/blob/0ba2dc2196e4c2c19a6075f8154afe7be137119d/plugins/out_stackdriver/stackdriver_conf.c#L317-L319) in each record,
  - [extract its value](https://github.com/fluent/fluent-bit/blob/0ba2dc2196e4c2c19a6075f8154afe7be137119d/plugins/out_stackdriver/stackdriver.c#L1845-L1850) if it exists,
  - and [use that as the `LOG_ID` in `logName`](https://github.com/fluent/fluent-bit/blob/0ba2dc2196e4c2c19a6075f8154afe7be137119d/plugins/out_stackdriver/stackdriver.c#L2015-L2021) 
      - 
- if `log_name_key` [isn't specified](https://github.com/fluent/fluent-bit/blob/0ba2dc2196e4c2c19a6075f8154afe7be137119d/plugins/out_stackdriver/stackdriver_conf.c#L322), it'll [use `logging.googleapis.com/logName`](https://github.com/fluent/fluent-bit/blob/0ba2dc2196e4c2c19a6075f8154afe7be137119d/plugins/out_stackdriver/stackdriver.h#L57) as the key name (this is the behavior that this PR uses).
- if there is no value for the key name, then it'll [use the tag](https://github.com/fluent/fluent-bit/blob/0ba2dc2196e4c2c19a6075f8154afe7be137119d/plugins/out_stackdriver/stackdriver.c#L2012-L2014) as the `LOG_ID`

-----------

### Example

The following config adds a new key `nameA` to each record, with value=`BLOOPsyslog`. The output plugin config has `log_name_key` set to `nameA`. This will result in `logName: "projects/<PROJECT_ID>/logs/BLOOPsyslog"`
```
[INPUT]
    Buffer_Chunk_Size 512k
    Buffer_Max_Size   5M
    DB                ${buffers_dir}/default_pipeline_syslog
    Key               message
    Mem_Buf_Limit     10M
    Name              tail
    Path              /var/log/messages
    Read_from_Head    True
    Rotate_Wait       30
    Skip_Long_Lines   On
    Tag               default_pipeline.syslog
    storage.type      filesystem

[FILTER]
    Add   nameA BLOOPsyslog
    Match default_pipeline.syslog
    Name  modify

[OUTPUT]
    Match_Regex       ^(default_pipeline\.syslog)$
    Name              stackdriver
    Retry_Limit       3
    stackdriver_agent Google-Cloud-Ops-Agent-Logging/2.7.1 (BuildDistro=el7;Platform=linux;ShortName=centos;ShortVersion=7.9.2009)
    tls               On
    tls.verify        Off
    workers           8
    log_name_key      nameA
```

**Logs Viewer**

image

qingling128:
Jocelyn beat me for this one. That's very accurate and a much better job of explaining things. Thank you so much, @jkschulz !

sophieyfang:
image

https://github.com/fluent/fluent-bit-docs/pull/637

PR #273, opened 2021-11-17
Install OpenSSL on Windows.

jkschulz:
http://b/206154592

Fluent Bit uses either mbedtls or OpenSSL for TLS, and will [default to OpenSSL if it's able to find](https://github.com/fluent/fluent-bit/blob/1d529d9fbc5de5559943ad5fd4e4c070fa319be7/CMakeLists.txt#L468-L473) its headers. There's a bug in mbedtls that's causing connection issues (see https://github.com/fluent/fluent-bit/issues/4300#issuecomment-964752078), so we'll build with OpenSSL instead. Our [Linux builds already install OpenSSL](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/Dockerfile), so this will also make Windows consistent.

Build and test info: http://b/206154592#comment2

### DIFF
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -68,8 +68,7 @@ RUN $ExpectedHash = \"1e62e6459835f61c418532a1544e295be5400498b3569aadfb6d0b5d7a
       throw \"OpenSSL hash mismatch. Expected: $ExpectedHash, Actual: $ActualHash\"`
     };
 
-RUN Start-Process msiexec.exe -Wait `
-    -ArgumentList '/i C:\local\Win64OpenSSL-1_1_1L.msi /quiet /norestart';
+RUN Start-Process msiexec.exe -Wait -ArgumentList '/i C:\local\Win64OpenSSL-1_1_1L.msi /quiet';
 
 #
 # Install winflexbison
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -61,14 +61,13 @@ RUN Write-Host ('Installing Visual C++ Redistributable Package'); `
 #
 ADD https://slproweb.com/download/Win64OpenSSL-1_1_1L.msi /local/Win64OpenSSL-1_1_1L.msi
 
-# Hash copied from https://github.com/slproweb/opensslhashes
-RUN $ExpectedHash = \"1e62e6459835f61c418532a1544e295be5400498b3569aadfb6d0b5d7a1c1dbc\";`
+# Verify SHA256 hash using expected value from https://github.com/slproweb/opensslhashes
+RUN $ExpectedHash = '1e62e6459835f61c418532a1544e295be5400498b3569aadfb6d0b5d7a1c1dbc';`
     $ActualHash = $(Get-FileHash /local/Win64OpenSSL-1_1_1L.msi).Hash.ToLower();`
     if ($ActualHash -ne $ExpectedHash) {`
       throw \"OpenSSL hash mismatch. Expected: $ExpectedHash, Actual: $ActualHash\"`
-    };
-
-RUN Start-Process msiexec.exe -Wait -ArgumentList '/i C:\local\Win64OpenSSL-1_1_1L.msi /quiet';
+    };`
+    Start-Process msiexec.exe -Wait -ArgumentList '/i C:\local\Win64OpenSSL-1_1_1L.msi /quiet';
 
 #
 # Install winflexbison
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -57,8 +57,16 @@ RUN Write-Host ('Installing Visual C++ Redistributable Package'); `
 
 #
 # Install OpenSSL
+# This must be done after installing Visual Studio
 #
 ADD https://slproweb.com/download/Win64OpenSSL-1_1_1L.msi /local/Win64OpenSSL-1_1_1L.msi
+
+# Hash copied from https://github.com/slproweb/opensslhashes
+RUN $ExpectedHash = \"1e62e6459835f61c418532a1544e295be5400498b3569aadfb6d0b5d7a1c1dbc\";`
+    $ActualHash = $(Get-FileHash /local/Win64OpenSSL-1_1_1L.msi).Hash.ToLower();`
+    if ($ActualHash -ne $ExpectedHash) {`
+      throw \"OpenSSL hash mismatch. Expected: $ExpectedHash, Actual: $ActualHash\"`
+    };
 
 RUN Start-Process msiexec.exe -Wait `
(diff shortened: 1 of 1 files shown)


qingling128 (approved):
LGTM

igorpeshansky (changes requested):

Dockerfile.windows
```diff
@@ -55,6 +55,15 @@ RUN Write-Host ('Installing Visual C++ Redistributable Package'); `
     Copy-Item -Path /Windows/System32/vccorlib140.dll -Destination /work/out/bin/; `
     Copy-Item -Path /Windows/System32/vcruntime140.dll -Destination /work/out/bin/;
 
+#
+# Install OpenSSL
+#
+ADD https://slproweb.com/download/Win64OpenSSL-1_1_1L.msi /local/Win64OpenSSL-1_1_1L.msi
```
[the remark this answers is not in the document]
quentinmit: Both mbedtls and OpenSSL are rarely used on Windows and this will be surprising to users either way; the Windows SSL library is called [Secure Channel](https://docs.microsoft.com/en-us/windows/win32/secauthn/secure-channel). Is there any way to use Secure Channel with fluentbit?

(For example, I don't think either off them will pick up enterprise certificates that the user has installed.)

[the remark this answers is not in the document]
jkschulz: > Is there any way to use Secure Channel with fluentbit?

Fluent Bit checks for OpenSSL specifically: https://github.com/fluent/fluent-bit/blob/1d529d9fbc5de5559943ad5fd4e4c070fa319be7/CMakeLists.txt#L468-L473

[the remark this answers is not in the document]
jkschulz: > If there's no preference, would using the [libraries from the `curl` project](https://github.com/curl/curl-for-win) be safer?

I'm not sure I understand the curl project -- what makes that safer? It looks like it'd also install other things we don't need. And I'm not sure that cmake's `find_package` (what [Fluent Bit uses](https://github.com/fluent/fluent-bit/blob/1d529d9fbc5de5559943ad5fd4e4c070fa319be7/CMakeLists.txt#L468-L473)) would work correctly if OpenSSL is a linked lib of something else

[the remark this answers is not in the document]
jkschulz: Added hash check!

Dockerfile.windows
```diff
@@ -55,6 +55,15 @@ RUN Write-Host ('Installing Visual C++ Redistributable Package'); `
...
+
+RUN Start-Process msiexec.exe `
+    -ArgumentList '/i C:\local\Win64OpenSSL-1_1_1L.msi ', '/quiet ', `
```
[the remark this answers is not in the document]
jkschulz: simplified!

Dockerfile.windows
```diff
@@ -57,12 +57,18 @@ RUN Write-Host ('Installing Visual C++ Redistributable Package'); `
...
 # Install OpenSSL
+# This must be done after installing Visual Studio
 #
 ADD https://slproweb.com/download/Win64OpenSSL-1_1_1L.msi /local/Win64OpenSSL-1_1_1L.msi
 
-RUN Start-Process msiexec.exe `
-    -ArgumentList '/i C:\local\Win64OpenSSL-1_1_1L.msi ', '/quiet ', `
-    '/norestart ' -NoNewWindow -Wait;
+# Hash copied from https://github.com/slproweb/opensslhashes
+RUN $ExpectedHash = \"1e62e6459835f61c418532a1544e295be5400498b3569aadfb6d0b5d7a1c1dbc\";`
```
igorpeshansky, line 65: You may be able to use single quotes to avoid the backslashes…

Dockerfile.windows
```diff
@@ -57,12 +57,18 @@ RUN Write-Host ('Installing Visual C++ Redistributable Package'); `
...
+    $ActualHash = $(Get-FileHash /local/Win64OpenSSL-1_1_1L.msi).Hash.ToLower();`
```
igorpeshansky, line 66: The above is a SHA256 hash, right? Want to specify that explicitly with `-Algorithm SHA256`?

Dockerfile.windows
```diff
@@ -57,12 +57,18 @@ RUN Write-Host ('Installing Visual C++ Redistributable Package'); `
...
+    if ($ActualHash -ne $ExpectedHash) {`
+      throw \"OpenSSL hash mismatch. Expected: $ExpectedHash, Actual: $ActualHash\"`
+    };
+
+RUN Start-Process msiexec.exe -Wait -ArgumentList '/i C:\local\Win64OpenSSL-1_1_1L.msi /quiet';
```
igorpeshansky, line 71: [Optional] Each `RUN` command adds a layer. Not a huge deal, but it might help to combine this one with the previous command…

Dockerfile.windows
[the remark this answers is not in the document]
igorpeshansky: The `curl` project is better known than Shining Light, I guess. But it looks like it's more of a hassle, and now that we check the hash, it should be safe.

igorpeshansky (changes requested):
Minor.

Dockerfile.windows
jkschulz: AH YOU'RE RIGHT! and since i changed "doesn't match" to "mismatch", i don't have to escape anything at alllll 🤸‍♀️

jkschulz: wait jk, it doesn't expand the vars with single quotes

Dockerfile.windows
jkschulz: yeah that's default -- added a comment

Dockerfile.windows
jkschulz: ooo you're right! tested it and it's a 9s diff! 

separate `RUN` commands
```
PS C:\Users\jschulz\ops-agent> Measure-Command { docker build -t testps -f .\dd . }

Days              : 0
Hours             : 0
Minutes           : 1
Seconds           : 14
Milliseconds      : 742
Ticks             : 747421059
TotalDays         : 0.000865070670138889
TotalHours        : 0.0207616960833333
TotalMinutes      : 1.245701765
TotalSeconds      : 74.7421059
TotalMilliseconds : 74742.1059
```

single `RUN` command
```
PS C:\Users\jschulz\ops-agent> Measure-Command { docker build -t testps -f .\dd . }

Days              : 0
Hours             : 0
Minutes           : 1
Seconds           : 5
Milliseconds      : 715
Ticks             : 657150537
TotalDays         : 0.000760590899305556
TotalHours        : 0.0182541815833333
TotalMinutes      : 1.095250895
TotalSeconds      : 65.7150537
TotalMilliseconds : 65715.0537
```

igorpeshansky (commented):
Already merged, but for the record, LGTM :shipit:

PR #275, opened 2021-11-19
receivers: add support for the Forward logging receiver

ridwanmsharif:
- Adds support for the `fluent_forward` logging receiver allowing
  Fluentd and Fluent Bit to be used as inputs to the Ops Agent
- Adds configgenerator unit tests


confgenerator/logging_receivers.go
```diff
@@ -200,6 +200,52 @@ func init() {
 	LoggingReceiverTypes.RegisterType(func() Component { return &LoggingReceiverTCP{} })
 }
 
+// A LoggingReceiverFluentForward represents the configuration for a Forward Protocol receiver.
+type LoggingReceiverFluentForward struct {
+	ConfigComponent `yaml:",inline"`
+
+	ListenHost string `yaml:"listen_host,omitempty" validate:"omitempty,ip"`
```
quentinmit, line 207: Why do we call it listen_host but validate that it's an IP and not a host?

quentinmit (dismissed):
I'm marking this "Request changes" just to make sure we resolve the questions from the design doc on what log schema(s) we'll support at ingest.

confgenerator/logging_receivers.go
jkschulz: we had the same discussion with the tcp receiver -- it's not the best name imo but we're staying consistent with `syslog` and `tcp` which already use it for that purpose

quentinmit: I think I'm not asking "can we rename it?" but "why can't we just allow hosts here too?"

jkschulz: ah, for consistency, and bc that's not supported by fluent bit. we could expand it to take both in the future if we decide the work is worth it

quentinmit: I just checked and it sure seems like fluent bit supports a hostname:

https://github.com/fluent/fluent-bit/blob/968cf967f938533aa4e3e0366310ef67dcf3d7df/src/flb_network.c#L1365

It's calling `getaddrinfo` on the user-supplied `listen` string, which supports both hostnames and IP addresses.

ridwanmsharif: This is a good point, I'll either add a commit to this PR for all of the `ListenHost` we have (`tcp`, `syslog` and `fluent_forward` are the only ones I see). Will also update the tests that assert on just IPs.

confgenerator/logging_receivers.go
```diff
@@ -249,6 +249,61 @@ func init() {
...
+	}
+	if r.ListenPort == 0 {
+		r.ListenPort = 24224
+	}
+
+	return []fluentbit.Component{{
+		Kind: "INPUT",
+		Config: map[string]string{
+			// https://docs.fluentbit.io/manual/pipeline/inputs/forward
+			"Name":   "forward",
```
qingling128, line 276: NIT: The spaces here in line 276 - 279 seem a bit adhoc.

confgenerator/add_lua_script.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+  "fmt"
+  "io/ioutil"
+  "os"
+  "path/filepath"
+)
+
+const addLogNameLuaScriptContents string = `
+function split (inputstr, sep)
```
qingling128, line 25: [No action needed]

Just surprised that LUA does not provide a built-in string splitting function LOL

confgenerator/confgenerator.go
```diff
@@ -164,8 +185,13 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
 					}
 					components = append(components, processor.Components(tag, strconv.Itoa(i))...)
 				}
-				components = append(components, setLogNameComponents(tag, rID)...)
-				tags = append(tags, regexp.QuoteMeta(tag))
+
+				// Logs ingested using the fluent_forward receiver already have the LogName
+				// set, using special rules because the LogName must preserve the existing
```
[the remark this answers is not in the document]
ridwanmsharif: Reworded

integration_test/ops_agent_test.go
```diff
@@ -621,6 +621,44 @@ func TestTCPLog(t *testing.T) {
...
+    pipelines:
+      fluent_pipeline:
+        receivers: [fluent_logs]
+        exporters: [google]
+`
+		if err := setupOpsAgent(ctx, logger, vm, config); err != nil {
+			t.Fatal(err)
+		}
+
+		if _, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", "echo '{\"msg\":\"test fluent forward log\"}' | /opt/google-cloud-ops-agent/subagents/fluent-bit/bin/fluent-bit --log_file=/var/log/forwarder.log -i stdin -o forward://127.0.0.1:24224"); err != nil {
```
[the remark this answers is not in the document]
ridwanmsharif: Ah was left behind from when I was debugging. Great catch. Removed.

confgenerator/add_lua_script.go
ridwanmsharif: Yeah I was surprised too actually :/

confgenerator/add_lua_script.go
```diff
@@ -0,0 +1,66 @@
...
+  return 2, timestamp, record
+end
+`
+
+// writeForwardScript writes the above Lua script to the given path so it can be
+// used by the logging subagent.
+//
+// TODO(ridwanmsharif): Replace this with in-config script when
+//   fluent/fluent-bit#4634 is supported.
+func writeForwardScript(path string) error {
```
qingling128, line 57: Per sync offline, we need more generic util for LUA script (needed by `modify_fields` as well). Will defer to @quentinmit and @ridwanmsharif to align on the desired file location and method name. I'm OK with a follow up PR as long as we are committed to getting it refactored in the near term.

confgenerator/logging_receivers.go
```diff
@@ -249,6 +249,61 @@ func init() {
...
+			// When the input plugin hits "mem_buf_limit", because we have enabled filesystem storage type, mem_buf_limit acts
+			// as a hint to set "how much data can be up in memory", once the limit is reached it continues writing to disk.
+			"Mem_Buf_Limit": "10M",
+		},
+	},
+		{
+			Kind: "FILTER",
+			Config: map[string]string{
+				"Name":   "lua",
+				"Match":  tag + ".*",
```
qingling128, line 296: Per discussion offline, we still wanna stick to the "LogEntry.logName: {RECEIVER_ID}.{ORIGINAL_TAG}" user experience.

There is some concern about the `{PIPELINE_ID}.{RECEIVER_ID}.*` match being too loose for the LUA filter here, and could potentially conflict with a `pipeline_id` that has `.` in it.

It is not a huge concern because `pipeline_id` and `receiver_id` need to be unique today. The chance of the corner case above is very low

AI:
* Double check whether the `.` between `pipeline_id` and `receiver_id` is an exact match or an `any character` match. (I thought it's an exact match, which made it less concerning)
* Double check whether a pipeline / receiver id allows `.` in them (My guess is yes. If no, then this whole problem might go away)
* Consider making the tag routing more specific for the forward plugin (e.g. We can use `fluent_forward.{PIPELINE_ID}.{RECEIVER_ID}` as the prefix instead, since the underlying routing is transparent to users.

confgenerator/logging_receivers.go
```diff
@@ -249,6 +249,61 @@ func init() {
...
+func (r LoggingReceiverFluentForward) Type() string {
+	return "fluent_forward"
+}
+
+func (r LoggingReceiverFluentForward) Components(tag string) []fluentbit.Component {
+	if r.ListenHost == "" {
+		r.ListenHost = "127.0.0.1"
+	}
+	if r.ListenPort == 0 {
+		r.ListenPort = 24224
```
qingling128, line 269: Needs a follow up effort / PR to ensure that if users try to use the same port in multiple pipelines, the config errors out. Priority wise, it's P2 ish. Users today will already get an error saying the port is occupied already.

The same issue applies to TCP plugin as well.

qingling128:
Note: I'm OOO from Feb 17 - 22. If the pending comment thread are addressed and approved by other reviewer(s) before I get back, feel free to merge.

ridwanmsharif:
image

Added a commit that ensures `fluent_forward` behaves sanely when used with receiver_ids and pipeline_ids that have "." in them. In the above image, this was run with a pipleine_id and receiver_id with "."s and things seem to work fine. The commit message gives some more info:

The receiver_id and pipeline_id currently allows "." to be used. This
creates a problem as thats what the Lua filter must rely on as a
delimeter when adding the old tag to the logname and removing the
pipeline_id from it. Additionally, allowing "." in the receiver_id
creates issues for when a receiver_id is a prefix of another one.
(its a very specific edge case where another receiver_id would have
another receiver_id followed by a ".")

Consider the following example:
```
logging:
  receivers:
    fluent_logs:
      type: fluent_forward
      listen_host: 1.2.3.4
      listen_port: 5678
    fluent_logs.collision:
      type: fluent_forward
      listen_host: 1.2.3.4
      listen_port: 5679
  service:
    pipelines:
      fluent_pipeline:
        receivers: [fluent_logs, fluent_logs.collision]
```

Notice that the receiver_id `fluent_logs.collision` completely contains the other
receiver_id `fluent_logs`. And so when we try to match on records
ingested by `fluent_logs`, since it will be suffixed by the tag of the
forwarder, we will need to match on `fluent_logs.*`. However because of
this, we will match on records from `fluent_logs.collision` as well. Which is not
desired. To get around this issue, we prefix the tag with a unique hash
that will deterministically ID the pipeline + receiver combination.

- Add a hash string as the prefix to the tag to prevent collisions
- Update lua script to only append existing tag to the LogName
- Have the Lua filter update logname for `fluent_forward` by appending
  the existing tag.
- Add test cases for the collision scenario

confgenerator/testdata/valid/linux/logging-receiver_forward/input.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+logging:
+  receivers:
+    fluent_logs:
```
qingling128, line 17: Should we unify the receiver / pipeline ID as well for the forward receiver? Similar to https://github.com/GoogleCloudPlatform/ops-agent/pull/384#issue-1123802764

aka `s/fluent_logs/fluent_forward` and `s/fluent_pipeline/fluent_forward`

qingling128 (approved):
LGTM modulo the open threads get resolved.

confgenerator/add_lua_script.go
ridwanmsharif: We haven't decided yet but this will be done in a follow-up PR. The config writing code will be folded in with this as well, and we'll have a common util that will write write files to disk (configs, Lua scripts, anything else we will need in the future). Keeping it separate from this PR so this PR doesn't get too unwieldy

confgenerator/logging_receivers.go
ridwanmsharif: We have b/220128869 to keep track of it

confgenerator/add_lua_script.go
```diff
@@ -0,0 +1,73 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package confgenerator
+
+import (
+  "fmt"
```
sophieyfang, line 18: I wonder why the `goimports` presubmit check didn't catch the off format in this file:

image

PR #276, opened 2021-11-22
Add Wildcard Refresh Interval

braydonk:
Adds new field wildcard_refresh_interval to the logging files receiver
in the ops-agent config.

### DIFF
--- a/confgenerator/config.go
+++ b/confgenerator/config.go
@@ -90,6 +90,8 @@ func (ve validationError) Error() string {
 		return fmt.Sprintf("%q must end with %q", ve.Field(), ve.Param())
 	case "ip":
 		return fmt.Sprintf("%q must be an IP address", ve.Field())
+	case "min":
+		return fmt.Sprintf("%q must be a minimum of %s", ve.Field(), ve.Param())
 	case "oneof":
 		return fmt.Sprintf("%q must be one of [%s]", ve.Field(), ve.Param())
 	case "required":
@@ -100,6 +102,8 @@ func (ve validationError) Error() string {
 		return fmt.Sprintf("%q must start with %q", ve.Field(), ve.Param())
 	case "url":
 		return fmt.Sprintf("%q must be a URL", ve.Field())
+	case "whole_time_denomination":
+		return fmt.Sprintf("%q must be a whole %s", ve.Field(), ve.Param())
 	}
 
 	return ve.FieldError.Error()
@@ -137,7 +141,7 @@ func newValidator() *validator.Validate {
 	v.RegisterValidationCtx("platform", func(ctx context.Context, fl validator.FieldLevel) bool {
 		return ctx.Value(platformKey) == fl.Param()
 	})
-	// duration validates that the value is a valid durationa and >= the parameter
+	// duration validates that the value is a valid duration and >= the parameter
 	v.RegisterValidation("duration", func(fl validator.FieldLevel) bool {
 		t, err := time.ParseDuration(fl.Field().String())
 		if err != nil {
@@ -148,6 +152,33 @@ func newValidator() *validator.Validate {
 			panic(err)
 		}
 		return t >= tmin
+	})
+	// time_denomination validates that the duration is whole to some time denomination
+	// eg. s, ms, m, h
+	v.RegisterValidation("whole_time_denomination", func(fl validator.FieldLevel) bool {
+		var t time.Duration
+		switch fl.Field().Interface().(type) {
+		case time.Duration:
+			t = fl.Field().Interface().(time.Duration)
+		case *time.Duration:
+			tPtr := fl.Field().Interface().(*time.Duration)
+			t = *tPtr
+		default:
+			panic(fmt.Sprintf("%s is not a valid time.Duration", fl.Field().String()))
+		}
+
+		var denomination time.Duration
+		switch fl.Param() {
+		case "second":
+			denomination = time.Second
+		default:
+			panic(fmt.Sprintf(
+				"whole_time_denomination tried to parse an unrecognized denomination: %s",
+				fl.Param(),
+			))
+		}
+
+		return t%denomination == 0
 	})
 	return v
 }
--- a/confgenerator/config.go
+++ b/confgenerator/config.go
@@ -156,15 +156,9 @@ func newValidator() *validator.Validate {
 	// time_denomination validates that the duration is whole to some time denomination
 	// eg. s, ms, m, h
 	v.RegisterValidation("whole_time_denomination", func(fl validator.FieldLevel) bool {
-		var t time.Duration
-		switch fl.Field().Interface().(type) {
-		case time.Duration:
-			t = fl.Field().Interface().(time.Duration)
-		case *time.Duration:
-			tPtr := fl.Field().Interface().(*time.Duration)
-			t = *tPtr
-		default:
-			panic(fmt.Sprintf("%s is not a valid time.Duration", fl.Field().String()))
+		t, ok := fl.Field().Interface().(time.Duration)
+		if !ok {
+			panic(fmt.Sprintf("whole_time_denomination: could not convert %s to time duration", fl.Field().String()))
 		}
 
 		var denomination time.Duration
@@ -172,10 +166,7 @@ func newValidator() *validator.Validate {
 		case "second":
 			denomination = time.Second
 		default:
-			panic(fmt.Sprintf(
-				"whole_time_denomination tried to parse an unrecognized denomination: %s",
-				fl.Param(),
-			))
+			panic(fmt.Sprintf("whole_time_denomination: unrecognized denomination %q", fl.Param()))
 		}
 
 		return t%denomination == 0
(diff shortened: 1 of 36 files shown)


confgenerator/config.go
```diff
@@ -149,6 +153,24 @@ func newValidator() *validator.Validate {
 		}
 		return t >= tmin
 	})
+	// time_denomination validates that the duration is whole to some time denomination
+	// eg. s, ms, m, h
+	v.RegisterValidation("whole_time_denomination", func(fl validator.FieldLevel) bool {
```
braydonk, line 158: The `panic` cases in this validator are not covered. Given that L148 (above in this file) is also not covered, do we accept the -0.7% fall in coverage due to these lines? We could think about adding tests that cover these lines, but not sure if that should live in this PR anyway since it's a new style of tests.

jefferbrecht: I wouldn't worry about missing coverage for panics, they'll fail the tests anyway if they get triggered which is fine. IMO panic==assert and it should never be possible to write test cases to trigger them in the first place. That seems to be the case here so all seems good.

qingling128: +1

confgenerator/testdata/invalid/linux/logging-receiver_files_refresh_interval_invalid_duration/golden_error
```diff
@@ -0,0 +1 @@
+cannot unmarshal uint64 into Go struct field UnifiedConfig.Logging of type time.Duration
```
qingling128, line 1: I wonder if we can reuse some similar validation / error message customization that `collection_interval` uses? In theory, the same type of validations should apply to both, except the minimum is different.

References: 
* https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/testdata/invalid/linux/metrics-receiver_invalid_parameter_malformed_collection_interval/golden_error
* https://github.com/GoogleCloudPlatform/ops-agent/blob/43187fcb01f8f3d27f3d735651a00ba617acb549/confgenerator/config.go#L87

braydonk: I have been trying to figure this out but haven't come up with a solution. That spot in `config.go` only seems to be called if it fails a validation, but this error is failing during the unmarshalling step. It looks like we also don't handle it for invalid port numbers, since those don't fail at the unmarshalling step as well. Reference: https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/testdata/invalid/linux/logging-receiver_tcp_type_invalid_parameter_listen_port_is_not_an_int/golden_error

qingling128: What about something like https://github.com/GoogleCloudPlatform/ops-agent/blob/8677126770c36441362348423fd9e5f0fc0f9410/confgenerator/config.go#L359

braydonk: This was only done as a band-aid fix for the fact that the `go-yaml` library couldn't unmarshal the `time.Duration` type. Now that it can (see https://github.com/goccy/go-yaml/pull/246) we'd rather use that than a string with a custom validator. 

The reason we didn't run into this problem with `CollectionInterval` is because it will unmarshal into a string, but fails on our `duration` validator. (FWIW this also resulted in an unhelpful error message of "did not meet minimum duration").

confgenerator/config.go
```diff
@@ -149,6 +153,24 @@ func newValidator() *validator.Validate {
...
+		t, ok := fl.Field().Interface().(time.Duration)
+		if !ok {
+			panic(fmt.Sprintf("whole_time_denomination: could not convert %s to time duration", fl.Field().String()))
+		}
+
+		var denomination time.Duration
+		switch fl.Param() {
+		case "second":
```
quentinmit, line 166: Why not just have the argument be "1s" and parse it as a duration?

quentinmit (approved):

confgenerator/config.go
```diff
@@ -149,6 +153,24 @@ func newValidator() *validator.Validate {
 		}
 		return t >= tmin
 	})
+	// time_denomination validates that the duration is whole to some time denomination
```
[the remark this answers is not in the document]
braydonk: I'm not sure how this would look when specifying the validation on the struct, since 1 second in nanoseconds is 1e+9. I could probably impose some kind of rule like having it work in seconds or milliseconds instead, but I worry that could become confusing for usage.

confgenerator/config.go
braydonk: I combined this idea and `multipleof` into a solution that should be more scalable than this original implementation.

confgenerator/config.go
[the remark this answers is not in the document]
quentinmit: I meant something like calling the validator `multipleof=1s`; you'd probably have to specialize to calling `time.ParseDuration` if the field is a `Duration`; then you can do the actual math on an arbitrary int value.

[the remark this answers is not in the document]
braydonk: Oh okay I see, that is sort of what I ended up doing. I made a single purpose version that is only for time values rather than making a validator that checks and parses either time or int. I thought it was easier to read that way, and we don't currently have a need for a `multipleof` validator on just ints. I can change it to be that way instead though if that's better.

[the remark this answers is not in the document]
quentinmit: Nah, I was just trying to code golf by keeping the name short and simple :) It's not very important.

confgenerator/testdata/invalid/linux/logging-receiver_files_refresh_interval_invalid_duration/golden_error
qingling128: Ah I see. Do I get it right that we are basically choosing between two options:

Option 1. Use a custom validator like `duration`.

Pros:
* Better user experience as the error message is much clearer (Sample errors in https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/testdata/invalid/linux/metrics-receiver_invalid_parameter_collection_interval_below_minimum/golden_error and https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/testdata/invalid/linux/metrics-receiver_invalid_parameter_malformed_collection_interval/golden_error)

Cons
* We need to maintain a custom validator `duration`.

Option 2. Use the upstream `time.Duration` validator

Pros
* No need to maintain a custom validator for `duration` parsing.

Cons
* Degraded user experience as the error message does not point out which line of the configuration is problematic with a concise reason. Users need to guess based on the internal config struct names.

Any other tradeoffs I missed?

If this is the main pros and cons, option 2 makes sense if we do get the pros (aka we can get rid of the maintenance of a custom validator). Otherwise it feels like we get degraded user experience but not much gains. Also for consistency and long-term maintenance, we should keep things consistent for similar types (aka between `collection_interval` and `wildcard_refresh_interval`.)

BTW we do eventually want to fix all the raw unmarshal error messages to make them more actionable and self descriptive. We haven't tackled them because the cost is relatively high, but it's not the ideal state that we'd like to stick with in the long term.

quentinmit: This isn't about validation; it's about whether the config struct contains a `time.Duration` or a `string`.

Actually I don't know why it does't have a better error message in option 2; the YAML parser is the thing generating the error message, so it certainly has enough info to emit the line and column number.

braydonk: I didn't get terribly far when I researched this, but what I managed to find was that if the library fails to assert a string on it, it returns the `typeError` which is the less helpful message we see here. If we put a different type of garbage input that does assert a string, we get the marginally more helpful `time: invalid duration "asdfjk34802934809"` which is still not ideal.

braydonk: >Also for consistency and long-term maintenance, we should keep things consistent for similar types (aka between collection_interval and wildcard_refresh_interval.)

FWIW I have a branch off this one that changes `CollectionInterval` to a time.Duration.

qingling128: Ah I see. A better unmarshal error message that emits the line and column number probably requires some upstream `go-yaml` changes. Sounds like it's a limitation at that level.

A follow up PR to change `CollectionInterval` sounds perfect. I'm ready to LGTM this one.

Note that we are now in a rare state of code chill / freeze, so feature PRs like this one needs to be held from merging for the time being. More details in the internal team email alias conversation.

confgenerator/config.go
qingling128: Looks like this one has been resolved. Thanks!

confgenerator/config.go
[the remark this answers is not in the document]
qingling128: Looks like this one has been resolved. Thanks!

qingling128 (approved):
LGTM, modulo a follow up PR to change `collection_interval` so that the two are consistent for long-term maintenance.

PR #277, opened 2021-11-23
-update_golden creating empty files

braydonk:
`updateOrCompareGolden` did not detect the case where a diff is empty
because the file doesn't exist yet. Added a check for the edge case that
creates the empty golden file if it doesn't exist yet.

### DIFF
--- a/confgenerator/confgenerator_test.go
+++ b/confgenerator/confgenerator_test.go
@@ -181,28 +181,23 @@ func updateOrCompareGolden(t *testing.T, testName string, goos string, expectedB
 	expected := strings.ReplaceAll(string(expectedBytes), "\r\n", "\n")
 	actual = strings.ReplaceAll(actual, "\r\n", "\n")
 	goldenPath := fmt.Sprintf(path, goos, testName)
-	if diff := cmp.Diff(expected, actual); diff != "" {
-		if *updateGolden {
+	diff := cmp.Diff(expected, actual)
+	if *updateGolden {
+		if diff != "" || !fileExists(goldenPath) {
 			// Update the expected to match the actual.
 			t.Logf("Detected -update_golden flag. Rewriting the %q golden file to apply the following diff\n%s.", goldenPath, cmp.Diff(actual, expected))
 			if err := ioutil.WriteFile(goldenPath, []byte(actual), 0644); err != nil {
 				t.Fatalf("error updating golden file at %q : %s", goldenPath, err)
 			}
-		} else {
-			t.Errorf("test %q: golden file at %s mismatch (-want +got):\n%s", testName, goldenPath, diff)
 		}
-	} else {
-		if *updateGolden {
-			// If the diff is empty and we need to update goldens, check whether the diff is
-			// empty because the file doesn't exist yet.
-			if _, err := os.Stat(goldenPath); err != nil {
-				t.Logf("Detected -update_golden flag. Creating %q golden file.", goldenPath)
-				if err := ioutil.WriteFile(goldenPath, []byte(""), 0644); err != nil {
-					t.Fatalf("error creating golden file at %q : %s", goldenPath, err)
-				}
-			}
-		}
+	} else if diff != "" {
+		t.Errorf("test %q: golden file at %s mismatch (-want +got):\n%s", testName, goldenPath, diff)
 	}
+}
+
+func fileExists(path string) bool {
+	_, err := os.Stat(path)
+	return err == nil
 }
 
 func TestGenerateConfigsWithInvalidInput(t *testing.T) {
--- a/confgenerator/confgenerator_test.go
+++ b/confgenerator/confgenerator_test.go
@@ -191,6 +191,17 @@ func updateOrCompareGolden(t *testing.T, testName string, goos string, expectedB
 		} else {
 			t.Errorf("test %q: golden file at %s mismatch (-want +got):\n%s", testName, goldenPath, diff)
 		}
+	} else {
+		if *updateGolden {
+			// If the diff is empty and we need to update goldens, we check whether the diff is
+			// empty because the file doesn't exist yet.
+			if _, err := os.Stat(goldenPath); err != nil {
+				t.Logf("Detected -update_golden flag. Creating %q golden file.", goldenPath)
+				if err := ioutil.WriteFile(goldenPath, []byte(""), 0644); err != nil {
+					t.Fatalf("error creating golden file at %q : %s", goldenPath, err)
+				}
+			}
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


confgenerator/confgenerator_test.go
```diff
@@ -191,6 +191,17 @@ func updateOrCompareGolden(t *testing.T, testName string, goos string, expectedB
 		} else {
 			t.Errorf("test %q: golden file at %s mismatch (-want +got):\n%s", testName, goldenPath, diff)
 		}
+	} else {
+		if *updateGolden {
+			// If the diff is empty and we need to update goldens, we check whether the diff is
```
braydonk, line 196:
```suggestion
			// If the diff is empty and we need to update goldens, check whether the diff is
```

braydonk: Actually I'll just commit this properly

confgenerator/confgenerator_test.go
```diff
@@ -191,6 +191,17 @@ func updateOrCompareGolden(t *testing.T, testName string, goos string, expectedB
 		} else {
 			t.Errorf("test %q: golden file at %s mismatch (-want +got):\n%s", testName, goldenPath, diff)
 		}
+	} else {
+		if *updateGolden {
+			// If the diff is empty and we need to update goldens, check whether the diff is
+			// empty because the file doesn't exist yet.
+			if _, err := os.Stat(goldenPath); err != nil {
+				t.Logf("Detected -update_golden flag. Creating %q golden file.", goldenPath)
+				if err := ioutil.WriteFile(goldenPath, []byte(""), 0644); err != nil {
```
igorpeshansky, line 200: Can we combine the code that writes the golden file from both branches? E.g., something like this:
```
	diff := cmp.Diff(expected, actual)
	if *updateGolden {
		if diff != "" || actual == "" {
			// Update the expected to match the actual.
			t.Logf("Detected -update_golden flag. Rewriting the %q golden file to apply the following diff\n%s.", goldenPath, cmp.Diff(actual, expected))
			if err := ioutil.WriteFile(goldenPath, []byte(actual), 0644); err != nil {
				t.Fatalf("error updating golden file at %q : %s", goldenPath, err)
			}
		}
	} else if diff != "" {
		t.Errorf("test %q: golden file at %s mismatch (-want +got):\n%s", testName, goldenPath, diff)
	}
```

igorpeshansky (changes requested):

confgenerator/confgenerator_test.go
braydonk: I got something similar to this to work; I modified it slightly from the example here because we only want it to write the file on empty diff if the file doesn't already exist. Let me know what you think.

confgenerator/confgenerator_test.go
```diff
@@ -181,19 +181,25 @@ func updateOrCompareGolden(t *testing.T, testName string, goos string, expectedB
 	expected := strings.ReplaceAll(string(expectedBytes), "\r\n", "\n")
 	actual = strings.ReplaceAll(actual, "\r\n", "\n")
 	goldenPath := fmt.Sprintf(path, goos, testName)
-	if diff := cmp.Diff(expected, actual); diff != "" {
-		if *updateGolden {
+	diff := cmp.Diff(expected, actual)
+	if *updateGolden {
+		if diff != "" || !fileExists(goldenPath) {
```
igorpeshansky, line 186: It's interesting that `readFileContent` assumes that any error from `ioutil.ReadFile` would be a "file not found" error. This means that if a non-empty file exists, but there's an intermittent error actually reading that file, the content will be assumed empty, and your code will not try to overwrite it, which would cause the tests to fail later (exactly what `-update_golden` is intended to prevent). If we want to catch that case, we should unconditionally attempt to overwrite the golden file when the content is empty (which will at least log a failure message in case of an intermittent error). Alternatively, we can fix line 168 to actually distinguish the various errors and only swallow the "file not found" ones… Does this make sense?

igorpeshansky (changes requested):

confgenerator/confgenerator_test.go
igorpeshansky: It's interesting that `readFileContent` assumes that any error from `ioutil.ReadFile` would be a "file not found" error. This means that if a non-empty file exists, but there's an intermittent error actually reading that file, the content will be assumed empty, and your code will not try to overwrite it, which would cause the tests to fail later (exactly what `-update_golden` is intended to prevent). If we want to catch that case, we should unconditionally attempt to overwrite the golden file when the content is empty (which will at least log a failure message in case of an intermittent error). Alternatively, we can fix line 168 to actually distinguish the various errors and only swallow the "file not found" ones… Does this make sense?

igorpeshansky (changes requested):
Sorry, should have responded on the original thread…

confgenerator/confgenerator_test.go
```diff
@@ -181,17 +182,29 @@ func updateOrCompareGolden(t *testing.T, testName string, goos string, expectedB
 	expected := strings.ReplaceAll(string(expectedBytes), "\r\n", "\n")
 	actual = strings.ReplaceAll(actual, "\r\n", "\n")
 	goldenPath := fmt.Sprintf(path, goos, testName)
-	if diff := cmp.Diff(expected, actual); diff != "" {
-		if *updateGolden {
+	diff := cmp.Diff(expected, actual)
+	if *updateGolden {
+		if diff != "" || !fileExists(t, testName, goldenPath) {
```
igorpeshansky, line 187: Now that line 170 checks for `os.ErrNotExist`, you can technically save yourself the `os.Stat` call, since you'd only have `diff == "" && actual == ""` (or `diff == "" && expected == ""`) when the file does not exist or was empty in the first place. The only difference between those is whether you'd want to overwrite (i.e., change the last modification time of) the empty file in the second case. With git, it usually doesn't matter, so we could go for the cheaper (and clearer) alternative. WDYT?

igorpeshansky (changes requested):
One suggested simplification, but let's discuss.

confgenerator/confgenerator_test.go
braydonk: True I guess if we're okay to make the tradeoff of overwriting the files when nothing needs to be written, then that would be clearer. The downside would be that if someone's doing a `-v` run they'll see golden files "changing" when they're not (the same empty files are just being overwritten). We would probably need to add a clarification for that somewhere if we did that. I might just be concerned over nothing though, and it would look clearer in the code.

braydonk: I've gone ahead and made that change, and if anything changes I can revert it or I can try to figure out where the clarification should go.

igorpeshansky: It might be worth adding a comment, e.g. `// When actual is empty, the file may not exist; overwrite it`… WDYT?

braydonk: I added a clarifying comment.

quentinmit (approved):

igorpeshansky (approved):
LGTM :shipit:

quentinmit (approved):
Include the new empty golden files this creates in the PR?

braydonk:
>Include the new empty golden files this creates in the PR?

This PR didn't end up making any new ones; the previous scenarios with empty files were already covered.

PR #278, opened 2021-11-29
Add Memcached Metrics Receiver

Mrod1598:
Added metrics receiver for memcached


docs/memcached.md
```diff
@@ -0,0 +1,54 @@
...
+| ---- | ----------- | ---- | ---- | ---------- |
+| memcached.bytes | Current number of bytes used by this server to store items. | By | Gauge | <ul> </ul> |
+| memcached.commands | Commands executed. | 1 | Sum | <ul> <li>command</li> </ul> |
+| memcached.current_connections | The current number of open connections. | connections | Gauge | <ul> </ul> |
+| memcached.current_items | Number of items currently stored in the cache. | 1 | Gauge | <ul> </ul> |
+| memcached.evictions | Cache item evictions. | 1 | Sum | <ul> </ul> |
+| memcached.network | Bytes transferred over the network. | by | Sum | <ul> <li>direction</li> </ul> |
+| memcached.operation_hit_ratio | Hit ratio for operations, expressed as a percentage value between 0.0 and 100.0. | % | Gauge | <ul> <li>operation</li> </ul> |
+| memcached.operations | Operation counts. | 1 | Sum | <ul> <li>type</li> <li>operation</li> </ul> |
+| memcached.rusage | Accumulated user and system time. | 1 | Sum | <ul> <li>state</li> </ul> |
```
quentinmit, line 48: What are the values for state? `user` and `system`?

Is `.rusage` the OT metric convention? (I assume it's short for "resource usage"?)

quentinmit (changes requested):

docs/memcached.md
```diff
@@ -0,0 +1,54 @@
...
+| memcached.threads | Number of threads used by the memcached instance. | 1 | Gauge | <ul> </ul> |
+| memcached.total_connections | Total number of connections opened since the server started running. | connections | Sum | <ul> </ul> |
```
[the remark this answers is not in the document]
Mrod1598: That does make more sense with the conventions we've established now.

docs/memcached.md
Mrod1598: it's how memcached [references it](https://docs.oracle.com/cd/E17952_01/mysql-5.6-en/ha-memcached-stats.html). and yes `user` and `system`, I'll add a table for attribute values.

docs/memcached.md
```diff
@@ -0,0 +1,54 @@
...
+
+| Name | Description | Unit | Type | Attributes |
+| ---- | ----------- | ---- | ---- | ---------- |
+| memcached.bytes | Current number of bytes used by this server to store items. | By | Gauge | <ul> </ul> |
+| memcached.commands | Commands executed. | 1 | Sum | <ul> <li>command</li> </ul> |
+| memcached.current_connections | The current number of open connections. | connections | Gauge | <ul> </ul> |
+| memcached.current_items | Number of items currently stored in the cache. | 1 | Gauge | <ul> </ul> |
+| memcached.evictions | Cache item evictions. | 1 | Sum | <ul> </ul> |
+| memcached.network | Bytes transferred over the network. | by | Sum | <ul> <li>direction</li> </ul> |
+| memcached.operation_hit_ratio | Hit ratio for operations, expressed as a percentage value between 0.0 and 100.0. | % | Gauge | <ul> <li>operation</li> </ul> |
```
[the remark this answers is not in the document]
Mrod1598: we have hit/miss on the operations metric already. so should we just remove this metric then?

docs/memcached.md
```diff
@@ -0,0 +1,54 @@
...
+          - memcached_metrics
+```
+
+## Metrics
+
+The Ops Agent collects the following metrics from your memcached servers.
+
+| Name | Description | Unit | Type | Attributes |
+| ---- | ----------- | ---- | ---- | ---------- |
+| memcached.bytes | Current number of bytes used by this server to store items. | By | Gauge | <ul> </ul> |
```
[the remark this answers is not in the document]
Mrod1598: This was an existing metric on the otel receiver. the memcached docs describe it as [`Current number of bytes used by this server to store items.` ](https://docs.oracle.com/cd/E17952_01/mysql-5.6-en/ha-memcached-stats-general.html). Do you have a suggestion for better name? perhaps `memcached.storage`?

docs/memcached.md
quentinmit: Ah, in the link memcached calls it `rusage_user` and `rusage_system` - in that context the name makes sense. But if you're just calling it `rusage` I thiink that is too generic - [`getrusage`](https://man7.org/linux/man-pages/man2/getrusage.2.html) is a system call that returns CPU usage, memory usage, and a number of other stats. Maybe we could call it `memcached.rusage.cpu`?

docs/memcached.md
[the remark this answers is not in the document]
quentinmit: I see from the link in your other comment thread that `bytes` is the name that memcached uses for this metric. I think that means we should probably keep it even if it's a little confusing, since that's what users coming from other monitoring systems will expect.

confgenerator/logging_receivers.go
```diff
@@ -125,7 +125,7 @@ func (r LoggingReceiverSyslog) Components(tag string) []fluentbit.Component {
 			"Mode":   r.TransportProtocol,
 			"Listen": r.ListenHost,
 			"Port":   fmt.Sprintf("%d", r.ListenPort),
-			"Parser": tag,
+			"Parser": "syslog-rfc5424",
```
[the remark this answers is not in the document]
quentinmit: If memcached outputs to syslog by default that probably means it writes to `/dev/log`, which means it will end up in `journald` and that's where you need to read the logs from. Using a syslog receiver in Ops Agent would mean that the user would have to reconfigure memcached to send logs to a different port that OA is listening on.

docs/memcached.md
Mrod1598: ok will update.

confgenerator/logging_receivers.go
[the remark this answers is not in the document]
Mrod1598: Ok, so we tested and it doesn't end up in journald just syslog, I've actually reverted this change because we originally thought that the parser was going to properly parse the different parts of the logs and double checked it and there doesn't seem to be any benefit.

docs/memcached.md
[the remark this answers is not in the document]
jsuereth: Yes, you can still give this in the dashboard via  query.

confgenerator/logging_receivers.go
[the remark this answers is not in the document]
qingling128: So with an out-of-the-box Ops Agent and an 
out-of-the-box Memcached today, logs are ingested but not parsed as structured logs, right?

Can we add a "format" parameter to the out-of-the-box ops agent syslog receiver and document sth like:

Memcached logs are collected automatically as part of syslog. To enable structured logging for all syslog, use:

```
logging:
  receivers:
    syslog:
      format:
        syslog-rfc5424
```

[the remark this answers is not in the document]
quentinmit: `syslog-rfc5424` is a network protocol; it applies to `type: syslog` receivers, which the default `syslog` receiver is not.

AFAICT memcached just sends its logs to stderr, so there's no `syslog`-type receiver needed. If it's running as a normal systemd service, you would want to use a journald receiver to extract its logs.

@Mrod1598 Maybe we can schedule a 1:1 where you walk me through what you tested?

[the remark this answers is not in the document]
Mrod1598: > So with an out-of-the-box Ops Agent and an out-of-the-box Memcached today, logs are ingested but not parsed as structured logs, right?
> 
> Can we add a "format" parameter to the out-of-the-box ops agent syslog receiver and document sth like:
> 
> Memcached logs are collected automatically as part of syslog. To enable structured logging for all syslog, use:
> 
> ```
> logging:
>   receivers:
>     syslog:
>       format:
>         syslog-rfc5424
> ```
> 
> This way it's backward compatible with existing syslog functionality.
> 
> To enable Memcached specific parsing, we can do an app specific processor to avoid duplicate logs while providing the functionality and maintaining backward compatibility. e.g.
> 
> ```
> logging:
>   receivers:
>     syslog:
>       format:
>         syslog-rfc5424
>   processors:
>     parse_memcached:
>         type: parse_memcached
>   pipelines:
>     default_pipeline:
>       processors:
>         parse_memcached
> ```
> 
> This is not a blocker for the metrics side of things though IMO.

These are the logs I'm getting with the new parser vs. no change

## No Change Logs
```
{
  "insertId": "rh0siyfvqwreh",
  "jsonPayload": {
    "message": "Dec 13 20:14:35 test-memcached-no-change-logs systemd-memcached-wrapper[13698]: <27 set foo 0 0 8"
  },
  "resource": {
    "type": "gce_instance",
    "labels": {
      "instance_id": "2603025410381497601",
      "zone": "us-central1-a"
    }
  },
  "timestamp": "2021-12-13T20:14:35.980433749Z",
  "logName": "projects/otel-agent-dev/logs/syslog",
  "receiveTimestamp": "2021-12-13T20:14:36.102239290Z"
}
```
## Updated Parser Logs
```
{
  "insertId": "1vzkvaifr7z3ak",
  "jsonPayload": {
    "message": "Dec 13 20:20:58 test-memcached--new-parser-logs systemd-memcached-wrapper[13467]: <27 set foo 0 0 8"
  },
  "resource": {
    "type": "gce_instance",
    "labels": {
      "instance_id": "5992099982425637625",
      "zone": "us-central1-a"
    }
  },
  "timestamp": "2021-12-13T20:21:40.089486747Z",
  "logName": "projects/otel-agent-dev/logs/syslog",
  "receiveTimestamp": "2021-12-13T20:21:41.041395584Z"
}
```
I think we were incorrect the first time we tested this. I don't see a difference and it just pulls it by default with the ops-agent

docs/memcached.md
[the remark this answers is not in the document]
Mrod1598: Added filter to exclude it.

docs/memcached.md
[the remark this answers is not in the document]
Mrod1598: Updated with the update to the operations collector

docs/memcached.md
Mrod1598: Updated to `memcached.cpu.usage` based on what otel wanted. is that fine for google or do we need to change it for ops-agent?

docs/memcached.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Metrics
+
+These are the metrics available for this scraper.
+
+| Name | Description | Unit | Type | Attributes |
+| ---- | ----------- | ---- | ---- | ---------- |
+| memcached.bytes | Current number of bytes used by this server to store items. | By | Gauge | <ul> </ul> |
+| memcached.commands | Commands executed. | 1 | Sum | <ul> <li>command</li> </ul> |
+| memcached.connections.current | The current number of open connections. | connections | Sum | <ul> </ul> |
```
quentinmit, line 42: `current` should be a Gauge (or an UpDownCounter if we're using that terminology in these docs) since it can both increase and decrease.

docs/memcached.md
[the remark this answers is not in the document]
quentinmit: Remove the docs then?

docs/memcached.md
```diff
@@ -0,0 +1,64 @@
...
+| memcached.connections.total | Total number of connections opened since the server started running. | connections | Sum | <ul> </ul> |
+| memcached.cpu.usage | Accumulated user and system time. | 1 | Sum | <ul> <li>state</li> </ul> |
+| memcached.current_items | Number of items currently stored in the cache. | 1 | Sum | <ul> </ul> |
```
quentinmit, line 45: This should also be a Gauge or UpDownCounter.

docs/memcached.md
Mrod1598: Updated to the new metric types after update.

docs/memcached.md
Mrod1598: updated on doc.

docs/memcached.md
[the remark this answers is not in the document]
Mrod1598: removed

docs/memcached.md
quentinmit: That name is fine with me

quentinmit (approved):
Please rebase after https://github.com/GoogleCloudPlatform/ops-agent/pull/300 so you don't have extra diffs before submitting

docs/memcached.md
```diff
@@ -0,0 +1,64 @@
...
+| `collection_interval` | `60s`                     | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
+
+Example Configuration:
+
+```yaml
+metrics:
+  receivers:
+    redis_metrics:
+      type: memcached
+      address: localhost:3306
```
qingling128, line 25: Is this field `endpoint` or `address`? Line 15 says `endpoint`. 

BTW, can we keep the example configuration with minimum customization, as users tend to copy paste it. We'd probably prefer them sticking with the default `60s` interval. In other words, we can remove these two lines?

```
address: localhost:3306
collection_interval: 30s
```

apps/memcached.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
+)
+
+type MetricsReceiverMemcached struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"omitempty,hostname_port|file"`
```
qingling128, line 13: Should we do `hostname_port|startswith=/` instead of `hostname_port|file` here as well?

See a similar conversation in https://github.com/GoogleCloudPlatform/ops-agent/pull/218/files/8620d7199eb6a617dc369657eade36956d49c609

integration_test/third_party_apps_data/test_config.yaml
```diff
@@ -41,3 +41,5 @@ per_application_overrides:
     platforms_to_skip: *common_skips
   redis:
     platforms_to_skip: *common_skips
+  memcached:
+    platforms_to_skip: *common_skips
```
qingling128, line 45: For newly added applications, we should not skip these distros. The older ones were released before we have cross distro tests. At some point, we'd need to go fix known issues in them as well.

docs/memcached.md
Mrod1598: Updated, however I'm pretty sure we require that collection interval is set as it is set to required in[ the validation](https://github.com/GoogleCloudPlatform/ops-agent/blob/d1c729ef424c1848b642838d5c32a16988290114/confgenerator/config.go#L375).  but updated to 60s and removed `address/endpoint` field in example. it is endpoint btw.

apps/memcached.go
Mrod1598: good point, updated.

docs/memcached.md
qingling128: You are right. I just realized today that the `collection_interval` is a required field in the implementation (It's supposed to be optional). We'll fix that separately as it's not directly related to this PR.

Resolving this thread.

integration_test/third_party_apps_data/test_config.yaml
qingling128: Looks like this has been addressed in https://github.com/GoogleCloudPlatform/ops-agent/pull/278/commits/99511cc178725c45d81ff56865a3c563ea8040a7

Mrod1598: yes, but haven't been able to verify it works.

confgenerator/logging_receivers.go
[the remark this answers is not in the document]
qingling128: Per sync offline, this PR will focus on metrics only. For logging, we'll rely on the out-of-the-box `syslog` receiver to collect Memcached logs and add processors as seen fit in a separate effort.

qingling128 (approved):
LGTM, modulo the integration tests pass.

qingling128:
https://github.com/GoogleCloudPlatform/ops-agent/pull/306 was just merged. This probably needs rebasing to get the tests pass.

Mrod1598:
skipping `ubuntu-minimal-1604-lts` & `ubuntu-1604-lts` because it seems the VM family doesn't exist.
```
2022/01/06 06:15:30 Running command: [gcloud compute instances create github-test-20220106-98f03-b8b5b5a5-9260-419e-8b7f-e6a68871ccdf --project=stackdriver-test-143416 --zone=us-central1-b --machine-type=e2-standard-4 --image-project=ubuntu-os-cloud --image-family=ubuntu-minimal-1604-lts --image-family-scope=global --network=default --format=json --metadata=serial-port-logging-enable=true,enable-oslogin=false,ssh-keys=test_user:ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDedDG4ocgscheKCUU0u+s1ykVUhof+9O+gQJPhKgUn5Ofl8tcGtj7HckuhLgtzg8v9j+iCPzlExX+t/Tm+4Rn1ib+wHG3kLvc2p4yjNAocjR7y9tleng5xdKLsyZQ+nmkMbhapQZXf6toKsXTS8r6TDvuq7yzT435/Iso0dcpoJRGPGEARpLJ0mtpc3UiUIHjXnm81htkTdGp0ORKIKJszDkry3J18JS0GUZkXAvWp2i8lfFg0TwfycADxUVEdBUZgfI6rfDQZt+/xqD4Zxq6C8560xAQQaPHo1IboOtoiKl/nMLUXWur0zKRhNESU8XTumtUKdU+bJoRV9BfPH0xH test_user
 --labels=kokoro_build_id=23fe48f7-61f2-4b2c-95c4-50040926f3ec --service-account=build-and-test-external@stackdriver-test-143416.iam.gserviceaccount.com]
2022/01/06 06:15:36 exit code: 1
2022/01/06 06:15:36 stdout: 
2022/01/06 06:15:36 stderr: ERROR: (gcloud.compute.instances.create) Could not fetch resource:
 - The resource 'projects/ubuntu-os-cloud/global/images/family/ubuntu-minimal-1604-lts' was not found
```

qingling128 (approved):
LGTM

qingling128:
Merging this since the `third party apps integration test` passed, and the failures in `ops agent integration test` are all related to sles 15 upstream repo reliability issues.

PR #280, opened 2021-11-29
Fix port in the example Apache config.

igorpeshansky:
b/207530308

### DIFF
--- a/docs/apache.md
+++ b/docs/apache.md
@@ -35,7 +35,7 @@ metrics:
   receivers:
     apache_metrics:
       type: apache
-      server_status_url: http://localhost:8080/server-status?auto
+      server_status_url: http://localhost:80/server-status?auto
       collection_interval: 30s
   service:
     pipelines:


docs/apache.md
```diff
@@ -35,7 +35,7 @@ metrics:
   receivers:
     apache_metrics:
       type: apache
-      server_status_url: http://localhost:8080/server-status?auto
+      server_status_url: http://localhost:80/server-status?auto
```
quentinmit, line 38: Why are you not fixing the code at the same time?

quentinmit (approved):

docs/apache.md
igorpeshansky: Good question. @qingling128, are you ok with adding a code fix to this PR?

qingling128: Do I get it right that:

The proposal is to change the default port in Apache implementation (https://github.com/GoogleCloudPlatform/ops-agent/blob/b125c158d7be96c182245dbc88666d6cf495fa00/apps/apache.go#L33) from `8080` to `80`?

Pros:
* Users with out-of-the-box Apache settings do not have to customize `server_status_url` to make things work.

Cons:
* This is technically a backward incompatible change and could potentially break some existing users. However we don't expect any Apache users to customize Apache to use the `8080` port. So in theory, most Apache users should have `server_status_url: http://localhost:80/server-status?auto` in their config today if they managed to make it work. With that being said, changing the default port should not have a big impact. 

I'm running some query to see how many users / VMs are collecting Apache logs today and could potentially be impacted.

quentinmit: That's correct; no one runs Apache on port 8080 by default, so anyone who is doing so expects to need to set the port in Ops Agent as well. (Also, our docs have been telling people to set it, so if they just copied + pasted they would be explicitly setting it anyway.)

qingling128: Makes sense. Let's fix the default port as proposed above.

igorpeshansky: @qingling128 Do we want to do that in this PR, or in a separate PR? That change is independent of changing the docs…

qingling128: Either way is fine.

qingling128 (approved):
LGTM

qingling128:
The changes in this PR are now obsolete, because the sample configuration in `docs/apache.md` has been removed. Instead we directly point to the cloud documentation site.

The port has already been fixed there:
image

The follow up change to fix the port in the implementation is here: https://github.com/GoogleCloudPlatform/ops-agent/pull/386

PR #282, opened 2021-12-02
Change Metrics Collection Interval to time.Duration

braydonk:
#276 updated the goccy/go-yaml library to a version that parses `time.Duration` directly, and the new time field used that. Update `MetricsReceiverShared.CollectionInterval` to be a `time.Duration` for consistency (also resolves [this TODO](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/config.go#L363)).


confgenerator/built-in-config-linux.yaml
```diff
@@ -13,7 +13,7 @@ metrics:
   receivers:
     hostmetrics:
       type: hostmetrics
-      collection_interval: 60s
+      collection_interval: 1m0s
```
igorpeshansky, line 16: Bleh. Is there no way we can force this to be serialized as seconds? These are user-visible sample configs, so it is not a great idea to change them…

Looks like the goccy/go-yaml library [does not support this](https://github.com/goccy/go-yaml/blob/master/encode.go#L581), but maybe a custom marshaler?

confgenerator/config.go
```diff
@@ -372,15 +358,7 @@ type MetricsReceiver interface {
...
 type MetricsReceiverShared struct {
-	CollectionInterval string `yaml:"collection_interval" validate:"required,duration=10s"` // time.Duration format
-}
-
-func (m MetricsReceiverShared) CollectionIntervalString() string {
-	// TODO: Remove when https://github.com/goccy/go-yaml/pull/246 is merged
-	if m.CollectionInterval != "" {
-		return m.CollectionInterval
-	}
-	return "60s"
```
igorpeshansky, line 383: Did we just remove the `60s` default? Does it matter?

confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_otel.conf
```diff
@@ -344,7 +344,7 @@ processors:
     - gce
 receivers:
   hostmetrics/default__pipeline_hostmetrics:
-    collection_interval: 60s
+    collection_interval: 1m0s
```
igorpeshansky, line 347: [Optional] This one is not ideal either, but less annoying than the user-visible sample config change.

confgenerator/testdata/invalid/linux/metrics-receiver_invalid_parameter_malformed_collection_interval/golden_error
```diff
@@ -1,9 +1 @@
...
-   2 |   receivers:
-   3 |     receiver_1:
-   4 |       type: hostmetrics
->  5 |       collection_interval: 24
-                                  ^
-   6 |   service:
-   7 |     pipelines:
-   8 |       default_pipeline:
\ No newline at end of file
+cannot unmarshal uint64 into Go struct field UnifiedConfig.Metrics of type time.Duration
```
igorpeshansky, line 1: [Optional] Sigh. This is a net negative for usability, but there's precedent for messages like this, so we can probably kick that can down the road and address all of those in one fell swoop at some point…

igorpeshansky (changes requested):

confgenerator/built-in-config-linux.yaml
quentinmit: ISTR last I looked into this it was not possible to implement a custom marshaler (hence why I had to fix it upstream in the first place). I think we just have to live with this; this is the canonical serialization of a `time.Duration(60000000000)` in Go.

confgenerator/testdata/invalid/linux/metrics-receiver_invalid_parameter_malformed_collection_interval/golden_error
quentinmit: Yeah, fixing this generically for all fields would be a great separate PR.

apps/builtinconf.go
```diff
@@ -39,7 +43,7 @@ var (
 				Receivers: map[string]cg.MetricsReceiver{
 					"hostmetrics": &MetricsReceiverHostmetrics{
 						ConfigComponent:       cg.ConfigComponent{Type: "hostmetrics"},
-						MetricsReceiverShared: cg.MetricsReceiverShared{CollectionInterval: "60s"},
+						MetricsReceiverShared: cg.MetricsReceiverShared{CollectionInterval: time.Second * 60},
```
quentinmit, line 46: Nit: It reads better if the number comes first.

```suggestion
						MetricsReceiverShared: cg.MetricsReceiverShared{CollectionInterval: 60 * time.Second},
```

confgenerator/config.go
quentinmit: +1

confgenerator/built-in-config-linux.yaml
braydonk: I recall a GitHub issue in the golang/go repo where custom `time.Duration` string formatting was discussed, and it was denied because of backwards compatibility. I'm going to keep looking for the issue, I wish I'd saved it. But the outcome of that was that we'll have to have some custom function to format this as `60s` if we want it that way.

The last intrusive way I can think to do that would be to keep the `CollectionIntervalString` function, check if the value is 60 seconds, and deliberately output `60s` in that scenario, otherwise just return the string formatted duration.

confgenerator/config.go
braydonk: I put 60 seconds as a value any time it looked like we relied on the default result of `CollectionIntervalString`, did I miss any?
I will change to using a constant instead of putting 60 seconds everywhere.

quentinmit: Yeah, a constant would be preferable, thanks!

confgenerator/built-in-config-linux.yaml
quentinmit: I think we just need to live with it; this is the canonical serialization of 60s.

PR #283, opened 2021-12-03
Aggregate away mountpoints using `max`, as there may be duplicates.

igorpeshansky:
The standalone monitoring agent discards duplicates: [`df.c:184-203`](https://github.com/Stackdriver/collectd/blob/stackdriver-agent-5.8.1/src/df.c#L184-L203).

b/204269280

### DIFF
--- a/apps/hostmetrics.go
+++ b/apps/hostmetrics.go
@@ -156,13 +156,14 @@ func (r MetricsReceiverHostmetrics) Pipelines() []otel.Pipeline {
 					"disk/bytes_used",
 					// change data type from int64 -> double
 					otel.ToggleScalarDataType,
-					// take sum over mode, mountpoint & type dimensions, retaining only device & state
-					otel.AggregateLabels("sum", "device", "state"),
+					// take max over mode, mountpoint & type dimensions, retaining only device & state
+					// there may be multiple mountpoints for the same device
+					otel.AggregateLabels("max", "device", "state"),
 				),
 				otel.RenameMetric(
 					"system.filesystem.utilization",
 					"disk/percent_used",
-					otel.AggregateLabels("sum", "device", "state"),
+					otel.AggregateLabels("max", "device", "state"),
 				),
 				otel.RenameMetric(
 					"system.memory.usage",
--- a/confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_otel.conf
@@ -154,7 +154,7 @@ processors:
       operations:
       - action: toggle_scalar_data_type
       - action: aggregate_labels
-        aggregation_type: sum
+        aggregation_type: max
         label_set:
         - device
         - state
@@ -163,7 +163,7 @@ processors:
       new_name: disk/percent_used
       operations:
       - action: aggregate_labels
-        aggregation_type: sum
+        aggregation_type: max
         label_set:
         - device
         - state
--- a/confgenerator/testdata/valid/linux/all-built_in_config/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/all-built_in_config/golden_otel.conf
@@ -159,7 +159,7 @@ processors:
       operations:
       - action: toggle_scalar_data_type
       - action: aggregate_labels
-        aggregation_type: sum
+        aggregation_type: max
         label_set:
         - device
         - state
@@ -168,7 +168,7 @@ processors:
       new_name: disk/percent_used
       operations:
       - action: aggregate_labels
-        aggregation_type: sum
+        aggregation_type: max
         label_set:
         - device
         - state
--- a/confgenerator/testdata/valid/linux/all-user_config_file_deleted/golden_otel.conf
+++ b/confgenerator/testdata/valid/linux/all-user_config_file_deleted/golden_otel.conf
@@ -159,7 +159,7 @@ processors:
       operations:
       - action: toggle_scalar_data_type
       - action: aggregate_labels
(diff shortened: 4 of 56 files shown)


quentinmit:
max is not the correct aggregator either; we need to actually report the mountpoints.

quentinmit:
@igorpeshansky and I discussed internally. `df` uses `/proc/self/mountinfo` to enumerate disks, not `/proc/mounts` or `/etc/mtab`. The former contains information about which mounts are bind mounts, which allows `df` to exclude them. `max` and `sum` both report (different) incorrect results for (different) filesystems. `max` does better with bind mounts, but it's not generally true that all mounts with the same reported "device" are the same filesystem. e.g. `iso9660` can mount multiple separate filesystems (sessions) from a single block device, tmpfs (currently broken due to a different bug) reports separate usage for every instance of the filesystem, etc.

The only way to actually report correct stats in (more) all cases is to keep the `mount` label that is present on this metric. (I say "more" because it's actually possible to have multiple filesystems mounted on the same mountpoint, though I believe you can only get usage information from the most recent of those mounts.)

qingling128:
Does that mean this requires some upstream `hostmetrics` receiver changes to retrieve additional info of the mounts? Or is the info already present and we just need to map them to a proper label? Do we need to add a new label `mount` to the `agent.googleapis.com/disk/percent_used` metrics? Right now it only has [`device` and `state`](https://cloud.google.com/monitoring/api/metrics_opsagent#agent-disk). The alternative is to append the `mount` to the end of `device` and put them together in the `device` label I guess. Thoughts? @igorpeshansky @quentinmit

igorpeshansky:
The upstream `hostmetrics` metric does include the `mount` label. The label is aggregated away in our config, because it's not present in `agent.googleapis.com/disk/percent_used`. That metric is broken with respect to representing multiple mounts that represent [different filesystems](https://xkcd.com/2531/). However, there is currently no mechanism to distinguish those mounts from bind mounts (which are not proper filesystems).

The long-term fix would be to add the `mount` label, but before we do that, we would also have to change the `hostmetrics` receiver to filter out bind mounts. This is also the reason we should not add mount info to the `device` label.
For the short term, my suggestion is to merge this PR as-is, which at least replicates the limitations of the standalone monitoring agent (each device gets one representative mount). My intuition is that most folks won't have ISO mounts on cloud VMs, and, since we don't yet support virtual devices like `tmpfs`, this should mitigate the common source of confusion caused by bind mounts. Adding correct support for `tmpfs` will require us to rethink that metric descriptor anyway.

quentinmit:
ISO was just one example of a filesystem other than tmpfs that has the ability to report multiple different capacities from a single notional "device". It's certainly not the only other one. I think I agree at this moment that max is correct in more cases than sum, but by merging this PR we're making the tmpfs bug harder to fix and I'm hoping that this can be the impetus we need to prioritize fixing both bugs correctly.

igorpeshansky:
Adding the `mount` label to the `agent.googleapis.com/disk/*` metrics would be sufficient for supporting `tmpfs` (or "fixing the `tmpfs` bug"). However, that would cause misleading data (and increased cardinality) for bind mounts. Fixing bind mounts would require making changes to the upstream `gopsutil` libraries and propagating them through multiple repos. And once we do add the `mount` label, the aggregation that this PR touches will go away entirely, so it shouldn't matter whether it uses `sum` or `max`, and it shouldn't make fixing the `tmpfs` issue any harder.

qingling128:
@quentinmit Could you help clarify the details around "by merging this PR we're making the tmpfs bug harder to fix"? It feels like it is not gonna make things worse than what it is today. But we might have missed something.

BTW can someone help summarize the discussion (once the dust settles) and file a bug for tracking the proper fix?

quentinmit:
I was expecting that the next step on that would be to change the collector to report `tmpfs` filesystems, which is a collector-only change (not touching `gopsutil`). Then `sum` would be the correct aggregator, to report the total bytes used in all tmpfses on the system. Including the `mount` label would make the aggregator irrelevant, as you note, but I expected that would be a subsequent change because of the relative complexity. Then hiding bind mounts is a separate task from either of those and I don't think blocks either of them.

igorpeshansky:
Since `sum` is the right aggregator for some filesystems, but not others, we would have to find a way to change aggregation based on the value of the device label anyway, whether or not this PR gets merged. One way to do this is to first clone the affected metric into two (one for filesystems that need `sum`, and one for those that need `max`), then apply the appropriate aggregations, and then use `combine` to merge them back into a single metric. This would probably have to be done with multiple applications of `metricstransform`, but it seems doable. So I would still argue for merging this PR as-is to fix the immediate issue.

Hiding bind mounts would actually block adding the `mount` label, IMO, because of the additional cardinality it would introduce (and because it would break the group-by behavior, since you'd no longer be able to have a sensible sum across all label values).

qingling128:
Sent a meeting invite for us to discuss the medium-term / future proper fix for `tmpfs and duplicate mounts`.

The current implementation is quite wrong, so I think it makes sense to get this PR in first so that customers' alerts do not constantly go off. The mixed `sum` and `max` as a temp workaround sounds a bit confusing. It would be great if we can avoid it. Ultimately we need to get `mount` labels and `tmpfs` working. If we need to switch the order to get the labels fixed first before `tmpfs` to avoid this, let's do that.

qingling128 (approved):
LGTM

qingling128:
Per sync offline with @quentinmit and @igorpeshansky, the decision is:

* [Short-term] Merge this PR as is because it makes things better than what it is today.
* [Medium-term] Work with upstream Open Telemetry to support filtering out bind mounts, add mount label to the metrics descriptor, and use no aggregation.

PR #284, opened 2021-12-03
Removed duplicated stackdriver output from fluentbit

rafaelwestphal:
Now aggregating the match regex on a single output driver instead
of two identical ones.

This will also reduce the number of workers allocated in fluentbit. Currently we have 8 works that Match_Regex ^(ops-agent-fluent-bit)$, and 8 workers that match the other regex we are interested in.


quentinmit (changes requested):
This was intentionally kept separate so that agent logs would still be delivered if something went wrong with the user pipelines. I'll leave it up to @qingling128 whether we want to keep that behavior.

rafaelwestphal:
Synced with qingling128 and the behavior will remain as is.

PR #315, opened 2022-01-11
Tomcat Metrics Receiver

Mrod1598:
Add Support for Tomcat Metrics. Additionally, there were two metrics that otel requested to be added `max_time` and `request_count`. These weren't originally in the scope doc, would we like to keep these metrics or filter them out?


apps/tomcat.go
```diff
@@ -0,0 +1,89 @@
...
+
+	config := map[string]interface{}{
+		"target_system":       targetSystem,
+		"collection_interval": r.CollectionIntervalString(),
+		"endpoint":            r.Endpoint,
+		"jar_path":            jarPath,
+	}
+
+	// Only set the username & password fields if provided
+	if r.Username != "" {
```
qingling128, line 66: What happens if only one of `username` or `password` exists? Is there an easy-to-diagnose error in the agent log? If not, can we add a validation and a unit test for the invalid case? Just to avoid support load in the long run.

apps/tomcat.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
+)
+
+type MetricsReceiverTomcat struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"omitempty,hostname_port|url"`
```
qingling128, line 30: If the url has to obey specific prefix. e.g. `service:jmx:`, we can use a `startswith` check. e.g.https://github.com/GoogleCloudPlatform/ops-agent/search?q=startswith

Mrod1598: Ok added additional Validation

docs/tomcat.md
```diff
@@ -0,0 +1,49 @@
...
+| ---                                                  | ---            | ---         | ---                            | ---         | 
+| workload.googleapis.com/tomcat.sessions              | Gauge          | sessions    |                                | The number of active sessions. |
+| workload.googleapis.com/tomcat.errors                | Cumulative     | errors      | proto_handler                  | The number of errors encountered. |
+| workload.googleapis.com/tomcat.processing_time       | Cumulative     | ms          | proto_handler                  | The total processing time. |
+| workload.googleapis.com/tomcat.traffic               | Cumulative     | by          | proto_handler, direction       | The number of bytes transmitted and received. |
+| workload.googleapis.com/tomcat.threads               | Gauge          | threads     | proto_handler, state           | The number of threads. |
+| workload.googleapis.com/tomcat.max_time              | Cumulative     | ms          | proto_handler                  | Maximum time to process a request. |
```
qingling128, line 46: Should `tomcat.max_time` be a Gauge instead? (I tried to find a reference in the scoping doc, but this metrics does not exist there).

apps/tomcat.go
qingling128: Great!

apps/tomcat.go
qingling128: Thanks!

qingling128 (commented):
Added one question for metrics type. Otherwise LGTM.

docs/tomcat.md
Mrod1598: It was an additional metric otel wanted and went back and checked and you are correct in it being a gauge. Opened a PR downstream to update it but if we don't want this additional metric we can also just remove it.

Mrod1598: Updated the metric as the downstream change got merged already.

qingling128: Great

qingling128 (approved):
LGTM

Mrod1598:
> LGTM

The third party int tests passed, I believe the ops-agent failing is unrelated.

qingling128:
Yup, the failure of `ops agent integration test ` is `--- FAIL: TestSyslogTCP/debian-10 (169.22s)`.

One thing that caught attention is the unit test coverage for tomcat seems relatively low. Could we take a quick look to see why and whether we are missing something there:

image

Mrod1598:
> Yup, the failure of `ops agent integration test ` is `--- FAIL: TestSyslogTCP/debian-10 (169.22s)`.
> 
> One thing that caught attention is the unit test coverage for tomcat seems relatively low. Could we take a quick look to see why and whether we are missing something there:
> 
> image

Added more unit tests to improve coverage

qingling128 (approved):
LGTM

PR #318, opened 2022-01-12
Removing old style of exporting fluentbit metrics

rafaelwestphal:
This style of exporting fluentbit metrics is not being used right now.
Since we are switching a new input, removing this old style.

### DIFF
--- a/confgenerator/fluentbit/service.go
+++ b/confgenerator/fluentbit/service.go
@@ -31,12 +31,6 @@ func (s Service) Component() Component {
 			// Log_File is set by Fluent Bit systemd unit (e.g. /var/log/google-cloud-ops-agent/subagents/logging-module.log).
 			"Log_Level": s.LogLevel,
 
-			// https://docs.fluentbit.io/manual/administration/monitoring
-			// Enable a built-in HTTP server that can be used to query internal information and monitor metrics of each running plugin.
-			"HTTP_Server": "On",
-			"HTTP_Listen": "0.0.0.0",
-			"HTTP_PORT":   "2020",
-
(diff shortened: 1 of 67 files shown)


confgenerator/fluentbit/service.go
```diff
@@ -31,12 +31,6 @@ func (s Service) Component() Component {
 			// Log_File is set by Fluent Bit systemd unit (e.g. /var/log/google-cloud-ops-agent/subagents/logging-module.log).
 			"Log_Level": s.LogLevel,
 
-			// https://docs.fluentbit.io/manual/administration/monitoring
-			// Enable a built-in HTTP server that can be used to query internal information and monitor metrics of each running plugin.
-			"HTTP_Server": "On",
```
qingling128, line 36: The answer is probably yes. But just checking: Is this `Off` by default?

qingling128 (approved):
LGTM

confgenerator/fluentbit/service.go
rafaelwestphal: Yes, I double checked and the behavior of ommiting and setting to "Off" is the same.

qingling128: Great. Thanks for checking!

qingling128:
Regarding the right timing to merge this one: I assume that we'd like this and https://github.com/GoogleCloudPlatform/ops-agent/pull/319 to go into the same Ops Agent release, so it kinda depends on when https://github.com/GoogleCloudPlatform/ops-agent/pull/319 is ready to be merged. @rafaelwestphal - Will defer to you for pulling the trigger on this one.

qingling128:
Obsolete by https://github.com/GoogleCloudPlatform/ops-agent/pull/330

PR #319, opened 2022-01-12
Added new fluentbit-metrics receiver

rafaelwestphal:
The receiver will be responsible for making the fluentbit internal
metrics available.

### DIFF
--- a/confgenerator/fluentbit/service.go
+++ b/confgenerator/fluentbit/service.go
@@ -55,3 +55,15 @@ func (s Service) Component() Component {
 		},
 	}
 }
+
+func (s Service) MetricsComponent() Component {
+	return Component{
+		Kind: "INPUT",
+		Config: map[string]string{
+			// https://docs.fluentbit.io/manual/pipeline/inputs/systemd
+			"Name":            "fluentbit_metrics",
+			"Scrape_On_Start": "True",
+			"Scrape_Interval": "30",
+		},
+	}
+}
--- a/confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_fluent_bit_main.conf
+++ b/confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_fluent_bit_main.conf
@@ -16,6 +16,11 @@
     storage.sync              normal
 
 [INPUT]
+    Name            fluentbit_metrics
+    Scrape_Interval 30
+    Scrape_On_Start True
+
+[INPUT]
     Buffer_Chunk_Size 512k
     Buffer_Max_Size   5M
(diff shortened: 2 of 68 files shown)


confgenerator/logging_receivers.go
```diff
@@ -272,3 +272,28 @@ func (r LoggingReceiverSystemd) Components(tag string) []fluentbit.Component {
...
+type LoggingReceiverMetric struct {
+	ConfigComponent `yaml:",inline"`
+}
+
+func (r LoggingReceiverMetric) Components(tag string) []fluentbit.Component {
+	return []fluentbit.Component{{
+		Kind: "INPUT",
+		Config: map[string]string{
+			// https://docs.fluentbit.io/manual/pipeline/inputs/systemd
+			"Name":            "fluentbit_metrics",
```
qingling128, line 286: Similar to before, we do not want to expose the implementation details of Fluent Bit to users. Instead, we make the self metrics enabled by default and users can not turn it off. This way we guarantee:

* Ops Agent self metrics are always ingested 
* In the future, we have flexibility to swap the underlying agents (e.g. Fluent Bit or Open Telemetry Collector) if needed.

rafaelwestphal: Changed implementation to hardcode the data as discussed

confgenerator/fluentbit/service.go
```diff
@@ -55,3 +55,15 @@ func (s Service) Component() Component {
 		},
 	}
 }
+
+func (s Service) MetricsComponent() Component {
+	return Component{
+		Kind: "INPUT",
+		Config: map[string]string{
+			// https://docs.fluentbit.io/manual/pipeline/inputs/systemd
```
qingling128, line 63: We probably wanna update this URL?

confgenerator/testdata/valid/linux/all-backward_compatible_with_explicit_exporters/golden_fluent_bit_main.conf
```diff
@@ -15,6 +15,11 @@
     storage.metrics           on
     storage.sync              normal
 
+[INPUT]
+    Name            fluentbit_metrics
+    Scrape_Interval 30
```
qingling128, line 20: Ops Agent's default collection interval is 60s. We might wanna match that.

qingling128 (approved):
Two minor comments. Otherwise LGTM.

qingling128:
As discussed offline, we'd like to bundle the lated PRs into one commit when merging to master to ensure things can be rolled back together if needed, and the change to remove port 2020 metrics and expose new metrics end up in the same agent release.

qingling128:
Obsolete by https://github.com/GoogleCloudPlatform/ops-agent/pull/330

PR #320, opened 2022-01-12
Postgresql integration

dehaansa:
Add support for postgresql metrics and postgresql_general logs.

Example log:
image

Concerns:
Currently setting default TLS Insecure to true, is that desired or do we want to use the default behavior of the receiver which is assume TLS when using a TCP connection.


apps/postgresql.go
```diff
@@ -0,0 +1,219 @@
...
+func (r MetricsReceiverPostgresql) Type() string {
+	return "postgresql"
+}
+
+func (r MetricsReceiverPostgresql) Pipelines() []otel.Pipeline {
+	transport := "tcp"
+	if r.Endpoint == "" {
+		transport = "unix"
+		r.Endpoint = defaultpostgresqlUnixEndpoint
+	} else if strings.HasPrefix(r.Endpoint, "/") {
```
qingling128, line 56: Now that multiple application have identical logic for `endpoint`. Can we use a shared helper? Something similar to `confgenerator.MetricsReceiverShared`. So that future applications can use the same helpful. This helps boost consistency across applications. And it's overall easier to maintain in the long term.

apps/postgresql.go
```diff
@@ -0,0 +1,219 @@
...
+
+	cfg := map[string]interface{}{
+		"collection_interval": r.CollectionIntervalString(),
+		"endpoint":            r.Endpoint,
+		"username":            r.Username,
+		"password":            r.Password,
+		"transport":           transport,
+	}
+
+	if transport == "tcp" {
```
qingling128, line 74: Similar here, multiple application will have logic for `TLS`. Can we use a shared helper? Something similar to `confgenerator.MetricsReceiverShared`. So that future applications can use the same helpful. This helps boost consistency across applications. And it's overall easier to maintain in the long term.

apps/postgresql.go
```diff
@@ -0,0 +1,219 @@
...
+func (r LoggingReceiverPostgresql) Components(tag string) []fluentbit.Component {
+	if len(r.IncludePaths) == 0 {
+		r.IncludePaths = []string{
+			// Default log paths for Debain / Ubuntu
+			"/var/log/postgresql/postgresql*.log",
+			// Default log paths for SLES
+			"/var/lib/pgsql/data/log/postgresql*.log",
+			// Default log paths for CentOS / RHEL
+			"/var/lib/pgsql/*/data/log/postgresql*.log",
+			// "/var/lib/pgsql/14/data/log/postgresql*.log",
```
qingling128, line 204: Are these comments removable? Or is there something implicit that we are trying to point out?

docs/postgresql.md
```diff
@@ -0,0 +1,139 @@
...
+## Configuration
+
+Following the guide for [Configuring the Ops Agent](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/configuration#file-location), add the required elements for your postgresql configuration.
+
+To configure a receiver for your postgresql metrics, specify the following fields:
+
+| Field                   | Default                         | Description |
+| ---                     | ---                             | ---         |
+| `type`                  | required                        | Must be `postgresql`. |
+| `endpoint`              | `/var/run/postgresql/.s.PGSQL.5432`   | The hostname:port or socket path used to connect to postgresql |
```
qingling128, line 18:
```suggestion
| `endpoint`              | `/var/run/postgresql/.s.PGSQL.5432`   | The hostname:port or socket path starting with `/` used to connect to postgresql |
```

docs/postgresql.md
```diff
@@ -0,0 +1,139 @@
...
+| `collection_interval`   | `60s`                           | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
+| `username`              | `postgres`                          | The username used to connect to the server. |
```
qingling128, line 20: Can we ensure invalid config combinations throw errors as expected?

e.g. username present but password not or vice versa, or if any combo in the TLS configs are not valid.

docs/postgresql.md
```diff
@@ -0,0 +1,139 @@
...
+| `password`              |                                 | The password used to connect to the server. |
+| `insecure`              | true                            | Signals whether to use a secure TLS connection or not. If insecure is true TLS will not be enabled. |
```
qingling128, line 22: There are multiple different valid combinations of how the TLS related fields could be used. Can we add some unit tests for them? Our users refer to the valid test case folder when trying to find sample configs.

docs/postgresql.md
```diff
@@ -0,0 +1,139 @@
...
+
+TCP connection with a username and password:
+
+```yaml
+metrics:
+  receivers:
+    postgresql_metrics:
+      type: postgresql 
+      endpoint: localhost:3306
+      collection_interval: 30s
```
qingling128, line 52:
```suggestion
      collection_interval: 60s
```

docs/postgresql.md
```diff
@@ -0,0 +1,139 @@
...
+| `key_file`              |                             | Path to the TLS key to use for TLS required connections. |
+| `ca_file`               |                             | Path to the CA cert. As a client this verifies the server certificate. If empty, uses system root CA. |
+
+Example Configuration:
+
+```yaml
+metrics:
+  receivers:
+    postgresql_metrics:
+      type: postgresql
```
qingling128, line 34: I thought `collection_interval` is required? (It's a bug that is to be fixed on our side to make it optitonal)

docs/postgresql.md
```diff
@@ -0,0 +1,139 @@
...
+
+The `postgresql` receiver defaults to connecting to a local postgresql server using a Unix socket and Unix authentication as the `postgres` user.
+
+## Configuration
+
+Following the guide for [Configuring the Ops Agent](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/configuration#file-location), add the required elements for your postgresql configuration.
+
+To configure a receiver for your postgresql metrics, specify the following fields:
+
+| Field                   | Default                         | Description |
```
qingling128, line 15: Can we add a column to indicate whether the config parameter is required or optional?

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_postgresql
```diff
@@ -0,0 +1,45 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+metrics:
+  receivers:
+    postgresql_metrics:
+      type: postgresql
+      collection_interval: 30s
+      username: postgres
```
qingling128, line 24: Does Postgres require username and password out of the box?

apps/postgresql.go
dehaansa: Those can be removed, I was attempting to lay out the expected paths signified by the asterisk in the default paths for CentOS/RHEL, but can remove.

apps/postgresql.go
dehaansa: The logic between the applications is different in this case. We could build a structure that specifies default transport & endpoint, but the rest isn't as standardized between applications. I'll take a pass at it.

apps/postgresql.go
dehaansa: The only thing that might differ between applications is default TLS, so I'll try to factor that in.

confgenerator/testdata/valid/linux/logging-receiver_mysql/logging-receiver_mysql_refresh_interval/golden_fluent_bit_main.conf
```diff
@@ -0,0 +1,633 @@
+@SET buffers_dir=/var/lib/google-cloud-ops-agent/fluent-bit/buffers
```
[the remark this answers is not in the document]
dehaansa: We're up to date vs master, I'll see why this showed up.

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_postgresql
dehaansa: Out of the box behavior differs for different distributions. Debian/CentOS is Ident based (logged in user) auth for unix socket, password auth via local TCP, password via external TCP. SLES is Ident based for unix of local TCP, no auth via external TCP.

apps/postgresql.go
dehaansa: I wasn't able to build something clean that reduced complexity.

docs/postgresql.md
dehaansa: I don't see a way to use the validator library we're using to do something like "if insecure=true, do not allow this field" for ca_file/cert_file/key_file.

Password is not always required even when username is provided, because of the Ident/OS authentication mode of auth.

apps/postgresql.go
qingling128: No worries.

apps/postgresql.go
qingling128: SG

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_postgresql
qingling128: 😵‍💫

confgenerator/testdata/valid/linux/logging-receiver_mysql/logging-receiver_mysql_refresh_interval/golden_fluent_bit_main.conf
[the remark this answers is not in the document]
dehaansa: Figured this out, had been accidentally dragged into the wrong folder & duplicated.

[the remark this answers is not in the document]
qingling128: Ah I see. Thanks for taking care of it!

confgenerator/testdata/invalid/linux/metrics-receiver_postgresql_insecure_plus/golden_error
```diff
@@ -0,0 +1 @@
+when insecure is true, no other TLS fields may be configured. Values were found for field(s) insecure_skip_verify, cert_file, ca_file, key_file for receiver postgresqlmetrics
```
qingling128, line 1: Slight adjustment for error message convention compliance:

“insecure_skip_verify“, ”cert_file“, ”ca_file“, ”key_file“ are not allowed when "insecure" is true, which indicates TLS is disabled for receiver "postgresqlmetrics".

dehaansa: Updated.

Issue #288, opened 2021-12-09
jvm.memory.pool.used emits values in Double then Integer


sophieyfang:
**Description**
`jvm.memory.pool.used` emits Double then Integer. As the metrics descriptor is automatically created per the first time it receives the timeseries. For this case, the initial points are having value 0.00  which will create the metrics descriptor in valueType of Double see below. However, the points following are integers which get denied by the GCM API with below error msgs 
```
Nov  2 14:36:39 app-admin-01 otelopscol[5791]: 2021-11-02T14:36:39.337Z#011info#011exporterhelper/queued_retry.go:231#011Exporting failed. Will retry the request after interval.#011{"kind": "exporter", "name": "googlecloud", "error": "rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Points must be written in order. One or more of the points specified had an older end time than the most recent point.: timeSeries[0-31]; Value type for metric workload.googleapis.com/jvm.memory.pool.used must be DOUBLE, but is INT64.: timeSeries[32-37]\nerror details: name = Unknown  desc = total_point_count:38 errors:{status:{code:3} point_count:38}", "interval": "22.291635104s"}
```

Note: I have saved all points before it sent out, See **Additional context** for the behavior of initial 0.00 points then integer values.

## Metrics Descriptor - jvm.memory.pool.used
```
{
  "name": "projects/sophieyfang-test/metricDescriptors/workload.googleapis.com/jvm.memory.pool.used",
  "labels": [
    {
      "key": "name"
    }
  ],
  "metricKind": "GAUGE",
  "valueType": "DOUBLE",
  "type": "workload.googleapis.com/jvm.memory.pool.used",
  "monitoredResourceTypes": [
    "anthos_l4lb",
    "apigee.googleapis.com/Environment",
    "apigee.googleapis.com/Proxy",
    "apigee.googleapis.com/ProxyV2",
    "apigee.googleapis.com/TargetV2",
    "aws_alb_load_balancer",
    "aws_cloudfront_distribution",
    ........
}

```

**Steps to reproduce**
### Inside the GCE VM
```
sophieyfang_google_com@ubuntu-1804-jvm:~$ sudo apt update
sophieyfang_google_com@ubuntu-1804-jvm:~$ sudo apt install -y default-jdk
sophieyfang_google_com@ubuntu-1804-jvm:~$ cat <<EOF > hello.java
> class HelloWorld {
[...]

PR #321, opened 2022-01-13
Update otel JMX component

dehaansa:
Based off of research completed by @cpheps https://github.com/GoogleCloudPlatform/ops-agent/issues/288#issuecomment-1011730111 we believe an updated jmx-metrics-jar should resolve issues with JMX metrics appearing as doubles when they should be integers.


dehaansa:
Being updated in https://github.com/GoogleCloudPlatform/ops-agent/pull/315 already

PR #322, opened 2022-01-13
Add service level config option to specify credentials file path

ridwanmsharif:
This change adds a config option to let users specify the credentials
file for a custom service account. This would let people configure the
ops agent (specifically the logging submodule) to use custom service
accounts.

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -168,13 +168,13 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
 			out = append(out, s.components...)
 		}
 		if len(tags) > 0 {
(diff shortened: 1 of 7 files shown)


quentinmit:
Should this field be located at the top level of the config instead of inside `logging`? (And, for that matter, implemented for OT as well. Presumably if we move to OT for logging it won't be possible to have separate credentials anyway.)

ridwanmsharif:
My thoughts here were that I'd also add this feature for OTEL as well before we launch them both. As for whether this should be one level up, that is very reasonable. I thought it could be useful to control them separately for now as we do with `log_level` for example

qingling128:
Follow up will be tracked in the upstream Fluent Bit feature request instead. This one is then obsolete, right?

PR #329, opened 2022-01-14
Add logging submodule uptime metrics.

rafaelwestphal:
Redirect the metrics from fluentbit to an output.


confgenerator/loggingmetrics.go
```diff
@@ -0,0 +1,64 @@
...
+			Type: "prometheus",
+			Config: map[string]interface{}{
+				"config": map[string]interface{}{
+					"scrape_configs": []map[string]interface{}{{
+						"job_name":        "logging-collector",
+						"scrape_interval": "1m",
+						"metrics_path":    "/metrics",
+						"static_configs": []map[string]interface{}{{
+							// TODO(b/196990135): Customization for the port number
+							"targets": []string{"0.0.0.0:2222"},
```
qingling128, line 38: Should this be 20202 instead given the latest implementation of https://github.com/GoogleCloudPlatform/ops-agent/pull/330?

confgenerator/loggingmetrics.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2021 Google LLC
```
qingling128, line 1: Agent self metrics related logic are mainly in https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/agentmetrics.go. Do we need a separate file (if so, let's rename the other file so it's clearly only doing the metrics subagent side of things), or can we just combine them as one file?

BTW, https://github.com/GoogleCloudPlatform/ops-agent/blob/2ea1dd954bff3d6b5279a3323c4361c88888749f/confgenerator/agentmetrics.go#L22 is designed to work for both logging subagent and metrics subagent's VERSION label. We could potentially re-use it.

qingling128:
To make the PR title more user friendly (because it will be used to generate release notes), can we rename it to something similar to:

```
Add logging submodule uptime metrics.
```

We can include the detailed description in the PR description.

confgenerator/loggingmetrics.go
rafaelwestphal: Fixed this in the new version

confgenerator/loggingmetrics.go
rafaelwestphal: I think keeping on the same file makes it a bit simpler, so I adopted the suggestion and put the implementation on agentmetrics.go . I think we still need to different pipelines since the VERSION label is different.

confgenerator/confgenerator.go
```diff
@@ -43,9 +45,13 @@ func (uc *UnifiedConfig) GenerateOtelConfig(hostInfo *host.InfoStat) (string, er
 	}
 
 	pipelines["agent"] = MetricsReceiverAgent{
-		Version: versionLabel,
+		Version: metricVersionLabel,
```
qingling128, line 48: For readability, can we adjust the naming a bit:

```
	pipelines["agent"] = AgentSelfMetrics{
		Version: metricVersionLabel,
	}.MetricsSubmodulePipeline()

	pipelines["fluentbit"] = AgentSelfMetrics{
		Version: loggingVersionLabel,
	}.LoggingSubmodulePipeline()
```

Another alternative is:

```
	pipelines["agent"] = AgentSelfMetrics{
		LoggingSubmoduleVersion: loggingVersionLabel,
		MetricsSubmoduleVersion: metricVersionLabel,
	}.Pipelines() // make this return a list
```

But that requires more refactoring.

Either way is fine with me.

quentinmit: It seems like this new pipeline is completely separate from the existing one - I would rather see a new struct be introduced for fluentbit instead of having two independent methods on the same struct. I think I even prefer that over @qingling128's second alternative above.

rafaelwestphal: Adopted the first suggestion.

confgenerator/confgenerator.go
```diff
@@ -42,9 +44,13 @@ func (uc *UnifiedConfig) GenerateOtelConfig(hostInfo *host.InfoStat) (string, er
 		}
 	}
 
-	pipelines["agent"] = MetricsReceiverAgent{
-		Version: versionLabel,
-	}.Pipeline()
+	pipelines["agent"] = AgentSelfMetrics{
```
qingling128, line 47:
```suggestion
	pipelines["otel"] = AgentSelfMetrics{
```

qingling128 (approved):
LGTM modulo one minor suggestion. Feel free to submit once resolved (Might need to update golden)

qingling128:
The failures in `third party apps integration test` are tomcat issues. This can be merged.

confgenerator/loggingmetrics.go
qingling128: Thanks!

PR #331, opened 2022-01-18
fix:Update to use [:upper:] [:lower:] to support accents and foreign alphabets

bluPhy:
…alphabets

Update to use [:upper:] [:lower:] to support accents and foreign alphabets. The output from lsb_release most likely never produce non-ASCII characters but as a matter of best practice I propose to change it.


qingling128:
In fact, on a second thought, all the build VMs we use today to build the agent comes with English by default. So this change might be a noop.

qingling128:
The integration tests passed before:
image

Merging

qingling128 (approved):
LGTM

igorpeshansky:
This change is a no-op; everyone else seems to assume that `/etc/lsb-release` will contain ASCII data (in fact, the Python implementation uses the `ascii` codec for everything: https://lists.debian.org/debian-lsb/2012/06/msg00000.html).

PR #332, opened 2022-01-19
Copy over tests and testing libraries from Google's code silo.

martijnvans:
These files are not yet picked up by Kokoro. They run, mostly
successfully, with the following commands, which do need some setup:
1. PROJECT needs to be replaced with some valid project
3. TRANSFERS_BUCKET needs to be set to a GCS bucket to use to upload files 
2. GOOGLE_APPLICATION_CREDENTIALS needs to be set to something
    that can operate on that project and upload to the bucket.

PROJECT=todo \
  TRANSFERS_BUCKET=todo \
  GOOGLE_APPLICATION_CREDENTIALS="${HOME}/credentials.json" \
  ZONE=us-central1-b \
  PLATFORMS=debian-10 \
  AGENTS_TO_TEST=ops-agent \
  go test -v ops_agent_test.go \
  -test.parallel=1000 \
  -tags=integration_test \
  -timeout=4h

PROJECT=todo \
  TRANSFERS_BUCKET=todo \
  GOOGLE_APPLICATION_CREDENTIALS="${HOME}/credentials.json" \
  ZONE=us-central1-b \
  PLATFORMS=debian-10 \
  SCRIPTS_DIR=third_party_apps_data \
  AGENTS_TO_TEST=ops-agent \
  go test -v third_party_apps_test.go \
  -test.parallel=1000 \
  -tags=integration_test \
  -timeout=4h

This PR is intended to be publicly visible, and should be close to its
final form, but it's not intended to be merged yet. First I will make
Kokoro able to pick up the files and run with those instead of pulling
them from Google's code silo. Once that is working, we can consider
merging this PR or something like it.


jkschulz (approved):

integration_test/logging/logging_test.go
```diff
@@ -83,8 +87,18 @@ func TestInvalidFile(t *testing.T) {
...
 
+	defer func() {
+		if err := dirLog.Close(); err != nil {
+			t.Errorf("dirLog.Close() failed with err=%v", err)
+		}
+	}()
+
 	// This will result in an error because /etc is already a directory.
 	invalidPath := "../../../../../../../../../etc"
+	if runtime.GOOS == "windows" {
```
jkschulz, line 98: Can you add a comment here that explains that this check is 
1. necessary only because this test is bundled into the other ops agent tests (which runs on both linux and windows), but in practice, the logger is only run on linux, and 
2. that future work could be to figure out how to set up github workflows/actions to separate out this test so that it's only run when changes are made to the logger itself, and only on linux (meaning this check could be deleted)

martijnvans: done, I mostly just copy-pasted what you wrote because it sounded good to me.

PR #333, opened 2022-01-19
Added validation for logging processor for requirement of time_format when time_key is present.

franciscovalentecastro:
- Updated logging_processors.go so time_format is required when time_key is present and the other way around.
- Added input.yaml and generated golden_error files in the /testdata/invalid/ folder.
- Updated /valid/logging-processor_parse_json_and_parse_regex_types/input.yaml to remove the cases where time_format or time_key are missing.

### DIFF
--- a/confgenerator/testdata/invalid/linux/logging-processor_parse_json_type_missing_time_format_required_with_time_key/golden_error
+++ b/confgenerator/testdata/invalid/linux/logging-processor_parse_json_type_missing_time_format_required_with_time_key/golden_error
@@ -1,0 +1,1 @@
+Key: 'ParserShared.time_format' Error:Field validation for 'time_format' failed on the 'required_with' tag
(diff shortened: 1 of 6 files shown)


confgenerator/testdata/invalid/linux/logging-processor_parse_json_type_missing_time_format_required_with_time_key/golden_error
```diff
@@ -0,0 +1 @@
+"time_format" is required when "TimeKey" is set
```
qingling128, line 1: [Might not be directly related to this PR, and not a huge deal] But should `TimeKey` be `time_key` instead? It could be related to the common util: https://github.com/GoogleCloudPlatform/ops-agent/blob/11dc4f1cf20e99cbcd09c3c298b5845b4e7b0f59/confgenerator/config.go#L103

Ignore unless it's a low hanging fruit.

qingling128 (approved):
LGTM

confgenerator/testdata/invalid/linux/logging-processor_parse_json_type_missing_time_format_required_with_time_key/golden_error
franciscovalentecastro: After exploring the validator library, I've found this is a common request of getting the "Field" name (snake case) from the "Param" as discussed here : https://github.com/go-playground/validator/issues/597 . 

The other way is converting `TimeKey` to SnakeCase (`time_key`) using a go library, but i don't think that's a good solution. 

I think that an update to the https://github.com/go-playground/validator library would be the best solution. I will leave it as it is in this PR.

sophieyfang (approved):

PR #337, opened 2022-01-21
Fix tomcat download URL

dehaansa:
This is a temporary fix to get integration tests passing for tomcat.

It appears the CDN we're using to download tomcat for integration tests only offers one version and there wasn't an obvious mirror with older versions remaining available. Longer term fix will be to download from an appropriate package repo or find a better CDN.


dehaansa:
One sles failure on current kokoro run.

qingling128 (approved):
LGTM

PR #342, opened 2022-01-21
Use a tomcat archive version instead of latest version cdn to fix flaky tests and extend the default log file paths.

dehaansa:
Where the previous temporary fix fixed tests, this one points at a repository that appears to not only be latest version & won't require updating intermittently.

PR #344, opened 2022-01-24
Update Tomcat Install scripts

Mrod1598:
The current Tomcat integration tests are flawed in how the tars are pulled. the mirror deletes the previous version when a new one is uploaded. So, changed to using packages instead. Also, added sleep statement at beginning of exercise script to help with flakiness of not being able to hit the endpoint sometimes.


apps/tomcat.go
```diff
@@ -165,7 +165,7 @@ type SystemLoggingReceiverTomcat struct {
 func (r SystemLoggingReceiverTomcat) Components(tag string) []fluentbit.Component {
 	if len(r.IncludePaths) == 0 {
 		r.IncludePaths = []string{
-			"/opt/tomcat/logs/catalina.out",
+			"/var/log/tomcat9/catalina.out",
```
qingling128, line 168: What impact does this change have? Looks like we are changing the default log file path. Does any distro use `/opt/tomcat/logs/catalina.out` instead? If so, should we include both?

ridwanmsharif:
The sleep part of this PR will reduce the flakiness we're seeing in the third party integration tests. This is currently blocking our 2.9.0 release of the Ops Agent

apps/tomcat.go
Mrod1598: I'll Add both, it seems the package manager uses what I changed it to and I'm seeing more and more the `/var/log` but I've seen both and tomcat doesn't explicitly call it out in the docs that I can tell so I'll add both.

quentinmit: What about tomcat10? Or tomcat8? I think we're going to need multiple file paths to cover different versions and distros.

integration_test/third_party_apps_data/applications/tomcat/centos_rhel/install
```diff
@@ -1,39 +1,27 @@
 set -e
 
 sudo yum install -y curl java
-sudo mkdir -p /opt/tomcat/stage
-sudo curl "https://dlcdn.apache.org/tomcat/tomcat-9/v9.0.58/bin/apache-tomcat-9.0.58.tar.gz" -o /opt/tomcat/stage/tomcat.tgz
-sudo tar -xvzf /opt/tomcat/stage/tomcat.tgz -C /opt/tomcat --strip 1
-
-cat << EOF > tomcat.service
+sudo yum install tomcat -y
+sudo cat << EOF | sudo tee /usr/lib/systemd/system/tomcat.service
```
quentinmit, line 5: Does the tomcat package not come with a systemd unit? Either way you shouldn't be overwriting anything in `/usr/lib`. If you need to add something to the unit file, you should be writing an override file in `/etc/systemd/system` (and there should be a good reason to do that).

quentinmit (changes requested):

Mrod1598:
Clossing This PR because https://github.com/GoogleCloudPlatform/ops-agent/pull/342 seems to fix it. I'll address the Include path issues in a separate PR.

quentinmit:
I think this PR is preferable to #342

Mrod1598:
> I think this PR is preferable to #342

yes, however, if this is blocking a release I'd rather have the other one merged while I investigate a better way to handle this as it is still failing on this branch.

qingling128:
+1 that #344 is preferable to #342. I'd prefer the user visible changes (aka the `includePaths` changes`) to go in this release as well. The installation related changes are fine to get in after the release.

integration_test/third_party_apps_data/applications/tomcat/centos_rhel/install
Mrod1598: Updated and there needs to be updated to the service file to enable jmx.

Mrod1598:
> +1 that #344 is preferable to #342. I'd prefer the user visible changes (aka the `includePaths` changes`) to go in this release as well. The installation related changes are fine to get in after the release.

reopened and updated log paths.

docs/tomcat.md
```diff
@@ -57,15 +57,15 @@ To configure a receiver for your tomcat system logs, specify the following field
 | Field                 | Default                           | Description |
 | ---                   | ---                               | ---         |
 | `type`                | required                          | Must be `tomcat_system`. |
-| `include_paths`       | `[/opt/tomcat/logs/catalina.out]` | A list of filesystem paths to read by tailing each file. A wild card (`*`) can be used in the paths; for example, `/var/log/apache*/*.log`.
+| `include_paths`       | `[/var/log/tomcat9/catalina.out]` | A list of filesystem paths to read by tailing each file. A wild card (`*`) can be used in the paths; for example, `/var/log/apache*/*.log`.
```
qingling128, line 60: We changed this to wildcard in #342, right?

Mrod1598: yes, looks like the doc update was missed

qingling128:
The Tomcat tests seem to be failing: 
image

integration_test/third_party_apps_data/applications/tomcat/centos_rhel/install
```diff
@@ -1,39 +1,27 @@
 set -e
 
 sudo yum install -y curl java
-sudo mkdir -p /opt/tomcat/stage
-sudo curl "https://archive.apache.org/dist/tomcat/tomcat-9/v9.0.58/bin/apache-tomcat-9.0.58.tar.gz" -o /opt/tomcat/stage/tomcat.tgz
-sudo tar -xvzf /opt/tomcat/stage/tomcat.tgz -C /opt/tomcat --strip 1
-
-cat << EOF > tomcat.service
+sudo yum install tomcat -y
+sudo cat << EOF | sudo tee /etc/systemd/system/tomcat.service
```
quentinmit, line 5: At least this is in /etc instead of /usr, but please use a systemd drop-in and just customize the one thing you need (the environment variables?). See https://www.freedesktop.org/software/systemd/man/systemd.unit.html#id-1.14.3

quentinmit (dismissed):

integration_test/third_party_apps_data/applications/tomcat/centos_rhel/install
qingling128: Thanks!

qingling128 (approved):
LGTM modulo the integration tests pass

qingling128:
Gentle ping on this PR. We are trying to do a release today, and it's blocked by the Tomcat failures.

Mrod1598:
CentOS script has to be left as is as the only available package is tomcat7 which we don't support

integration_test/third_party_apps_data/applications/tomcat/centos_rhel/install
qingling128: This one seems obsolete now. It later turned out that the only available package is tomcat7 on CentOS, which we don't support. So this change has been reverted.

PR #346, opened 2022-01-24
Solr Metrics Receiver

StefanKurek:
Adds ability to collect Solr metrics to ops-agent.

[Scope doc](https://docs.google.com/document/d/1GiSjVfshBnTr4-838bLF2EmCY_SUkz9oakk9bbJP9V0/edit#)

### DIFF
--- a/apps/solr.go
+++ b/apps/solr.go
@@ -1,0 +1,81 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apps
+
+import (
+	"log"
+
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
+)
+
+type MetricsReceiverSolr struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+
+	Endpoint string `yaml:"endpoint" validate:"omitempty,hostname_port"`
+	Username string `yaml:"username" validate:"required_with=Password"`
+	Password string `yaml:"password" validate:"required_with=Username"`
+}
+
+const defaultSolrEndpoint = "localhost:18983"
+
+func (r MetricsReceiverSolr) Type() string {
+	return "solr"
+}
+
+func (r MetricsReceiverSolr) Pipelines() []otel.Pipeline {
+	if r.Endpoint == "" {
+		r.Endpoint = defaultSolrEndpoint
+	}
+
+	jarPath, err := FindJarPath()
+	if err != nil {
+		log.Printf(`Encountered an error discovering the location of the JMX Metrics Exporter, %v`, err)
+	}
+
+	targetSystem := "solr"
+
+	config := map[string]interface{}{
+		"target_system":       targetSystem,
+		"collection_interval": r.CollectionIntervalString(),
+		"endpoint":            r.Endpoint,
+		"jar_path":            jarPath,
+	}
+
+	// Only set the username & password fields if provided
+	if r.Username != "" && r.Password != "" {
+		config["username"] = r.Username
+		config["password"] = r.Password
+	}
+
+	return []otel.Pipeline{{
+		Receiver: otel.Component{
+			Type:   "jmx",
+			Config: config,
+		},
+		Processors: []otel.Component{
+			otel.NormalizeSums(),
+			otel.MetricsTransform(
+				otel.AddPrefix("workload.googleapis.com"),
+			),
+		},
+	}}
+}
+
+func init() {
+	confgenerator.MetricsReceiverTypes.RegisterType(func() confgenerator.Component { return &MetricsReceiverSolr{} })
+}
--- a/apps/solr.go
+++ b/apps/solr.go
@@ -15,8 +15,6 @@
 package apps
 
 import (
-	"log"
-
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
 )
@@ -26,9 +24,7 @@ type MetricsReceiverSolr struct {
 
 	confgenerator.MetricsReceiverShared `yaml:",inline"`
 
-	Endpoint string `yaml:"endpoint" validate:"omitempty,hostname_port"`
-	Username string `yaml:"username" validate:"required_with=Password"`
-	Password string `yaml:"password" validate:"required_with=Username"`
+	confgenerator.MetricsReceiverSharedJVM `yaml:",inline"`
 }
 
 const defaultSolrEndpoint = "localhost:18983"
@@ -38,42 +34,19 @@ func (r MetricsReceiverSolr) Type() string {
 }
 
 func (r MetricsReceiverSolr) Pipelines() []otel.Pipeline {
-	if r.Endpoint == "" {
-		r.Endpoint = defaultSolrEndpoint
-	}
-
-	jarPath, err := FindJarPath()
-	if err != nil {
-		log.Printf(`Encountered an error discovering the location of the JMX Metrics Exporter, %v`, err)
-	}
-
 	targetSystem := "solr"
 
-	config := map[string]interface{}{
-		"target_system":       targetSystem,
-		"collection_interval": r.CollectionIntervalString(),
-		"endpoint":            r.Endpoint,
-		"jar_path":            jarPath,
-	}
-
-	// Only set the username & password fields if provided
-	if r.Username != "" && r.Password != "" {
-		config["username"] = r.Username
-		config["password"] = r.Password
-	}
-
-	return []otel.Pipeline{{
-		Receiver: otel.Component{
-			Type:   "jmx",
-			Config: config,
-		},
-		Processors: []otel.Component{
+	return r.MetricsReceiverSharedJVM.JVMConfig(
+		targetSystem,
+		defaultSolrEndpoint,
+		r.CollectionIntervalString(),
+		[]otel.Component{
 			otel.NormalizeSums(),
 			otel.MetricsTransform(
 				otel.AddPrefix("workload.googleapis.com"),
 			),
 		},
-	}}
+	)
 }
 
 func init() {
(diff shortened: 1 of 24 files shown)


apps/solr.go
```diff
@@ -0,0 +1,81 @@
...
+func (r MetricsReceiverSolr) Type() string {
+	return "solr"
+}
+
+func (r MetricsReceiverSolr) Pipelines() []otel.Pipeline {
+	if r.Endpoint == "" {
+		r.Endpoint = defaultSolrEndpoint
+	}
+
+	jarPath, err := FindJarPath()
```
qingling128, line 45: Can we refactor this part a bit to make use of the common util in https://github.com/GoogleCloudPlatform/ops-agent/pull/370?

confgenerator/testdata/invalid/linux/metrics-receiver_solr_malformed_url/input.yaml
```diff
@@ -0,0 +1,25 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+metrics:
+  receivers:
+    solr_metrics:
```
qingling128, line 17: It's preferred for the receiver and pipeline IDs to stay the same as the receiver type in the sample configuration. E.g.  "s/{APP}_metrics/{APP}", "s/{APP}_pipeline/{APP}"

confgenerator/testdata/valid/linux/metrics-receiver_solr/input.yaml
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+
+metrics:
+  receivers:
+    solr_metrics:
+      type: solr
+      # exercising default endpoint
+      username: otelu
+      password: otelp
+      collection_interval: 30s
```
qingling128, line 22: Remove `collection_interval` here?

For the context: There should at least be one valid unit test that verifies the absolute minimum user configuration. All optional configurations should be absent for this case including "collection_interval".

docs/solr.md
```diff
@@ -0,0 +1,50 @@
...
+| `type`                | required           | Must be `solr`. |
+| `endpoint`            | `localhost:18983`  | The [JMX Service URL](https://docs.oracle.com/javase/8/docs/api/javax/management/remote/JMXServiceURL.html) or host and port used to construct the Service URL. Must be in the form of `host:port`. Values in `host:port` form will be used to create a Service URL of `service:jmx:rmi:///jndi/rmi://<host>:<port>/jmxrmi`. |
+| `username`            | not set by default | The configured username if JMX is configured to require authentication. |
+| `password`            | not set by default | The configured password if JMX is configured to require authentication. |
+| `collection_interval` | `60s`              | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
+
+
+Example Configuration:
+
+```yaml
```
qingling128, line 22: It's preferred for the receiver and pipeline IDs to stay the same as the receiver type in the sample configuration. E.g.  "s/{APP}_metrics/{APP}", "s/{APP}_pipeline/{APP}"

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_solr
```diff
@@ -0,0 +1,33 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+metrics:
+  receivers:
+    solr_metrics:
+      type: solr
+      endpoint: localhost:18983
+      collection_interval: 60s
```
qingling128, line 25: Remove `collection_interval`?

integration_test/third_party_apps_data/agent/ops-agent/linux/supported_applications.txt
```diff
@@ -1,10 +1,11 @@
 apache
 cassandra
+elasticsearch
```
qingling128, line 3: There are still some pending PRs for `elasticsearch`. Would this trigger a test failure by including `elasticsearch` early?

integration_test/third_party_apps_data/applications/solr/centos_rhel/install
```diff
@@ -0,0 +1,40 @@
...
+SELINUX=disabled
+SELINUXTYPE=targeted
+EOF
+
+sudo yum install -y \
+    java-1.8.0-openjdk java-1.8.0-openjdk-devel curl lsof
+
+curl -L -o \
+    solr-8.11.1.tgz \
+    https://downloads.apache.org/lucene/solr/8.11.1/solr-8.11.1.tgz
```
qingling128, line 15: Is it possible to use regular yum/apt-get/zypper install?

Chances are a bit higher that users will use the default package manager (e.g. yum, apt-get, zyypper) to install a package instead of a pinned version of the application. The install (aka integration_test/third_party_apps_data/applications/{APP}/{DISTRO}/install) script should try reflect the native installation methods as much as possible.

Same for the other `install` scripts.

integration_test/third_party_apps_data/agent/ops-agent/linux/supported_applications.txt
StefanKurek: @qingling128 i didn't actually add elasticsearch. i just moved it to an alphabetical spot

integration_test/third_party_apps_data/applications/solr/centos_rhel/install
StefanKurek: @qingling128 it doesn't look like this option is available for solr. their official install instructions just show using a tarball

StefanKurek:
@qingling128 thanks for the feedback! can you take another look when you have a moment?

integration_test/third_party_apps_data/applications/solr/centos_rhel/install
quentinmit: Debian 9 and 10 appear to have packages for solr: https://packages.debian.org/buster/solr-tomcat
as do Ubuntu focal and bionic: https://packages.ubuntu.com/focal/solr-tomcat

integration_test/third_party_apps_data/applications/solr/centos_rhel/install
```diff
@@ -0,0 +1,40 @@
...
+tar -xzf \
+    solr-8.11.1.tgz \
+    solr-8.11.1/bin/install_solr_service.sh \
+    --strip-components=2
+
+sudo bash ./install_solr_service.sh solr-8.11.1.tgz -n
+
+sudo chown -R solr:solr /opt/solr*
+
+cat <<EOF | sudo tee /etc/default/solr.in.sh
```
quentinmit, line 26: Can you edit the existing file in place (e.g. with `sed`) instead of copying the default values here? That will ensure that we are compatible with defaults even if the defaults change.

integration_test/third_party_apps_data/applications/solr/centos_rhel/post
```diff
@@ -0,0 +1,5 @@
+set -e
+
+sudo systemctl restart solr
+
+sudo su - solr -c "/opt/solr/bin/solr create -c otel"
```
quentinmit, line 5: Is this necessary for the metrics to exist? Can you add a comment describing what this is doing? (maybe `# Create a database called otel so there are some metrics to scrape`?)

docs/solr.md
qingling128: Thanks!

integration_test/third_party_apps_data/agent/ops-agent/linux/supported_applications.txt
qingling128: Oops. Missed the last line. I see. Resolving this.

qingling128 (approved):
LGTM modulo the remaining open threads get resolved.

integration_test/third_party_apps_data/applications/solr/centos_rhel/post
StefanKurek: @quentinmit it is. i can throw a comment in there.

integration_test/third_party_apps_data/applications/solr/centos_rhel/install
StefanKurek: @quentinmit I don't believe those are what we want to use (and conferred with our dev ops team about this) for a base Solr installation. We are following the instructions of what is in the official install doc https://solr.apache.org/guide/7_0/installing-solr.html

integration_test/third_party_apps_data/applications/solr/centos_rhel/post
StefanKurek: @quentinmit done

StefanKurek:
@quentinmit thanks for the feedback. I have updated this so please take another look when you have a chance to resolve the remaining conversations. Thanks!

integration_test/third_party_apps_data/applications/solr/centos_rhel/post
quentinmit: Actually I unresolved this, this also seems like something that's important enough to put in the Markdown docs too (something like "Note: No metrics will be reported if the Solr instance does not contain any cores.")

integration_test/third_party_apps_data/applications/solr/centos_rhel/install
quentinmit: Debian's docs say that `apt install solr-tomcat` will give you a running Solr with zero configuration: https://wiki.debian.org/Solr

I'll leave it up to @qingling128 what we want to do here. I'd love if we could test both installation methods - the most important part of these integration tests is testing distro-specific details, since otherwise we could just do all the tests on a single distro with manually-installed software. So maybe use the packages on the Debian and Ubuntu releases that have them, and the tarball on the other distros, so we cover them both.

integration_test/third_party_apps_data/applications/solr/centos_rhel/post
StefanKurek: @quentinmit I can add that. But basically Solr is unusable without any cores added, so I don't think a customer will ever set it up without one

integration_test/third_party_apps_data/applications/solr/centos_rhel/install
StefanKurek: @quentinmit so I tried out that package, but the final installation state was very different than when using the tarball. Also, the documentation that you linked to (https://wiki.debian.org/Solr) refers to features like solr.xml which was deprecated in 2014 and unsupported in 2015. I think it's installing a pretty old version of Solr.

integration_test/third_party_apps_data/applications/solr/centos_rhel/post
StefanKurek: @quentinmit added a comment in the docs

qingling128: Thanks!

integration_test/third_party_apps_data/applications/solr/centos_rhel/install
qingling128: Started a side conversation around `Test Ops Agent Supported Application Versions`. Shared the doc. Feel free to chime in there. I think we can keep discussing this separately (aka not a blocker for this PR).

qingling128:
The open threads seem to have all been resolved.

Waiting for the integration tests for a final check before merging.

PR #350, opened 2022-01-24
Add Hbase metrics Receiver

Mrod1598:
Add Hbase metrics Receiver. There were some requested metrics additions upstream, Including adding `mean`, `min`, and `median` to all operations.


apps/hbase.go
```diff
@@ -0,0 +1,87 @@
...
+func (r MetricsReceiverHbase) Type() string {
+	return "hbase"
+}
+
+func (r MetricsReceiverHbase) Pipelines() []otel.Pipeline {
+	if r.Endpoint == "" {
+		r.Endpoint = defaultHbaseEndpoint
+	}
+
+	jarPath, err := FindJarPath()
```
qingling128, line 48: Can we refactor this part a bit to use the common util by https://github.com/GoogleCloudPlatform/ops-agent/pull/370?

integration_test/third_party_apps_data/applications/hbase/centos_rhel/install
```diff
@@ -0,0 +1,36 @@
+set -e
+
+sudo yum -y install \
+    java-1.8.0-openjdk java-1.8.0-openjdk-devel wget
+
+sudo mkdir /opt/hbase
+wget http://apache.mirror.gtcomm.net/hbase/stable/hbase-2.4.9-bin.tar.gz
```
qingling128, line 7: Chances are a bit higher that users will use the default package manager (e.g. yum, apt-get, zyypper) to install a package instead of a pinned version of the application.

Can we adjust it to do something like `yum install hadoop` instead? Similar requests for the other `install` scripts.

docs/hbase.md
```diff
@@ -0,0 +1,90 @@
...
+| `username`   | not set by default | The configured username if JMX is configured to require authentication. |
+| `password`   | not set by default | The configured password if JMX is configured to require authentication. |
+| `collection_interval` | `60s`    | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
+
+Example Configuration:
+
+```yaml
+metrics:
+ receivers:
+ hbase_metrics:
```
qingling128, line 25: It's preferred for the receiver and pipeline IDs to stay the same as the receiver type in the sample configuration. E.g.  "s/{APP}_metrics/{APP}", "s/{APP}_pipeline/{APP}"

docs/hbase.md
```diff
@@ -0,0 +1,90 @@
...
+  type: hbase
+  collection_interval: 60s
```
qingling128, line 27: Remove `collection_interval` as it's optional now.

confgenerator/testdata/valid/linux/metrics-receiver_hbase/input.yaml
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+
+metrics:
+  receivers:
+    hbase_metrics:
+      type: hbase
+      # exercising default endpoint
+      username: otelu
+      password: otelp
+      collection_interval: 30s
```
qingling128, line 22: Remove `collection_interval` as it's no longer required?

integration_test/third_party_apps_data/applications/hbase/sles/post
```diff
@@ -0,0 +1,4 @@
+set -e
+
+sudo systemctl restart hbase 
+sleep 30
```
qingling128, line 4: For a metrics side only PR, the Fluent Bit submodule should not be updated.

image

integration_test/third_party_apps_data/applications/hbase/centos_rhel/install
Mrod1598: There doesn't seem to be one for hbase either.

integration_test/third_party_apps_data/applications/hbase/sles/post
Mrod1598: Changes Removed

confgenerator/testdata/valid/linux/metrics-receiver_hbase/input.yaml
Mrod1598: Removed

docs/hbase.md
Mrod1598: Removed

docs/hbase.md
Mrod1598: Updated

apps/hbase.go
Mrod1598: Updated

docs/hbase.md
```diff
@@ -0,0 +1,89 @@
...
+  receivers:
+   - hbase
+```
+
+## Metrics
+In addition to Hbase specific metrics, by default Hbase will also report [JVM metrics](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/docs/jvm.md#metrics)
+
+| Metric                                                 | Data Type  | Unit           | Labels                       | Description |
+| ---                                                    | ---        | ---            | ---                          | ---   |  
+| `hbase.master.region_server.count` | Gauge | `{servers}` | `state` | The number of region servers. |
```
quentinmit, line 39: Document the values of the `state` label?

docs/hbase.md
```diff
@@ -0,0 +1,89 @@
+# `hbase` Metrics Receiver
+
+The `hbase` metrics receiver can fetch stats from a Hbase server's Java Virtual Machine (JVM) via [JMX](https://www.oracle.com/java/technologies/javase/javamanagement.html).
```
quentinmit, line 3: Please explain in the docs how the cluster metrics are collected (does it only collect the local region server's metrics? or metrics for all the region servers in the cluster?).

docs/hbase.md
```diff
@@ -0,0 +1,89 @@
...
+## Metrics
+In addition to Hbase specific metrics, by default Hbase will also report [JVM metrics](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/docs/jvm.md#metrics)
+
+| Metric                                                 | Data Type  | Unit           | Labels                       | Description |
+| ---                                                    | ---        | ---            | ---                          | ---   |  
+| `hbase.master.region_server.count` | Gauge | `{servers}` | `state` | The number of region servers. |
+| `hbase.master.regions_in_transition.count` | Gauge | `{regions}` |  | The number of regions that are in transition. |
+| `hbase.master.regions_in_transition.over_threshold` | Gauge | `{regions}` |  | The number of regions that have been in transition longer than a threshold time. |
+| `hbase.master.regions_in_transition.oldest_age` | Gauge | `ms` |  | The age of the longest region in transition. |
+| `hbase.region_server.region.count` | Gauge | `{regions}` | `region_server` | The number of regions hosted by the region server. |
```
quentinmit, line 43: These metrics all show a `region_server` label but the scoping doc says that label is a resource label and so won't be presentt.

integration_test/third_party_apps_data/applications/hbase/centos_rhel/install
qingling128: Got it. Guess we have to maintain this piece of code then. NVM

docs/hbase.md
Mrod1598: Due to limitation in the JMX collector, it does not distinguish between normal and resource attributes. So, it will be visible to the customer as a normal attribute. This may change down the round as the collector evolves.

docs/hbase.md
Mrod1598: Added

docs/hbase.md
Mrod1598: Added

docs/hbase.md
quentinmit: We should strip the label using OT config then, if the label always has a single value.

docs/hbase.md
```diff
@@ -0,0 +1,102 @@
+# `hbase` Metrics Receiver
+
+The `hbase` metric receiver can fetch stats from a HBase server's Java Virtual Machine (JVM) via [JMX](https://www.oracle.com/java/technologies/javase/javamanagement.html). It collects metrics specific to the local region server, as well as metrics presented by the Master node if the node being monitored is indeed the Master.
+
+For High Availability configurations it is recommended to monitor the active Master node only. Monitoring the active Master and its replicants will result in duplicate metrics for the Master node.
```
quentinmit, line 5: Please try to make this paragraph clearer to explain that it only applies to master nodes; I read it as saying "you should only configure the receiver on one node", which isn't true - you should configure it on "one master node and  all region server nodes".

If the "if the node being monitored is indeed the Master" is accurate, can't you just configure it on all of the nodes and there just won't be master metrics on the replicants until the replicant becomes the master?

docs/hbase.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Prerequisites
+
+In order to expose a JMX endpoint, you must set the `com.sun.management.jmxremote.port` system property. It is recommended to also set the `com.sun.management.jmxremote.rmi.port` system property to the same port. To expose JMX endpoint remotely, you must also set the `java.rmi.server.hostname` system property. By default, these properties are set in a Hbase deployment's hbase-env.sh file and the default Hbase installation requires no JMX authentication with JMX exposed locally on 127.0.0.1:8050.
```
quentinmit, line 9: I don't think this "By default" sentence is accurate - your integration test has to manually enable JMX, and also the default port seems to be :10101 and not :8050.

qingling128:
BTW: There is some on-going conversation in the `Regional metrics and master metrics` section of the `Duplicate / Overlapping Metrics Issue from Clustered Applications` doc that is related. Just shared it.

docs/hbase.md
qingling128: Hbase has some pending discussions going on in the `Category 3 - Regional metrics and master metrics` section of the `Duplicate / Overlapping Metrics Issue from Clustered Applications` doc.

The phrasing here might be adjusted based on the decision. We'll need to resolve that conversion before releasing this.

qingling128 (approved):
LGTM modulo 2 minor comments.

After syncing to master, let's trigger an integration tests for a sanity check before merging, as the last green integration test run had been there for a while.

docs/hbase.md
```diff
@@ -0,0 +1,102 @@
...
+| `endpoint`   | `localhost:10101` | The [JMX Service URL](https://docs.oracle.com/javase/8/docs/api/javax/management/remote/JMXServiceURL.html) or host and port used to construct the Service URL. Must be in the form of `service:jmx:<protocol>:<sap>` or `host:port`. Values in `host:port` form will be used to create a Service URL of `service:jmx:rmi:///jndi/rmi://<host>:<port>/jmxrmi`. |
+| `collect_jvm_metrics` | true    | Should the set of support [JVM metrics](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/docs/jvm.md#metrics) also be collected |
+| `username`   | not set by default | The configured username if JMX is configured to require authentication. |
+| `password`   | not set by default | The configured password if JMX is configured to require authentication. |
+| `collection_interval` | `60s`    | A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`. |
+
+Example Configuration:
+
+```yaml
+metrics:
```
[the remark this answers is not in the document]
Mrod1598: Updated

docs/hbase.md
Mrod1598: Updated

PR #351, opened 2022-01-24
Add Hadoop metrics receiver

Mrod1598:
Add hadoop metrics receiver


confgenerator/testdata/invalid/linux/metrics-receiver_hadoop_malformed_url/input.yaml
```diff
@@ -0,0 +1,25 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+metrics:
+  receivers:
+    hadoop_metrics:
```
qingling128, line 17: It's preferred for the receiver and pipeline IDs to stay the same as the receiver type in the sample configuration. E.g.  "s/{APP}_metrics/{APP}", "s/{APP}_pipeline/{APP}"

confgenerator/testdata/valid/linux/metrics-receiver_hadoop/input.yaml
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+
+metrics:
+  receivers:
+    hadoop_metrics:
+      type: hadoop
+      # exercising default endpoint
+      username: otelu
+      password: otelp
+      collection_interval: 30s
```
qingling128, line 22: We could remove `collect_interval` from here as it's optional.

For the context: There should at least be one valid unit test that verifies the absolute minimum user configuration. All optional configurations should be absent for this case including "collection_interval".

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
```diff
@@ -0,0 +1,144 @@
...
+
+sudo yum -y install \
+    java-1.8.0-openjdk java-1.8.0-openjdk-devel wget
+
+sudo mkdir -p \
+    /opt/hadoop/logs \
+    /opt/volume/datanode \
+    /opt/volume/namenode
+
+wget https://dlcdn.apache.org/hadoop/common/hadoop-3.3.1/hadoop-3.3.1.tar.gz
```
qingling128, line 11: Chances are a bit higher that users will use the default package manager (e.g. yum, apt-get, zyypper) to install a package instead of a pinned version of the application.

Can we adjust it to do something like `yum install hadoop` instead? Similar requests for the other `install` scripts.

apps/hadoop.go
```diff
@@ -0,0 +1,87 @@
...
+func (r MetricsReceiverHadoop) Type() string {
+	return "hadoop"
+}
+
+func (r MetricsReceiverHadoop) Pipelines() []otel.Pipeline {
+	if r.Endpoint == "" {
+		r.Endpoint = defaultHadoopEndpoint
+	}
+
+	jarPath, err := FindJarPath()
```
qingling128, line 48: Can we refactor this part a bit to use the common util by https://github.com/GoogleCloudPlatform/ops-agent/pull/370?

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
Mrod1598: there doesn't seem to be an available package for hadoop.

confgenerator/testdata/valid/linux/metrics-receiver_hadoop/input.yaml
Mrod1598: Updated

apps/hadoop.go
Mrod1598: Updated

confgenerator/testdata/invalid/linux/metrics-receiver_hadoop_malformed_url/input.yaml
Mrod1598: Updated

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
quentinmit: I agree, I don't see any packaging for Hadoop. This seems like a fairly common thing for Java apps - hopefully we can get away from manual installation when we move on to non-Java apps.

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
```diff
@@ -0,0 +1,144 @@
...
+sudo tar -xf hadoop-3.3.1.tar.gz -C /opt/hadoop --strip 1
+
+JAVA_HOME="$(dirname "$(dirname "$(readlink "$(readlink "$(which javac)")")")")"
+
+cat <<EOF | sudo tee /opt/hadoop/env
```
quentinmit, line 16: Use an `EnvironmentFile` in the systemd unit so this isn't duplicated so many times?

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
qingling128: Ah NVM for this app then.

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_hadoop
```diff
@@ -0,0 +1,33 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+metrics:
+  receivers:
+    hadoop:
+      type: hadoop
+      endpoint: service:jmx:rmi:///jndi/rmi://127.0.0.1:8004/jmxrmi
+      collection_interval: 60s
```
qingling128, line 25: Remove `collection_interval` here as well?

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
```diff
@@ -0,0 +1,144 @@
...
+EOF
+
+getent group hadoop || sudo groupadd hadoop
+id -u hadoop || sudo useradd -s /bin/bash -d /opt/hadoop hadoop -g hadoop
+
+sudo chown -R hadoop:hadoop /opt/hadoop
+sudo chown -R hadoop:hadoop /opt/volume
+
+sudo -u hadoop bash -c "cat /opt/hadoop/env >> ~/.bashrc"
+sudo -u hadoop bash -c "echo -e 'y\n' | ssh-keygen -b 2048 -t rsa -f ~/.ssh/id_rsa -q -N \"\""
```
quentinmit, line 138: ... I don't want to know why the hadoop user needs to be able to ssh to itself, do I?

integration_test/third_party_apps_data/applications/hadoop/debian_ubuntu/install
```diff
@@ -0,0 +1,144 @@
+set -e
+
+sudo apt-get update
+sudo DEBIAN_FRONTEND=noninteractive apt-get install -y default-jdk default-jdk-headless wget
+
+sudo mkdir -p \
```
quentinmit, line 6: These scripts are nearly identical, can we refactor to share common code?

qingling128 (approved):
LGTM modulo the remaining open threads are addressed.

integration_test/third_party_apps_data/applications/hadoop/debian_ubuntu/install
Mrod1598: So, what is actually copied into the environment? is it the whole hadoop folder?

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
Mrod1598: I promise it's a [requirement](https://phoenixnap.com/kb/install-hadoop-ubuntu).

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
Mrod1598: We can't use an environment file because the environment variables are used in three spots `systemd unit`, `core-site.xml`, and `hadoop users bashrc` and the environement file, I don't believe, allows the export for the environment file.

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_hadoop
Mrod1598: Removed

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
quentinmit: Aha, I found it: https://stackoverflow.com/a/20477333
Apparently it uses ssh to start the daemons on data nodes. I'm kinda surprised that is needed on a single-node cluster.

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
quentinmit: I think you could put the export statement on a separate line e.g.
```
A=1
B=2
export A B
```
and systemd would ignore the export line. I haven't tested it though.

integration_test/third_party_apps_data/applications/hadoop/debian_ubuntu/install
quentinmit: I don't know the answer to that :/
@martijnvans perhaps?

qingling128: I was thinking about something similar. It would likely require some refactoring of the testing infra though to support a shared snippet.

```
- hadoop
   - install
      - common
      - debian_ubuntu
      - centos_rhel
      - sles
```

A follow up PR for this refactoring is better IMO instead of trying to doing that in this PR. So that this PR gets unblocked first, and we could keep the refactoring separate from a new feature.

martijnvans: ling is right, we'd need to introduce some kind of "common" script that gets run for all platforms. Such a change should be pretty easy.

qingling128:
Integration tests seem to be failing for Hadoop:

image

integration_test/third_party_apps_data/applications/hadoop/debian_ubuntu/install
qingling128: Cool. I'll leave this thread unresolved for easy reference (GitHub UI makes it hard to find resolved comments). It does not block this PR from being merged.

integration_test/third_party_apps_data/applications/hadoop/centos_rhel/install
Mrod1598: Haven't been able to apply this successfully. I can't seem to find anything online as an example of anyone doing this either. Seems that once we add in the ability to refactor out this into a separate common script this will be a lot cleaner anyways as a side effect, if that was your concern.

apps/hadoop.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
+)
+
+type MetricsReceiverHadoop struct {
+	confgenerator.ConfigComponent          `yaml:",inline"`
+	confgenerator.MetricsReceiverShared    `yaml:",inline"`
+	confgenerator.MetricsReceiverSharedJVM `yaml:",inline"`
+
+	CollectJVMMetics *bool `yaml:"collect_jvm_metrics"`
```
qingling128, line 29: Now that the Elasticsearch PR is merged, can we make use of the common util here as well? https://github.com/GoogleCloudPlatform/ops-agent/blame/c145e38a9640f55eb46761398ad5446ecc9cea07/confgenerator/config.go#L479

Since there is a planned follow up PR for Hadoop, and this change is not user visible, I'm okay with doing it in the follow up PR as well.

qingling128 (approved):
LGTM

apps/hadoop.go
Mrod1598: Updated!

qingling128: Thanks!

qingling128 (approved):
LGTM. Waiting for the integration tests before merging

qingling128:
`third party apps integration test (Debian)` is failing for debian 10 on tomcat, which is a known issue that is being fixed by a separate PR https://github.com/GoogleCloudPlatform/ops-agent/pull/344.

`third party apps integration test (CentOS)` passed in few other PRs but failed for this one. However the error seems to be a service account key issue. Might not be related to this PR.

docs/hadoop.md
```diff
@@ -0,0 +1,48 @@
+# `hadoop` Metrics Receiver
+
+The `hadoop` metrics receiver can fetch stats from a Hadoop server's Java Virtual Machine (JVM) via [JMX](https://www.oracle.com/java/technologies/javase/javamanagement.html).
+
```
qingling128, line 4: Just realized that Hadoop has some pending discussions going on in the `Category 3 - Regional metrics and master metrics` section of the `Duplicate / Overlapping Metrics Issue from Clustered Applications` doc.

We'll need to resolve that conversion before releasing this.

qingling128 (changes requested):

Mrod1598:
> `third party apps integration test (Debian)` is failing for debian 10 on tomcat, which is a known issue that is being fixed by a separate PR #344.
> 
> `third party apps integration test (CentOS)` passed in few other PRs but failed for this one. However the error seems to be a service account key issue. Might not be related to this PR.


It did seem to be a one off issue.
image

docs/hadoop.md
qingling128: Per the recent decision in the Duplicate / Overlapping Metrics Issue from Clustered Applications, let's add some context here:

> It collects metrics specific to the local region server, as well as metrics presented by the Master node if the node being monitored is indeed the Master. For High Availability configurations, it is recommended for every master node to report cluster metrics, even though they are identical, to avoid single point of failures when one master goes down.

qingling128 (approved):
LGTM modulo 2 minor comments.

apps/hadoop.go
```diff
@@ -0,0 +1,58 @@
...
+
+const defaultHadoopEndpoint = "localhost:8004"
+
+func (r MetricsReceiverHadoop) Type() string {
+	return "hadoop"
+}
+
+func (r MetricsReceiverHadoop) Pipelines() []otel.Pipeline {
+	targetSystem := "hadoop"
+	if r.CollectJVMMetrics == nil || *r.CollectJVMMetrics {
```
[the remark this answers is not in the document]
Mrod1598: Added

docs/hadoop.md
Mrod1598: Added

qingling128 (approved):
LGTM

Issue #363, opened 2022-01-26
nginx third_party_apps_test fails when run against a fresh project


martijnvans:
The error looks like:

2022/01/26 16:31:27 Attempt 1 of debian-10 test of nginx finished with err=WaitForMetric(metric="workload.googleapis.com/nginx.requests", extraFilters=[]): rpc error: code = NotFound desc = Can not find metric resource_container_ids: [REDACTED] request_context: REQUEST_CONTEXT_READ user_name: [REDACTED] filter: "type = \"workload.googleapis.com/nginx.requests\"", retryable=false

I think i know what is causing this and i have two fixes in mind:

- a sleep in the post install script for nginx
- improved retrying for this kind of error

PR #364, opened 2022-01-26
Fixes for nginx test when running against a fresh project

martijnvans:
See https://github.com/GoogleCloudPlatform/ops-agent/issues/363.

This PR was tested with the following command:

PROJECT=[REDACTED] \
TRANSFERS_BUCKET=[REDACTED] \
GOOGLE_APPLICATION_CREDENTIALS="${HOME}/credentials.json" \
ZONE=us-central1-b \
PLATFORMS=debian-10,centos-7 \
SCRIPTS_DIR=third_party_apps_data \
go test -v third_party_apps_test.go  -test.parallel=1000  -tags=integration_test  -timeout=4

And, it didn't pass, for various reasons. Mostly that my test file is at head but it's testing the stable agent, which is substantially behind head. But I believe the failures were unrelated to this change. (and nginx passed.) The interesting part of this change was changing the retry logic for metrics queries, and none of the failures that I saw were due to a metrics query not being retried when it was supposed to.


qingling128 (approved):
LGTM

integration_test/gce/gce_testing.go
```diff
@@ -345,7 +345,9 @@ func isRetriableLookupMetricError(err error) bool {
 	return strings.Contains(err.Error(), "Internal error") ||
 		strings.Contains(err.Error(), "The metric referenced by the provided filter is unknown") ||
 		// workload.googleapis.com/* domain metrics are created on first write, and may not be immediately queryable.
-		strings.Contains(err.Error(), "failed to look up metric workload.googleapis.com")
+		// The error doesn't always look the same, hopefully looking for "code = NotFound" will catch all variations.
+		strings.Contains(err.Error(), "failed to look up metric workload.googleapis.com") ||
+		strings.Contains(err.Error(), "code = NotFound")
```
quentinmit, line 350: Is it possible to check the code directly instead of checking the string version?

martijnvans: (with grouchiness) this api doesn't seem to document what kind of error object is being returned, and looking at the code didn't help either: https://github.com/googleapis/google-cloud-go/blob/monitoring/v1.2.0/monitoring/apiv3/metric_client.go#L362

it could be some kind of gax-ian error object, or it could be decided by c.metricClient.ListTimeSeries, which just links back to itself, very unhelpful.

I agree that matching on the error string is gross but don't have a better solution at the moment. suggestions welcome.

quentinmit: I would expect it returns an error compatible with [`status.FromError`](https://pkg.go.dev/google.golang.org/grpc/status#FromError), which returns a `Status` object that has a `Code()` method that you can compare against `codes.NotFound`.

I would also expect "failed to look up metric" to be paired with a code (perhaps even `NotFound`), though maybe you want to explicitly match on `workload.googleapis.com` for some reason?

martijnvans: OK I changed it to just look for NotFound (and Internal). Much cleaner and hopefully more robust to tiny changes in the error message. It seems to work.

PR #365, opened 2022-01-26
Delete some unused scripts

martijnvans:
Specifically, I'm deleting:

integration_test/third_party_apps_data/agent/metrics/...           # collectd support scripts
integration_test/third_party_apps_data/applications/noop/...    # dummy application that does nothing

These scripts were only ever checked in to get us off the ground, by providing an example to start from. We don't need them anymore.


qingling128 (approved):
LGTM

PR #376, opened 2022-01-31
Fix go tests to support rocky-linux

martijnvans:


igorpeshansky (approved):
This subsumes one of the commits on #377. I'll rebase that one once this is in.

LGTM :shipit:

PR #378, opened 2022-02-02
Updates for logging processor requirement of time_format when time_key is present.

franciscovalentecastro:
- Updated TestProcessorOrder in integration_test/ops_agent_test.go, since the config used is missing time_format.
- Updated logging_processors.go so time_format is required when time_key is present and the other way around.
- Added input.yaml and generated golden_error files in the /testdata/invalid/ folder.
- Updated /valid/logging-processor_parse_json_and_parse_regex_types/input.yaml to remove the cases where time_format or time_key are missing.

b/212428534 - This PR has the same updates as #333 (which was reverted) and adding required updates to the integration tests.

### DIFF
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -357,7 +357,7 @@ func TestProcessorOrder(t *testing.T) {
         receivers: [mylog_source]
         processors: [json1, json2]
         exporters: [google]
-`, logPath, "%Y-%m-%dT%H:%M:%S.%L%Z")
+`, logPath, "%Y-%m-%dT%H:%M:%S.%L%z")
 
 		if err := setupOpsAgent(ctx, logger, vm, config); err != nil {
 			t.Fatal(err)
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -347,6 +347,7 @@ func TestProcessorOrder(t *testing.T) {
       type: parse_json
       field: message
       time_key: time
+      time_format: "%s"
     json2:
       type: parse_json
       field: log
@@ -356,7 +357,7 @@ func TestProcessorOrder(t *testing.T) {
         receivers: [mylog_source]
         processors: [json1, json2]
         exporters: [google]
-`, logPath)
+`, logPath, "%Y-%m-%dT%H:%M:%S.%L%Z")
 
 		if err := setupOpsAgent(ctx, logger, vm, config); err != nil {
 			t.Fatal(err)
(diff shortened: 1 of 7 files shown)


integration_test/ops_agent_test.go
```diff
@@ -356,13 +357,13 @@ func TestProcessorOrder(t *testing.T) {
...
         exporters: [google]
-`, logPath)
+`, logPath, "%Y-%m-%dT%H:%M:%S.%L%z")
 
 		if err := setupOpsAgent(ctx, logger, vm, config); err != nil {
 			t.Fatal(err)
 		}
 
-		line := fmt.Sprintf(`{"log":"{\"level\":\"info\",\"message\":\"start\"}\n","time":"%s"}`, time.Now().Format(time.RFC3339Nano)) + "\n"
+		line := fmt.Sprintf(`{"log":"{\"level\":\"info\",\"message\":\"start\"}\n","time":"%s"}`, time.Now().UTC().Format(time.RFC3339Nano)) + "\n"
```
franciscovalentecastro, line 366: When not using UTC time stamps (e.g. 2022-02-10T08:07:29.018896832-08:00) the parsing fails in windows (b/218888265). This is one way to avoid the error until the bug is fixed, another way could be skipping the test for windows.

martijnvans: I thinkn this deserves a comment in the code, otherwise the next person is going to have no idea why this is like that, and especially why it is different from the other test that does something similar without UTC and with Z.

qingling128:
I'll defer this one to @sophieyfang and @martijnvans

integration_test/ops_agent_test.go
franciscovalentecastro: I found out that [TestCustomLogFormat](https://github.com/GoogleCloudPlatform/ops-agent/blob/bb8a03e38c2c71d2c5261a8e148ed6b4380c4e05/integration_test/ops_agent_test.go#L279) also throws the error

```
[error] [parser] cannot parse '2022-02-09T17:59:21.323823326Z' after %L
[ warn] [parser:my_pipeline.mylog_source.0] invalid time format %Y-%m-%dT%H:%M:%S.%L%Z for '2022-02-09T17:59:21.323823326Z'
```

The difference is that `parse_regex` sends the log with a "current time" timestamp (this is expected [out stackdriver plugin](https://github.com/fluent/fluent-bit/blob/1fa0e94a09e4155f8a6d8a0efe36a5668cdc074e/plugins/out_stackdriver/stackdriver.c#L2120)) and `parse_json` doesn't in this case. I've tested this (b/219090004) and seems to happen only with "%Z".

I could update both tests in another PR to use "%z", but it may need also to use UTC.

franciscovalentecastro: Update : Added comment in code explaining difference with test TestCustomLogFormat.

PR #383, opened 2022-02-03
Retry registercloudguest failures in third_party_apps_test

martijnvans:
prepareSLES() failures indicate something is wrong with the SLES servers or the VM itself. We should try harder to paper over these issues by retrying them, with a fresh VM if need be. Retrying 5 times might help if the SLES servers are unresponsive, and changing the error type from nonRetryable -> retryable makes the test try again with a fresh VM if that fails.


qingling128 (approved):
LGTM as the changes are SLES specific and those tests have passed.

PR #384, opened 2022-02-04
Unify receiver and pipeline IDs in 3rd party app sample configs.

qingling128:
# Background

It's preferred for the receiver and pipeline IDs to stay the same as the receiver type in the sample configuration, enable_<APP> files and the main `confgenerator/testdata/valid/{LINUX_OR_WINDOWS}/{logging_or_metrics}-receiver_{APP}/input.yaml` test case, because some users look at the `valid` cases for sample configurations as well. Additional more complex test cases could use more complex receiver or pipeline ID. 

The reason behind that is that Log names are derived from logging receiver IDs, and the dropdown in UI + sample query / dashboard relies on the log name. So the sample dashboards and queries can use that. If users customize it, they can then adjust dashboards and queries accordingly. For consistency, the same convention applies to the sample configurations on the metrics side.

# Included changes

* `s/{APP}_metrics/{APP}`, `s/{APP}_pipeline/{APP}`, `s/{APP}_default/{APP}` in the receiver / pipeline IDs in the following files:
  * Sample configuration in markdowns
  * `enable_<APP>` files
  * The main `confgenerator/testdata/valid/{LINUX_OR_WINDOWS}/{logging_or_metrics}-receiver_{APP}/input.yaml` test case
* In the `confgenerator/testdata/valid/{LINUX_OR_WINDOWS}` folder, make sure that the `{logging_or_metrics}-receiver_{APP}/input.yaml` test case is the minimum configuration of the application. Split the test case into a separate `{logging_or_metrics}-receiver_{APP}_custom/input.yaml` when necessary to test customized configuration
* Make sure the `enable_<APP>` files are testing the minimum configuration as well. e.g. Remove unnecessary `collection_interval` and `endpoint` etc. optional config parameter settings.

# To review locally

The golden file changes are not important for purpose of this PR, thus can be excluded to reduce the noise.
```
git clone --recurse-submodules git@github.com:GoogleCloudPlatform/ops-agent.git
cd ops-agent
git checkout lingshi-id
git diff origin/master..HEAD -- . ':(exclude)*golden*'
```


confgenerator/testdata/valid/linux/metrics-receiver_apache_custom/input.yaml
```diff
@@ -14,11 +14,11 @@
 
 metrics:
   receivers:
-    apache_metrics:
+    apache:
       type: apache
-      collection_interval: 30s
+      server_status_url: http://localhost/server-status?auto
```
ridwanmsharif, line 19: Can we add the `collection_interval` back for this custom test case? With the lowest possible value? I was looking at https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/apache#apache-metrics and it doesn't specify what the lowest value is, is it 30s?

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_apache
```diff
@@ -43,4 +42,4 @@ logging:
 EOF
 
 sudo service google-cloud-ops-agent restart
-sleep 30
+sleep 60
```
ridwanmsharif, line 45: Why did the sleep need to be increased? 60s looks good, just wondering if there was a known bug or something

ridwanmsharif (approved):
LGTM

integration_test/third_party_apps_data/agent/ops-agent/linux/enable_apache
qingling128: Oh, because I removed the customized `collection_interval 30s` from all the configurations that are used by the integration tests. So they'll default to `60s`, which is closer to what most users would do.

In theory our tests should know to wait and retry as well. But waiting for 1 min here does not hurt.

confgenerator/testdata/valid/linux/metrics-receiver_apache_custom/input.yaml
qingling128: Sure thing. Will add it back and push a commit once the integration tests finish (Trying to avoid triggering integration tests again. I'll have the unit tests run)

qingling128: Hmm... I just realized the previous integration tests were not triggered properly. So I just made the change to `10s` (the minimum) and re-triggered them.

qingling128:
Tests look okay:

image

Debian seems to be failing for an unrelated reason: `tomcat install on debian 10 fails`. It is being addressed separately by https://github.com/GoogleCloudPlatform/ops-agent/pull/344. 

Gonna rebase then merge.

PR #394, opened 2022-02-09
Add Couchdb logging receiver

JonathanWamsley:
Added logging receiver for couchdb.

A single receiver that supports 2 log formats as access logs or error logs:
Logs are received from a single source /var/log/couchdb/couchdb.log

Sample log example:
```yaml
{
  "insertId": "152cslufvw01o4",
  "jsonPayload": {
    "message": "admin GET /forbidden 404 ok 1",
    "status_message": "ok",
    "remote_user": "admin",
    "pid": "0.25387.37",
    "path": "/forbidden",
    "host": "127.0.0.1",
    "node": "couchdb"
  },
  "httpRequest": {
    "requestMethod": "GET",
    "status": 404,
    "responseSize": "1",
    "remoteIp": "127.0.0.1",
    "serverIp": "localhost:5984"
  },
  "resource": {
    "type": "gce_instance",
    "labels": {
      "instance_id": "1314100884383110088",
      "project_id": "otel-agent-dev",
      "zone": "us-central1-a"
    }
  },
  "timestamp": "2022-02-10T20:34:32.105337Z",
  "severity": "NOTICE",
  "logName": "projects/otel-agent-dev/logs/couchdb",
  "receiveTimestamp": "2022-02-10T20:34:34.937004216Z"
}
```

PR #379, opened 2022-02-02
Hadoop logging receiver

BinaryFissionGames:
Implements the logging receiver for Hadoop. 

Based off of #351 - will rebase onto master after hadoop metrics are merged.

PR #396, opened 2022-02-10
Only run tests relevant to current PR.

stati0n:
Creates a mode of testing for `go test -short` where we run a smaller slice of tests.

This is intended to be used with presubmits.  Postsubmits should omit the `-short` flag and run the full suite.

Determines what apps are impacted through extracting app name from the file path of modified files
and checking against the known list of apps available for test.


integration_test/third_party_apps_test.go
```diff
@@ -363,6 +364,16 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
 
+func modifiedFiles(t *testing.T) []string {
+	cmd := exec.Command("git", "diff", "--name-only", "upstream/master")
+	out, err := cmd.Output()
+	if err != nil {
+		t.Fatalf("got error calling `git diff`: %v", err)
+	}
+
+	return strings.Split(string(out), "\n")
+}
```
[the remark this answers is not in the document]
stati0n: Whoops.  I had it as a separate file but eventually decided that there wasn't enough in the separate file to warrant it.  I will remove modified.go.

integration_test/third_party_apps_test.go
```diff
@@ -377,57 +388,107 @@ func TestThirdPartyApps(t *testing.T) {
...
-	for _, platform := range testConfig.Platforms {
-		platform := platform // https://golang.org/doc/faq#closures_and_goroutines
-		t.Run(platform, func(t *testing.T) {
-			t.Parallel()
+	type test struct {
+		platform   string
+		app        string
+		skipReason string
+	}
+	tests := []test{}
```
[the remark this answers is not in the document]
stati0n: SGTM.

integration_test/third_party_apps_test.go
```diff
@@ -377,57 +388,107 @@ func TestThirdPartyApps(t *testing.T) {
...
+
+		} else if strings.HasPrefix(f, "integration_test") {
+			// Any other modified file in the directory probably
+			// means we should test everything.
+			impactedApps = make(map[string]bool)
+			break
+		}
+
+	}
+	for i := range tests {
```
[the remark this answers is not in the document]
stati0n: It did make a copy of each element and didn't update the original.  That's why I did it like this.  I had debated if I could do it with pointers, but it didn't seem worth it.

integration_test/third_party_apps_test.go
```diff
@@ -377,57 +388,107 @@ func TestThirdPartyApps(t *testing.T) {
...
+		} else if strings.HasPrefix(f, "integration_test/third_party_apps_data/applications/") {
+			// Folder names: integration_test/third_party_apps_data/applications/<app_name>
+			f := strings.TrimPrefix(f, "integration_test/third_party_apps_data/applications/")
+			f = strings.Split(f, "/")[0]
+			impactedApps[f] = true
+
+		} else if strings.HasPrefix(f, "integration_test") {
+			// Any other modified file in the directory probably
+			// means we should test everything.
+			impactedApps = make(map[string]bool)
```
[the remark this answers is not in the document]
stati0n: SGTM.  I thought this was cleaner.

integration_test/third_party_apps_test.go
```diff
@@ -377,57 +388,107 @@ func TestThirdPartyApps(t *testing.T) {
...
+			tests[i].skipReason = "Skipping test due to 'platforms_to_skip' entry in test_config.yaml"
+		}
+	}
 
-					if sliceContains(testConfig.PerApplicationOverrides[app].PlatformsToSkip, platform) {
-						t.Skip("Skipping test due to 'platforms_to_skip' entry in test_config.yaml")
+	// Execute tests
+	for _, tc := range tests {
+		platform := tc.platform // https://golang.org/doc/faq#closures_and_goroutines
+		app := tc.app           // https://golang.org/doc/faq#closures_and_goroutines
```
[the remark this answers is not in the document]
stati0n: Done.

integration_test/third_party_apps_test.go
```diff
@@ -377,57 +388,107 @@ func TestThirdPartyApps(t *testing.T) {
...
+		t.Run(platform, func(t *testing.T) {
```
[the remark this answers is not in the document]
stati0n: Done.

integration_test/third_party_apps_test.go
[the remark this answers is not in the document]
martijnvans: it's not about it being clean, there's a correctness issue. resetting it to empty doesn't do the same thing as forcing it to test everything. What if you decide to test everything and thereby set it to empty, then a later modified file makes you decide that you should test apache? Then the map will contain just apache, and you'll end up testing just apache instead of everything.

integration_test/third_party_apps_test.go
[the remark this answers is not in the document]
igorpeshansky: Yeah, see https://go.dev/ref/spec#RangeClause and https://stackoverflow.com/a/15952415/6752892…

integration_test/third_party_apps_test.go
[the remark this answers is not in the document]
stati0n: Resolved.

integration_test/third_party_apps_test.go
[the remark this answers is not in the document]
stati0n: Needs work.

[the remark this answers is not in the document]
stati0n: Done.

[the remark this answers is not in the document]
stati0n: Done.

integration_test/third_party_apps_test.go
```diff
@@ -363,6 +364,75 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+}
+
+func determineImpactedApps(mf []string) map[string]bool {
+	impactedApps := make(map[string]bool)
+	for _, f := range mf {
+		if strings.HasPrefix(f, "apps/") {
+			// File names: apps/<appname>.go
+			f := strings.TrimPrefix(f, "apps/")
+			f = strings.TrimSuffix(f, ".go")
+			impactedApps[f] = true
```
martijnvans, line 384: how does this deal with modified files like common_logging_processors.go? https://github.com/GoogleCloudPlatform/ops-agent/blob/master/apps/common_logging_processors.go

I think you need to check that the app is valid by checking it against integration_test/third_party_apps_data/applications/<app_name>. Unless that is being done later somewhere, in which case I missed it. Otherwise won't it try to run an app called common_logging_processors?

integration_test/third_party_apps_test.go
```diff
@@ -363,6 +364,75 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+
+func determineTestsToSkip(tests []test, impactedApps map[string]bool, testConfig testConfig) {
+	_, testAll := impactedApps["all"]
+	for i, test := range tests {
+		if testing.Short() {
+			if len(impactedApps) > 0 {
+				_, testApp := impactedApps[test.app]
+				if testAll == false && testApp == false {
+					tests[i].skipReason = fmt.Sprintf("skipping %v because it's not impacted by pending change", test.app)
+				} else if testAll == true {
```
martijnvans, line 419: This differs from what's in the doc. which might be ok. but the difference is, in the doc, certain changes would really trigger all applications on all platforms. For example, changing gce_testing.go would trigger everything. What you have here lacks that option, instead it will test all applications only on the default platform. Not sure if you intended this difference. I'm OK with it though, I can always omit -test.short and run the tests manually in that case. Just wanted to make sure you knew about this difference.

integration_test/third_party_apps_test.go
```diff
@@ -377,25 +447,53 @@ func TestThirdPartyApps(t *testing.T) {
...
-			if len(apps) == 0 {
-				t.Fatalf("Found no applications when testing agentType=%v, platform=%v", agentType, platform)
-			}
-			for _, app := range apps {
-				app := app // https://golang.org/doc/faq#closures_and_goroutines
-				t.Run(app, func(t *testing.T) {
-					t.Parallel()
+	// Filter tests
+	mf := modifiedFiles(t)
+	ia := determineImpactedApps(mf)
```
martijnvans, line 467: personally I would inline these variables that aren't used for anything later:

	determineTestsToSkip(tests, determineImpactedApps(modifiedFiles(t)), testConfig)

This is spelled out for C++ land in tip of the week 161

but this is an optional suggestion.

integration_test/third_party_apps_test.go
```diff
@@ -363,6 +364,75 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+
+type test struct {
+	platform   string
+	app        string
+	skipReason string
+}
+
+const defaultPlatform = "debian-10"
+
+func determineTestsToSkip(tests []test, impactedApps map[string]bool, testConfig testConfig) {
```
martijnvans, line 411: I feel like this needs a comment. Perhaps a link to the doc (which is already shared with observiq folks) and the doc made an effort to be publicly viewable.

integration_test/third_party_apps_test.go
stati0n: This is a gap.  Let me figure out what to do here.

integration_test/third_party_apps_test.go
stati0n: Comment forthcoming.

integration_test/third_party_apps_test.go
stati0n: This was an oversight.  Let me think about what to do about it.

integration_test/third_party_apps_test.go
stati0n: Done.

integration_test/third_party_apps_test.go
stati0n: Done.

integration_test/third_party_apps_test.go
stati0n: I think this is the right thing to do.  I think we'll cover the full test matrix in postsubmit.

integration_test/third_party_apps_test.go
```diff
@@ -327,6 +328,110 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+			f := strings.TrimPrefix(f, "integration_test/third_party_apps_data/applications/")
+			f = strings.Split(f, "/")[0]
+			// The directories here are already authoritative, no
+			// need to check against list.
+			impactedApps[f] = true
+
+		} else {
+			// Any other modified file means we should test
+			// everything.
+			impactedApps["all"] = true
```
martijnvans, line 392: I think this will result in "test everything" to occur all the time, since PRs that modify e.g. couchdb.go will also tend to modify various files outside of that directory. Examples:

https://github.com/GoogleCloudPlatform/ops-agent/pull/394/files
https://github.com/GoogleCloudPlatform/ops-agent/pull/379/files

I guess we might be able to ignore changes:
- under docs/
- under confgenerator/testdata/
- stuff like integration_test/third_party_apps_data/agent/ops-agent/linux/enable_hadoop, or better yet, rename this kind of file to integration_test/third_party_apps_data/applications/hadoop/enable

(hmm, not sure what to do about changes to integration_test/third_party_apps_data/agent/ops-agent/linux/supported_applications.txt, i guess ignore them and rely on the other heuristics instead)

It's tricky to set up the exclusions exactly right to avoid triggering "all" for PRs that touch an app... hmm

stati0n: Per our offline discussion, the latest revision always runs all apps for the default platform, and also runs all platforms for any app deemed to be "impacted".

integration_test/third_party_apps_test.go
```diff
@@ -327,6 +328,97 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+	app        string
+	skipReason string
+}
+
+var defaultPlatforms = map[string]bool{
+	"debian-10":    true,
+	"windows-2019": true,
+}
+
+// Mark some tests for skipping, based on test_config and impacted apps.
```
martijnvans, line 400: needs some mention of the -test.short flag as well.

integration_test/third_party_apps_test.go
```diff
@@ -327,6 +328,97 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+// Always test all apps against the default platform.  If a subset of apps is
+// determined to be impacted, also test all platforms for those apps.
+func determineTestsToSkip(tests []test, impactedApps map[string]bool, testConfig testConfig) {
+	for i, test := range tests {
+		if testing.Short() {
+			_, testApp := impactedApps[test.app]
+			_, defaultPlatform := defaultPlatforms[test.platform]
+			if defaultPlatform == false && testApp == false {
```
martijnvans, line 408: Why not:

if !defaultPlatform && !testApp {

I think we usually prefer ! to == false, similar to preferring <nothing> to == true.

integration_test/third_party_apps_test.go
```diff
@@ -341,57 +433,83 @@ func TestThirdPartyApps(t *testing.T) {
...
+		t.Run(tc.platform, func(t *testing.T) {
+			t.Parallel()
+			t.Run(tc.app, func(t *testing.T) {
+				t.Parallel()
+
+				if tc.skipReason != "" {
+					t.Skip(tc.skipReason)
+				}
+
+				// ctx, cancel := context.WithTimeout(context.Background(), gce.SuggestedTimeout)
```
martijnvans, line 466: stuff is commented out that shouldn't be.

martijnvans (changes requested):

integration_test/third_party_apps_test.go
stati0n: This is not the right place.  I will add a comment elsewhere.

integration_test/third_party_apps_test.go
```diff
@@ -327,6 +328,98 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+		if file.IsDir() {
+			allApps[file.Name()] = true
+		}
+	}
+	return allApps
+}
+
+func modifiedFiles(t *testing.T) []string {
+	cmd := exec.Command("git", "diff", "--name-only", "origin/master")
+	out, err := cmd.Output()
```
martijnvans, line 351: how do you feel about logging cmd.Output()? I think it would be helpful and not too noisy:
log.Printf("git diff output:\nstdout:%v\nstderr:%v\n", out, err)

particularly for the case where git diff returns an error, this innfo would be great. but I think it would be useful in all cases, if only as a clue that it is runnning git diff and doing stuff with the output in the first place.

integration_test/third_party_apps_test.go
```diff
@@ -327,6 +328,98 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+		} else if strings.HasPrefix(f, "integration_test/third_party_apps_data/applications/") {
+			// Folder names: integration_test/third_party_apps_data/applications/<app_name>
+			f := strings.TrimPrefix(f, "integration_test/third_party_apps_data/applications/")
+			f = strings.Split(f, "/")[0]
+			// The directories here are already authoritative, no
+			// need to check against list.
+			impactedApps[f] = true
+
+		}
+	}
```
martijnvans, line 385: how about logging here what we detected as impacted? Simply log.Printf("impacted apps: %v", impactedApps)

martijnvans (approved):

martijnvans (approved):

integration_test/third_party_apps_test.go
```diff
@@ -327,6 +328,101 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+	out, err := cmd.Output()
+	log.Printf("git diff output:\n\tstdout:%v\n\tstderr:%v\n", out, err)
+	if err != nil {
+		t.Fatalf("got error calling `git diff`: %v", err)
+	}
+
+	return strings.Split(string(out), "\n")
+}
+
+// Determine what apps are impacted by current code changes.
```
sophieyfang, line 361: For cases when a broken change is introduced in the code that are not in either these two categories, but the code change touches the flow of those third party integration. It will skip tests for such cases..?
I think the code for 3rd party isn’t secluded only to those two folders, it uses many from other places too.

sophieyfang: By `Postsubmits should omit the -short flag and run the full suite.` @stati0n do you mean the test being kicked before merging into master?

stati0n: Postsubmits means the build for release.  In this change, we are consciously choosing lower coverage in exchange for higher reliability and faster runtime.

PR #398, opened 2022-02-10
Shrink linux instance size: use 2 CPUs per VM instead of 4

martijnvans:
This is to help with CPU quota issues.

We used to run with 2 CPUs for all instances until we boosted them all to 4 about a year ago to improve Windows startup times.


igorpeshansky (approved):
LGTM :shipit:

martijnvans:
The remaining failures are flakes, particularly https://github.com/GoogleCloudPlatform/ops-agent/issues/407. Between the last few runs each test has turned up green. Merging.

PR #399, opened 2022-02-10
Change all 3rd party log receivers to use multiline input over post-input parser

dehaansa:
Noted in comment here: https://github.com/GoogleCloudPlatform/ops-agent/pull/219#issuecomment-1015939175
This pattern has been implemented already in elasticsearch & zookeeper.


quentinmit:
Do we know how this behaves when paired with non-file receivers?

i.e. if the user pairs these processors with syslog or journald, will we still parse the logs correctly?

dehaansa:
> Do we know how this behaves when paired with non-file receivers?
> 
> i.e. if the user pairs these processors with syslog or journald, will we still parse the logs correctly?

It's a good question. My understanding is that syslog is capable of transmitting multiline messages, but I'd have to try to set up some tests to find out if the applications we're supporting send multiline messages across syslog together. There does not appear to be a way to configure the syslog receiver in fuent bit to specify multiline as a possibility.

qingling128:
@sophieyfang had done some experiment for pairing with non-file receivers. TL;DR (Sophie can correct me) is that if a multi-line log entry happens to be split across to chunks, they would end up being parsed incorrectly. Each chunk is around 2 MiB.

The support to specify multiline directly in the input plugin is in the Fluent Bit roadmap.

dehaansa:
I've confirmed that multiline syslog messages come through correctly as single messages to the parser. If a syslog forwarder in the middle splits lines or something similar there could be a problem, but those edge cases seem like something solvable in the future when Fluent Bit is updated as Ling stated is on the roadmap.

quentinmit:
> I've confirmed that multiline syslog messages come through correctly as single messages to the parser. If a syslog forwarder in the middle splits lines or something similar there could be a problem, but those edge cases seem like something solvable in the future when Fluent Bit is updated as Ling stated is on the roadmap.

Hm, what was the test setup you used to generate/consume multiline syslog messages? Did you manually use a client library to generate messages, configure an app to write directly to syslog (which one?), or configure an rsyslogd to forward to fluentbit?

My understanding is that the only difference this PR buys us is that fluentbit otherwise doesn't properly parse multiline log entries that span across multiple log chunks, right? That seems like the root cause bug, which AIUI is not currently on Calyptio's roadmap to fix.

Instead of this change, which might break forwarding setups, I'm interested in opportunistically moving parsers into input plugins when the parser is immediately after the input plugin, which would preserve all of the current fluentbit behavior but fix the bug when possible.

PR #401, opened 2022-02-10
Fix postgresql sles integration test issues

dehaansa:
Seems that the package installer has changed in zypper and the service name changed & the requirements to initialize the server have also changed.


qingling128 (approved):
LGTM

qingling128:
Adding @jefferbrecht @ridwanmsharif as well who could help merge this once the integration tests pass in case I get stuck in meetings all day (@dehaansa is working on some fix, and it also needs rebasing). Right now it looks like SLES 15 postgresql is still failing:

image

qingling128:
Postgresql passed on SLES 12 and 15:
image

So it's OK to merge.

PR #409, opened 2022-02-14
Solr System Log fixes

StefanKurek:
Modifies Solr System Log regex to fix Severity not always being properly parsed (which also fixes multi line issues too)

Also translates severity properly (Solr documentation was incorrect on the various log severity levels)

### DIFF
--- a/apps/solr.go
+++ b/apps/solr.go
@@ -68,7 +68,7 @@ func (p LoggingProcessorSolrSystem) Components(tag string, uid string) []fluentb
 			Parsers: []confgenerator.RegexParser{
 				{
 					// Sample line: 2022-01-06 04:16:08.794 INFO  (qtp1489933928-64) [   x:gettingstarted] o.a.s.c.S.Request [gettingstarted]  webapp=/solr path=/get params={q=*:*&_=1641440398872} status=0 QTime=2
-					Regex: `^(?<timestamp>\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}:\d{2}\.\d{3,6})\s(?<level>[A-z]+)\s{2}\((?<thread>[^\)]+)\)\s\[c?:?(?<collection>[^\s]*)\ss?:?(?<shard>[^\s]*)\sr?:?(?<replica>[^\s]*)\sx?:?(?<core>[^\]]*)\]\s(?<source>[^\s]+)\s(?<message>(?:(?!\s\=\>)[\s\S])+)\s?=?>?(?<exception>[\s\S]*)`,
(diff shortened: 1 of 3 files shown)


qingling128 (commented):
Is there a test case that we can add to prevent potential regressions in the future?

StefanKurek:
> Is there a test case that we can add to prevent potential regressions in the future?

@qingling128 are you imagining on the integration level or more of a unit test that would check a number of logs specific to a technology are being parsed properly? In theory this could apply to many of the integrations if a regex was written without a large enough sample of logs.

The question of adding these types of app specific log tests might be larger than this PR (because I don't think we have done this before yet)? I'll defer to whatever you think though.

qingling128:
@StefanKurek - Yeah, I meant the https://github.com/GoogleCloudPlatform/ops-agent/blob/master/integration_test/third_party_apps_data/applications/solr/expected_logs.yaml test. Looks like it can take an array. I wonder if we could somehow trigger a multiline log via https://github.com/GoogleCloudPlatform/ops-agent/blob/master/integration_test/third_party_apps_data/applications/solr/debian_ubuntu/post and verify it in the YAML.

It's OK to do a follow up PR though.

qingling128 (approved):

StefanKurek:
> @StefanKurek - Yeah, I meant the https://github.com/GoogleCloudPlatform/ops-agent/blob/master/integration_test/third_party_apps_data/applications/solr/expected_logs.yaml test. Looks like it can take an array. I wonder if we could somehow trigger a multiline log via https://github.com/GoogleCloudPlatform/ops-agent/blob/master/integration_test/third_party_apps_data/applications/solr/debian_ubuntu/post and verify it in the YAML.
> 
> It's OK to do a follow up PR though.

@qingling128 ah. Part of the problem is that I can only generate these types of logs (related to this PR fix) by causing an issue with the Solr setup. This would generate both ERROR severity and multiline logs...but then we would have a broken Solr setup after ./post 😄

qingling128:
> @qingling128 ah. Part of the problem is that I can only generate these types of logs (related to this PR fix) by causing an issue with the Solr setup. This would generate both ERROR severity and multiline logs...but then we would have a broken Solr setup after ./post 😄

Is it possible to trigger an issue then revert it (aka the log might still go through)? If the effort is very high, let's punt it for now though.

Either way it's not a blocker for this PR. So I'm gonna merge it, assuming it has been tested manually.

qingling128:
Confirmed offline that this has been tested manually. @StefanKurek Can you help attach a sample log to the PR description when you get a sec?

PR #412, opened 2022-02-15
Document how to run tests outside of Kokoro

martijnvans:
This flow allows for much faster iteration when iterating on an integration test.

I removed an obsolete section talking about test_config.yaml as well, since that feature predates the whole test source code being available on GitHub.


integration_test/README.md
```diff
@@ -1,12 +1,89 @@
...
+ -tags=integration_test \
+ -timeout=4h
+```
+
+This will run the tests against the stable Ops Agent. To test against a pre-built
+but unreleased agent, you can use add the AGENT_PACKAGES_IN_GCS environment
+variable onto your command like this:
+
+```
+AGENT_PACKAGES_IN_GCS=gs://ops-agents-public-buckets-test-logs/prod/stackdriver_agents/testing/consumer/ops_agent/presubmit_github/debian/166/20220215-095636/agent_packages \
```
ridwanmsharif, line 79: [nit] I don't know where would be the best place to put this but it would be really useful to document instructions on how to find the prebuilt unreleased agent from presubmit runs.

ridwanmsharif (approved):
LGTM

integration_test/README.md
martijnvans: Done.

PR #414, opened 2022-02-16
Add resource_name to log entries using `modify` + `nest` filter

franciscovalentecastro:
b/169784211

### DIFF
--- a/confgenerator/logging.go
+++ b/confgenerator/logging.go
@@ -36,7 +36,9 @@ func setLogNameComponents(tag, logName string) []fluentbit.Component {
 
 // setInstanceNameComponent generates a component that inserts the hostname to all log entries.
 func setInstanceNameComponent(Hostname string) []fluentbit.Component {
-	if len(Hostname) == 0 {Hostname = ""}
+	if len(Hostname) == 0 {
+		Hostname = ""
+	}
 	return []fluentbit.Component{
 		{
 			Kind: "FILTER",
--- a/confgenerator/logging.go
+++ b/confgenerator/logging.go
@@ -34,6 +34,31 @@ func setLogNameComponents(tag, logName string) []fluentbit.Component {
 	}
 }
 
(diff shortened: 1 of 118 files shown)


confgenerator/logging.go
```diff
@@ -34,6 +34,33 @@ func setLogNameComponents(tag, logName string) []fluentbit.Component {
...
+			Kind: "FILTER",
+			Config: map[string]string{
+				"Match": "*",
+				"Add":   fmt.Sprintf("compute.googleapis.com/resource_name %s", Hostname),
+				"Name":  "modify",
+			},
+		},
+		{
+			Kind: "FILTER",
+			Config: map[string]string{
```
ridwanmsharif, line 53: We moved away from this approach but I realized something interesting:

Even if the match+nest performance wasn't too bad I don't think this approach would work. I don't think the Fluent Bit `nest` plugin supports using nest on a key that might already exist. Would it override the existing values? Or append to it if it exists? (I suspect the former).

I say this because users using `parse_json` could already have some labels nested under `logging.googleapis.com/labels` so this change would need to work well with that too.

Doesn't matter anymore since we're going in another direction

franciscovalentecastro: You're right. After a little searching in fluent-bit, what you say `Would it override the existing values?` it's probably true due to [this issue](https://github.com/fluent/fluent-bit/issues/3635).

franciscovalentecastro: Just checked by doing two consecutive `modify` + `nest` filters and it doesn't append or override the past labels, it just creates a new identical object with the new labels underneath : 

```
{ ... "logging.googleapis.com/labels"=>{"labela"=>"valuea"}, "logging.googleapis.com/labels"=>{"labelb"=>"valueb"}}
```
So as you say, it wouldn't work well for this application. The out_stackdriver plugin only gets the first labels object it finds.

PR #415, opened 2022-02-16
Move PrepareSLES() into the core library

martijnvans:
All tests, not just third_party_apps_test, should benefit from the full "retry dance" that we need to do to paper over SLES VM initialization issues.


qingling128 (approved):
LGTM

PR #420, opened 2022-02-18
Refactors the shared structs/methods which configure and return a pipeline for JMX Receivers

StefanKurek:
This refactor is necessary for the WildFly Receiver update in order to allow for a new defaultAdditionalJars config option. With this pattern we can easily add more defaults in the future without having to refactor the pipeline builder method each time.

Adds MetricsReceiverShared struct to MetricsReceiverSharedJVM struct since it will always need it.
This also allows the ConfigurePipelines method to remove the collection interval argument.
Then added a new method "WithDefaultEndpoint" for MetricsReceiverSharedJVM and removed the defaultEndpoint
argument from the ConfigurePipelines method. This allows for a more extensible pattern where
new defaults do not have to chnage the ConfigurePipeline's method signature.
Lastly made changes to the various application go files which currently make use of these methods/structs.

### DIFF
--- a/apps/activemq.go
+++ b/apps/activemq.go
@@ -22,8 +22,6 @@ import (
 type MetricsReceiverActivemq struct {
 	confgenerator.ConfigComponent `yaml:",inline"`
 
-	confgenerator.MetricsReceiverShared `yaml:",inline"`
-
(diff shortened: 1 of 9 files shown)


qingling128 (approved):
LGTM

StefanKurek:
@qingling128 rebased 👍

PR #431, opened 2022-03-01
Delete some now-useless features from test_config.yaml

martijnvans:
Two features are being deleted:
1. the number of retries. This is silly now because it's much more straightforward to just edit the retry loop by hand. Previously the retry loop itself was not open source so i introduced a config option for it. But I don't think we need it anymore.
2. platforms_override. This feature predates the change where each distro is split out into a separate Kokoro build. It doesn't really work anymore. Each kokoro build will build for distro X and then forcibly telling it to test on distros WXYZ won't work because W,Y, and Z can't run the package built for X. Also, if someone really needs to run a test on platform W, they can modify the go code (just like for retries) instead of modifying test_config.yaml.

This PR is a step towards getting rid of test_config.yaml completely, (https://github.com/GoogleCloudPlatform/ops-agent/pull/430), but I figured I'd start with deleting its useless features.


qingling128 (approved):
LGTM

PR #432, opened 2022-03-01
Remove duplicate entries from supported_applications.txt

martijnvans:
elasticsearch and hbase were in there twice, which results in them both being tested twice. :)

I also reordered a few entries to make the list fully alphabetical instead of mostly alphabetical.

### DIFF
--- a/integration_test/third_party_apps_data/agent/ops-agent/linux/supported_applications.txt
+++ b/integration_test/third_party_apps_data/agent/ops-agent/linux/supported_applications.txt
@@ -1,11 +1,9 @@
 activemq
 apache
 cassandra
-hadoop
-elasticsearch
 couchdb
-hbase
 elasticsearch
+hadoop
 hbase
 jvm
 kafka


integration_test/third_party_apps_data/agent/ops-agent/linux/supported_applications.txt
```diff
@@ -1,11 +1,9 @@
 activemq
 apache
 cassandra
-hadoop
```
igorpeshansky, line 4: Should we add validation to enforce alphabetical order in this file, to prevent this from happening in the future?

igorpeshansky (approved):
LGTM :shipit:

integration_test/third_party_apps_data/agent/ops-agent/linux/supported_applications.txt
martijnvans: I don't see much need for alphabetical order enforcement. Duplicate entry enforcement on the other hand... it's a good idea, I'll consider it for a follow up PR.

martijnvans: I split out the validation into https://github.com/GoogleCloudPlatform/ops-agent/pull/438 so that it can be reviewed separately.

martijnvans:
The failures seem to be build failures, unrelated to this PR. I think enough tests passed that we can be confident in this PR.

PR #438, opened 2022-03-01
Reject redundant entries in supported_applications.txt

martijnvans:
Raise a test failure if the same application appears more than once in supported_applications.txt. This is to prevent the situation that https://github.com/GoogleCloudPlatform/ops-agent/pull/432 is fixing from reoccurring.


integration_test/third_party_apps_test.go
```diff
@@ -73,6 +73,19 @@ func osFolder(platform string) string {
 	return "linux"
 }
 
+// rejectDuplicates looks for duplicate entries in the input slice and returns
+// an error if any is found.
+func rejectDuplicates(apps []string) error {
+	unique := make(map[string]bool)
+	for _, app := range apps {
+		if unique[app] {
+			return fmt.Errorf("application %q appears multiple times in supported_applications.txt", app)
```
igorpeshansky, line 82: [Optional] Would it make sense to list all duplicates?

integration_test/third_party_apps_test.go
```diff
@@ -73,6 +73,19 @@ func osFolder(platform string) string {
 	return "linux"
 }
 
+// rejectDuplicates looks for duplicate entries in the input slice and returns
+// an error if any is found.
+func rejectDuplicates(apps []string) error {
+	unique := make(map[string]bool)
```
igorpeshansky, line 79: The name `unique` implies that it would be true for items that are unique, and false for duplicates. Would `seen` be a better name?

igorpeshansky (changes requested):

integration_test/third_party_apps_test.go
igorpeshansky: I guess by resolving this you were saying that reporting one duplicate at a time is fine. Well, it was optional anyway. For the record, I was suggesting something like:
```
func rejectDuplicates(apps []string) error {
	seen := make(map[string]bool)
	var dups []string
	for _, app := range apps {
		if seen[app] {
			dups = append(dups, app)
		}
		seen[app] = true
	}
	if len(dups) > 0 {
		return fmt.Errorf("applications %q appear multiple times in supported_applications.txt", dups)
	}
	return nil
}
```

igorpeshansky (approved):
LGTM :shipit:

martijnvans:
yes, I both think that reporting one duplicate at a time is fine and that the code for reporting all duplicates is more complex and not worth it. So I'll go ahead and proceed with the current approach.

PR #433, opened 2022-03-01
Update fluentbit to v1.8.13

sophieyfang:

PR #452, opened 2022-03-05
More 3p fixes

dehaansa:
Fix hbase download path now that 2.4.10 replaced 2.4.9 in stable directory.
Change rabbitmq source for debian 11.

### DIFF
--- a/integration_test/third_party_apps_test.go
+++ b/integration_test/third_party_apps_test.go
@@ -425,7 +425,7 @@ func TestThirdPartyApps(t *testing.T) {
 		t.Fatal(err)
 	}
 	tests := []test{}
-	platforms := strings.Split(os.Getenv("PLATFORMS"), ",")
+	platforms := []string{"debian-11"} //strings.Split(os.Getenv("PLATFORMS"), ",")
 	for _, platform := range platforms {
 		apps, err := appsToTest(agentType, platform)
 		if err != nil {
--- a/integration_test/third_party_apps_test.go
+++ b/integration_test/third_party_apps_test.go
@@ -425,7 +425,7 @@ func TestThirdPartyApps(t *testing.T) {
 		t.Fatal(err)
 	}
 	tests := []test{}
-	platforms := []string{"debian-11"} //strings.Split(os.Getenv("PLATFORMS"), ",")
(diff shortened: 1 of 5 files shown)


integration_test/third_party_apps_test.go
```diff
@@ -425,7 +425,7 @@ func TestThirdPartyApps(t *testing.T) {
 		t.Fatal(err)
 	}
 	tests := []test{}
-	platforms := strings.Split(os.Getenv("PLATFORMS"), ",")
+	platforms := []string{"debian-11"} //strings.Split(os.Getenv("PLATFORMS"), ",")
```
sophieyfang, line 428: Is this a temp change for getting the test on debian11? It’s not intended to be in master branch, am I right?

dehaansa: Correct! I was hoping this would cause kokoro to run debian-11 tests. I ran the integration tests locally following the instructions we discussed previously, but I was hoping to force kokoro to run them in this current kokoro run on the PR

sophieyfang: I think this line of change will cause the 6 3p test running in the checks to have the title as e.g.  3rd party integration test (Centos 7) <— but inside it, it’s actually testing debian11. Not sure if it will cause it to be in chaotic way as some are in centos7 some are in debian11

sophieyfang: The code to turn on kokoro on debian11 is part of google internal code. It involves building the agent from this PR branch. Then copy paste it to gcs bucket. Bringing up corresponding VM then set up apps and agent for the test. Some happens with google internal stuff. Some are open sourced in this repo. Therefor, this line of change won’t test debian11 unfortunately. 

if you test it locally following the dev guide to trigger this third party test go file on debian11, that will be good enough for us to merge this change into master branch and close those GitHub issue 😊

Still we can’t turn on debian11 on presubmit test yet. As those apps using multiline parser crash immediately on starting up with current master code which uses 1.8.12. Fluentbit 1.8.13 though fixes the issue however, it has regression and fails our soak test… therefore, I am holding on this PR https://github.com/GoogleCloudPlatform/ops-agent/pull/433 till fluentbit fixes the regression.

sophieyfang (approved):

sophieyfang (approved):

PR #361, opened 2022-01-26
Try reenabling mysql tests

martijnvans:
mysql tests were completely broken, let's see if they are fixed now.

PR #455, opened 2022-03-08
Reenable MySQL tests again

martijnvans:
https://github.com/GoogleCloudPlatform/ops-agent/pull/361 reenabled mysql tests, but looks like they were accidentally re-disabled due to a bad merge in https://github.com/GoogleCloudPlatform/ops-agent/pull/396.


stati0n (approved):

martijnvans:
merging, because the last run was clean and the latest run is still going to be running for an hour.

PR #472, opened 2022-03-16
Test for upgrading the ops agent

martijnvans:
This is a test for the upgrade path for the Ops Agent (buganizer number 214272935). First the stable Ops Agent is installed, ignoring AGENT_PACKAGES_IN_GCS and REPO_SUFFIX, then the Ops Agent is installed from AGENT_PACKAGES_IN_GCS/REPO_SUFFIX, and after each installation we do some basic checks to make sure that logging and monitoring are working.

Implementation Note: there were already some nearly-suitable liveness checkers for logging and monitoring for Test{Logging,Metrics}AgentCrashRestart, so I factored them out and reused them. I had to change the logging one to generate a UUID so that it was suitable for running multiple times consecutively.

This PR was tested:

- on Kokoro [PASSED]
- using "go test", with an old REPO_SUFFIX (20220210-1) [PASSED]
- using "go test", with a new REPO_SUFFIX (20220315-1) [PASSED]
- using an old AGENT_PACKAGES_IN_GCS [PASSED]


integration_test/ops_agent_test.go
```diff
@@ -1164,6 +1166,41 @@ func TestWindowsMonitoringAgentConflict(t *testing.T) {
...
+
+func TestUpgradeOpsAgent(t *testing.T) {
+	t.Parallel()
+	gce.RunForEachPlatform(t, func(t *testing.T, platform string) {
+		t.Parallel()
+
+		ctx, logger, vm := agents.CommonSetup(t, platform)
+
+		// This will install the stable (last-released) Ops Agent.
+		if err := installOpsAgentFromRapture(ctx, logger.ToMainLog(), vm, ""); err != nil {
```
davidbtucker, line 1183: Two questions:

1. Do we want to make a parameter for the first version installed? (It could still default to installing the last-released.) That would make it possible to test arbitrary combinations of versions, or to reproduce failures.

2. I notice the second install below uses `setupOpsAgent()`, which does an extra sleep. Do we need that sleep here too?

martijnvans: I wrote a response, then decided to rewrite the code so that the logic is more unified and everything (except some tests for standalone agents) funnels through setupOpsAgent. Here's what I wrote before I decided to rewrite stuff:

1. I think that the parameter you're asking for is already in place? The last argument to installOpsAgentFromRapture() is the repo_suffix to install. This callsite passes in "", meaning stable, and other callsites pass os.Getenv("REPO_SUFFIX").
2. hmm. As it turns out, the sleep is only actually useful for one or two tests, the ones that log to TCP/UDP ports. Those tests really need the agent to be totally ready by the time that they log anything to the port. The other tests log to files or the windows event log, and the ops agent I guess knows how to get caught up on stuff that it missed in those cases. (actually I don't know that it does for the windows event log... maybe the sleep IS helpful for windows... last time I tried removing the sleep i think the windows tests passed, not sure though.)  Anyway, the sleep isn't needed in this case, .... actually it's kind of gross how these helper functions are organized, let me rewrite the PR.

davidbtucker: Oh sorry, I meant could we make an env var for the first version installed, like `REPO_SUFFIX_PREV`. Then we could test arbitrary upgrades by setting `REPO_SUFFIX_PREF` and `REPO_SUFFIX`. Does that make sense at all?

davidbtucker (approved):

integration_test/ops_agent_test.go
martijnvans: sure, added. At first I felt like it was YAGNI, but I think it's harmless too, so I added it.

davidbtucker (approved):
LGTM

PR #476, opened 2022-03-17
Upgrade hbase

jefferbrecht:
<s>2.4.9 seems to have been fully replaced by 2.4.10.</s> Actually let's try Martijn's idea of automatically grabbing the latest stable version.

### DIFF
--- a/integration_test/third_party_apps_data/applications/hbase/centos_rhel/install
+++ b/integration_test/third_party_apps_data/applications/hbase/centos_rhel/install
@@ -1,13 +1,16 @@
 set -e
+set -o pipefail
 
-HBASE_VERSION=$(curl https://apache.mirror.gtcomm.net/hbase/stable/ | grep --extended-regexp 'hbase-[0-9.]+-bin.tar.gz' | grep --extended-regexp --only-matching '[0-9]+\.[0-9]+\.[0-9]+' | head -1)
+HBASE_VERSION=$(curl https://apache.mirror.gtcomm.net/hbase/stable/ \
+    | grep --extended-regexp --only-matching 'hbase-[0-9.]+-bin.tar.gz' \
+    | grep --extended-regexp --only-matching --max-count=1 --color=never '[0-9]+\.[0-9]+\.[0-9]+')
 
 sudo yum -y install \
     java-1.8.0-openjdk java-1.8.0-openjdk-devel wget
 
 sudo mkdir /opt/hbase
-wget http://apache.mirror.gtcomm.net/hbase/$HBASE_VERSION/hbase-$HBASE_VERSION-bin.tar.gz
-sudo tar -xvf hbase-$HBASE_VERSION-bin.tar.gz -C /opt/hbase --strip 1
+wget http://apache.mirror.gtcomm.net/hbase/"$HBASE_VERSION"/hbase-"$HBASE_VERSION"-bin.tar.gz
+sudo tar -xvf hbase-"$HBASE_VERSION"-bin.tar.gz -C /opt/hbase --strip 1
 
 sudo tee /opt/hbase/conf/hbase-env.sh <<EOF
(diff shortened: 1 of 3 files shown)


dehaansa (approved):

integration_test/third_party_apps_data/applications/hbase/centos_rhel/install
```diff
@@ -1,11 +1,13 @@
 set -e
 
+HBASE_VERSION=$(curl https://apache.mirror.gtcomm.net/hbase/stable/ | grep --extended-regexp 'hbase-[0-9.]+-bin.tar.gz' | grep --extended-regexp --only-matching '[0-9]+\.[0-9]+\.[0-9]+' | head -1)
```
martijnvans, line 3: I think --only-matching makes sense for the first grep command as well, in case there happen to be two version number-looking things on that line.

martijnvans, line 3: I can't see `head -1` documented, could we use `head --lines=1` instead? or `grep --max-count=1`?

martijnvans, line 3: Since we have pipes in this script, the script should probably have `set -o pipefail` at the top. Details: https://stackoverflow.com/questions/68465355/what-is-the-meaning-of-set-o-pipefail-in-bash-script

otherwise if there's an error along the pipeline the script will keep going and HBASE_VERSION will just be empty. :/ tested with:

```
set -e

HBASE_VERSION=$(curl https://wwasldfkj.net/hbase/stable/ \
    | grep --extended-regexp 'hbase-[0-9.]+-bin.tar.gz' \
    | grep --extended-regexp --only-matching '[0-9]+\.[0-9]+\.[0-9]+' \
    | head -1)

echo 1111111 $HBASE_VERSION  # prints 1111111 

set -o pipefail

HBASE_VERSION2=$(curl https://wwasldfkj.net/hbase/stable/ \
    | grep --extended-regexp 'hbase-[0-9.]+-bin.tar.gz' \
    | grep --extended-regexp --only-matching '[0-9]+\.[0-9]+\.[0-9]+' \
    | head -1)

echo 2222222 $HBASE_VERSION2  # never runs due to set -e combined with pipefail
```

martijnvans, line 3: IMO line wrapping with \ will make this more readable:

```
HBASE_VERSION=$(curl https://apache.mirror.gtcomm.net/hbase/stable/ \
    | grep --extended-regexp 'hbase-[0-9.]+-bin.tar.gz' \
    | grep --extended-regexp --only-matching '[0-9]+\.[0-9]+\.[0-9]+' \
    | head -1)
```

it works, I tried it.

integration_test/third_party_apps_data/applications/hbase/centos_rhel/install
```diff
@@ -1,11 +1,13 @@
...
+
 sudo yum -y install \
     java-1.8.0-openjdk java-1.8.0-openjdk-devel wget
 
 sudo mkdir /opt/hbase
-wget http://apache.mirror.gtcomm.net/hbase/2.4.9/hbase-2.4.9-bin.tar.gz
-sudo tar -xvf hbase-2.4.9-bin.tar.gz -C /opt/hbase --strip 1
+wget http://apache.mirror.gtcomm.net/hbase/$HBASE_VERSION/hbase-$HBASE_VERSION-bin.tar.gz
```
martijnvans, line 9: FYI:

The google bash style guide says to always double quote variables: `"$HBASE_VERSION"`. IIUC this is to handle variable values that contain spaces "as you expect" instead of "how bash is implemented". There is also a linter for this in google3.

Since this is not in google3 this is optional, but IMO it's a good habit.

martijnvans (approved):
Looks good, the only important thing is it really needs pipefail, see my comment below.

jefferbrecht:
All tests passed before updating to master, which only affected docs, so this is OK to merge.

PR #478, opened 2022-03-18
Added request_count logging self-metrics to ops-agent

rafaelwestphal:
Added the self metrics request_count to ops-agent.
This relays on metrics exposed by the fluentbit prometheus endpoint.


rafaelwestphal:
Implemented on another PR

PR #479, opened 2022-03-18
Added log_entry_count and request_count self-metrics to ops-agent

rafaelwestphal:
Added the self metrics log_entry_count and request_count to ops-agent. This relays on metrics exposed by the
    fluentbit prometheus endpoint.


qingling128 (commented):
Can we add some integration tests for this? Simply uncommenting https://github.com/GoogleCloudPlatform/ops-agent/blob/9645814fa28d3edf854813a89310999581d91f32/integration_test/ops_agent_test.go#L901-L904 might be sufficient

qingling128 (approved):
LGTM

PR #483, opened 2022-03-21
Use archive for hbase

jefferbrecht:
Archive links should last longer than what we had before.


martijnvans (approved):

jefferbrecht:
Debian 10 failure looks unrelated (mysql). Linux hbase tests passed so it looks good to merge.

PR #487, opened 2022-03-22
Upgrade fluentbit to v1.8.14

sophieyfang:
Soak-tested in b/222146442#comment13
All distros e2e test b/222146442#comment14


qingling128:
Looking at the soak test:

There has been some minor CPU usage improvement from Fluent Bit `1.8.12` to `1.8.14`, but a big memory usage regression:

image

image

PR #488, opened 2022-03-22
Also try keyserver.ubuntu.com for MySQL key

martijnvans:
This is to help with the occasional flake where pgp.mit.edu is inaccessible.


integration_test/third_party_apps_data/applications/mysql/debian_ubuntu/install
```diff
@@ -4,7 +4,7 @@ sudo apt update
 sudo apt install -y wget
 
 # From https://dev.mysql.com/doc/mysql-apt-repo-quick-guide/en/#repo-qg-apt-repo-manual-setup
-sudo apt-key adv --keyserver pgp.mit.edu --recv-keys 3A79BD29
+sudo apt-key adv --keyserver pool.sks-keyservers.net --recv-keys 3A79BD29
```
igorpeshansky, line 7: Maybe use keyserver.ubuntu.com instead? See https://github.com/rvm/rvm/issues/5096.

igorpeshansky (changes requested):

integration_test/third_party_apps_data/applications/mysql/debian_ubuntu/install
martijnvans: done. the latest snapshot tries both locations. sadface

integration_test/third_party_apps_data/applications/mysql/debian_ubuntu/install
```diff
@@ -3,8 +3,11 @@ set -e
 sudo apt update
 sudo apt install -y wget
 
-# From https://dev.mysql.com/doc/mysql-apt-repo-quick-guide/en/#repo-qg-apt-repo-manual-setup
-sudo apt-key adv --keyserver pgp.mit.edu --recv-keys 3A79BD29
+function install_key() {
```
igorpeshansky, line 6: [Optional] `add_apt_key`?

igorpeshansky (changes requested):

jefferbrecht (approved):

integration_test/third_party_apps_data/applications/mysql/debian_ubuntu/install
martijnvans: done

igorpeshansky (approved):
LGTM :shipit:

quentinmit:
Honestly you should just switch wholesale; `pgp.mit.edu` is very very flaky.

(Or better, you should import the key from an https URL instead of using a keyserver)

martijnvans:
Quentin, I'm not having luck finding such a URL and I'm not sure how to pass one in even if I found it. I feel like progress on this PR has stalled and I could use some more help from you to get it going again, or I could merge it as is and admit that it's not the best possible implementation and we can fix it later if needed. Which would you prefer?

quentinmit:
> Quentin, I'm not having luck finding such a URL and I'm not sure how to pass one in even if I found it. I feel like progress on this PR has stalled and I could use some more help from you to get it going again, or I could merge it as is and admit that it's not the best possible implementation and we can fix it later if needed. Which would you prefer?

The first bullet point at https://dev.mysql.com/doc/mysql-apt-repo-quick-guide/en/#repo-qg-apt-repo-manual-setup says

> Download the MySQL GPG Public key (see [Signature Checking Using GnuPG](https://dev.mysql.com/doc/refman/8.0/en/checking-gpg-signature.html) on how to do that) and save it to a file

I agree that I can't seem to find that as a separate file directly... one option would be

```
curl https://dev.mysql.com/doc/refman/8.0/en/checking-gpg-signature.html | perl -nle 's/^.*(-----BEGIN)/\1/; print if /BEGIN PGP PUBLIC KEY BLOCK/../END PGP PUBLIC KEY BLOCK/'
```

Another option appears to be to just install https://dev.mysql.com/get/mysql-apt-config_0.8.22-1_all.deb which will configure the apt repo and load the appropriate key.

jefferbrecht:
Tests all passed and it's up to date with master. This is a substantial improvement that'll help avoid many test failures so I'll go ahead and merge it now. If there are any outstanding threads of discussion here then a follow-up PR would be appropriate.

quentinmit:
> Tests all passed and it's up to date with master. This is a substantial improvement that'll help avoid many test failures so I'll go ahead and merge it now. If there are any outstanding threads of discussion here then a follow-up PR would be appropriate.

Yes, there are two suggestions above for making this more reliable that are waiting for reply.

martijnvans:
I'm taking those suggestions under consideration in https://github.com/GoogleCloudPlatform/ops-agent/pull/510.

PR #491, opened 2022-03-23
Print interleaved stdout+stderr in logs and errors.

martijnvans:
Consider a sequence where 1 and 2 were printed to stdout, then 3 was printed to stderr, then 4 was printed to stdout.

Previously, this looked like this in the logs:

```
stdout:
1
2
4
stderr:
3
```

After this change, it will look like:

```
stdout+stderr:
1
2
3
4
```

Which is a lot less confusing. stdout and stderr are still captured separately for programmatic access.

### DIFF
--- a/integration_test/gce/gce_testing.go
+++ b/integration_test/gce/gce_testing.go
@@ -519,9 +519,19 @@ type CommandOutput struct {
 	Stderr string
 }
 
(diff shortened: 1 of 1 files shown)


jefferbrecht (approved):
+1000, this will be a huge help for sifting through build logs! LGTM

PR #494, opened 2022-03-24
Support mysql changes introduced in opentelemetry-operations-collector v0.45.1

dehaansa:
v0.45.1 has not yet been pulled into opentelemetry-operations-collector (see https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/85), but once it is I wanted these changes to be ready to go, to support maximizing backwards compatibility while also grabbing bug fixes as appropriate.

See https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/7924 for the relevant changes to the mysql receiver.

Net changes after updating to v0.45.1 or greater & merging this PR:
- "buffer_pool_pages" label values will go from [data, dirty, flushed, free, misc, total] to [data, free, misc]
- new "buffer_pool_data_pages" metric with label values [dirty, flushed]
- new "buffer_pool_page_flushes" metric (the value of "total" from the previous "buffer_pool_pages" metric)
- "buffer_pool_size" label values will go from [data, dirty, total] to [data, dirty]
- new "buffer_pool_limit" metric (the value of "total" from the previous "buffer_pool_size" metric)


integration_test/third_party_apps_data/applications/mysql/expected_metrics.yaml
```diff
@@ -9,13 +9,29 @@
...
+  value_type: INT64
+  kind: GAUGE
+  monitored_resource: gce_instance
+  labels:
+    kind: dirty|flushed
+- type: workload.googleapis.com/mysql.buffer_pool_page_flushes
+  value_type: INT64
+  kind: CUMULATIVE
+  monitored_resource: gce_instance
+  labels: []
```
jefferbrecht, line 23: Empty `labels` should be `{}` (there's another one below for `buffer_pool_limit`)

integration_test/third_party_apps_data/applications/mysql/expected_metrics.yaml
```diff
@@ -9,13 +9,29 @@
   kind: GAUGE
   monitored_resource: gce_instance
   labels:
-    kind: data|dirty|flushed|free|misc|total
+    kind: data|free|misc
+- type: workload.googleapis.com/mysql.buffer_pool_data_pages
```
jefferbrecht, line 13: Nit: prefer sorting metrics alphabetically by type so that they don't get rearranged when generate_expected_metrics gets run

integration_test/third_party_apps_data/applications/mysql/expected_metrics.yaml
```diff
@@ -1,21 +1,37 @@
...
-    kind: data|dirty|flushed|free|misc|total
+    kind: data|free|misc
 - type: workload.googleapis.com/mysql.buffer_pool_size
   value_type: DOUBLE
   kind: GAUGE
   monitored_resource: gce_instance
   labels:
-    kind: data|dirty|total
+    kind: data|dirty
 - type: workload.googleapis.com/mysql.commands
```
jefferbrecht, line 35: If this is the PR that pulls in the fixed command labels then we should add them here too. If I've been following correctly, it should contain the values from `agent.googleapis.com/mysql/command_count` in the discrepancy doc (i.e. the super long list of like 100 different values). cc @igorpeshansky to confirm

dehaansa: I don't believe that those changes have yet been made upstream, but will wait for Igor's confirmation.

jefferbrecht:
LGTM as far as the coding and tests are concerned, modulo the thread on `mysql.commands`, but I think @igorpeshansky should also review this PR just in case since he's closer to the original issue.

integration_test/third_party_apps_data/applications/mysql/expected_metrics.yaml
qingling128: I recall there was a conversation and the conclusion is that `total` is a buggy label and mess up the chart aggregation. So the decision is to treat it as a bug fix and remove `total`

qingling128: Confirmed with Igor offline. We can remove `total`.

igorpeshansky: The `commands` metric has not yet been fixed — it still only pulls in prepared statement metrics.

apps/mysql.go
```diff
@@ -67,6 +67,23 @@ func (r MetricsReceiverMySql) Pipelines() []otel.Pipeline {
...
 			otel.NormalizeSums(),
 			otel.MetricsTransform(
+				// The following changes are here to ensure maximum backwards compatibility after the fixes
+				// introduced https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/7924
+				otel.ChangePrefix("mysql\\.buffer_pool\\.", "mysql.buffer_pool_"),
+				otel.UpdateMetric("mysql.buffer_pool_pages",
+					otel.ToggleScalarDataType,
+				),
+				otel.UpdateMetric("mysql.buffer_pool_data_pages",
+					otel.RenameLabelValues("status", map[string]string{"clean": "flushed"}),
```
igorpeshansky, line 77: As discussed in the doc, this rename is not appropriate. I would just keep "clean", since it's a new metric, anyway.

apps/mysql.go
```diff
@@ -67,6 +67,23 @@ func (r MetricsReceiverMySql) Pipelines() []otel.Pipeline {
...
+				),
+				otel.UpdateMetric("mysql.threads",
+					otel.ToggleScalarDataType,
+				),
+				otel.RenameMetric("mysql.buffer_pool_usage", "mysql.buffer_pool_size",
+					otel.RenameLabel("status", "kind"),
+					otel.ToggleScalarDataType,
+					otel.RenameLabelValues("kind", map[string]string{"clean": "data"}),
```
igorpeshansky, line 85: As discussed in the doc, these are not the same value, so this rename is misleading. I would just keep it as "clean", and let the "data" label remain in projects that used the old agent versions. Eventually, any time series with non-null values in the "data" label will expire, and it'll just obsolete (and absent in newer projects).

igorpeshansky (dismissed):

qingling128:
Any remaining action items / open questions for this one?

apps/mysql.go
jefferbrecht: Resolving since this appears to have been addressed.

qingling128:
Sync'd offline. There is a separate conversation going on for how Ops Agent handles upstream otel metrics changes like this in the long-term. As for this PR, it's ok to be merged first as it's aligned with what we have been doing (aka adding a bandage for backward compatibility and hold on syncing back to upstream until the next major releasee), so that the release is unblocked first.

We might adjust the strategy in the coming future. If so, we will revisit Mysql receiver at that time.

BTW `third party apps integration test (Windows)` seems to have failed (everything else passed). Looking at the history, it passed in some previous runs though.

PR #496, opened 2022-03-25
implement SAP hana logs receiver

dehaansa:
Logs receiver for SAP HANA trace logs is relatively straightforward.

image

Below is an outdated image (source_file), but shows multiline support
image


PhilBrammer:
Are the multi-line log entries being captured correctly?

PhilBrammer:
AFAIK, without using the multiline regex processor, I would expect each line in the multiline log to be treated as a separate log entry, to be parsed with your regex as individual lines.  Recommend using LoggingProcessorParseMultilineRegex.  Mysql logging implements this: https://github.com/GoogleCloudPlatform/ops-agent/blob/05186978760c75e34477cae3cfc63f66adebcce1/apps/mysql.go#L154

dehaansa:
@PhilBrammer the pattern I'm following here is the more recent pattern that we've been using (see elasticsearch, couchdb, hadoop, rabbitmq, zookeeper & wildfly). The pattern in mysql can encounter issues with with logs being separated by chunks in fluent bit and ending up not being parsed together via the multiline parser, though the pattern we're following here where the multiline is attached to the input also has potential issues (unverified) with syslog forwarding.

apps/saphana.go
```diff
@@ -0,0 +1,135 @@
...
+				"thread_id":             "int",
+				"connection_id":         "int",
+				"transaction_id":        "int",
+				"update_transaction_id": "int",
+			},
+		},
+	}.Components(tag, uid)
+
+	c = append(c,
+		fluentbit.TranslationComponents(tag, "severity_flag", "logging.googleapis.com/severity", true,
```
ridwanmsharif, line 57: We should use the same modify fields processor to add both the translation components and the instrumentation source label. See the following for an example.

https://github.com/GoogleCloudPlatform/ops-agent/blob/487639ed8656db052376242ca92d9815ec891644/apps/apache.go#L95-L122

apps/saphana.go
```diff
@@ -0,0 +1,135 @@
...
+				{"w", "WARNING"},
+				{"e", "ERROR"},
+				{"f", "ALERT"},
+			},
+		)...,
+	)
+
+	// If a log is not associated with a connection/transaction the related
+	// fields will be "-1", and we do not want to report those fields
+	for _, field := range []string{
```
ridwanmsharif, line 70: Instead of having 3 modify filters (which will have a large dip in performance), we should use the `modify_fields` processor. The `omit_if` syntax should fit right in for this use case too.

https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/configuration#logging-processor-modify-fields

apps/saphana.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+		{
+			StateName: "cont",
+			NextState: "cont",
+			Regex:     `^(?!\[\d+\]\{-?\d+\})`,
+		},
+	}
+
+	c := r.LoggingReceiverFilesMixin.Components(tag)
+	c = append(c, r.LoggingProcessorSapHanaTrace.Components(tag, "saphana_trace")...)
```
ridwanmsharif, line 128: Please use `p.Type()` instead of `saphana_trace`

apps/saphana.go
```diff
@@ -16,9 +16,117 @@ package apps
...
+		// Sample line: [18048]{-1}[-1/-1] 2020-11-10 12:24:20.988943 e commlib          commlibImpl.cpp(00986) : ERROR: comm::connect to Host: 127.0.0.1, port: 30001, Error: exception  1: no.2110017  (Basis/IO/Stream/impl/NetworkChannel.cpp:2989)
+		// 				System error: SO_ERROR has pending error for socket. rc=111: Connection refused. channel={<NetworkChannel>={<NetworkChannelBase>={this=139745749931736, fd=21, refCnt=1, local=127.0.0.1/58654_tcp, remote=127.0.0.1/30001_tcp, state=ConnectWait, pending=[----]}}}
+		// 				exception throw location:
+		// 				 1: 0x00007f1937d9095c in .LTHUNK27.lto_priv.2256+0x558 (libhdbbasis.so)
+		// 				 ...
+		//				 25: 0x0000563f44888831 in _GLOBAL__sub_I_setServiceStarting.cpp.lto_priv.239+0x520 (hdbnsutil)
+		Regex: `^\[(?<thread_id>\d+)\]\{(?<connection_id>-?\d+)\}\[(?<transaction_id>-?\d+)\/(?<update_transaction_id>-?\d+)\]\s+(?<time>\d{4}-\d{2}-\d{2}\s+\d{2}:\d{2}:\d{2}\.\d{3,6}\d+)\s+(?<severity_flag>\w+)\s+(?<component>\w+)\s+(?<source_file>\S+)\s+:\s+(?<message>[\s\S]+)`,
```
ridwanmsharif, line 42: For the `source_file` field here, I wonder if we can use the `LogEntry` sourceLocation field. We have an example here:
https://github.com/GoogleCloudPlatform/ops-agent/blob/9d9f04c2d9b1ae24521d60185b77f36a28ef0581/confgenerator/logging_receivers.go#L420-L432

Would be a bonus if we can reuse the `modify_fields` processor below but I don't see a `nest` option

apps/saphana.go
```diff
@@ -16,9 +16,117 @@ package apps
...
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
 )
 
+type LoggingProcessorSapHanaTrace struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+}
+
+func (LoggingProcessorSapHanaTrace) Type() string {
+	return "saphana_trace"
```
ridwanmsharif, line 28: Are there other logs that are not `trace` that make its way to `syslog`? Or that could be configured in the future?
I'm wondering why we don't just call this `saphana` like we do with other receivers that have only a singular input

ridwanmsharif (approved):
LGTM modulo a couple of comments that need addressing.

apps/saphana.go
dehaansa: Sure, we can change it.

confgenerator/testdata/valid/linux/logging-receiver_saphana_custom/golden_34dece3a06a12dbed7c1e431af1e84b7.lua
```diff
@@ -0,0 +1,87 @@
...
+end
+record["logging.googleapis.com/sourceLocation"]["file"] = value
+end)(v)
+local v = __field_6;
+(function(value)
+if record["logging.googleapis.com/sourceLocation"] == nil
+then
+record["logging.googleapis.com/sourceLocation"] = {}
+end
+record["logging.googleapis.com/sourceLocation"]["line"] = value
```
ridwanmsharif, line 84: See here. The Lua code already nests it and so the additional filter is a no-op essentially

apps/saphana.go
```diff
@@ -16,9 +16,136 @@ package apps
...
+				},
+				`sourceLocation.line`: {
+					MoveFrom: "jsonPayload.source_line",
+				},
+				InstrumentationSourceLabel: instrumentationSourceValue(p.Type()),
+			},
+		}.Components(tag, uid)...,
+	)
+
+	c = append(c, fluentbit.Component{
```
ridwanmsharif, line 92: This isn't necessary is it? I believe `modify_fields` already nests this under `sourceLocation`. See the other comment for the actual LUA code that the modify creates. Certainly looks like a nesting is happening already.

ridwanmsharif:
Looks like everything passes except for `FAIL: TestThirdPartyApps/sql-std-2019-win-2019/mssql` which is a known unrelated issue being investigated in b/237276968.

PR #499, opened 2022-03-29
Testing: improve Windows test logs

martijnvans:
The idea is to parse the XML returned by powershell over stderr, and extract the human readable part and print that.

To do this, I had to separate out the part of runCommand() that logs the command and the raw stdout/stderr into a function runAndLog(), and for runRemotelyWindows(), get it to call runCommand() directly, skipping runAndLog() in order to do its own logging.

This whole thing is just a workaround for the fact that powershell ignores -OutputFormat Text when run noninteractively.

Before: https://storage.cloud.google.com/ops-agents-public-buckets-test-logs/prod/stackdriver_agents/testing/consumer/ops_agent/presubmit_github/windows/449/20220322-104940/logs/TestLoggingAgentCrashRestart_windows-2019/main_log.txt

After: https://storage.googleapis.com/ops-agents-public-buckets-test-logs/prod/stackdriver_agents/testing/consumer/ops_agent/test/windows/1043/20220808-084805/logs/TestLoggingAgentCrashRestart_windows-2019/main_log.txt

I also checked that when there is something useful on stderr, like a "missing cmdlet" error, it does show up right, and I checked that if the XML fails to parse, that the raw XML is dumped to the log, as desired.


integration_test/gce/gce_testing.go
```diff
@@ -604,15 +618,47 @@ func runRemotelyWindows(ctx context.Context, logger *log.Logger, vm *VM, command
...
 		return CommandOutput{}, err
 	}
-	return runCommand(ctx, logger, "",
+	output, exitCode, err := runCommand(ctx, "",
 		[]string{winRM(),
 			"--host=" + vm.IPAddress,
 			"--username=" + vm.WindowsCredentials.Username,
 			"--password=" + vm.WindowsCredentials.Password,
-			fmt.Sprintf("--command=powershell -NonInteractive -encodedcommand %q", base64.StdEncoding.EncodeToString([]byte(encoded))),
+			fmt.Sprintf("--command=powershell -NonInteractive -OutputFormat Text -EncodedCommand %q", base64.StdEncoding.EncodeToString([]byte(encoded))),
```
martijnvans, line 626: adding -OutputFormat Text does nothing, but I included it here because:

1. it's always bothered me how this seemingly-useful option is not called out in our code anywhere
2. if microsoft ever fixes powershell's behavior, we'll start using the fix automatically.

integration_test/gce/gce_testing.go
```diff
@@ -16,69 +16,64 @@
 
 /*
 Package gce holds various helpers for testing the agents on GCE.
+
 To run a test based on this library, you can either:
 
 * use Kokoro by triggering automated presubmits on your change, or
 * use "go test" directly, after performing the setup steps described
-in README.md.
+  in README.md.
```
igorpeshansky, line 24: These indents area erroneous (https://tip.golang.org/doc/comment#mistakes). Let's undo them? Or is this a merge issue?

integration_test/gce/gce_testing.go
```diff
@@ -16,69 +16,64 @@
...
-	go test -v ops_agent_test.go \
-	  -test.parallel=1000 \
-	  -tags=integration_test \
-	  -timeout=4h
+    Ops Agent code, first build and upload a package to Rapture. Then look up
+    the REPO_SUFFIX for that build and add it as an environment variable to the
+    command below; for example: REPO_SUFFIX=20210805-2. You can also use
+	AGENT_PACKAGES_IN_GCS, for details see README.md.
+
+PROJECT=dev_project \
```
igorpeshansky, line 37: Isn't this a command? It should be indented with 1 tab (https://tip.golang.org/doc/comment#code). We can do all lines with the same indent.

I thought we've addressed these formatting issues in #719?

integration_test/gce/gce_testing.go
```diff
@@ -359,14 +353,6 @@ func IsWindows(platform string) bool {
 	return strings.HasPrefix(platform, "windows-") || strings.HasPrefix(platform, "sql-")
 }
 
-// PlatformKind returns "linux" or "windows" based on the given platform.
```
igorpeshansky, line 362: This definitely looks like a merge issue — it undoes charges from #719.

integration_test/gce/gce_testing.go
```diff
@@ -1285,7 +1303,7 @@ sudo zypper --non-interactive refresh test-vendor`
 	installCmd := `set -ex
 
 ` + repoSetupCmd + `
-sudo zypper --non-interactive install --force-resolution --capability 'python>=3.6'
```
igorpeshansky, line 1288: I think this was also a recent change. Do you need to merge in master?

martijnvans: bad merge

integration_test/gce/gce_testing.go
martijnvans: bad merge

integration_test/gce/gce_testing.go
martijnvans: bad merge

integration_test/gce/gce_testing.go
martijnvans: bad merge

PR #500, opened 2022-03-29
Add expected_logs

sophieyfang:
Add expected_logs and merge expected_logs and expected_metrics as one metadata.yaml file.

Details about the expected_logs are stated in the integration_test/README.md


integration_test/third_party_apps_test.go
```diff
@@ -172,65 +172,61 @@ func installAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.
...
 
-// expectedLogs encodes a series of assertions about what data we expect
-// to see in the logging backend.
-type expectedLogs struct {
-	// Note on tags: the "yaml" tag specifies the name of this field in the
-	// .yaml file.
-	LogEntries []expectedLog `yaml:"log_entries"`
+type logFields struct {
+	Name        string `yaml:"name" validate:"required"`
+	Value       string `yaml:"value" validate:"required"`
```
jefferbrecht, line 177: Since this is still treated as a regex, a name like `Matcher` or `ValueRegex` might be more appropriate.

jefferbrecht: There's also a check in `constructQuery` for `Value != ""`, so you could actually make this optional as well. Or, for consistency with expected_metrics, we could default all `Value`s to `.*`. (The README says it's optional FWIW.)

integration_test/third_party_apps_test.go
```diff
@@ -172,65 +172,61 @@ func installAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.
...
+	Description string `yaml:"description" validate:"optional,omitempty"`
 }
+
 type expectedLog struct {
-	LogName string `yaml:"log_name"`
-	// Map of field name to a regex that is expected to match the field value.
-	// For example, {"jsonPayload.message": ".*access denied.*"}.
-	FieldMatchers map[string]string `yaml:"field_matchers"`
+	LogName string       `yaml:"log_name" validate:"required"`
+	Fields  []*logFields `yaml:"fields" validate:"required"`
```
jefferbrecht, line 184: These don't need to be pointers

integration_test/third_party_apps_test.go
```diff
@@ -172,65 +172,61 @@ func installAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.
...
+}
+
+type integrationMetadata struct {
+	ExpectedLogs    []*expectedLog          `yaml:"expected_logs"`
```
jefferbrecht, line 188: Same, these don't need to be pointers

integration_test/third_party_apps_test.go
```diff
@@ -172,65 +172,61 @@ func installAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.
...
+	ExpectedMetrics []common.ExpectedMetric `yaml:"expected_metrics"`
 }
 
 // constructQuery converts the given map of:
 //   field name => field value regex
```
jefferbrecht, line 193: Update comment since it's not a map anymore

integration_test/third_party_apps_test.go
```diff
@@ -172,65 +172,61 @@ func installAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.
...
-func runMetricsTestCases(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM, testCaseBytes []byte) error {
-	var metrics []common.ExpectedMetric
+func runMetricsTestCases(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM, metrics []common.ExpectedMetric) error {
 	var err error
-	if err = yaml.UnmarshalStrict(testCaseBytes, &metrics); err != nil {
-		return fmt.Errorf("could not unmarshal contents of expected_metrics.yaml: %v", err)
-	}
 	if err = common.ValidateMetrics(metrics); err != nil {
-		return fmt.Errorf("expected_metrics.yaml failed validation: %v", err)
+		return fmt.Errorf("expectedMetrics field failed validation: %v", err)
```
jefferbrecht, line 227:
```suggestion
		return fmt.Errorf("expected_metrics failed validation: %v", err)
```

integration_test/third_party_apps_test.go
```diff
@@ -172,65 +172,61 @@ func installAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.
...
 	}
-	logger.ToMainLog().Printf("Parsed expected_metrics.yaml: %+v", metrics)
+	logger.ToMainLog().Printf("Parsed expectedMetrics: %+v", metrics)
```
jefferbrecht, line 229:
```suggestion
```

integration_test/third_party_apps_test.go
```diff
@@ -397,19 +393,26 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
-			return nonRetryable, err
+		logger.ToMainLog().Printf("Parsed metadata.yaml: %+v", metadata)
+		if metadata.ExpectedLogs != nil {
+			logger.ToMainLog().Println("found expectedLogs, running logging test cases...")
+			if err = runLoggingTestCases(ctx, logger, vm, metadata.ExpectedLogs); err != nil {
+				return nonRetryable, err
+			}
+		}
+		if metadata.ExpectedMetrics != nil {
+			logger.ToMainLog().Println("found expectedMetrics, running metrics test cases...")
```
jefferbrecht, line 412:
```suggestion
			logger.ToMainLog().Println("found expected_metrics, running metrics test cases...")
```

integration_test/third_party_apps_test.go
jefferbrecht: Also, how does it behave if `expected_logs` or `expected_metrics` is missing? The nil check in `runSingleTest` suggests this should be allowed (i.e. they are optional) but there's no `validate:"optional"` here. Not sure if the marshaler will give an error or just give you an empty slice.

integration_test/README.md
```diff
@@ -176,13 +176,41 @@ Then, inside `applications/<application>/`:
 1.  (if you want to test logging) `expected_logs.yaml`
 1.  (if you want to test metrics) `expected_metrics.yaml`
 
+### expected_logs
+
+We use `expected_logs` field inside `metadata.yaml` file both as a test artifact and as a source for documentation, e.g. [Apache(httpd) public doc](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/apache#monitored-logs). All logs ingested from the integration should be documented here.
```
jefferbrecht, line 181:
```suggestion
We use `expected_logs` inside `metadata.yaml` both as a test artifact and as a source for documentation, e.g. [Apache(httpd) public docs](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/apache#monitored-logs). All logs ingested from the integration should be documented here.
```

integration_test/third_party_apps_test.go
```diff
@@ -172,65 +172,61 @@ func installAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.
...
-// to see in the logging backend.
-type expectedLogs struct {
-	// Note on tags: the "yaml" tag specifies the name of this field in the
-	// .yaml file.
-	LogEntries []expectedLog `yaml:"log_entries"`
+type logFields struct {
+	Name        string `yaml:"name" validate:"required"`
+	Value       string `yaml:"value" validate:"required"`
+	Type        string `yaml:"type" validate:"required"`
+	Description string `yaml:"description" validate:"optional,omitempty"`
```
jefferbrecht, line 179: `Description` is documented as required in the README.

integration_test/README.md
```diff
@@ -176,13 +176,41 @@ Then, inside `applications/<application>/`:
...
+    type: string
+    description: Contents of the Host header
+  - name: jsonPayload.user
+    type: string
+    description: Authenticated username for the request
+```
+- `name`: required, it will be used in e2e searching for the matching logs
+- `type`: required, informational
+- `description`: required, informational
+- `value`: optional, the value of the logEntry field will be used in e2e searching for the matching logs
```
jefferbrecht, line 203: s/`logEntry`/`LogEntry`

Also, since `value` behaves like a regex, that should be mentioned here.

integration_test/README.md
```diff
@@ -176,13 +176,41 @@ Then, inside `applications/<application>/`:
...
+ 
+
 ### expected_metrics.yaml
 
-We use `expected_metrics.yaml` both as a test artifact and as a source for documentation. All metrics ingested from the integration should be documented here.
+We use `expected_metrics` field inside `metadata.yaml` file both as a test artifact and as a source for documentation. All metrics ingested from the integration should be documented here.
```
jefferbrecht, line 208:
```suggestion
We use `expected_metrics` inside `metadata.yaml` both as a test artifact and as a source for documentation. All metrics ingested from the integration should be documented here.
```

jefferbrecht:
This looks really good! Just some minor comments about docs/naming/optional/requiredness.

integration_test/third_party_apps_test.go
qingling128: `Description` should be required I guess? The public documentation is derived from this, right?

qingling128 (approved):
LGTM modulo the open threads get addressed (mainly minor comments so I'm gonna LGTM in advance)

integration_test/third_party_apps_test.go
sophieyfang: For activemq which only has metrics but no logs at this time, it will not fall under thiis code snippet so it will be no testing for logs
```
			if err = runMetricsTestCases(ctx, logger, vm, metadata.ExpectedMetrics); err != nil {
				return nonRetryable, err
			}
		}
```
There is actually no built-in "optional" tag from golang validator: https://pkg.go.dev/github.com/go-playground/validator/v10. So i will remove the optional tag i put in else where

integration_test/third_party_apps_test.go
jefferbrecht: Yeah, it would be great IMO to keep it required. The validator should be changed from `optional` to `required` then.

integration_test/third_party_apps_test.go
sophieyfang: TL,DR: I will change Value to ValueRegex and make Value a required field

> [Optional. Not a very strong preference]
> 
> Making the field required could serve as a reminder for developers to verify the value in e2e whenever it is possible. It could potentially help boost test coverage. Developers can set it explicitly to `.*` when necessary.
> 
> This does mean the README needs to be changed to make it required though.
This is actually a very good point of view that i haven't thought of. I intended to make Value optional and description required but got it wrong..


> Since this is still treated as a regex, a name like `Matcher` or `ValueRegex` might be more appropriate.
`ValueRegex` SGTM

> There's also a check in `constructQuery` for `Value != ""`, so you could actually make this optional as well. Or, for consistency with expected_metrics, we could default all `Value`s to `.*`. (The README says it's optional FWIW.)
This is no build-in optional from https://pkg.go.dev/github.com/go-playground/validator/v10. For string, if the yaml file doesn't have it, it will default to "". If it's pointer, it will default to nil. Thinking that there will never be a case that logEntry field is intentionally set as "" and to be tested.

integration_test/third_party_apps_test.go
sophieyfang: hmmm initially i thought it might be more efficient to use pointer. codes in confgenerator seem to all use pointer. As the struct is not actually big one, maybe i am being over optimized.

integration_test/third_party_apps_test.go
sophieyfang: done

integration_test/README.md
sophieyfang: done

integration_test/third_party_apps_test.go
jefferbrecht: How granular are the generated docs intended to be? I wonder if it would make sense to do per-leaf-field entries to solve the problem of objects, for example if something writes to `httpRequest` then have one entry for each field inside of `httpRequest`:

```yaml
- name: httpRequest.requestMethod
  value: GET|PUT|POST|...
- name: httpRequest.requestUrl
  value: .*
...
```

But that is a more complex question I think and we don't have to answer it here on the first cut.

I hadn't considered objects before so now I'm back to whether we should just make `ValueRegex` optional -- how about if the user doesn't specify it (or if they specify `""`), then instead of skipping the test, we check for field existence using e.g. `httpRequest:*`? This works for all kinds of fields, doesn't matter if it's object, string, int, etc.

integration_test/third_party_apps_data/applications/couchdb/metadata.yaml
```diff
@@ -43,3 +44,34 @@
...
+    description: Process ID
+  - name: jsonPayload.message
+    type: string
+    description: Log message
+  - name: jsonPayload.host
+    type: string
+    description: host instance name
+  - name: httpRequest
+    type: object
+    description: Information about the HTTP request associated with this log entry, if applicable. It's availble in access logs.
```
jefferbrecht, line 77: Typo in `It's availble in access logs`, but I'm wondering what that part is trying to say. Is it saying that the information originated from couchdb's access logs? Or that the user can refer to the access logs for more information? I feel like the part before this sentence sufficiently explains the `httpRequest` anyway so maybe this whole chunk can be removed. Up to you.

integration_test/third_party_apps_data/applications/elasticsearch/metadata.yaml
```diff
@@ -102,3 +103,43 @@
...
     thread_pool_name: .*
+expected_logs:
+- log_name: elasticsearch_json
+  fields:
+  - name: jsonPayload.message
+    value_regex: initialized
+    type: string
+    description: 'Log message'
+  - name: jsonPayload.component
+    value_regex: o.e.n.Node
```
jefferbrecht, line 114: These are intended to be literal `.` characters, so they need to be escaped.

integration_test/third_party_apps_test.go
```diff
@@ -172,65 +172,61 @@ func installAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.
...
 //   field name => field value regex
 // into a query filter to pass to the logging API.
-func constructQuery(fieldMatchers map[string]string) string {
+func constructQuery(fields []logFields) string {
 	var parts []string
-	for field, matcher := range fieldMatchers {
-		parts = append(parts, fmt.Sprintf("%s=~%q", field, matcher))
+	for _, field := range fields {
+		if field.ValueRegex != "" {
+			parts = append(parts, fmt.Sprintf("%s=~%q", field.Name, field.ValueRegex))
```
jefferbrecht, line 199: I noticed that some of the `value_regex` contain spaces so the regex needs to be quoted:

```suggestion
			parts = append(parts, fmt.Sprintf(`%s=~"%q"`, field.Name, field.ValueRegex))
```

jefferbrecht (changes requested):
I'm doing another pass and just leaving more minor/optional comments, but I did spot one thing we need to fix in the code, which is that the regex needs to be quoted.

integration_test/third_party_apps_data/applications/elasticsearch/metadata.yaml
jefferbrecht: (Since it was like that before, I'd be OK with doing a follow-up PR that does a deeper pass over the YAML files to resolve these kinds of issues. We don't need to hold up this PR for things that were already there.)

jefferbrecht (approved):
LGTM modulo a resolution to the `value_regex` thread 👍

PR #505, opened 2022-04-01
RFC: Add support for complex capture group names in parse_regex.

igorpeshansky:
This will allow capturing special fields using `parse_regex`.

b/227803515

### DIFF
--- a/confgenerator/logging_processors.go
+++ b/confgenerator/logging_processors.go
@@ -17,6 +17,7 @@ package confgenerator
 import (
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/filter"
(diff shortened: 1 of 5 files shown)


confgenerator/logging_processors.go
```diff
@@ -62,6 +63,52 @@ func init() {
 	LoggingProcessorTypes.RegisterType(func() Component { return &LoggingProcessorParseJson{} })
 }
 
+// rewriteComplexCaptures translates disallowed capture group names into placeholders and a transformation to rename them back in the log record.
+func rewriteComplexCaptures(regex, tag string) (string, []fluentbit.Component) {
+	// Short-circuit regexes that don't have disallowed capture group names
+	disallowed := regexp.MustCompile(`\(\?P<(?:[A-Za-z0-9_]*[^A-Za-z0-9_>])+(?:[A-Za-z0-9_]*)>`)
```
quentinmit, line 69: I don't think it's possible to parse a regex with a regex, because of the nesting parentheses problem. You would need a smarter parser/lexer. We may yet have to go down that road to adopt Stanza, but I really don't want to if we can avoid it...

confgenerator/logging_processors.go
```diff
@@ -62,6 +63,52 @@ func init() {
...
+		}
+		l = i[3]
+	}
+	r = append(r, regex[l:])
+	// Reconstruct the regex
+	regex = strings.Join(r, "")
+	// Rename all captured fields
+	oc := make([][2]string, len(rewrites))
+	for i, g := range rewrites {
+		oc = append(oc, [2]string{"Rename", fmt.Sprintf("__%d %q", i+1, g)})
```
quentinmit, line 99: This doesn't allow users to set complicated fields they might want to set such as `httpRequest.status`. I think you need to use `modify_fields` here to perform the rename operation so that it can parse all the field names correctly.

confgenerator/logging_processors.go
igorpeshansky: In general, you're correct. Luckily, this isn't actually parsing a regex — it's only looking for the named capturing groups, which have a much simpler syntax.

confgenerator/logging_processors.go
igorpeshansky: Right, it doesn't, but this wasn't the intended goal. This was supposed to match the `parse_json` functionality. As I understand it, `parse_json` would not transform `{"httpRequest.status": 200}` into `LogEntry(httpRequest={status=200})` either… For that, you'd need something like the `modify_fields` processor you're adding in #474.

igorpeshansky (commented):
Thanks for the comments. PTAL.

confgenerator/logging_processors.go
quentinmit: We should not match `parse_json`'s behavior here. It intentionally has a different input shape.

confgenerator/logging_processors.go
quentinmit: I'm pretty sure I could construct a regex that this code does not parse correctly. e.g. I don't think it would correctly handle

```
(?P<groupone>\(?P<notagroup>content\))
```

which would match the string

```
P<notagroup>content
```

igorpeshansky: That's fair. It's not really a nesting problem, as much as an escaping problem, and a potential limitation of Go's regexp engine (which, e.g., does not allow lookbehind). It may be possible to slurp up escaped backslashes prefixed by the start anchor or a non-backslash character, which will take care of that. Let me fix the regex (and I can add your example to the test to make sure it stays untouched).

confgenerator/logging_processors.go
jefferbrecht: What about:

1.  `(?<logging.googleapis.com/labels/hello>...)`

2.  `(?<"logging.googleapis.com/labels".hello>...)`

The second form is basically the filter syntax, except it uses a non-LogEntry structure which would be confusing...

If `severity` is the only one that doesn't need nesting, and `severity` is the one causing the most trouble for users, what if we just forget about nested fields? Seems like it would still add value even if `severity` is the only one that really works with the syntax.

PR #508, opened 2022-04-04
Add log_file_path label using modify_fields.

franciscovalentecastro:
- Enable "Reserve_Data : true" for regex and json parsers.
- Add "Path_Key : agent.googleapis.com/log_file_path" to get path to tailed log.
- Use "modify_fields" to add as label.


franciscovalentecastro:
The new #666 PR adds the same changes, but it uses `modify_fields` now that is merged to the `master` branch. Closing this PR.

PR #510, opened 2022-04-04
Install a newer version of mysql-apt-config

martijnvans:
This avoids the need to fetch the updated repo key in a separate step.


quentinmit (approved):

martijnvans:
third party apps tests passed, the ops agent test failure looks like an unrelated flake.

PR #520, opened 2022-04-06
Add supported app version to 3rd party apps and validation test

alison0211:
Issue: https://b.corp.google.com/issues/225742908


integration_test/third_party_apps_data/applications/memcached/metadata.yaml
```diff
@@ -4,6 +4,7 @@ long_name: Memcached
 minimum_supported_agent_version:
   metrics: 2.8.0
 expected_metrics:
+supported_app_version: ["1.6"]
```
sophieyfang, line 7: Test failed because it should've been:
```
supported_app_version: ["1.6"]
expected_metrics:
```

sophieyfang (approved):

ViliusS:
I see that this change added ActiveMQ Artemis 2.x to Ops Agent supported applications however I never managed it to work with Artemis, only with classic ActiveMQ. All the integration tests also only test classic ActiveMQ 5.16.x series.

Did anyone actually try Ops Agent with Artemis?

PR #521, opened 2022-04-07
expected_metrics follow-up

jefferbrecht:
This is a follow-up PR to address optional feedback from the last PR (#456), as well as to update the generator script to use the new metadata.yaml format.


jefferbrecht:
I don't feel strongly about including the regenerated metadata.yaml files and I'd be OK with reverting it if others feel they should be left unchanged. My rationale was to minimize diffs when the script is run in the future, but the downside is that it'll cause merge conflicts for anyone actively working on them, and I'm not sure how much use the generator script gets anyway.

integration_test/third_party_apps_data/applications/apache/metadata.yaml
```diff
@@ -1,4 +1,4 @@
-public_url: "https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/apache"
+public_url: https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/apache
 short_name: apache
```
sophieyfang, line 2: There are quite some changes in the per integration metadata.yaml files. Some are reordering. May i ask what's the reason to reorder them?🤔

jefferbrecht: This was just the result of running the generate script on the existing yaml files -- the marshaller doesn't preserve ordering by default so the result is that things got out of order when reading and rewriting the file.

I would be fine with reverting all of these yaml changes in this PR since it is indeed quite a lot of changes. The caveat is that the next time someone runs the generate script, it'll change the ordering anyway.

sophieyfang: The thing with the current code is that it changes too many files unnecessarily. It will need reviewer carefully skim over one by one to see that nothing gets wrong.

This shall preserve the ordering: https://stackoverflow.com/questions/33639269/preserving-order-of-yaml-maps-using-go

jefferbrecht: I think I tried `MapSlice` and ran into some walls with being able to continue using structured types with it. I also tried going deeper using v3 of the package which lets you access the underlying token stream but it still formats things slightly differently (arrays are indented in v3 but not in v2 and there's not enough control over it).

So for now I think a better solution is to just pull out all of these yaml changes to reduce the diff size.

integration_test/common/common.go
```diff
@@ -29,55 +29,71 @@ type ExpectedMetric struct {
...
+	Fields  []LogFields `yaml:"fields" validate:"required"`
+}
+
+type MinimumSupportedAgentVersion struct {
+	Logging string `yaml:"logging,omitempty"`
+	Metrics string `yaml:"metrics,omitempty"`
+}
+
+type IntegrationMetadata struct {
+	PublicUrl                    string                       `yaml:"public_url,omitempty"`
```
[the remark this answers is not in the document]
jefferbrecht: `omitempty` here is for `yaml` so that it doesn't need to serialize empty fields. I added `omitempty` to all of the optional fields to make the new generated yaml more similar to the existing ones but we don't need it anymore if I'm removing them from the diff. I can revert these.

integration_test/third_party_apps_data/applications/apache/metadata.yaml
sophieyfang: SGTM!
Didn't know in golang it needs this much work and still can't get it to work.. in python, there is oyaml that will just do the trick.

Another way i can think of is we can make generate.go only print out expected_metrics in the terminal, and letting the code submitter manually modify the metadata.yaml with the provided info showed in the terminal

jefferbrecht: True, that would make the code much simpler too... I'll give that a try, good suggestion!

integration_test/common/common.go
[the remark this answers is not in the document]
sophieyfang: My bad. For all times, i was thinking of something different..

I think having omitempty in 'yaml' specifier is fine, as it won't set the field if it's missing from the file.  It doesn't after the e2e test much, as set or not set, golang always has it as empty string in either case. The current test does not use the field for querying if the value shows as empty string: https://github.com/GoogleCloudPlatform/ops-agent/blob/19e1eb67f06bf4497678df9a1f111cd1ca32f279/integration_test/third_party_apps_test.go#L217
But in general i think it's good to have omitempty：）
image

sophieyfang (approved):
LGTM after the comment about adding back omitempty:)

integration_test/common/common.go
```diff
@@ -29,55 +29,71 @@ type ExpectedMetric struct {
 	// The metric type, for example workload.googleapis.com/apache.current_connections.
 	Type string `yaml:"type"`
```
sophieyfang, line 30: Shall we make this required and also below fields.

jefferbrecht: Good catch, will do.

jefferbrecht: Updated

integration_test/common/common.go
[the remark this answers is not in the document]
jefferbrecht: Updated

sophieyfang (approved):

PR #524, opened 2022-04-08
Makefile: create Makefile

braydonk:
I got tired of typing these commands out, so thought it wouldn't hurt to
get a Makefile started.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
+# IMPORTANT: This Makefile is intended for development use only. Its targets
+# are meant for convenience in common development tasks. Please do not build
+# any dependencies on this file as everything is subject to change.
+
 test:
 	go test ./...
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,11 @@
+test:
+	go test ./...
+
+test_confgenerator:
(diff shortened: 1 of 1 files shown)


igorpeshansky:
I thought we used the `Dockerfile` as a replacement for a `Makefile`, but I'll let @quentinmit comment…

qingling128:
Having these for convenience should be fine I think. Can we add a declamation / notice / comment to make it clear that this is for internal development convenience and are subject to changes without further notice. So that customers don't build dependencies on them. Or they can do so but with known risks that we might change / remove the interface in the future.

braydonk:
Makefile here is meant mostly as a way to run common tasks with `make` tags and not meant for actually building. We could try another tool for running these sorts of things (i.e. [tusk](https://github.com/rliebz/tusk)) but `make` is pretty ubiquitous for this in other Go projects (including across OpenTelemetry repos). I can add a comment to the top of the Makefile that signifies that it is meant for running tasks in Development and not for production/release builds.

qingling128 (approved):
LGTM modulo some disclamation / comment is added as mentioned above.

PR #527, opened 2022-04-12
Add a 20s sleep before installing Windows services.

igorpeshansky:
This is to accommodate delayed service deletion on upgrades.

b/227212680


qingling128 (approved):
LGTM

PR #512, opened 2022-04-04
Varnish Metrics Receiver

JonathanWamsley:
Add Support for Varnish Metrics.

[Scope Doc](https://docs.google.com/document/d/1wMdSs4DRGq1LOC0LNMgmtS-KbCXl1yCDTpV9YyQnFrQ/edit)

PR #528, opened 2022-04-13
Add Varnish logging receiver

JonathanWamsley:
This PR depends builds upon Varnish Metrics Receiver [PR](https://github.com/GoogleCloudPlatform/ops-agent/pull/512)

Sample log output
```json
{
  "insertId": "tz6bioffz0ksc",
  "jsonPayload": {},
  "httpRequest": {
    "requestMethod": "GET",
    "requestUrl": "http://127.0.0.1:6081/forbidden.html",
    "status": 404,
    "responseSize": "166",
    "userAgent": "curl/7.74.0",
    "remoteIp": "127.0.0.1",
    "protocol": "HTTP/1.1"
  },
  "resource": {
    "type": "gce_instance",
    "labels": {
      "instance_id": "5669575632407692159",
      "zone": "us-central1-a"
    }
  },
  "timestamp": "2022-05-04T17:58:01Z",
  "labels": {
    "compute.googleapis.com/resource_name": "varnish-debian-access"
  },
  "logName": "projects/otel-agent-dev/logs/varnish_access",
  "receiveTimestamp": "2022-05-04T17:58:02.315735128Z"
}
```


apps/varnish.go
```diff
@@ -52,3 +53,29 @@ func (r MetricsReceiverVarnish) Pipelines() []otel.Pipeline {
 func init() {
 	confgenerator.MetricsReceiverTypes.RegisterType(func() confgenerator.Component { return &MetricsReceiverVarnish{} })
 }
+
+type LoggingProcessorVarnish struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+}
+
+func (LoggingProcessorVarnish) Type() string {
+	return "varnish"
```
ridwanmsharif, line 62: What do you think of having this be a receiver instead? And we would have the files receiver and the processor implemented here in the PR. That way the user won't have to configure a files receiver to use this processor.

apps/varnish.go
```diff
@@ -52,3 +53,29 @@ func (r MetricsReceiverVarnish) Pipelines() []otel.Pipeline {
...
+}
+
+func (p LoggingProcessorVarnish) Components(tag string, uid string) []fluentbit.Component {
+	c := confgenerator.LoggingProcessorParseRegex{
+		// Logging documentation: https://github.com/varnishcache/varnish-cache/blob/04455d6c3d8b2d810007239cb1cb2b740d7ec8ab/doc/sphinx/reference/varnishncsa.rst#format
+		// Sample line: 127.0.0.1 - - [02/Mar/2022:15:55:05 +0000] "GET http://localhost:8080/test HTTP/1.1" 404 273 "-" "curl/7.64.0"
+		Regex: `^(?<http_request_serverIp>[^ ]*) ([^ ]*) (?<http_request_remoteIp>[^ ]*) \[(?<time>[^\]]*)\] "(?<http_request_requestMethod>\S+)(?: +(?<http_request_requestUrl>[^\"]*?)(?: +(?<http_request_protocol>\S+))?)?" (?<http_request_status>[^ ]*) (?<http_request_responseSize>[^ ]*)(?: "(?<http_request_referer>[^\"]*)" "(?<http_request_userAgent>[^\"]*)")?$`,
```
ridwanmsharif, line 69: For the user to use this plugin, what will they have to do to set up Varnish? It looks like we can ask them to use a specific logging format (https://docs.varnish-software.com/tutorials/enabling-logging-with-varnishncsa/#step-3-customize-options-1) and then reuse that here.

apps/varnish.go
dehaansa: The reason we didn't initially create a varnish receiver is that there is no default log location for varnish. The varnish docs lay out the process for [configuring varnishncsa](https://docs.varnish-software.com/tutorials/enabling-logging-with-varnishncsa/) which is what we're generally following in the integration tests for varnish in order to validate logging.

As there is no default log location, we thought the right path was to provide the processor for parsing, but not pre-define a receiver. If that's not the right path, we're happy to change it.

apps/varnish.go
dehaansa: To minimize user configuration we were leaving the varnishncsa logging format as default, is there something valuable you're seeing in the custom log format that appears to be worth the extra configuration requirement?

apps/varnish.go
ridwanmsharif: I think for consistency with the other receivers we should go with this being modelled as a receiver. We could default to reading from `/var/log/varnish/varnishncsa.log` and explicitly call that out in our docs. If the user chooses to configure varnish to log to other locations, they can always update the `include_paths` param in the receiver.

There also seems to be some prior art for this. Our agent should document the configuration requirements that it requires from Varnish to work as intended and that should be enough.

apps/varnish.go
ridwanmsharif: Ack. Not a whole lot of benefit, given that this configuration lives outside the purview of the Ops Agent, it feels natural to me that we provide a log option pattern that the Ops Agent expects. If the user configures the logs to be structured differently, their Ops Agent logs will become unstructured and that user surprise is what I was hoping to minimize. 

But our agent could also just document the configuration requirements that it requires from Varnish to work as intended and that should be enough.

apps/varnish.go
quentinmit: -1, I don't think we should invent a path unless that's actually something that's commonly used out there.

Do you think `/var/log/varnish/varnishncsa.log` is actually a common path? Do any of the distros ship sample configs to generate it, for instance?

ridwanmsharif: The log file mentioned is from their official docs here: https://docs.varnish-software.com/tutorials/enabling-logging-with-varnishncsa/

And there is prior art for this pattern and this log file specifically. See https://github.com/DataDog/integrations-core/tree/master/varnish#log-collection for an example.

Maybe there is more research required to see what the common log file is but this path seems to be the only one I've found being used

apps/varnish.go
```diff
@@ -52,3 +53,35 @@ func (r MetricsReceiverVarnish) Pipelines() []otel.Pipeline {
 func init() {
 	confgenerator.MetricsReceiverTypes.RegisterType(func() confgenerator.Component { return &MetricsReceiverVarnish{} })
 }
+
+type LoggingProcessorVarnishAccess struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+}
+
+func (LoggingProcessorVarnishAccess) Type() string {
+	return "varnish_access"
```
ridwanmsharif, line 62: I don't see any other varnish logs, is there a reason this isn't just `varnish`? Are there other types of logs that we aren't reporting/collecting yet?

apps/varnish.go
```diff
@@ -52,3 +53,35 @@ func (r MetricsReceiverVarnish) Pipelines() []otel.Pipeline {
...
+}
+
+func (p LoggingProcessorVarnishAccess) Components(tag string, uid string) []fluentbit.Component {
+	return genericAccessLogParser(tag, uid)
```
ridwanmsharif, line 66: Can we actually use the generic access log parser here? From what I remembered I thought we needed a custom regex rule because varnish logs had fields like `httpRequest.serverIp`?

docs/varnish.md
```diff
@@ -49,3 +20,21 @@ The Ops Agent collects the following metrics from your varnish instance.
...
+
+| Field                       | Type                                                                                                                            | Description                                                   |
+|-----------------------------|---------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------|
+| `httpRequest.protocol`      | string                                                                                                                          | Protocol used for the request                                 |
+| `httpRequest.referer`       | string                                                                                                                          | Contents of the `Referer` header                              |
+| `httpRequest.remoteIp`      | string                                                                                                                          | Client IP address                                             |
```
ridwanmsharif, line 32: Are we missing `httpRequest.serverIp` from the scope doc?

apps/varnish.go
ridwanmsharif: Synced offline, moving forward with `/var/log/varnish/varnishncsa.log` is okay since there is some convention

apps/varnish.go
JonathanWamsley: updated to `varnish` since it is the only log file

apps/varnish.go
JonathanWamsley: yeah, a generic access log parser can be used. I added the sample line back in if you want to try. The varnishncsa defaults to to [client mode](https://varnish-cache.org/docs/6.0/reference/varnishncsa.html#modes) , so the [varnishncsa log format](https://github.com/varnishcache/varnish-cache/blob/04455d6c3d8b2d810007239cb1cb2b740d7ec8ab/doc/sphinx/reference/varnishncsa.rst#format) reflects the [access log format](https://httpd.apache.org/docs/current/logs.html#accesslog). It was my mistake for putting serverIp since varnishncsa defaults to client mode, not backend mode.

docs/varnish.md
JonathanWamsley: no, see above. I will update the scope doc accordingly.

integration_test/third_party_apps_data/applications/varnish/metadata.yaml
```diff
@@ -5,12 +5,13 @@ description: |-
...
   as well as the number of sessions and backend connections.
-  The integration also collects Varnish access logs and parses them
+  The integration also collects Varnish logs and parses them
   into a standardized json payload.
-
   For more information about Varnish, see
   [Varnish HTTP Cache](https://varnish-cache.org/).
 supported_app_version: ["6.x", "7.0.x"]
+configure_integration: |-
+  The Varnish logging processor processes logs using [varnishncsa](https://varnish-cache.org/docs/6.0/reference/varnishncsa.html). Varnish can enable varnishncsa logging by following [this](https://docs.varnish-software.com/tutorials/enabling-logging-with-varnishncsa/) guide depending on os. Logs are expected to be in the default format and a logrotation should be setup.
```
ridwanmsharif, line 14: Can we also specify in here that the Ops Agent by default expects the configuration of Varnish to write to `/var/log/varnish/varnishncsa.log`. If another file is chosen, the user will have to also update the receiver configuration

ridwanmsharif, line 14: Can we also specify in here that the Ops Agent by default expects the configuration of Varnish to write to `/var/log/varnish/varnishncsa.log`. If another file is chosen, the user will have to also update the receiver configuration

ridwanmsharif (approved):
LGTM

integration_test/third_party_apps_data/applications/varnish/metadata.yaml
JonathanWamsley: updated

PR #529, opened 2022-04-14
Add Active Directory Domain Services logging receiver

BinaryFissionGames:
- Adds an Active Directory Domain Services (active_directory_ds) logs receiver
- Tweaks the existing windows event log receiver (which is used for the active_directory_ds receiver) to parse the timestamp from the `TimeGenerated` field 
- Tweaks the third party integration tests to allow for restarting after the installation script runs (ADDS requires a system restart after installation)

Example log record:
```json
{
  "insertId": "lxl8wdfhuvnw9",
  "jsonPayload": {
    "RecordNumber": 250,
    "ComputerName": "brandon-testing-windows.testing.local",
    "TimeGenerated": "2022-04-12 18:58:42 +0000",
    "Message": "Microsoft Active Directory Domain Services startup complete, version 6.3.9600.20296 \r\n",
    "Sid": "S-1-5-7",
    "SourceName": "NTDS General",
    "EventID": 1000,
    "TimeWritten": "2022-04-12 18:58:42 +0000",
    "Qualifiers": 16384,
    "StringInserts": [
      "6.3.9600.20296",
      ""
    ],
    "Data": "",
    "EventType": "Information",
    "EventCategory": 12,
    "Channel": "Directory Service"
  },
  "resource": {
    "type": "gce_instance",
    "labels": {
      "instance_id": "7509714225315374123",
      "zone": "us-central1-a"
    }
  },
  "timestamp": "2022-04-12T18:58:42Z",
  "severity": "INFO",
  "logName": "projects/otel-agent-dev/logs/active_directory_ds",
  "receiveTimestamp": "2022-04-12T18:58:44.596591648Z"
}
```


martijnvans:
Hmm, this is a difficult one. I dug into the serial port logs (which I don't think that ObservIQ folks have access to, sadly) and found this error:

```
WinRM is already set up to receive requests on this computer.
Set-WSManQuickConfig : <f:WSManFault
xmlns:f=\"http://schemas.microsoft.com/wbem/wsman/1/wsmanfault\"
Code=\"2150858909\" Machine=\"github-test-20220422-1ae75-3d807eff-c03b-4548-ad30-f
b61172d433f.test.example.com\"><f:Message>WinRM cannot process the request. The
following error with errorcode 0x80090350 occurred while using Negotiate
authentication: An unknown security error occurred.
 Possible causes are:
  -The user name or password specified are invalid.
  -Kerberos is used when no authentication method and no user name are
specified.
  -Kerberos accepts domain user names, but not local user names.
  -The Service Principal Name (SPN) for the remote computer name and port does
not exist.
  -The client and remote computers are in different domains and there is no
trust between the two domains.
 After checking for the above issues, try the following:
  -Check the Event Viewer for events related to authentication.
  -Change the authentication method; add the destination computer to the WinRM
TrustedHosts configuration setting or use HTTPS transport.
 Note that computers in the TrustedHosts list might not be authenticated.
   -For more information about WinRM configuration, run the following command:
winrm help config. </f:Message></f:WSManFault>
At line:116 char:17
+                 Set-WSManQuickConfig -force
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Set-WSManQuickConfig], In
   validOperationException
    + FullyQualifiedErrorId : WsManError,Microsoft.WSMan.Management.SetWSManQu
   ickConfigCommand
windows-startup-script-ps1 exit status 1
Finished running startup scripts.
```

This would indeed make the STARTUP_SCRIPT_DONE not appear in the serial port output. The startup script wasn't ready for `Enable-PSRemoting` to fail. Could you try simply removing this line? https://github.com/GoogleCloudPlatform/ops-agent/blob/25abc2cebef7dff1cb261cfbae774d2ba6ca0d84/integration_test/gce/gce_testing.go#L860

if memory serves, we don't actually need that line for anything, it was just there to dump some debugging info. If that doesn't work, we can use "trap" in that startup script such that STARTUP_SCRIPT_DONE is still emitted even when `Enable-PSRemoting` fails. I'll post a followup comment with specifics about that.

martijnvans:
Re using trap, I think this will work but haven't tried it:

```
metadataCopy["windows-startup-script-ps1"] = `
trap {
  "Error occurred: $_";

  $port = new-Object System.IO.Ports.SerialPort 'COM3'
  $port.Open()
  $port.WriteLine("STARTUP_SCRIPT_DONE")
  $port.Close()
}

Enable-PSRemoting  # Might help to diagnose b/185923886.
```

but to be clear, I think we should try removing Enable-PSRemoting first.

martijnvans:
I've run a couple of tests and it looks like removing `Enable-PSRemoting` is safe, so I'll proceed with https://github.com/GoogleCloudPlatform/ops-agent/pull/548 and in the meantime you can remove it in your PR and rerun tests.

BinaryFissionGames:
Thanks for looking into this @martijnvans!

BinaryFissionGames:
Rebased; Kokoro tests have passed previously:
Screen Shot 2022-04-26 at 10 41 10 AM

integration_test/gce/gce_testing.go
```diff
@@ -1215,6 +1214,20 @@ func StartInstance(ctx context.Context, logger *log.Logger, vm *VM) error {
...
+// RestartInstance stops and starts the instance.
+// It also waits for the instance to be reachable over ssh post-restart.
+func RestartInstance(ctx context.Context, logger *logging.DirectoryLogger, vm *VM) error {
+	fileLogger := logger.ToFile("VM_Restart.txt")
+
+	if err := StopInstance(ctx, fileLogger, vm); err != nil {
+		return fmt.Errorf("Failed to stop instance: %w", err)
+	}
+
+	time.Sleep(30 * time.Second)
```
jefferbrecht, line 1226: Nit: `30 * time.Second` should also have its own `vmSomethingBackoffDuration`

integration_test/third_party_apps_data/applications/active_directory_ds/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+    type: string
+    description: "The type of event."
+  - name: jsonPayload.Qualifiers
+    type: number
+    description: "A qualifier number that is used for event identification."
+  - name: jsonPayload.EventCategory
+    type: number
+    description: "The category of the event."
+  - name: jsonPayload.Channel
+    value_regex: '^Directory Service$'
```
jefferbrecht, line 48: Does this need to be `value_regex: '^Directory Service|Active Directory Web Services$'`?

BinaryFissionGames: My understanding is that all these regexes should be as specific as possible to match only a single entry.

I can add another test case for a log from the Active Directory Web Services channel, if that's the concern.

integration_test/gce/gce_testing.go
BinaryFissionGames: Good callout, this sleep actually shouldn't even be necessary, I'll just remove it altogether.

integration_test/third_party_apps_data/applications/active_directory_ds/metadata.yaml
BinaryFissionGames: I added in another case to ensure we get messages from the web services channel

apps/active_directory_ds.go
```diff
@@ -0,0 +1,26 @@
+package apps
```
jefferbrecht, line 1: Nit: add copyright header (just copy from any of the other apps)

BinaryFissionGames: done!

jefferbrecht (approved):

PR #532, opened 2022-04-18
Add integration description

sophieyfang:
b/229635463

### DIFF
--- a/integration_test/third_party_apps_data/applications/activemq/metadata.yaml
+++ b/integration_test/third_party_apps_data/applications/activemq/metadata.yaml
@@ -1,6 +1,13 @@
 public_url: "https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/activemq"
 short_name: activemq
 long_name: Apache ActiveMQ
+description: -|
+  The Apache ActiveMQ integration collects storage usage and message metrics.
+  Storage metrics include memory and disk usage. Message metrics include number of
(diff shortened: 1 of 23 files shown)


integration_test/third_party_apps_data/applications/activemq/metadata.yaml
```diff
@@ -1,6 +1,13 @@
 public_url: "https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/activemq"
 short_name: activemq
 long_name: Apache ActiveMQ
+description: -|
+  The Apache ActiveMQ integration collects storage usage and message metrics.
+  Storage metrics include memory and disk usage. Message metrics include number of
+  waiting messages, average wait time, and expired messages.
+
+  For more information about ActiveMQ, see
+  [activemq.apache.org/](https://activemq.apache.org/){:class="external"}.
```
qingling128, line 10: Is the `{:class="external"}` part needed here, or can it be emitted or added later on internally.

Same for the other files.

qingling128 (approved):
LGTM modulo the open thread is addressed.

integration_test/third_party_apps_data/applications/activemq/metadata.yaml
sophieyfang: Removed the external as staging the doc with or without external tag there is no difference in the look

qingling128 (approved):
LGTM

PR #533, opened 2022-04-18
Migrate existing 3rd party log processors to `modify_fields` where it makes sense

quentinmit:


quentinmit:
Closing in favor of #627

PR #535, opened 2022-04-19
Fix handling of missing fields in filters

quentinmit:
In Lua, `string.lower(nil)` throws an exception, so if `omit_if` or `exclude_logs` is used with a filter that references a field that doesn't exist, the log processing fails and fluent-bit drops the entry on the floor. This is a regression from the previous `modify`-based implementation of `exclude_logs`.

This commit adds an `== nil` guard, as well as explicitly coerces the field to a string (in case it is an integer, or a table, or something more exotic), so the comparison can never fail.

Finally, it generates the Lua code for `omit_if` and multiple filters in a deterministic order, to prevent test flakes.


qingling128 (approved):
LGTM

Kudos for adding the test case as well!

PR #538, opened 2022-04-19
Normalize fluent-bit files paths on Windows to use `\` everywhere.

igorpeshansky:
This allows using `/` in wildcard expressions in the `files` receiver configs regardless of the platform. Today, fluent-bit's wildcard handler only accepts `\` on Windows, while non-wildcard paths accept either `\` or `/`.

b/223795873


confgenerator/testdata/valid/windows/logging-receiver_files_refresh_interval/input.yaml
```diff
@@ -16,11 +16,13 @@ logging:
   receivers:
     log_source_id1:
       type: files
-      wildcard_refresh_interval: 10s
       include_paths:
-      - /path/to/log/1/a
+      - /path/to/log/1/*
```
davidbtucker, line 20: Might be worth splitting this out into a separate test, or at least leaving a comment that this file is testing extra functionality.

davidbtucker (approved):

confgenerator/testdata/valid/windows/logging-receiver_files_refresh_interval/input.yaml
igorpeshansky: The old test was invalid, because it was purporting to test the wildcard refresh interval configuration without a wildcard.

confgenerator/config.go
```diff
@@ -346,7 +346,9 @@ type Logging struct {
 
 type LoggingReceiver interface {
 	Component
-	Components(tag string) []fluentbit.Component
+	// Components returns fluentbit components that implement this receiver.
+	// tag is the log tag that should be produced by those components, and platform is the target platform.
+	Components(tag string, platform string) []fluentbit.Component
```
quentinmit, line 351: Adding a string parameter like this is a major change, and it's going to be even more painful if we have to add more later. If we decided we need to go down this road, we should add a `context.Context` parameter instead so that we only have to do this once.

confgenerator/logging_receivers.go
```diff
@@ -24,13 +24,31 @@ import (
...
 	return path.Join("${buffers_dir}", dir)
 }
 
+// normalizePath replaces all occurrences of `/` in the path with the system path separator.
+// This is a no-op on Linux, but will normalize everything to use `\` on Windows.
+func normalizePath(path string, platform string) string {
+	if platform != "windows" {
+		return path
+	}
+	return strings.ReplaceAll(path, `/`, `\`)
```
quentinmit, line 40: This is not generally safe on Windows. `/` and `\` have subtly different meanings. For example, `\\.\C:\foo/bar\baz.txt` != `\\.\C:\foo\bar\baz.txt`

quentinmit (commented):
This seems like an awfully large amount of work to work around a fluent-bit bug. Why are we not just asking them to fix their path parsing?

PR #541, opened 2022-04-20
integration_test: Add configuration_options field to metadata

ridwanmsharif:
- Adds configuration_options field to metadata.yaml for all 3rd party Application
- Makes this a required field moving forward

### DIFF
--- a/integration_test/third_party_apps_data/applications/tomcat/metadata.yaml
+++ b/integration_test/third_party_apps_data/applications/tomcat/metadata.yaml
@@ -119,14 +119,28 @@ expected_logs:
     description: Authenticated username for the request
 configuration_options:
   logs:
-  - type: ""
+  - type: tomcat_system
     fields:
     - name: type
       default: null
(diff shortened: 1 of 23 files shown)


integration_test/third_party_apps_data/applications/tomcat/metadata.yaml
```diff
@@ -107,3 +107,54 @@ expected_logs:
   - name: jsonPayload.user
     type: string
     description: Authenticated username for the request
+configuration_options:
+  logs:
+  - type: tomcat_system
```
sophieyfang, line 112: image
There is some unmarshall error

ridwanmsharif: Fixed.

sophieyfang (approved):

igorpeshansky:
Why would we not generate this info from the agent code? That way there's less possibility for it to fall out of sync?

PR #534, opened 2022-04-19
Update IIS metrics receiver with v2 metrics

Mrod1598:
Update the existing metrics receiver to use the new opentelemtry version of the IIS metrics receiver. Additionally, update the operations-collector to enable this.

PR #542, opened 2022-04-20
Added SQL Server Receiver Metrics.

StefanKurek:
Adds SQL Server metrics receiver to ops agent.


qingling128:
The public SQL server images are documented here: https://cloud.google.com/compute/docs/images/os-details#sql_server.

It looks like our test framework should [already handles sql images](https://github.com/GoogleCloudPlatform/ops-agent/blob/2c2538ab50b5766f890066a95329f2aa7d2244a1/integration_test/gce/gce_testing.go#L351).

Reading https://github.com/GoogleCloudPlatform/ops-agent/blob/2b52f26b5b5fc5411f442509c5965f2b9af9b7ae/integration_test/third_party_apps_test.go#L115, it might already trigger MSSQL tests on SQL server images today. (CC @martijnvans in case I get it wrong)

StefanKurek:
> The public SQL server images are documented here: https://cloud.google.com/compute/docs/images/os-details#sql_server.
> 
> It looks like our test framework should [already handles sql images](https://github.com/GoogleCloudPlatform/ops-agent/blob/2c2538ab50b5766f890066a95329f2aa7d2244a1/integration_test/gce/gce_testing.go#L351).
> 
> Reading
> 
> https://github.com/GoogleCloudPlatform/ops-agent/blob/2b52f26b5b5fc5411f442509c5965f2b9af9b7ae/integration_test/third_party_apps_test.go#L115
> 
> , it might already trigger MSSQL tests on SQL server images today. (CC @martijnvans in case I get it wrong)

@qingling128 gotcha. So if tests are already running on sql server images, then all I'd have to do is add another line to remove "sqlserver" from the apps as well (when it's a windows machine but not a sql server one)

StefanKurek:
@qingling128 so I'm not sure that I actually have the ability to add new sql images to test against (I may be mistaken but I think this is set in the PLATFORMS environment variable), but given that someone adds to this, I think this code should be ready for it.

martijnvans:
i gave it a shot but I did it wrong, so I need to wait another hour for results. Will update here once the latest run has finished.

martijnvans:
OK I sent a change to add sql-std-2019-win-2019 to our tested images list. That is making its way through our presubmit tests now.

martijnvans:
Oh and by the way, we sould include https://github.com/GoogleCloudPlatform/ops-agent/pull/547 into this PR or merge it into this PR, your choice. If we don't have that change, then our tests will try to run the `sqlserver` application on all versions of windows. [Which is bad, right? I might not understand the situation right.]

martijnvans:
Tests passed, but it didn't run sqlserver. This is because sqlserver isn't in here in your PR:

https://github.com/GoogleCloudPlatform/ops-agent/blob/master/integration_test/third_party_apps_data/agent/windows/supported_applications.txt

StefanKurek:
> Tests passed, but it didn't run sqlserver. This is because sqlserver isn't in here in your PR:
> 
> https://github.com/GoogleCloudPlatform/ops-agent/blob/master/integration_test/third_party_apps_data/agent/windows/supported_applications.txt

@martijnvans I'm pretty sure it's there https://github.com/GoogleCloudPlatform/ops-agent/blob/74f646aaa1f5c64af117ad09682d158ebd32dd82/integration_test/third_party_apps_data/agent/windows/supported_applications.txt

martijnvans:
I'm sorry, the real reason it didn't run is because my pending change to actually run tests on sql server windows hasn't been accepted yet. I'll ping some people and see if we can get that in.

confgenerator/confgenerator.go
```diff
@@ -84,6 +84,9 @@ func (uc *UnifiedConfig) GenerateOtelConfig(hostInfo *host.InfoStat) (string, er
 					"instrumentation_library_labels": false,
 					// Omit service labels, which break agent metrics.
 					"service_resource_labels": false,
+					"resource_filters": []map[string]interface{}{
+						{"prefix": "sqlserver.database.name"},
+					},
```
igorpeshansky, line 89: Why are we adding this across the board? Is there really no way to confine this change to the sqlserver receiver? Does this not break existing metrics?

confgenerator/testdata/invalid/linux/metrics-receiver_sqlserver_non_windows/golden_error
```diff
@@ -0,0 +1 @@
+metrics receiver with type "sqlserver" is not supported. Supported metrics receiver types: [activemq, apache, cassandra, couchdb, elasticsearch, hadoop, hbase, hostmetrics, jvm, kafka, memcached, mongodb, mysql, nginx, postgresql, rabbitmq, redis, solr, tomcat, wildfly, zookeeper].
```
igorpeshansky, line 1: Other tests of this kind are named `metrics-receiver_invalid_type_*` (e.g., `metrics-receiver_invalid_type_mssql`). For consistency, this should be named `metrics-receiver_invalid_type_sqlserver`.

I can see us renaming these wholesale to `metrics-receiver_sqlserver_unsupported` (`metrics-receiver_iis_unsupported`, `metrics-receiver_mssql_unsupported`, `logging-receiver_windows_event_log_unsupported`) to ensure consistent test prefix, but I'd strongly suggest doing that in a separate PR, and making this one consistent with the existing tests.

apps/sqlserver.go
```diff
@@ -0,0 +1,52 @@
...
+func (r MetricsReceiverSqlServer) Pipelines() []otel.Pipeline {
+	cfg := map[string]interface{}{
+		"collection_interval": r.CollectionIntervalString(),
+	}
+
+	return []otel.Pipeline{{
+		Receiver: otel.Component{
+			Type:   "sqlserver",
+			Config: cfg,
+		},
```
igorpeshansky, line 41: Do you need this local? Can you just inline the `map` below?

```suggestion
	return []otel.Pipeline{{
		Receiver: otel.Component{
			Type:   "sqlserver",
			Config: map[string]interface{}{
				"collection_interval": r.CollectionIntervalString(),
			},
		},
```

docs/sqlserver.md
```diff
@@ -0,0 +1,55 @@
+# `sqlserver` Metrics Receiver
+
+The `sqlserver` metrics receiver can fetch stats from a SQL Server via the [Windows Performance Counters](https://docs.microsoft.com/en-us/windows/win32/perfctrs/performance-counters-portal).
+
+## Prerequisites
+Because this `sqlserver` receiver makes use of the windows perforamnce counters, it is required for it to run locally on the same machine with the SQL Server installation.
```
igorpeshansky, line 6: Typo: `perforamnce`

igorpeshansky, line 6: `it is required for it` is a bit jarring. How about just `it is required to run`?

martijnvans:
OK if you try again, it should run MS SQL server this time. I'll kick it off.

docs/sqlserver.md
```diff
@@ -0,0 +1,55 @@
+# `sqlserver` Metrics Receiver
+
+The `sqlserver` metrics receiver can fetch stats from a SQL Server via the [Windows Performance Counters](https://docs.microsoft.com/en-us/windows/win32/perfctrs/performance-counters-portal).
```
igorpeshansky, line 3: I'll let @qingling128 make the final determination here, but does it make sense to call out the differences between this receiver and `mssql`? Otherwise it's going to be very confusing to customers to have two receivers with seemingly the same functionality. If we do call out the difference, we should add some text to the mssql docs as well.

docs/sqlserver.md
```diff
@@ -0,0 +1,55 @@
...
+| ---                                                  | ---            | ---         | ---         | ---         | 
+| workload.googleapis.com/sqlserver.user.connection.count | Gauge       | connections |             | Number of users connected to the SQL Server. |
+| workload.googleapis.com/sqlserver.lock.wait_time.avg | Gauge          | ms          |             | Average wait time for all lock requests that had to wait. |
+| workload.googleapis.com/sqlserver.lock.wait.rate     | Gauge          | requests/s  |             | Number of lock requests resulting in a wait. |
+| workload.googleapis.com/sqlserver.batch.request.rate | Gauge          | requests/s  |             | Number of batch requests received by SQL Server. |
+| workload.googleapis.com/sqlserver.batch.sql_compilation.rate | Gauge  | compilations/s |          | Number of SQL compilations needed. |
+| workload.googleapis.com/sqlserver.batch.sql_recompilation.rate | Gauge | compilations/s |         | Number of SQL recompilations needed. |
+| workload.googleapis.com/sqlserver.page.buffer_cache.hit_ratio | Gauge | %           |             | Pages found in the buffer pool without having to read from disk. |
```
igorpeshansky, line 42: `Percent of pages` or `Fraction of pages`?

docs/sqlserver.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Metrics
+The Ops Agent collects the following metrics from SQL Server.
+Note: In order for any metrics to be retrieved, the receiver must run locally on the same server as the SQL Server installation.
+
+| Metric                                               | Data Type      | Unit        | Labels      | Description |
+| ---                                                  | ---            | ---         | ---         | ---         | 
+| workload.googleapis.com/sqlserver.user.connection.count | Gauge       | connections |             | Number of users connected to the SQL Server. |
+| workload.googleapis.com/sqlserver.lock.wait_time.avg | Gauge          | ms          |             | Average wait time for all lock requests that had to wait. |
+| workload.googleapis.com/sqlserver.lock.wait.rate     | Gauge          | requests/s  |             | Number of lock requests resulting in a wait. |
```
igorpeshansky, line 38: The descriptions of these metrics say `Number of …`, but they are actually rates. Do we need to mention something about that in the description?

Alternatively, should (can?) we ingest these as sums and let the users compute the rates at query time?

docs/sqlserver.md
```diff
@@ -0,0 +1,55 @@
...
+| workload.googleapis.com/sqlserver.page.buffer_cache.hit_ratio | Gauge | %           |             | Pages found in the buffer pool without having to read from disk. |
+| workload.googleapis.com/sqlserver.page.life_expectancy | Gauge        | s           |             | Time a page will stay in the buffer pool. |
+| workload.googleapis.com/sqlserver.page.split.rate    | Gauge          | pages/s     |             | Number of pages split as a result of overflowing index pages. |
+| workload.googleapis.com/sqlserver.page.lazy_write.rate | Gauge        | writes/s    |             | Number of lazy writes moving dirty pages to disk. |
+| workload.googleapis.com/sqlserver.page.checkpoint.flush.rate | Gauge  | pages/s     |             | Number of pages flushed by operations requiring dirty pages to be flushed. |
+| workload.googleapis.com/sqlserver.page.operation.rate | Gauge         | operation/s | type        | Number of physical database page operations issued. |
+| workload.googleapis.com/sqlserver.transaction_log.growth.count | Gauge | growths    | database    | Total number of transaction log expansions for a database. |
```
igorpeshansky, line 48: Shouldn't this be Sum (or Cumulative)? Ditto for the next metric…

docs/sqlserver.md
```diff
@@ -0,0 +1,55 @@
...
+| workload.googleapis.com/sqlserver.transaction_log.shrink.count | Gauge | shrinks    | database    | Total number of transaction log shrinks for a database. |
+| workload.googleapis.com/sqlserver.transaction_log.usage | Gauge       | %           | database    | Percent of transaction log space used. |
```
igorpeshansky, line 50: `percent_used`?

integration_test/third_party_apps_data/applications/sqlserver/metadata.yaml
```diff
@@ -0,0 +1,125 @@
...
+  The Microsoft SQL Server integration collects metrics from your SQL Server
+  instances and databases. Page, batch, lock, and user connection count information
+  is provided at the instance level. Transaction and transaction log information
+  is provided at the database level. 
+
+  For more information on SQL Server, see the
+  [SQL Server](https://docs.microsoft.com/en-us/sql/sql-server)
+  documentation.
+minimum_supported_agent_version:
+  metrics: 2.10.0
```
igorpeshansky, line 14: This needs to be updated, doesn't it?

igorpeshansky (changes requested):
Some of my comments about metric names and types should probably have been raised earlier in the scoping doc. If that's the case, please feel free to refer me to where that discussion happened in that doc.

confgenerator/confgenerator.go
StefanKurek: @igorpeshansky that's a good call out that I am aware of. I could minimize it a bit possibly by only including that filter if the sql server pipeline exists....but it still would apply to everything being exported in that circumstance because there is only a single google exporter for all apps (as far as I know).

StefanKurek: To add a little more to this, the work that was done to the google exporter (to be able to handle resource attributes) allows it to be configured with filters in order to translate the OTEL resource attributes to labels. Outside of this, nothing was added to account for the issues of OTEL receivers with resource attributes.

So unless I'm missing something, this is the best way I could think of dealing with the issue without coming up with a significant change to how the ops agent handles pipelines (which I know is not great because this change is on a global exporter which affects all pipelines...hopefully mitigated by good resource attribute naming).

martijnvans:
Test failed, I see this error in [ops_agent_logs.txt](https://storage.cloud.google.com/ops-agents-public-buckets-test-logs/prod/stackdriver_agents/testing/consumer/third_party_apps/presubmit_github/windows/773/20220422-104541/logs/TestThirdPartyApps_sql-std-2019-win-2019_sqlserver/ops_agent_logs.txt)

```
4/22/2022 6:43:31 PM  1 Error            failed to generate config files: logging receiver with type "sqlserver" is    
                                         not supported. Supported logging receiver types: [apache_access,              
                                         apache_error, cassandra_debug, cassandra_gc, cassandra_system, couchdb,       
                                         elasticsearch_gc, elasticsearch_json, files, fluent_forward, hadoop,          
                                         hbase_system, iis_access, kafka, mongodb, mysql_error, mysql_general,         
                                         mysql_slow, nginx_access, nginx_error, postgresql_general, rabbitmq, redis,   
                                         solr_system, syslog, tcp, tomcat_access, tomcat_system, wildfly_system,       
                                         windows_event_log, zookeeper_general].                
```

docs/sqlserver.md
StefanKurek: @igorpeshansky Unfortunately all of the sql server windows performance counters are gauges. The descriptions were mostly pulled verbatim from Microsoft's documentation for the counters (even though again they are gauges), but we could change the descriptions easily enough if need be.

docs/sqlserver.md
StefanKurek: @qingling128 and @jefferbrecht as approvers of the scope doc, what are you opinions?

docs/sqlserver.md
jefferbrecht: Similar to the [new `iis` metrics](https://github.com/GoogleCloudPlatform/ops-agent/pull/534/files#diff-98d6406bd3b368722a5ecde561e446b313751caddc7c593b5ae130467728cbcfR37), this should be under the existing `mssql` receiver with a `receiver_version: 2` field, rather than as a separate `sqlserver` receiver.

integration_test/third_party_apps_data/applications/sqlserver/metadata.yaml
jefferbrecht: This should be 2.15.0 (2.14.0 is likely going out before this is merged).

docs/sqlserver.md
StefanKurek: This was not quite called out in the scope doc "This will not replace the existing mssql metric functionality and will be able to be run alongside it." But I'll wait for @qingling128's feedback and if you're all in agreement then I'll move forward with this pattern.

integration_test/third_party_apps_data/applications/sqlserver/metric_name.txt
```diff
@@ -0,0 +1 @@
+workload.googleapis.com/sqlserver.user.connection.count
```
jefferbrecht, line 1: This file isn't used anymore and can be deleted. (The `representative: true` metric under `expected_metrics` replaces the old behaviour.)

confgenerator/testdata/invalid/linux/metrics-receiver_sqlserver_non_windows/golden_error
StefanKurek: fixed

apps/sqlserver.go
StefanKurek: changed

docs/sqlserver.md
StefanKurek: fixed

StefanKurek: changed

docs/sqlserver.md
StefanKurek: changed

docs/sqlserver.md
StefanKurek: changed

integration_test/third_party_apps_data/applications/sqlserver/metadata.yaml
StefanKurek: updated

integration_test/third_party_apps_data/applications/sqlserver/metric_name.txt
StefanKurek: deleted

confgenerator/confgenerator.go
igorpeshansky: @StefanKurek I may be missing something about how metrics are modified, but it looks like these resource fields would be added indiscriminately. If not, a comment with a pointer to why this works would be really helpful.

@jefferbrecht I suspect we might just not be running the expected metrics tests with the `sqlserver` pipeline present. One way to be absolutely sure is to build a prototype package from this branch, install it, enable the `sqlserver` receiver, and see if it starts getting API failures due to extraneous labels. I can try to do that tomorrow unless someone beats me to it.

For the record, the substitution of `.` to `_` happens [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/metric/metric.go#L641-L643).

StefanKurek: @igorpeshansky I believe this is the area of code you're looking for https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/836e6fe9832c9f19146dabab29a8b55d9511459f/exporter/collector/monitoredresource.go#L201

jefferbrecht: IIUC the labels are only added if the prefix matches the attribute key. It will be added indiscriminately in the sense that it's not receiver-specific, but effectively this is the only receiver whose attribute keys contain that prefix. There's some documentation [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/3a2c57209765f4752bd140a007612a81f58489c1/exporter/collector/config.go#L92-L94), and the code for matching/adding is [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/836e6fe9832c9f19146dabab29a8b55d9511459f/exporter/collector/monitoredresource.go#L201).

The `sqlserver` tests were running correctly last I checked:

```
2022/04/22 17:49:32 nonEmptySeries() iterator supplied err <nil> and series metric:{type:"workload.googleapis.com/sqlserver.user.connection.count"}  resource:{type:"gce_instance"  labels:{key:"instance_id"  value:"9188105535274022384"}  labels:{key:"project_id"  value:"stackdriver-test-143416"}  labels:{key:"zone"  value:"us-central1-b"}}  metric_kind:GAUGE  value_type:DOUBLE  points:{interval:{end_time:{seconds:1650674969  nanos:104023000}  start_time:{seconds:1650674969  nanos:104023000}}  value:{double_value:0}}
2022/04/22 17:49:32 Found representative metric, sleeping before checking remaining metrics
...
2022/04/22 17:50:42 Skipping optional or representative metric workload.googleapis.com/sqlserver.user.connection.count
2022/04/22 17:50:42 nonEmptySeries() iterator supplied err <nil> and series metric:{type:"workload.googleapis.com/sqlserver.transaction_log.flush.data.rate"  labels:{key:"sqlserver_database_name"  value:"master"}}  resource:{type:"gce_instance"  labels:{key:"instance_id"  value:"9188105535274022384"}  labels:{key:"project_id"  value:"stackdriver-test-143416"}  labels:{key:"zone"  value:"us-central1-b"}}  metric_kind:GAUGE  value_type:DOUBLE  points:{interval:{end_time:{seconds:1650675029  nanos:74760000}  start_time:{seconds:1650675029  nanos:74760000}}  value:{double_value:0}}  points:{interval:{end_time:{seconds:1650674969  nanos:103151000}  start_time:{seconds:1650674969  nanos:103151000}}  value:{double_value:0}}
```

Or are you saying that if _any_ attribute key matches that prefix, then _all_ metrics passing through that exporter, regardless of _their_ attributes, might _also_ get labels added? If that's what you're saying then I assume we can just check the agent metrics from the above test run and confirm that they don't have extraneous labels, since agent metrics pass through the same exporter:

```
{
  "timeSeries": [
    {
      "metric": {
        "labels": {
          "version": "google-cloud-ops-agent-logging/2.14.0-windows-ltsc2019"
        },
        "type": "agent.googleapis.com/agent/uptime"
      },
      "resource": {
        "type": "gce_instance",
        "labels": {
          "instance_id": "9188105535274022384",
          "zone": "us-central1-b",
          "project_id": "stackdriver-test-143416"
        }
      },
      ...
```

But you're right that the tests don't check for this currently.

docs/sqlserver.md
jefferbrecht: IMO the types and units make it obvious enough that these are rates. I don't view the descriptions as contradictory but I'm not against changing it to something like `Number of lock requests per second resulting in a wait.` if we want it to be more clear or if it's more consistent with how we document other such receivers.

docs/sqlserver.md
jefferbrecht: While `percent_used` is clearer IMO, I don't feel strongly about renaming this; there's precedent either way (e.g. `agent.googleapis.com/memory/percent_used` vs. `workload.googleapis.com/activemq.memory.usage`).

I'll defer to @igorpeshansky, so unless there's an objection from the other reviewers then I'm fine with changing it to `percent_used`.

jefferbrecht:
```
2022/04/25 11:12:10 Attempt 1 of sql-std-2019-win-2019 test of sqlserver finished with err=workload.googleapis.com/sqlserver.transaction_log.shrink.count: kind: expected CUMULATIVE but got GAUGE; workload.googleapis.com/sqlserver.transaction_log.growth.count: kind: expected CUMULATIVE but got GAUGE, retryable=false
```

I just reset the descriptors for those two so they should pass if you re-run it.

StefanKurek:
@qingling128 @jefferbrecht @igorpeshansky FYI I have created a 2nd PR for merging the functionality of this PR into the existing mssql application depending on which route we go with (we can just close the PR that no longer makes sense). https://github.com/GoogleCloudPlatform/ops-agent/pull/554

StefanKurek:
So I believe the tests are failing because I changed two metrics to be monotonic sums bug in the OTEL receiver they are still Gauges (I have a PR in OTEL to change this as well). Can you all verify that this would cause a problem, or is there a way for this ops agent PR to move forward before the OTEL change is made?

jefferbrecht:
> So I believe the tests are failing because I changed two metrics to be monotonic sums bug in the OTEL receiver they are still Gauges (I have a PR in OTEL to change this as well). Can you all verify that this would cause a problem, or is there a way for this ops agent PR to move forward before the OTEL change is made?

Any change between GAUGE/CUMULATIVE is breaking on the Ops Agent side, so it would cause problems if we merged this and released them as GAUGE but changed them to CUMULATIVE later -- e.g. any users who pick up a release with GAUGE would need to delete their descriptors for a new release with CUMULATIVE to work correctly.

With 2.14.0 having just been released it'll probably be at least ~1 week before the next feature release, which I would hope is enough time for the upstream OTel change to go through. Otherwise the only way forward I can think of would be to disable those metrics for now and then do a follow-up PR to re-enable them once the OTel change is merged.

StefanKurek:
@jefferbrecht would the casttosumprocessor do the trick? I added this to the pipeline here, but either it doesn't do what I was hoping it would do, or I'm using it incorrectly.

jefferbrecht:
> @jefferbrecht would the casttosumprocessor do the trick? I added this to the pipeline here, but either it doesn't do what I was hoping it would do, or I'm using it incorrectly.

Hmm, you're right, I just took a look through the logs and it looks like it's actually sending it as CUMULATIVE and it's only failing because the test project still has the old GAUGE descriptors:

```
Metric kind for metric workload.googleapis.com/sqlserver.transaction_log.shrink.count must be GAUGE, but is CUMULATIVE.;
Metric kind for metric workload.googleapis.com/sqlserver.transaction_log.growth.count must be GAUGE, but is CUMULATIVE.
```

I just reset the descriptors again so re-running the tests should pass this time.

Yeah, casttosumprocessor should be fine IMO, I hadn't thought of that.

docs/sqlserver.md
StefanKurek: I went ahead and just changed this. at this point I think all open issues can be resolved

jefferbrecht (approved):
> @jefferbrecht looks like tests passed. I'm ok with moving the focus to https://github.com/GoogleCloudPlatform/ops-agent/pull/554 if you don't think there will be any challenge from @qingling128 on it. I think you might have to reset the descriptors over there so I can run those tests again too.
> 
> Only outstanding items are whether we should have a metric name change and whether or not some descriptions should have different wording.

Sounds good. LGTM modulo resolutions to the items you mentioned.

StefanKurek:
> > @jefferbrecht looks like tests passed. I'm ok with moving the focus to #554 if you don't think there will be any challenge from @qingling128 on it. I think you might have to reset the descriptors over there so I can run those tests again too.
> > Only outstanding items are whether we should have a metric name change and whether or not some descriptions should have different wording.
> 
> Sounds good. LGTM modulo resolutions to the items you mentioned.

@jefferbrecht in case you didn't notice I made changes for every outstanding item. At the point the last thing remaining is whether to use this PR or to use https://github.com/GoogleCloudPlatform/ops-agent/pull/554 (which adds the new functionality to the existing mssql similar to the iis PR). If everyone agrees that is the route we should go, it would be great if everyone jumped over to that one and hopefully give approvals if all looks good 😄

qingling128:
Catching up on this. Yeah, the decision is to extend the existing receiver as V2. See the doc `[PUBLIC] Ops Agent: new IIS/MSSQL receivers` for details. We could probably link that doc somewhere in the IIS and MSSQL scope docs.

StefanKurek:
> Catching up on this. Yeah, the decision is to extend the existing receiver as V2. See the doc `[PUBLIC] Ops Agent: new IIS/MSSQL receivers` for details. We could probably link that doc somewhere in the IIS and MSSQL scope docs.

@qingling128 ok I will close this PR and we'll just use https://github.com/GoogleCloudPlatform/ops-agent/pull/554 from now on.

StefanKurek:
Attempting to merge https://github.com/GoogleCloudPlatform/ops-agent/pull/554 instead

PR #544, opened 2022-04-20
confgenerator: add resource_name label using modify fields.

franciscovalentecastro:
b/169784211


confgenerator/logging.go
```diff
@@ -21,12 +21,15 @@ import (
...
 
 // setLogNameComponents generates a series of components that rewrites the tag on log entries tagged `tag` to be `logName`.
-func setLogNameComponents(tag, logName, receiverType string) []fluentbit.Component {
+func setLogNameComponents(tag, logName, receiverType string, hostName string) []fluentbit.Component {
 	return LoggingProcessorModifyFields{
 		Fields: map[string]*ModifyField{
 			"logName": {
 				DefaultValue: &logName,
 			},
+			`labels."agent.googleapis.com/resource_name"`: {
```
quentinmit, line 30: I think from reading the tea leaves (https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/dce3e4b53ec4ff35e72d3c5fac1cf38428048e9d/lib/fluent/plugin/common.rb#L23 and https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/dce3e4b53ec4ff35e72d3c5fac1cf38428048e9d/lib/fluent/plugin/out_google_cloud.rb#L1126) that this should actually be

```
compute.googleapis.com/resource_name
```

to match the legacy logging agent.

confgenerator/logging.go
```diff
@@ -21,12 +21,15 @@ import (
...
+				StaticValue: &hostName,
```
quentinmit, line 31: Use `DefaultValue` so that users can override it.

confgenerator/confgenerator.go
```diff
@@ -209,7 +209,7 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
 					}
 					components = append(components, processor.Components(tag, strconv.Itoa(i))...)
 				}
-				components = append(components, setLogNameComponents(tag, rID, receiver.Type())...)
+				components = append(components, setLogNameComponents(tag, rID, receiver.Type(), hostInfo.Hostname)...)
```
quentinmit, line 212: hostname isn't necessarily the same as the instance name - the resource name has to come from the GCE metadata server

confgenerator/logging.go
franciscovalentecastro: That's right! Fixed.

confgenerator/logging.go
franciscovalentecastro: Done!

confgenerator/confgenerator.go
franciscovalentecastro: I went back to explore the initial discussion ( b/169784211#comment18 ) with @igorpeshansky and @qingling128 about this task and realized why it was ok to use `hostInfo.Hostname`. 

The [value that is used](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/dce3e4b53ec4ff35e72d3c5fac1cf38428048e9d/lib/fluent/plugin/out_google_cloud.rb#L1126) in `google-fluentd` (which this feature wants to mirror) is actually the [`Socket.gethostname`](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/lib/fluent/plugin/common.rb#L196
) which is equivalent of getting `os.Hostname` in go with `gopsutil`.

integration_test/ops_agent_test.go
```diff
@@ -679,6 +679,50 @@ func TestModifyFields(t *testing.T) {
...
+		config := fmt.Sprintf(`logging:
+  receivers:
+    f1:
+      type: files
+      include_paths:
+      - %s
+  processors:
+    json:
+      type: parse_json
+  exporters:
```
qingling128, line 698: [Optional] `exporters` are not required and can be inferred from the configuration if they are not specified.

qingling128 (approved):
LGTM

integration_test/ops_agent_test.go
quentinmit: They're not only not required, but they're completely ignored if specified.

PR #546, opened 2022-04-21
Revert back to old exporter

jefferbrecht:
We need the ability to skip CreateMetricDescriptor for system metrics only, which we can't feasibly do with the new exporter. Reverting back to the old exporter until we come up with a plan.


jefferbrecht:
Un-reverting because it turns out the previous exporter never calls CreateMetricDescriptor anyway.

We want to be in state where it _does_ get called for workload metrics, but that's a separate discussion -- as far as breaking changes are concerned, we're fine for now with the new exporter.

PR #560, opened 2022-04-27
confgenerator: add stringer implementation for UnifiedConfig

ridwanmsharif:
[b/235585629](http://b/235585629) | P3 | Confgenerator serializes and de-serializes the config unnecessarily

- Add stringer implementation to the unified config
- Removes the need for the extraneous serialization and deserialization
- Cleans up some dead code
- Makes it so that an empty list for include_paths is a valid use case


sophieyfang:
Centos 7 regex shall be activemq:
image

confgenerator/config.go
```diff
@@ -64,6 +64,15 @@ func (uc *UnifiedConfig) DeepCopy(platform string) (UnifiedConfig, error) {
 	return fromYaml, nil
 }
 
+func (uc *UnifiedConfig) String() string {
+	marshalledConfig, err := yaml.Marshal(uc)
+	if err != nil {
+		return fmt.Sprintf("failed to convert Unified config to yaml %v", err)
```
braydonk, line 70:
```suggestion
		return fmt.Sprintf("failed to convert Unified config to yaml: %v", err)
```

confgenerator/files.go
```diff
@@ -35,6 +35,10 @@ func ReadUnifiedConfigFromFile(path, platform string) (UnifiedConfig, error) {
 	if err != nil {
 		return uc, err
 	}
+	if err = uc.Validate(platform); err != nil {
```
braydonk, line 38: Doesn't change anything functionally, but conventionally statements like this will have an `err` just for this scope.
```suggestion
	if err := uc.Validate(platform); err != nil {
```

confgenerator/confgenerator_test.go
```diff
@@ -172,18 +172,18 @@ func generateConfigs(platform platformConfig, testDir string) (got map[string]st
...
 		apps.BuiltInConfStructs,
 	)
 	if err != nil {
 		return
 	}
-	got[builtinConfigFileName] = string(builtInConfBytes)
+	got[builtinConfigFileName] = apps.BuiltInConfStructs[platform.OS].String()
 
-	uc, err := confgenerator.ParseUnifiedConfigAndValidate(confBytes, platform.OS)
+	err = uc.Validate(platform.OS)
```
braydonk, line 186: Every time `Validate` is called, or really any of the `UnifiedConfig` methods, we have to constantly pass in the platform. Maybe it should just be a private field?

Also maybe an aside, but while looking at this I noticed `uc.Logging.Validate` takes in a platform but doesn't use it. Also, I don't know why we have a `uc.ValidateMetrics` but also `uc.Logging.Validate` I don't understand why that API is different. Might be out of scope for this PR though.

confgenerator/confgenerator_test.go
```diff
@@ -172,18 +172,18 @@ func generateConfigs(platform platformConfig, testDir string) (got map[string]st
...
+	uc, err := confgenerator.MergeConfFiles(
 		filepath.Join("testdata", testDir, inputFileName),
 		platform.OS,
 		apps.BuiltInConfStructs,
 	)
 	if err != nil {
 		return
 	}
-	got[builtinConfigFileName] = string(builtInConfBytes)
+	got[builtinConfigFileName] = apps.BuiltInConfStructs[platform.OS].String()
```
braydonk, line 184: Maybe built in config could be a field on the Unified Config instead of always needing to look up from `apps`? That or `apps` could have a public function that gets the built in configuration for that platform.

confgenerator/testdata/invalid/linux/logging-receiver_files_type_invalid_parameter_include_paths_is_empty_list/golden/error
```diff
@@ -1,9 +0,0 @@
-[3:15] "include_paths" is a required field
```
braydonk, line 1: Why is this test valid now?

cmd/ops_agent_windows/run_windows.go
```diff
@@ -125,18 +125,18 @@ func (s *service) checkForStandaloneAgents(unified *confgenerator.UnifiedConfig)
...
+	uc, err := confgenerator.MergeConfFiles(s.userConf, "windows", apps.BuiltInConfStructs)
 	if err != nil {
 		return err
 	}
 
-	s.log.Info(1, fmt.Sprintf("Built-in config:\n%s", builtInConfig))
-	s.log.Info(1, fmt.Sprintf("Merged config:\n%s", mergedConfig))
-	uc, err := confgenerator.ParseUnifiedConfigAndValidate(mergedConfig, "windows")
+	s.log.Info(1, fmt.Sprintf("Built-in config:\n%s", apps.BuiltInConfStructs["windows"]))
+	s.log.Info(1, fmt.Sprintf("Merged config:\n%s", uc))
```
jefferbrecht, line 134: This is a good opportunity to refactor these magic numbers out to a named constant.

cmd/google_cloud_ops_agent_diagnostics/main.go
```diff
@@ -22,17 +22,17 @@ import (
...
 
 func getUnifiedConfigAndValidate(userConfPath, platform string) (confgenerator.UnifiedConfig, error) {
-	_, mergedConfig, err := confgenerator.MergeConfFiles(userConfPath, platform, apps.BuiltInConfStructs)
+	uc, err := confgenerator.MergeConfFiles(userConfPath, platform, apps.BuiltInConfStructs)
 	if err != nil {
 		return confgenerator.UnifiedConfig{}, err
 	}
 
-	uc, err := confgenerator.ParseUnifiedConfigAndValidate(mergedConfig, platform)
+	err = uc.Validate(platform)
```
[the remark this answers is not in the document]
ridwanmsharif: So I actually think the behaviour now is strictly better than before. I didn't see any good reason why the empty list for `include_paths` was ever prohibited, it doesn't break anything. Our validation on the field says it is `required` to be valid, not that it must be non-empty. If we do want that as well, we should add that as a validator and not rely on marshalling with `omitempty` and then unmarshalling again to validate this one validation rule.

Also the behaviour before was kinda broken. It would surface errors on the merged config and give yaml line numbers and show the error on the merged config. Which is never visible to the users anyway, so the errors were less than ideal too. The merged config should be an internal thing, and users should only interact with the user specified config.

Let me know if you think I should add a validator to ensure the old behaviour regarding us requiring non-empty `include_paths` (which we should also document) or whether we should loosed the requirements as we do here.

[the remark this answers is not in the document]
jefferbrecht: YAML and validation are orthogonal. That we inline validation with unmarshalling is a convenience and not a requirement: it is absolutely possible to run struct validation entirely on its own without having to marshal/unmarshal anything to/from YAML.

The marshal/unmarshal round-trip should be a no-op. YAML is only an intermediate representation and should be state-preserving. The YAML tags for `include_paths` allow valid things to become invalid after a round-trip test, therefore `omitempty` was the wrong tag to use. If the original intent was that `include_paths` must not be empty then that should have been encoded entirely within the `validate` tag, e.g. `min=1` instead of `required`.

What is the use case for `include_paths` being empty? "That it doesn't break things" isn't the bar we should meet; the bar should at least be "we don't want the user to do the wrong thing", and an empty `include_paths` is clearly the wrong thing IMO, and should be disallowed.

Regarding where to put the struct-level validation: I would go the other way and call `uc.Validate` from within `MergeConfFiles`, which has the additional benefit of simplifying all of the call sites. I can't think of a reason why the intermediate state after-merge-but-before-validate would be useful so it should be OK to combine those steps.

For the missing struct-level validation, I would still be OK with assuming that valid inputs always produce a valid output during `MergeConfFiles`, which implies that struct-level validation should be performed inside there and issue a `panic` on failure.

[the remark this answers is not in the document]
jefferbrecht: Looks good 👍 We might also be able to get away with moving `original.Validate` to the panic as well, since `original` is just the built-in struct rather than something that's user-provided.

cmd/google_cloud_ops_agent_engine/main.go
```diff
@@ -40,24 +39,16 @@ func main() {
...
 	if err != nil {
 		return err
 	}
 
 	// Log the built-in and merged config files to STDOUT. These are then written
 	// by journald to var/log/syslog and so to Cloud Logging once the ops-agent is
 	// running.
-	log.Printf("Built-in config:\n%s", builtInConfig)
-	log.Printf("Merged config:\n%s", mergedConfig)
+	log.Printf("Built-in config:\n%s\n", apps.BuiltInConfStructs["linux"])
```
braydonk, line 50: Sorry, I was wrong; `fmt.Printf` doesn't add the new lines but the `log` package actually prints everything on a new line regardless. Sorry!

confgenerator/files.go
```diff
@@ -24,18 +24,32 @@ import (
...
-func ReadUnifiedConfigFromFile(path, platform string) (UnifiedConfig, error) {
-	uc := UnifiedConfig{}
+// ReadUnifiedConfigFromFile reads the user config file and returns a UnifiedConfig.
+// If the user config file does not exist, it returns nil.
+func ReadUnifiedConfigFromFile(path, platform string) (*UnifiedConfig, error) {
+	if _, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			// If the user config file does not exist, we don't want any overrides.
+			return nil, nil
+		} else {
```
braydonk, line 34: this else isn't necessary

cmd/google_cloud_ops_agent_diagnostics/main.go
```diff
@@ -21,21 +21,20 @@ import (
...
-		return confgenerator.UnifiedConfig{}, confgenerator.UnifiedConfig{}, err
+	if userUc == nil {
+		userUc = &confgenerator.UnifiedConfig{}
 	}
 
-	mergedUc, err = confgenerator.ParseUnifiedConfigAndValidate(mergedConfig, platform)
+	mergedUc, err := confgenerator.MergeConfFiles(userConfPath, platform, apps.BuiltInConfStructs)
 	if err != nil {
-		return confgenerator.UnifiedConfig{}, confgenerator.UnifiedConfig{}, err
+		return &confgenerator.UnifiedConfig{}, &confgenerator.UnifiedConfig{}, err
```
braydonk, line 37: here as well

confgenerator/files.go
```diff
@@ -24,17 +24,30 @@ import (
...
+			// If the user config file does not exist, we don't want any overrides.
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to retrieve the user config file %q: %w \n", path, err)
+	}
 
+	uc := &UnifiedConfig{}
 	data, err := ioutil.ReadFile(path)
 	if err != nil {
 		return uc, err
```
braydonk, line 41: If you just return nil here then you don't need to instantiate on line 38

confgenerator/files.go
```diff
@@ -24,17 +24,30 @@ import (
...
 	}
-	uc, err = UnmarshalYamlToUnifiedConfig(data, platform)
+	*uc, err = UnmarshalYamlToUnifiedConfig(data, platform)
```
braydonk, line 43: UnmarshalYamlToUnifiedConfig really should return a reference, should we do that in this PR?

ridwanmsharif: Might as well. Done.

ridwanmsharif:
I wonder if I'll ever see any of my PRs have only green checks. Or will the windows and sles gods always cast a shadow on me. Investigating the failures now. Hoping they are unrelated.

ridwanmsharif:
Okay the test failures do appear to be unrelated and flaky. Don't think they are blocking the PR.

braydonk (approved):

PR #564, opened 2022-04-27
Fix Varnish Metadata

sophieyfang:
Introduced from https://github.com/GoogleCloudPlatform/ops-agent/pull/512


sophieyfang:
This breaks our doc generation script, we need to find a way to enforce if expected_logs|metrics is defined there must be at least one metrics or logs below

sophieyfang:
> This breaks our doc generation script, we need to find a way to enforce if expected_logs|metrics is defined there must be at least one metrics or logs below

Actually what broke the doc generation script is the missing configuration_options for the metrics..

PR #565, opened 2022-04-28
metadata: add configuration_options for varnish and add validation

ridwanmsharif:
This change adds in the configuration option for Varnish and adds validation enforcement for new receivers.

### DIFF
--- a/integration_test/third_party_apps_data/applications/varnish/metadata.yaml
+++ b/integration_test/third_party_apps_data/applications/varnish/metadata.yaml
@@ -64,3 +64,20 @@ expected_metrics:
   labels:
     operation: .*
 expected_logs:
+configuration_options:
+  metrics:
+  - type: varnish
+    fields:
+    - name: type
+      default: null
+      description: This value must be varnish.
+    - name: cache_dir
+      default: null
+      description: This specifies the cache dir instance name to use when collecting metrics. If not specified, this will default to the host. 
+    - name: exec_dir
+      default: null
+      description: The directory where the varnishadm and varnishstat executables are located. If not provided, will default to relying on the executables being in the user's PATH.
+    - name: collection_interval
+      default: 60s
+      description: A time.Duration value, such as 30s or 5m.
+
--- a/integration_test/third_party_apps_data/applications/varnish/metadata.yaml
+++ b/integration_test/third_party_apps_data/applications/varnish/metadata.yaml
@@ -1,5 +1,15 @@
 short_name: varnish
 long_name: Varnish HTTP Cache
+description: |-
+  The Varnish integration collects cache and session metrics.
+  It monitors the number of objects entering and exiting the cache,
+  as well as the number of sessions and backend connections.
+  The integration also collects Varnish access logs and parses them
+  into a standardized json payload.
+
+  For more information about Varnish, see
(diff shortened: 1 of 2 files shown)


integration_test/third_party_apps_data/applications/varnish/metadata.yaml
```diff
@@ -1,5 +1,6 @@
 short_name: varnish
 long_name: Varnish HTTP Cache
+supported_app_version: ["6.0+", "7.0+"]
```
ridwanmsharif, line 3: @JonathanWamsley Could you confirm that this is the correct supported version for Varnish?

integration_test/third_party_apps_test.go
```diff
@@ -379,6 +380,12 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
 		logger.ToMainLog().Printf("Parsed metadata.yaml: %+v", metadata)
 	}
 
+	// Validate the metadata.
+	validate := validator.New()
+	if err = validate.Struct(metadata); err != nil {
+		return nonRetryable, err
+	}
+
```
sophieyfang, line 388: I am also working on this :) It needs to do quite much to get it work in the we intended it to be.😮‍💨
metadata parsed by unmarshallstrict will allocate any substruct as {} instead of nill thus the required tag won't work. I will send a PR soon. For now you can restore this file back, i will handle the work in my PR :)

integration_test/third_party_apps_data/applications/varnish/metadata.yaml
JonathanWamsley: We don't use + but rather x to indicate all the versions in the series. So `["6.x", "7.0.x"]` like in the [scope](https://docs.google.com/document/d/1WyQ42YPsRodVKbk1_kvEBTZT7RDvzhXzHtlgEWOxaE8/) doc. This is is what other receivers like [wildfly](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/integration_test/third_party_apps_data/applications/wildfly/metadata.yaml#L32) does too.

integration_test/third_party_apps_test.go
sophieyfang: Here is the PR: https://github.com/GoogleCloudPlatform/ops-agent/pull/567

integration_test/third_party_apps_data/applications/varnish/metadata.yaml
```diff
@@ -1,5 +1,6 @@
 short_name: varnish
 long_name: Varnish HTTP Cache
```
sophieyfang, line 2: It also needs Description from the scope doc, it will start failing the test once https://github.com/GoogleCloudPlatform/ops-agent/pull/567 is in

```
The Varnish integration collects cache and session metrics. It monitors the number of objects entering and exiting the cache, as well as the number of sessions and backend connections. The integration also collects Varnish access logs and parses them into a standardized json payload.
```

ridwanmsharif: Added

integration_test/third_party_apps_data/applications/varnish/metadata.yaml
ridwanmsharif: Done.

PR #566, opened 2022-04-28
Fix for test failures on SLES12

martijnvans:
The errors were like:

```
Problem: python36-3.6.8-2.13.x86_64 requires python36-base = 3.6.8, but this requirement cannot be provided
 uninstallable providers: python36-base-3.6.8-2.13.x86_64[SUSE_Linux_Enterprise_Server_x86_64:SLES12-SP5-Pool]
```

I also tried this as an alternate fix, but it doesn't work on SLES 15:

```
sudo zypper --non-interactive install python36 python3-certifi
```


igorpeshansky (approved):
LGTM :shipit:

> I also tried this as an alternate fix, but it doesn't work on SLES 15:
> ```
> sudo zypper --non-interactive install python36 python3-certifi
> ```

Right, because on SLES 15, there is no `python36` package, just `python3` (which is at version 3.6, as I recall).

PR #568, opened 2022-04-28
Fix LuaAccessor when write=false.

franciscovalentecastro:


confgenerator/filter/internal/ast/ast.go
```diff
@@ -148,7 +148,8 @@ then
 			fmt.Fprintf(&out, `record[%s] = {}
 `, p)
 		} else {
-			fmt.Fprintf(&out, `return nil`)
+			fmt.Fprintf(&out, `return nil
```
quentinmit, line 151: Nit: I think this would be slightly clearer as `"return nil\n"`

quentinmit (approved):

confgenerator/filter/internal/ast/ast.go
franciscovalentecastro: Fixed.

quentinmit (approved):
New unit test LGTM

franciscovalentecastro:
Third party apps integration test (Windows) are failing due to an unrelated issue.
Screen Shot 2022-04-29 at 15 30 35

PR #573, opened 2022-05-02
Add Jetty logging

Mrod1598:
Add a system log processor and access log receiver.

System Log
image

Access Log
image


ridwanmsharif (commented):
Overall the main point I'm concerned about is maintaining the logging processor for Jetty forever. If the logs are already collected by the `syslog` receiver (admittedly unstructured), we should move forward without adding a processor for now. Eventually we will have a way to sctructure all syslog messages and update their lognames, and when that effort lands - this application, along with Aerospike, activemq and memcached will benefit.

apps/jetty.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
+)
+
+type LoggingProcessorJettySystem struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+}
+
+func (LoggingProcessorJettySystem) Type() string {
```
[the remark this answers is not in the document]
Mrod1598: They are collected by default for syslog.

Based on that effort, I've just removed the system logs to be revisited at a later date. I noted in the docs that system logs are collected by default.

integration_test/third_party_apps_data/applications/jetty/metadata.yaml
```diff
@@ -0,0 +1,61 @@
...
+  access logs. Access logs are parsed into a JSON payload
+  focused on request details. 
+
+  For more information about Jetty, see
+  [The Jetty Docs](https://www.eclipse.org/jetty/documentation/jetty-11/operations-guide/index.html).
+minimum_supported_agent_version:
+  metrics: 2.9.0
+  logging: 2.9.0
+supported_app_version: ["11.x", "10.x", "9.4.x"]
+expected_logs:
```
ridwanmsharif, line 15: Can we also add the expected log for the `jetty_system` logs? See https://github.com/GoogleCloudPlatform/ops-agent/blob/230384c4fdd2a502439c6b8f172381e07a13079a/integration_test/third_party_apps_data/applications/activemq/metadata.yaml#L90 for an example for when an application sends logs to `syslog`

ridwanmsharif (approved):
LGTM

ridwanmsharif:
Looks like the integrations tests are failing for Jetty. Lets fix that and then we can merge

ridwanmsharif (approved):

integration_test/third_party_apps_data/applications/jetty/metadata.yaml
```diff
@@ -0,0 +1,68 @@
...
+    description: ''
+  - name: httpRequest.status
+    type: string
+    description: ''
+- log_name: syslog
+  fields:
+  - name: jsonPayload.message
+    value_regex: '.*oejs\.Server:main: Started Server\@.*\{STARTING\}.*'
+    type: string
+    description: null
```
ridwanmsharif, line 51: All the other descriptions default to `''` instead of `null`. We should do that here too

ridwanmsharif:
Uggggh have to rebase again and will merge then (feel free to merge if you can)

integration_test/third_party_apps_data/applications/jetty/metadata.yaml
Mrod1598: Gotcha, updated.

integration_test/third_party_apps_data/applications/jetty/metadata.yaml
```diff
@@ -0,0 +1,68 @@
...
+long_name: Jetty
+description: |-
+  The Jetty integration collects
+  access logs. Access logs are parsed into a JSON payload
+  focused on request details. 
+
+  For more information about Jetty, see
+  [The Jetty Docs](https://www.eclipse.org/jetty/documentation/jetty-11/operations-guide/index.html).
+minimum_supported_agent_version:
+  metrics: 2.9.0
```
sophieyfang, line 12: This shouldn't be set up. Our release job will assign the actual release version when we trigger our ops-agent release.

sophieyfang: minimum_supported_agent_version for new integration or existing integration with newly added logging/metrics support doesn't need to be manually set.
Our release job will take care of.

Mrod1598: Got it. Would you like me to make a quick PR to fix it?

PR #581, opened 2022-05-04
Backfill 3rd party apps logo path

sophieyfang:


qingling128 (approved):
LGTM

Per sync offline, while not ideal, this is the best we could do given the limitations with Cloud documentation site image rendering policies and the potential complexity we need to introduce if moving this to a separate internal pipeline.

Nice to have: It would be great if we can somehow make it clear that these fields are not required and will be filled in later in the process.

PR #600, opened 2022-05-10
Gracefully handle previous version install failure in upgrade test.

igorpeshansky:
Before the initial release on a given platform, this test is effectively guaranteed to fail, because there is no stable package to install yet. So we skip this test if the pre-upgrade version fails to install. Since `setupOpsAgentFrom` will retry a few times, the the test should only be skipped if all retries fail consistently.


alison0211 (approved):

integration_test/ops_agent_test.go
```diff
@@ -1425,6 +1425,10 @@ func TestUpgradeOpsAgent(t *testing.T) {
 		// a default value of "", which means stable.
 		firstVersion := packageLocation{repoSuffix: os.Getenv("REPO_SUFFIX_PREVIOUS")}
 		if err := setupOpsAgentFrom(ctx, logger, vm, "", firstVersion); err != nil {
+			// Installation from stable may fail before the first release.
+			if firstVersion.repoSuffix == "" && (strings.HasPrefix(err.Error(), "installOpsAgent() failed to run googet") || strings.HasPrefix(err.Error(), "installOpsAgent() error running repo script")) {
+				return
```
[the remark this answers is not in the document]
igorpeshansky: Actually, I should've used `t.Skip()` anyway, which would also log. Done.

Subbarker (approved):
LGTM

integration_test/ops_agent_test.go
```diff
@@ -1425,6 +1425,10 @@ func TestUpgradeOpsAgent(t *testing.T) {
 		// a default value of "", which means stable.
 		firstVersion := packageLocation{repoSuffix: os.Getenv("REPO_SUFFIX_PREVIOUS")}
 		if err := setupOpsAgentFrom(ctx, logger, vm, "", firstVersion); err != nil {
+			// Installation from stable may fail before the first release.
+			if firstVersion.repoSuffix == "" && (strings.HasPrefix(err.Error(), "installOpsAgent() failed to run googet") || strings.HasPrefix(err.Error(), "installOpsAgent() error running repo script")) {
+				t.Skip(fmt.Printf("Installing stable agent failed with error %v; assuming first release.", err))
```
martijnvans, line 1430: there's a t.Skipf function that does the fmt stuff for you.

martijnvans (approved):

integration_test/ops_agent_test.go
igorpeshansky: TIL. Done.

PR #611, opened 2022-05-13
Testing: Use permalink for zookeeper download

martijnvans:
This kind of link will (hopefully) not disappear on us as often, breaking our testing setup.


martijnvans:
Integration test failure looks like an unrelated flake: fluent-bit for some reason doesn't shut down within the 30 seconds that we give it, leading to the test failure. Not sure if this is a known issue (fluent-bit being slow to shut down).

igorpeshansky (approved):
LGTM :shipit:

Just have to remember to bump it periodically…

PR #623, opened 2022-05-20
Testing: add support for SAP HANA image and app

martijnvans:
This PR does everything we need to start supporting a custom image with SAP HANA and an as-yet undeveloped app called 'saphana'. Ideally gce_testing.go wouldn't need to know about this custom image, but I don't see a way around it.

The most complex part of the PR is to set up the skipping logic. We want to introduce a new SAP HANA specific platform, and we want to:
1.  only run the app 'saphana' on that platform, and
2. for that platform, skip running the other apps. (this decision is optional, technically we could try to run e.g. nginx tests on this HANA image but that would be weird)


Subbarker:
I'm wondering if we can do this in a less hard-coded way. What would it look like if we had to support several custom images?

integration_test/third_party_apps_test.go
```diff
@@ -565,7 +578,18 @@ func TestThirdPartyApps(t *testing.T) {
 			for attempt := 1; attempt <= 4; attempt++ {
 				logger := gce.SetupLogger(t)
 				logger.ToMainLog().Println("Calling SetupVM(). For details, see VM_initialization.txt.")
-				vm := gce.SetupVM(ctx, t, logger.ToFile("VM_initialization.txt"), gce.VMOptions{Platform: tc.platform, MachineType: agents.RecommendedMachineType(tc.platform)})
+				options := gce.VMOptions{
```
jkschulz, line 581: mega-nit: does this need a space?

```suggestion
				options := gce.VMOptions {
```

jkschulz (approved):

integration_test/third_party_apps_test.go
martijnvans: it definitely doesn't need one in order to compile. I think it's a stylistic choice; i can't find any official rule about this. But I think people tend to write structs without the extra whitespace.

martijnvans:
corey gave this a green light in an offline conversation.

PR #626, opened 2022-05-24
confgenerator: disable retry_on_failure by default

braydonk:
Based on discussions with the OpenTelemetry team at Google, we want this
config disabled by default due to a Monitoring API behaviour that
conflicts with the default retry logic in the Google Cloud Exporter.

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -71,6 +71,9 @@ func (uc *UnifiedConfig) GenerateOtelConfig(hostInfo *host.InfoStat) (string, er
 		Exporter: otel.Component{
 			Type: "googlecloud",
 			Config: map[string]interface{}{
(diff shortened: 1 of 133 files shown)


confgenerator/confgenerator.go
```diff
@@ -71,6 +71,9 @@ func (uc *UnifiedConfig) GenerateOtelConfig(hostInfo *host.InfoStat) (string, er
 		Exporter: otel.Component{
 			Type: "googlecloud",
 			Config: map[string]interface{}{
+				"retry_on_failure": map[string]interface{}{
```
davidbtucker, line 74: Can we add a comment here linking to the bug which has more context?

davidbtucker (approved):

davidbtucker (approved):

PR #627, opened 2022-05-24
Populate `logging.googleapis.com/instrumentation_source` for app integrations

quentinmit:
b/223614097


integration_test/third_party_apps_test.go
```diff
@@ -186,13 +186,20 @@ func installAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.
...
 	var parts []string
 	for _, field := range fields {
 		if field.ValueRegex != "" {
 			parts = append(parts, fmt.Sprintf(`%s=~"%s"`, field.Name, field.ValueRegex))
 		}
 	}
+
+	if logName != "syslog" {
+		// verify instrumentation_source label
+		val := fmt.Sprintf("agent.googleapis.com/%s", logName)
```
quentinmit, line 199: Is the log name always guaranteed to equal the receiver type in these tests?

qingling128: Yes

qingling128: Discussed offline. The options we considered

Option 1: Stick with what the PR is doing right now. No additional manual work to add information into metadata.yaml for each application. Always compare the value of `logging.googleapis.com/instrumentation_source` to `agent.googleapis.com/<receiver_type>` and retrieve the receiver_type from the `log_name` field in metadata.yaml.

Option 2: Have each application add a block like below. No change to the test framework
```
expected_logs:
- log_name: apache_access
  fields:
  - name: labels."logging.googleapis.com/instrumentation_source"
    value_regex: agent.googleapis.com/apache_access
    type: string
```

Option 3: Have each application add a block like below. Change the test framework to get the expected value from the receiver_type  field of the metadata.yaml file
```
expected_logs:
- log_name: apache_access
- receiver_type: apache_access
```

Decision: Option 1. Users could potentially chain a 3rd party app receiver with a modify_fields processor to set `log_name` to a different value from `receiver_type`, but that is not what our tests aim to cover. So it's fine to go with option 1 and save some manual toil for developers to add per application metadata.

confgenerator/logging_modify_fields.go
```diff
@@ -36,14 +36,15 @@ type ModifyField struct {
 	omitVar string `yaml:"-"`
 
 	// Operations to perform
-	MapValues map[string]string `yaml:"map_values"`
-	Type      string            `yaml:"type" validate:"omitempty,oneof=integer float"`
-	OmitIf    string            `yaml:"omit_if" validate:"omitempty,filter"`
+	MapValues          map[string]string `yaml:"map_values"`
+	MapValuesExclusive bool              `yaml:"map_values_exclusive" validate:"excluded_without=MapValues"`
```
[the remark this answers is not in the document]
qingling128: Filed related feature and doc trackers:
* b/235508077 | modify_fields: Support unsetting the destination field if the source value does not match any keys in map_values
* b/235510044 | Public documentation for the new map_values_exclusive feature in modify_fields processor

Resolving this thread

PR #631, opened 2022-05-26
Set FluentBit self logs to buffer in memory instead of filesystem

avilevy18:
Parameterized `storage.type` config with variable `BufferInMemory`. Edit component to always use in memory buffering for self logs

Following: http://b/226668416, this PR changes self logs to buffer in memory as opposed to the filesystem.

Soak Test:
image


qingling128 (approved):
LGTM

qingling128:
SLES 12 was failing, but it's not related to this PR.

confgenerator/logging_receivers.go
```diff
@@ -57,6 +57,7 @@ type LoggingReceiverFilesMixin struct {
 	ExcludePaths            []string        `yaml:"exclude_paths,omitempty"`
 	WildcardRefreshInterval *time.Duration  `yaml:"wildcard_refresh_interval,omitempty" validate:"omitempty,min=1s,multipleof_time=1s"`
 	MultilineRules          []MultilineRule `yaml:"-"`
+	BufferInMemory          bool
```
quentinmit, line 60: Mark it as `yaml:"-"` so it can't be set by user config.

quentinmit (commented):
Is there a doc somewhere I can read on why we want to do this?

(Also, presumably fluent-bit is still writing state somewhere on disk so it doesn't reprocess the same log lines, right?)

confgenerator/logging_receivers.go
avilevy18: Changed in: [f67580f](https://github.com/GoogleCloudPlatform/ops-agent/pull/631/commits/f67580f4f76efa0303dc9986a7867855b006bd27)

PR #633, opened 2022-05-26
Turned off data integrity check

avilevy18:
Turned off data integrity check when writing and reading logs from the filesystem for FluentBit

http://b/233109163


confgenerator/fluentbit/service.go
```diff
@@ -36,8 +36,8 @@ func (s Service) Component() Component {
 			// https://docs.fluentbit.io/manual/administration/buffering-and-storage#service-section-configuration
 			// storage.path is set by Fluent Bit systemd unit (e.g. /var/lib/google-cloud-ops-agent/fluent-bit/buffers).
 			"storage.sync": "normal",
-			// Enable the data integrity check when writing and reading data from the filesystem.
-			"storage.checksum": "on",
+			// Turned off due to empty spaces being considered as corrupted. Most users do not enforce checksum. See:b/233109163 for more info
```
igorpeshansky, line 39: How about:
```suggestion
			// Disable data integrity check to avoid spurious corruption errors due to empty spaces.
			//  Most users do not enforce checksum.
```
I don't believe there's a need to mention the bug in the code — `git blame` can trace it to this PR, which already references it.

igorpeshansky (commented):
One suggested change.

confgenerator/fluentbit/service.go
avilevy18: Changed in: [adf27bc](https://github.com/GoogleCloudPlatform/ops-agent/pull/633/commits/adf27bc0caf9fb8e5795223c563f4f49eb24419c)

igorpeshansky: Note: `Disable`, not `Disabled` (because it describes what the option does, not what this PR did)…

avilevy18: Thanks for the feedback.

Changed in: [d177711](https://github.com/GoogleCloudPlatform/ops-agent/pull/633/commits/d17771141c3235bf6630d9ee1d2407b01d18750a)

qingling128 (approved):
LGTM

PR #643, opened 2022-06-02
Add Jetty minimum supported metrics version

EvanSimpson:
Adds missing field in jetty metadata file.


integration_test/third_party_apps_data/applications/jetty/metadata.yaml
```diff
@@ -8,6 +8,7 @@ description: |-
   integration also collects access and system logs. Access logs are parsed
   into a JSON payload focused on request details.
 minimum_supported_agent_version:
+  metrics: 2.17.0
```
sophieyfang, line 11: We shouldn't manually add this version manually. It should be taken care by the post ops agent release job. Any reason on adding this before we actually released 2.17.0 today..? 🤔

PR #644, opened 2022-06-02
Added validation tests for metadata.yaml

avilevy18:
Created common_test.go. Developed system for declaring expected errors from test yaml files and verifies expected outcome with actual errors from validation framework.

http://b/230789204

### DIFF
--- a/integration_test/common/common_test.go
+++ b/integration_test/common/common_test.go
@@ -17,8 +17,10 @@
 package common
 
 import (
+	"flag"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
 	"strings"
 	"testing"
@@ -28,30 +30,23 @@ import (
 )
 
 const (
-	//relative path to testdata folder
-	testdataDir   = "testdata"
-	inputYamlName = "input.yaml"
+	// relative path to testdata folder
+	testdataDir     = "testdata"
+	inputYamlName   = "input.yaml"
+	goldenErrorName = "golden_error"
 )
 
-type testCase struct {
-	dirName        string
-	expectedErrors map[fieldError]struct{}
-}
+var (
+	updateGolden = flag.Bool("update_golden", false, "Whether to update the expected golden confs if they differ from the actual generated confs.")
+)
 
-type fieldError struct {
-	field string
-	tag   string
-}
-
-func getInputYamlBytes(t *testing.T, dirName string) []byte {
-	yamlFilePath := path.Join(testdataDir, dirName, inputYamlName)
-	contents, err := ioutil.ReadFile(yamlFilePath)
+func getTestFile(t *testing.T, dirName, fileName string) string {
+	filePath := path.Join(testdataDir, dirName, fileName)
+	contents, err := ioutil.ReadFile(filePath)
 	if err != nil {
-		t.Fatal("could not read dirName: " + yamlFilePath)
+		t.Fatal("could not read dirName: " + filePath)
 	}
-	escapedYamlStr := strings.ReplaceAll(string(contents), "\r\n", "\n")
-
-	return []byte(escapedYamlStr)
+	return strings.ReplaceAll(string(contents), "\r\n", "\n")
 }
 
 func UnmarshallAndValidate(t *testing.T, bytes []byte, i interface{}) error {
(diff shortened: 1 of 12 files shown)


integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+	dirName        string
+	expectedErrors map[fieldError]struct{}
+}
+
+type fieldError struct {
+	field string
+	tag   string
+}
+
+func getYaml(t *testing.T, dirName string) []byte {
```
braydonk, line 46: `getInputYamlBytes` would be a more descriptive name, or `getTestCaseYamlBytes` (probably the first one)

integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+	dir := path.Join(testdataDir, dirName, inputYamlName)
```
braydonk, line 47: `yamlFilePath` maybe? Cause this is a file not a directory

integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+	contents, err := ioutil.ReadFile(dir)
+	if err != nil {
+		t.Fatal("could not read dirName: " + dir)
+	}
+	str := strings.ReplaceAll(string(contents), "\r\n", "\n")
```
braydonk, line 52: `escapedYamlStr`

integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	return v.Struct(i)
+}
+
+func TestAll(t *testing.T) {
+	table := []testCase{
+		{
+			"pass",
```
braydonk, line 69: I think we should put the struct field names here i.e. `dirName: "pass"` but for every field in this slice

integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+			"integration-metadata_unique_supported-app-version",
+			map[fieldError]struct{}{
+				{"SupportedAppVersion", "unique"}: {},
+			},
+		},
+	}
+
+	for _, test := range table {
+		t.Run(test.dirName, func(t *testing.T) {
+			testDir(t, test)
```
braydonk, line 114: I would suggest adding a comment here that we put this in another function so we could `t.Parallel()` the test cases

integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+		})
+	}
+}
+
+func testDir(t *testing.T, test testCase) {
```
braydonk, line 119: Probably needs a more descriptive name, `testMetadataValidation` because this takes in a whole test case not just a directory

integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+	t.Parallel()
+	fmt.Println("test.dirName: " + test.dirName)
+	bytes := getYaml(t, test.dirName)
+	actualErrors := UnmarshallAndValidate(t, bytes, &IntegrationMetadata{})
+
+	//testdata/pass
+	if test.expectedErrors == nil {
+		if actualErrors == nil {
+			return
+		} else {
```
braydonk, line 129: this else statement is superfluous

integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+
+		for i := range actualErrorsMap {
+			delete(expectedErrorsMap, i)
+		}
+
+		if len(expectedErrorsMap) > 0 {
+			t.Fatal(fmt.Sprintf("Unexcpected errors detected: \n Expected error: %v but got %v", test.expectedErrors, actualErrorsMap))
+		}
+	} else {
```
braydonk, line 153: I think it would be better to do this check first, `if actualErrors == nil` then fatal with expected to fail instead of an if else

integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	}
+
+	if actualErrors != nil {
+		fieldErrors := actualErrors.(validator.ValidationErrors)
+		if len(fieldErrors) != len(test.expectedErrors) {
+			t.Fatal("Expecting validation errors to equal expected errors")
+		}
+		expectedErrorsMap := test.expectedErrors
+		actualErrorsMap := map[fieldError]struct{}{}
```
braydonk, line 140: Not a huge deal, but I don't think this needs to be a set. Probably a bit more readable as a simple slice

integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+		for _, err := range fieldErrors {
```
braydonk, line 141: renaming this to `fieldErr` would probably be a good idea; `err` is a commonly understood name for go variables to represent something that actually failed, so could spook people at a glance

integration_test/common/common_test.go
```diff
@@ -0,0 +1,156 @@
...
+	if test.expectedErrors == nil {
+		if actualErrors == nil {
+			return
+		} else {
+			t.Fatal("Expecting no errors")
+		}
+	}
+
+	if actualErrors != nil {
+		fieldErrors := actualErrors.(validator.ValidationErrors)
```
braydonk, line 135: Should do `fieldErrors, ok := <cast>` and check that the cast actually worked first

avilevy18: Changed in: [47fc576](https://github.com/GoogleCloudPlatform/ops-agent/pull/644/commits/47fc5762328e00791b7d5467846eca6db23e67a6)

integration_test/common/common_test.go
```diff
@@ -0,0 +1,163 @@
...
+	v := validator.New()
+	err := yaml.UnmarshalStrict(bytes, i)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return v.Struct(i)
+}
+
+func TestAll(t *testing.T) {
+	table := []testCase{
```
qingling128, line 67: Have we considered using a `golden_error` file in the same directory? Similar to https://github.com/GoogleCloudPlatform/ops-agent/tree/master/confgenerator/testdata/invalid/linux/logging-processor_exclude_logs_invalid_field

The potential benefit is to skip the step to update this list. "invalid case" and "errors" will be easily found together. And if users forgets to update this list, the test will still executes. With the current approach, if someone adds an invalid `input.yaml` but forget or is not aware that they need to update it here, the tests might be skipped by accident.

There might be downside I'm not seeing though.

qingling128 (approved):
LGTM

PR #632, opened 2022-05-26
Add Flink logging receiver

JonathanWamsley:
Example output

```json
{
  "insertId": "148gda2f1jvvvu",
  "jsonPayload": {
    "source": "org.apache.flink.runtime.entrypoint.ClusterEntrypoint",
    "message": "        [] -  Starting StandaloneSessionClusterEntrypoint (Version: 1.14.4, Scala: 2.11, Rev:895c609, Date:2022-02-25T11:57:14+01:00)"
  },
  "resource": {
    "type": "gce_instance",
    "labels": {
      "instance_id": "8710493455276707297",
      "project_id": "otel-agent-dev",
      "zone": "us-central1-a"
    }
  },
  "timestamp": "2022-05-26T17:12:02.616Z",
  "severity": "INFO",
  "labels": {
    "compute.googleapis.com/resource_name": "flink-debian2"
  },
  "logName": "projects/otel-agent-dev/logs/flink",
  "receiveTimestamp": "2022-05-26T17:12:22.426334126Z"
}
```

PR #651, opened 2022-06-06
Add Apache Flink Metricsreceiver

JonathanWamsley:
Add flink metrics receiver. This builds on #632

### DIFF
--- a/integration_test/third_party_apps_data/applications/flink/metadata.yaml
+++ b/integration_test/third_party_apps_data/applications/flink/metadata.yaml
@@ -1,4 +1,3 @@
-public_url: "https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/flink"
 app_url: "https://flink.apache.org/"
 short_name: Flink
 long_name: Apache Flink


apps/flink.go
```diff
@@ -0,0 +1,158 @@
...
+)
+
+type MetricsReceiverFlink struct {
+	confgenerator.ConfigComponent       `yaml:",inline"`
+	confgenerator.MetricsReceiverShared `yaml:",inline"`
+	Endpoint                            string `yaml:"endpoint" validate:"omitempty,url,startswith=http:"`
+}
+
+func (MetricsReceiverFlink) Type() string {
+	return "flinkmetrics"
```
ridwanmsharif, line 30: Can we rename the metrics type to just `flink`?

integration_test/third_party_apps_data/applications/flink/metadata.yaml
```diff
@@ -0,0 +1,60 @@
...
+app_url: https://flink.apache.org/
+short_name: Flink
+long_name: Apache Flink
+logo_path: ""
+description: The Apache Flink integration collects client, jobmanager and taskmanager
+  logs and parses them into a JSON payload. The result includes fields for source,
+  level, and message.
+supported_app_version: ["1.13.6", "1.14.4"]
+configure_integration: ""
+expected_metrics:
```
ridwanmsharif, line 11: This list seems incomplete? There should be more expected metrics right?

apps/flink.go
JonathanWamsley: updated. changed to flink

integration_test/third_party_apps_data/applications/flink/metadata.yaml
JonathanWamsley: updated. added rest of the metrics in.

ridwanmsharif (approved):

JonathanWamsley:
The flink.tgz file had a patch bump from 1.14.4 to 1.14.5 and the current version no longer exists in the install scripts. A [solution](https://github.com/GoogleCloudPlatform/ops-agent/pull/674) to use an archive source instead should fix the running tests.

qingling128:
Previously the only failed test was Windows, which is a known issue that was recently fixed.

OK to merge. Please rebase though.

integration_test/third_party_apps_data/applications/flink/metadata.yaml
```diff
@@ -1,10 +1,255 @@
 public_url: "https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/flink"
```
ridwanmsharif, line 1: Can we get rid of the public_url? See https://github.com/GoogleCloudPlatform/ops-agent/pull/632#discussion_r905486897

It will be set by the post agent release job that will run

PR #658, opened 2022-06-08
Use correct httpRequest key in all integrations

braydonk:
The original PR that adjusted the Ops Agent to use this feature was made before these integrations were available, so I overlooked them when I went to finally merge it. This PR adjusts the http request key for the new integrations I had forgotten.

### DIFF
--- a/apps/couchdb.go
+++ b/apps/couchdb.go
@@ -112,7 +112,7 @@ func (p LoggingProcessorCouchdb) Components(tag string, uid string) []fluentbit.
 			"Match":         tag,
 			"Operation":     "nest",
 			"Wildcard":      "http_request_*",
-			"Nest_under":    "logging.googleapis.com/http_request",
+			"Nest_under":    confgenerator.HttpRequestKey,
 			"Remove_prefix": "http_request_",
 		},
 	})
--- a/confgenerator/filter/internal/ast/ast.go
+++ b/confgenerator/filter/internal/ast/ast.go
@@ -52,7 +52,10 @@ var logEntryRootStructMapToFluentBit = map[string]string{
 	"labels":         "logging.googleapis.com/labels",
 	"operation":      "logging.googleapis.com/operation",
 	"sourceLocation": "logging.googleapis.com/sourceLocation",
-	"httpRequest":    "logging.googleapis.com/httpRequest",
+	// TODO: This needs to be the same as confgenerator.HttpRequestKey. Importing
+	// that package here results in a circular import. That should move somewhere
+	// better, and once it does we can use that here.
+	"httpRequest": "logging.googleapis.com/httpRequest",
 }
 
 func (m Target) fluentBitPath() ([]string, error) {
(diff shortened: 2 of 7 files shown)


apps/couchdb.go
```diff
@@ -112,7 +112,7 @@ func (p LoggingProcessorCouchdb) Components(tag string, uid string) []fluentbit.
 			"Match":         tag,
 			"Operation":     "nest",
 			"Wildcard":      "http_request_*",
-			"Nest_under":    "logging.googleapis.com/http_request",
+			"Nest_under":    confgenerator.HttpRequestKey,
```
igorpeshansky, line 115: This is fine for a short-term fix, but this copy-pasting of config components is ridiculous — longer term, we should have a common function that every receiver can call to set this up. Probably not worth taking the time to refactor it now. @qingling128 for thoughts…

confgenerator/filter/internal/ast/ast.go
```diff
@@ -52,7 +52,7 @@ var logEntryRootStructMapToFluentBit = map[string]string{
 	"labels":         "logging.googleapis.com/labels",
 	"operation":      "logging.googleapis.com/operation",
 	"sourceLocation": "logging.googleapis.com/sourceLocation",
-	"httpRequest":    "logging.googleapis.com/http_request",
+	"httpRequest":    "logging.googleapis.com/httpRequest",
```
igorpeshansky, line 55: Why not `confgenerator.HttpRequestKey` here as well?

braydonk: It results in a circular import; `confgenerator` imports `filter`, so `filter` can't import `confgenerator`. Not sure where a better spot for that constant would be than `confgenerator`.

apps/couchdb.go
ridwanmsharif: +1 to having something added here. Every receiver in the Ops Agent (which most 3rd Party integrations are) are expected to create a `LogEntry` since processors down in the pipeline expect that contract. Having some common util here, that would be more easily maintainable that would define and codify that contract would be my preference. But thats orthogonal and non-blocking to this PR.

braydonk: Took a look at `common_logging_processors.go` and what's there is very specific to access logs. We could have some pretty easy wins if we make some common processor utilities a little more universal.

igorpeshansky: I'd vote for deferring this to a follow-up PR.

confgenerator/filter/internal/ast/ast.go
braydonk: Yeah I think I agree with all of that. I added the explanatory comment.

A lot of projects have something like an `env` package for module-wide constants, unsure if that's the way we want to go or not.

apps/couchdb.go
qingling128: +1

qingling128 (approved):
LGTM

qingling128:
The related IIS and CouchDB tests passed. Merging this first to reduce the number of failed tests in master HEAD. Will rebase https://github.com/GoogleCloudPlatform/ops-agent/pull/660 to add test coverage for CouchDB.

apps/couchdb.go
braydonk: I will leave it out of this PR, but I am interested in looking deeper into this.

PR #661, opened 2022-06-08
Update ops_agent_test.go to require exactly one source from which to install the Ops Agent

Subbarker:


integration_test/ops_agent_test.go
```diff
@@ -24,11 +24,14 @@ to the ones mentioned at the top of gce_testing.go:
 
 PLATFORMS: a comma-separated list of distros to test, e.g. "centos-7,centos-8".
 
-The following variables are optional:
+Exactly one of the following variables must be set:
```
martijnvans, line 27: I think, for convenience, it would be great to have an easy way to install the latest stable agent. I'd like to have special handling for REPO_SUFFIX=stable, where it would get turned into "" at some level, perhaps inside installOpsAgent().

integration_test/ops_agent_test.go
```diff
@@ -24,11 +24,14 @@ to the ones mentioned at the top of gce_testing.go:
...
 
 REPO_SUFFIX: If provided, what package repo suffix to install the ops agent from.
```
martijnvans, line 29: I think we should we do the same thing to third_party_apps_test.go so that they are consistent.

PR #669, opened 2022-06-10
Ensuring that each ExpectedMetric.Type is unique in IntegrationMetadata

avilevy18:
Add validation to ensure ExpectedMetric.Type is unique in IntegrationMetadata and create appropriate tests.

Rename TestAll in common_test.go to TestMetadataValidation

Added to http://b/235595896


integration_test/common/testdata/integration-metadata_unique_expected-metrics-type/input.yaml
```diff
@@ -0,0 +1,117 @@
+public_url: "https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/activemq"
```
igorpeshansky, line 1: Do we need such a large test file for this? And does it need real-world data? Should we pare it down to a few entries with generic names?

braydonk: It probably doesn't need real world data, but the whole file is validated at once so it has to pass all validations present in `IntegrationMetadata`

avilevy18: Changed in: [a3b3e13](https://github.com/GoogleCloudPlatform/ops-agent/pull/669/commits/a3b3e13d3e634f380c64b51626a7316e97822d3e)

igorpeshansky: Ok, this is no longer real world data, but do we need a file this large to test the validation? Wouldn't it be enough to have 2 (maybe 3) entries?

That said, it does seem to be consistent with the other tests, which also have overgrown test inputs, so maybe we can defer.

avilevy18: I believe we should defer, I initially wanted to trim the test input but I thought it would be better for consistency.

avilevy18: Changed in: [f9abb2d](https://github.com/GoogleCloudPlatform/ops-agent/pull/669/commits/f9abb2d6359abf815e008557f2c4c1a72669acf7)

igorpeshansky (approved):
LGTM :shipit:

PR #670, opened 2022-06-10
Testing: add a way to adjust the log URL

martijnvans:
This is to fix an issue with internally-run tests where they are pointing at the public bucket (incorrectly).


qingling128 (approved):
LGTM

PR #672, opened 2022-06-19
Add Vault Audit logging

Mrod1598:
Example Output:
image


antonblock (approved):

apps/vault.go
```diff
@@ -0,0 +1,67 @@
...
+	r.MultilineRules = []confgenerator.MultilineRule{
+		{
+			StateName: "start_state",
+			NextState: "cont",
+			Regex:     `^{.*`,
+		},
+		{
+			StateName: "cont",
+			NextState: "cont",
+			Regex:     `^[^{].*[,}]$`,
```
ridwanmsharif, line 57: Can we add an example of a multiline log line? 

Other receivers use the pattern of having the `cont` regex be the negation match of the `start_state` regex. Is there a reason this one doesn't follow that pattern?
https://github.com/GoogleCloudPlatform/ops-agent/blob/487639ed8656db052376242ca92d9815ec891644/apps/cassandra.go#L106-L116

apps/vault.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+	}
+
+	c := r.LoggingReceiverFilesMixin.Components(tag)
+	return append(c, r.LoggingProcessorVaultJson.Components(tag, "vault_audit")...)
```
ridwanmsharif, line 62: Please use `p.Type()` instead of `vault_audit`

ridwanmsharif (approved):
LGTM modulo a couple comments

integration_test/third_party_apps_data/applications/vault/metadata.yaml
```diff
@@ -0,0 +1,140 @@
+public_url: "https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/vault"
```
ridwanmsharif, line 1: We should remove the public_url now. It returns a 404 as is. It will be set by the post release job that will be run.

https://github.com/GoogleCloudPlatform/ops-agent/pull/632#discussion_r905486897

Mrod1598: Updated

apps/vault.go
Mrod1598: Updated

Mrod1598:
kokoro has passed
image

ridwanmsharif (approved):

integration_test/third_party_apps_data/applications/vault/metadata.yaml
```diff
@@ -0,0 +1,139 @@
+app_url: "https://www.vaultproject.io/"
+short_name: Vault
+long_name: Hashicorp Vault
+description: |-
+  Vault is an identity-based secrets and encryption management system. This integration Collects Vault's audit logs.
+minimum_supported_agent_version:
```
sophieyfang, line 6: minimum_supported_agent_version shall be set by jobs not manually

PR #673, opened 2022-06-21
Custom validator to ensure there is only one Representative in Expected Metric

avilevy18:
Add custom validator to ensure there is only one ExpectedMetric.Representative in IntegrationMetadata.

Refactor third_party_apps_test.go to use new validator.


integration_test/common/testdata/integration-metadata_oneof_field_expected-metric/golden_error
```diff
@@ -0,0 +1 @@
+Key: 'IntegrationMetadata.ExpectedMetrics' Error:Field validation for 'ExpectedMetrics' failed on the 'oneof_field' tag
```
igorpeshansky, line 1: I'm confused — shouldn't `oneof` [produce a more sensible error message](https://github.com/go-playground/validator/blob/master/translations/en/en.go#L1279-L1291) like `monitored_resource must be one of ["gce_instance"]`? Where is this one coming from?

igorpeshansky, line 1: Can you please explain how this test case is related to the code changes in this CL? Don't we instead need a test case that has more than one representative metric?

integration_test/common/testdata/integration-metadata_oneof_field_expected-metric/input.yaml
```diff
@@ -0,0 +1,112 @@
+public_url: "https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/activemq"
```
igorpeshansky, line 1: Which of these fields causes the golden error? Should we reduce the size of this test input, just like we did in #669, and add a comment on the line that should trigger the error?

integration_test/common/common.go
```diff
@@ -121,3 +101,33 @@ func SliceContains(slice []string, toFind string) bool {
...
+			return true
+		}
+
+		representativeCount := 0
+		for _, m := range metrics {
+			if m.Representative {
+				representativeCount += 1
+			}
+		}
+		return representativeCount == 1
```
igorpeshansky, line 130: Do we also need a `RegisterTranslation` call to define an error message for this validator?

igorpeshansky (commented):
Mostly questions and one implementation suggestion.

integration_test/common/testdata/integration-metadata_oneof_field_expected-metric/input.yaml
avilevy18: Changed in: [59870d7](https://github.com/GoogleCloudPlatform/ops-agent/pull/673/commits/59870d7c770575b84c837fcb1d7e1c705dd5405d)

integration_test/common/common.go
```diff
@@ -121,3 +101,33 @@ func SliceContains(slice []string, toFind string) bool {
 	}
 	return false
 }
+
+func NewIntegrationMetadataValidator() *validator.Validate {
+	v := validator.New()
+	_ = v.RegisterValidation("onerepresentative", func(fl validator.FieldLevel) bool {
```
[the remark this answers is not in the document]
avilevy18: [59870d7](https://github.com/GoogleCloudPlatform/ops-agent/pull/673/commits/59870d7c770575b84c837fcb1d7e1c705dd5405d)

integration_test/common/common.go
```diff
@@ -121,3 +101,33 @@ func SliceContains(slice []string, toFind string) bool {
...
+
+		fieldStr := fl.Field().String()
+		if fieldStr == "" {
+			// Ignore the case where this field is not actually specified or is left empty.
```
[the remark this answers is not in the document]
avilevy18: Changed in: [59870d7](https://github.com/GoogleCloudPlatform/ops-agent/pull/673/commits/59870d7c770575b84c837fcb1d7e1c705dd5405d)

integration_test/common/common.go
sophieyfang: That was my initial thought too. The current error msg isn't clear on the root cause on why it fails the test, it only says it fails on the tag, thus the `onetrue` tag needs to have a comment saying on what this tag expected. Such comment might be better preserved in a well explained error message.

We have this in ops-agent validator. That is client facing thus we must explain it well while this is internal, maybe a comment might suffice if the effort on implementing the error message needs more than moderate effort.

integration_test/common/testdata/integration-metadata_onetrue_expected-metric/input.yaml
```diff
@@ -0,0 +1,31 @@
...
+    broker: .*
+    destination: .*
+  optional: true
+expected_logs:
+- log_name: examplelog
+  fields:
+  - name: jsonPayload.message
+    value_regex: '(example):'
+    type: string
+    description: null
```
sophieyfang, line 25: We can remove this part on expected_logs, as this test is testing the metrics side validation, thus less irrelevant contents in the test file, better the focus is on the metrics side

integration_test/common/testdata/integration-metadata_onetrue_expected-metric/input.yaml
```diff
@@ -0,0 +1,31 @@
+app_url: "https://example.google.com/"
+short_name: OpsAgent
+long_name: Cloud OpsAgent
+description: |-
+  Example description
+minimum_supported_agent_version:
+  metrics: 1.0.0
+  logging: 1.0.0
+supported_app_version: ["Example0"]
+expected_metrics:
```
sophieyfang, line 10: This test file is testing that none of the expected metrics has the representative field. There shall be another test that is on when there are more than one expected metrics that have representative field. Thus it's matching `there must be exactly one metric with representative: true`, cite from: https://github.com/GoogleCloudPlatform/ops-agent/blob/2e5fe210186b12e5c175fc9344d26e7bac151759/integration_test/common/common.go#L111

integration_test/common/testdata/integration-metadata_onetrue_expected-metric/input.yaml
avilevy18: Changed in: [ca458e4](https://github.com/GoogleCloudPlatform/ops-agent/pull/673/commits/ca458e452065931cb991b63a1550a0ba7cf88245)

integration_test/common/common.go
igorpeshansky: @avilevy18 and I had an offline discussion about it. Doing this via the translations mechanism turns out to be extremely cumbersome and counter-intuitive. We can do the same kind of switch as we have in the config validator, but that would have to be a follow-up change.

integration_test/common/testdata/integration-metadata_oneof_field_expected-metric/golden_error
igorpeshansky: Discussed offline. These default messages require one to explicitly register the supplied translations, which is nontrivial. We will likely add a switch similar to the [one in the config validator](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/config.go#L90-L129), but in a follow-up change.

integration_test/common/common.go
```diff
@@ -104,30 +107,47 @@ func SliceContains(slice []string, toFind string) bool {
...
+	_ = v.RegisterValidation("onetrue", func(fl validator.FieldLevel) bool {
+		field := fl.Field()
+		param := fl.Param()
 
-		fieldStr := fl.Field().String()
-		if fieldStr == "" {
-			// Ignore the case where this field is not actually specified or is left empty.
-			return true
+		if param == "" {
+			panic(fmt.Sprintf("onetrue must contain an argument"))
```
igorpeshansky, line 115: This doesn't need the `fmt.Sprintf`, does it?

integration_test/common/common.go
```diff
@@ -104,30 +107,47 @@ func SliceContains(slice []string, toFind string) bool {
...
-		if !ok {
-			return false
-		}
+		switch field.Kind() {
 
-		if len(metrics) == 0 {
-			return true
-		}
+		case reflect.Slice, reflect.Array:
+			elem := field.Type().Elem()
```
igorpeshansky, line 121: Where is this used? Were you planning to check that the resulting type is a struct that has the requested field?

integration_test/common/common.go
```diff
@@ -104,30 +107,47 @@ func SliceContains(slice []string, toFind string) bool {
...
+			for i := 0; i < field.Len(); i++ {
+				elm := reflect.Indirect(field.Index(i))
+
+				f := elm.FieldByName(param)
+
+				if !f.IsValid() {
+					panic(fmt.Sprintf("Invalid field name %s", param))
+				}
+
+				if !f.IsZero() {
```
igorpeshansky, line 141: Should you also check that the field is a boolean?

integration_test/common/common.go
```diff
@@ -104,30 +107,47 @@ func SliceContains(slice []string, toFind string) bool {
...
+				elem = elem.Elem()
 			}
+
+			count := 0
+			for i := 0; i < field.Len(); i++ {
+				elm := reflect.Indirect(field.Index(i))
+
+				f := elm.FieldByName(param)
+
+				if !f.IsValid() {
```
igorpeshansky, line 137: I would do this check via the type, rather than on each element of the slice. E.g., something like this:
```Go
		case reflect.Slice, reflect.Array:
			// Ignore the case where this field is not actually specified or is left empty.
			if field.Len() == 0 {
				return true
			}

			elem := field.Type().Elem()
			if elem.Kind() == reflect.Ptr {
				elem = elem.Elem()
			}

			sf, ok := elem.FieldByName(param)
			if !ok {
				panic(fmt.Sprintf("Invalid field name %s", param))
			}
			if sfTyp := sf.Type; sfTyp.Kind() != reflect.Bool {
				panic(fmt.Sprintf("Field %s is %s, not bool", param, sfTyp))
			}

			count := 0
			for i := 0; i < field.Len(); i++ {
				if !reflect.Indirect(field.Index(i)).FieldByName(param).IsZero() {
					count++
				}
			}

			return count == 1
```

integration_test/common/testdata/integration-metadata_expected-metric_onetrue_no_represenetative/golden_error
```diff
@@ -1 +1 @@
-Key: 'IntegrationMetadata.ExpectedMetrics' Error:Field validation for 'ExpectedMetrics' failed on the 'oneof_field' tag
\ No newline at end of file
+Key: 'IntegrationMetadata.ExpectedMetrics' Error:Field validation for 'ExpectedMetrics' failed on the 'onetrue' tag
```
igorpeshansky, line 1: Looks like there's a typo in the directory name (`represenetative`).

integration_test/common/testdata/integration-metadata_expected-metric_onetrue_two_represenetative/golden_error
```diff
@@ -0,0 +1 @@
+Key: 'IntegrationMetadata.ExpectedMetrics' Error:Field validation for 'ExpectedMetrics' failed on the 'onetrue' tag
```
igorpeshansky, line 1: Looks like there's a typo in the directory name (`represenetative`).

integration_test/common/testdata/integration-metadata_expected-metric_onetrue_two_represenetative/input.yaml
```diff
@@ -0,0 +1,32 @@
...
+expected_metrics:
+- type: example.googleapis.com/example.example1
+  value_type: INT64
+  kind: GAUGE
+  monitored_resource: gce_instance
+  labels:
+    broker: .*
+    destination: .*
+  representative: true
+- type: example.googleapis.com/example.example1
```
igorpeshansky, line 19: `example2`?

integration_test/common/testdata/integration-metadata_expected-metric_onetrue_two_represenetative/input.yaml
```diff
@@ -0,0 +1,32 @@
...
+  value_type: INT64
+  kind: GAUGE
+  monitored_resource: gce_instance
+  labels:
+    broker: .*
+    destination: .*
+  representative: true
```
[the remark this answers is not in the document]
avilevy18: Changed in: [42da351](https://github.com/GoogleCloudPlatform/ops-agent/pull/673/commits/42da351c6e2089dcedf53a561fe57305c67ce00c)

integration_test/common/testdata/integration-metadata_expected-metric_onetrue_two_represenetative/input.yaml
avilevy18: Changed in: [42da351](https://github.com/GoogleCloudPlatform/ops-agent/pull/673/commits/42da351c6e2089dcedf53a561fe57305c67ce00c)

integration_test/common/testdata/integration-metadata_expected-metric_onetrue_two_represenetative/golden_error
avilevy18: Changed in: [42da351](https://github.com/GoogleCloudPlatform/ops-agent/pull/673/commits/42da351c6e2089dcedf53a561fe57305c67ce00c)

integration_test/common/testdata/integration-metadata_expected-metric_onetrue_no_represenetative/golden_error
avilevy18: Changed in: [6e34c18](https://github.com/GoogleCloudPlatform/ops-agent/pull/673/commits/6e34c187f07f69928e7153b38e56ca0d1f9430b7)

igorpeshansky (approved):
LGTM :shipit:

PR #676, opened 2022-06-22
Refer to archive for solr install

dehaansa:
Missed one using downloads, rather than dlcdn. This should be the last one, but apparently today was release day for all kinds of projects.


dehaansa:
Not sure what caused the CentOS 7 failure (Cento-OS? that typo is new here...) but it shouldn't be related to this PR.

qingling128 (approved):
LGTM

PR #694, opened 2022-06-27
MSSQL: Fix Integration Test

schmikei:
These were not run during merge and because of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9691, they needed to become different types.

Apologies for not catching it sooner! Should unblock anything that was relying on this being fixed!

http://b/237276968


integration_test/third_party_apps_data/applications/mssql/enable
```diff
@@ -9,11 +9,13 @@ metrics:
     mssql_v2:
       type: mssql
       receiver_version: 2
+      collection_interval: 30s
```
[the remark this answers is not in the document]
schmikei: Not necessarily I can remove!

integration_test/third_party_apps_data/applications/mssql/metadata.yaml
```diff
@@ -38,11 +38,13 @@ expected_metrics:
   monitored_resource: gce_instance
   labels: {}
 # v2 metrics
-- type: workload.googleapis.com/sqlserver.user.connection.count # TODO: look into this metric type change
-  value_type: DOUBLE
+- type: workload.googleapis.com/sqlserver.user.connection.count
```
qingling128, line 41: For the records: we are ok with changing the expected type from DOUBLE to INT because the V2 mssql metrics was never released.

qingling128 (approved):
LGTM

integration_test/third_party_apps_data/applications/mssql/metadata.yaml
```diff
@@ -38,11 +38,13 @@ expected_metrics:
...
+  value_type: INT64
   kind: GAUGE
   monitored_resource: gce_instance
   labels: {}
+  # if no user connection may not be present
```
igorpeshansky, line 46: It may be best to make this an end-of-line comment, to avoid it being accidentally separated from the value.

igorpeshansky (commented):
Minor stuff, otherwise LGTM.

integration_test/third_party_apps_data/applications/mssql/metadata.yaml
igorpeshansky: Actually, can/should we force a user connection to happen as part of the test? This may be more of a follow-up thing, though…

integration_test/third_party_apps_data/applications/mssql/metadata.yaml
```diff
@@ -38,12 +38,11 @@ expected_metrics:
   monitored_resource: gce_instance
   labels: {}
 # v2 metrics
-- type: workload.googleapis.com/sqlserver.user.connection.count
+- type: workload.googleapis.com/sqlserver.user.connection.count # if no user connection may not be present
```
igorpeshansky, line 41: [Optional] You could have left it after `optional: true` as well.

igorpeshansky (approved):
LGTM :shipit:

integration_test/third_party_apps_data/applications/mssql/metadata.yaml
schmikei: From my testing the current integration test does create a single user connection on the instance, I just noticed this while doing some deeper testing on my own sqlserver instance that I spun up!

schmikei: Actually I am going to remove the optional, just to prevent future confusion

integration_test/third_party_apps_data/applications/mssql/metadata.yaml
schmikei: @igorpeshansky just removed this I think everything should be good to go now 👍

igorpeshansky:
Both of the failed Kokoro checks were flakes. We had a successful run of the third_party tests, and I've double-checked that it included `mssql`:
```
    --- PASS: TestThirdPartyApps/sql-std-2019-win-2019/mssql (739.41s)
```
I'm going to merge this.

PR #702, opened 2022-06-28
Add Vault Metrics Receiver

Mrod1598:
Add the vault metrics receiver using the prometheus receiver as a base and applying transformations.


apps/vault.go
```diff
@@ -1,10 +1,251 @@
...
+
+	for _, operation := range operations {
+		operationTimeName := "vault.storage.operation." + operation + ".time"
+		operationCountName := "vault.storage.operation." + operation + ".count"
+		newMetrics = append(newMetrics, operationTimeName)
+		newMetrics = append(newMetrics, operationCountName)
+		for _, storage := range storages {
+			oldMetric := "vault_" + storage + "_" + operation
+
+			queries = append(queries, otel.SummaryCountValToSum(oldMetric, "cumulative", true))
```
dehaansa, line 156: Should we be dropping the summary metric after we've gotten the parts of it that we want separated out?

apps/vault.go
```diff
@@ -1,10 +1,251 @@
...
+			Config: map[string]interface{}{
+				"config": map[string]interface{}{
+					"scrape_configs": []map[string]interface{}{
+						scrapeConfig,
+					},
+				},
+			},
+		},
+		Processors: []otel.Component{
+			storageMetricTransforms,
```
dehaansa, line 96: Unless there's a reason I'm missing, should refactor to only create one Transform processor for the two different sets of actions

apps/vault.go
Mrod1598: it gets filtered out because it's specified to only keep new metrics

apps/vault.go
Mrod1598: Updated

Mrod1598:
Vault seems to be passing but couchbase is broken?

integration_test/third_party_apps_data/applications/vault/metadata.yaml
```diff
@@ -4,6 +4,117 @@ long_name: Hashicorp Vault
...
+  monitored_resource: gce_instance
+  value_type: INT64
+  labels: 
+    cluster: .*
+- type: workload.googleapis.com/vault.core.leader.duration
+  optional: true
+  kind: GAUGE
+  monitored_resource: gce_instance
+  value_type: INT64
+  labels: {}
```
ridwanmsharif, line 19: The scope doc says this needs to have a `cluster` label? Does this of the scope doc need to be updated?

apps/vault.go
```diff
@@ -1,10 +1,299 @@
...
+	OldName     string
+	NewName     string
+	Description string
+	Unit        string
+}
+
+func (r MetricsReceiverVault) addStorageMetrics() (transforms []otel.TransformQuery, newMetrics []string) {
+	storageLabel := "storage"
+
+	storages := []string{
```
ridwanmsharif, line 156: Can we define all constants at the top of the file, makes it easier to update and less error prone.

ridwanmsharif (approved):

integration_test/third_party_apps_data/applications/vault/metadata.yaml
```diff
@@ -4,6 +4,117 @@ long_name: Hashicorp Vault
 description: |-
   Vault is an identity-based secrets and encryption management system. This integration Collects Vault's audit logs.
```
mtabasko, line 5: collects

mtabasko, line 5: Since we're adding the metrics receiver here, shouldn't this summary also say something about metrics. Perhaps something like this:

... audit logs. The integration also collects token, memory, and storage metrics.

integration_test/third_party_apps_data/applications/vault/metadata.yaml
```diff
@@ -139,5 +250,38 @@ configuration_options:
     - name: wildcard_refresh_interval
       default: 60s
       description: The interval at which wildcard file paths in include_paths are refreshed. Given as a time duration, for example 30s or 2m. This property might be useful under high logging throughputs where log files are rotated faster than the default interval.
```
mtabasko, line 252: `include_paths`  (also at line 246, which I couldn't comment on for some reason).

mtabasko, line 252: `30s` or `2m`.

integration_test/third_party_apps_data/applications/vault/metadata.yaml
```diff
@@ -139,5 +250,38 @@ configuration_options:
...
+  metrics:
+  - type: vault
+    fields: 
+    - name: type
+      default: null
+      description: This value must be vault.
```
mtabasko, line 258: `vault`.

integration_test/third_party_apps_data/applications/vault/metadata.yaml
```diff
@@ -139,5 +250,38 @@ configuration_options:
...
+    - name: endpoint
+      default: localhost:8200
+      description: The hostname:port used by vault
+    - name: token
+      default: localhost:8200
+      description: Token used for authentication.
+    - name: metrics_path
+      default: /v1/sys/metrics
+      description: the path for metrics collection.
```
mtabasko, line 267: The

integration_test/third_party_apps_data/applications/vault/metadata.yaml
```diff
@@ -139,5 +250,38 @@ configuration_options:
...
+    - name: collection_interval
+      default: 60s
+      description: A time.Duration value, such as 30s or 5m.
```
mtabasko, line 270: A [time.Duration](https://pkg.go.dev/time#ParseDuration) value, such as `30s` or `5m`.

integration_test/third_party_apps_data/applications/vault/metadata.yaml
```diff
@@ -139,5 +250,38 @@ configuration_options:
...
+    - name: insecure
+      default: true
+      description: Sets whether or not to use a secure TLS connection. If set to false, then TLS is enabled.
+    - name: insecure_skip_verify
+      default: false
+      description: Sets whether or not to skip verifying the certificate. If insecure is set to true, then the insecure_skip_verify value is not used.
```
mtabasko, line 276: `true`

mtabasko, line 276: `insecure_skip_verify`  (or just "the value of this field is not used.")

mtabasko, line 276: `insecure`

mtabasko (changes requested):
An assortment of editorial changes for public documentation.

integration_test/third_party_apps_data/applications/vault/metadata.yaml
Mrod1598: Updated

Mrod1598: Updated

Mrod1598: Updated

integration_test/third_party_apps_data/applications/vault/metadata.yaml
Mrod1598: Updated

integration_test/third_party_apps_data/applications/vault/metadata.yaml
Mrod1598: Updated

integration_test/third_party_apps_data/applications/vault/metadata.yaml
Mrod1598: Updated

Mrod1598: Updated

apps/vault.go
Mrod1598: Updated

integration_test/third_party_apps_data/applications/vault/metadata.yaml
Mrod1598: Updated the scope doc and it was also a DOUBLE not an INT so updated that as well.

Mrod1598:
@ridwanmsharif This should be ready for review when you get a second.

apps/vault.go
ridwanmsharif: [nit] Can we also move the list of storages to the top aswell and alphabetize it? Instead of reverse ahphabetical order? That is if the order doesn't matter.

ridwanmsharif (approved):

apps/vault.go
Mrod1598: Updated

mtabasko (commented):
Looks good from my end. Thank you!!

mtabasko (approved):

Mrod1598:
@ridwanmsharif should be good now.

PR #721, opened 2022-07-06
Added validation for duplicated network port for logging receivers

LujieDuan:
[b/220128869](http://b/220128869) | P3 | Add validation failure at the Ops Agent level when there are two receivers listening on the same port
The validation added here can check for both cases: 1. two receivers using the same port and 2. one receiver used by two pipelines. 
Originally considered to use custom validator but that would require traverse the fields in the yaml tree between `service` and `receivers` sections to check receivers' port and whether the receiver is used by any pipeline. 
This current proposed solution added a new interface LoggingNetworkReceiver by extending the LoggingReceiver interface and provide the method to access the port taken by this receiver. The the validation is actually happening in the `Validate()` on the `Logging` instance.  
To validate, loop over all services/pipelines and keep track of ports used, and return the error if there is port conflict. 
Hostname is not considered in this validation; i.e., if two syslog receivers listening on `127.0.0.1:5540` and `0.0.0.0:5540` we still consider there is conflict.

### DIFF
--- a/confgenerator/config.go
+++ b/confgenerator/config.go
@@ -392,6 +392,12 @@ func (m *loggingReceiverMap) UnmarshalYAML(unmarshal func(interface{}) error) er
 	return nil
 }
 
+// Logging receivers that listen on a port of the host
+type LoggingNetworkReceiver interface {
+	LoggingReceiver
+	GetListenPort() uint16
+}
+
 type LoggingProcessor interface {
 	Component
 	// Components returns fluentbit components that implement this procesor.
@@ -698,6 +704,7 @@ func (l *Logging) Validate(platform string) error {
 	if l.Service == nil {
 		return nil
 	}
+	portTaken := map[uint16]string{} // port -> receiverId map
 	for _, id := range sortedKeys(l.Service.Pipelines) {
 		p := l.Service.Pipelines[id]
 		if err := validateComponentKeys(l.Receivers, p.ReceiverIDs, subagent, "receiver", id); err != nil {
@@ -717,6 +724,10 @@ func (l *Logging) Validate(platform string) error {
 			return err
 		}
 		if _, err := validateComponentTypeCounts(l.Processors, p.ProcessorIDs, subagent, "processor"); err != nil {
+			return err
+		}
+		// portTaken will be modified/updated by the validation function
+		if _, err := validateReceiverPorts(portTaken, l.Receivers, p.ReceiverIDs); err != nil {
 			return err
 		}
 		if len(p.ExporterIDs) > 0 {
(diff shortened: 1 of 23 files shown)


confgenerator/testdata/invalid/linux/logging-receiver_duplicated_port/input.yaml
```diff
@@ -0,0 +1,18 @@
+logging:
```
qingling128, line 1: Can we also add a unit test case like this for completeness?

```
logging:
  receivers:
    s1:
      type: syslog
      transport_protocol: tcp
      listen_host: 0.0.0.0
      listen_port: 5140      
    s2:
      type: syslog
      transport_protocol: tcp
      listen_host: 0.0.0.0
      listen_port: 5140  
  service:
    pipelines:
      p1:
        receivers: [s1, s2]
```

confgenerator/config.go
```diff
@@ -865,6 +882,32 @@ func validateComponentTypeCounts(components interface{}, refs []string, subagent
 	return r, nil
 }
 
+// Validate that no two receivers are using the same port; adding new port usage to the input map `taken`
+func validateReceiverPorts(taken map[uint16]string, components interface{}, refs []string) (map[uint16]string, error) {
```
qingling128, line 886: [Optional] I'm trying to think how to better distinguish the IDs from `refs` and the IDs from `taken[port]` to help with readability by naming them in a more descriptive way.

Some brainstorming ideas are: `pipelineRIDs` and `portRID`. e.g.

```
for _, pipelineRID := range pipelineRIDs {
...
   if portRID, ok := taken[port]; ok { 
        if portRID == pipelineRID
```
But they might be long per Golang naming standards. Will defer to you whether to make any adjustment.

confgenerator/config.go
```diff
@@ -776,6 +787,12 @@ var (
 		"iis":                     1,
 		"mssql":                   1,
 	}
+
+	receiverPortLimits = map[string]bool{
```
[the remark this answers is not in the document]
LujieDuan: Good point! Replaced this with a slice and updated the validation function to loop over the slice to find if the type is in this slice.

confgenerator/config.go
LujieDuan: Yes agree those are not good names here to distinguish those ids. Updated the naming as suggested.

confgenerator/testdata/invalid/linux/logging-receiver_duplicated_port/input.yaml
LujieDuan: Test case added as `logging-receiver_duplicated_port_same_service`

qingling128 (approved):
LGTM

PR #722, opened 2022-07-06
Testing: make metadata.yaml required

martijnvans:
All applications have a `metadata.yaml` file now. Using a zero-value `metadata.yaml` file could cause confusion and tests that look like they're passing when they're not.


martijnvans:
This was actually included in https://github.com/GoogleCloudPlatform/ops-agent/pull/723, I forgot to mention that this PR was meant to come first, as a way to get one thing out of the way first. But it got included in the other PR, so I'll close this one.

PR #719, opened 2022-07-05
Extended core metrics tests

avilevy18:
Extended testing coverage of core metrics tests to include `value_type`, `kind`, `monitored_resource` and `labels`

Refactored AssertMetrics and assertMetricsLabel into common.go

Added expected metrics yaml to integration_test/agent_metrics

http://b/230790404

PR #723, opened 2022-07-06
Testing: replace supported_applications.txt with a field in metadata.yaml

martijnvans:
This is action item #1 from this document: https://docs.google.com/document/d/1TFJ-OhIm2KmATv-sam3aLlZz8T5vMp7zsHSMxp0kVtw/edit?usp=sharing&resourcekey=0-JWBhMrd2SdUg1VYHvwuNtw

The idea is to make it impossible to forget to update `supported_applications.txt`, instead requiring each `metadata.yaml` file to contain an additional field called `supported_operating_systems`, containing either `linux`, `windows`, or `linux_and_windows`.

This PR involved a lot of reshuffling in `third_party_apps_test.go` in order to get the parsed `metadata.yaml` to be accessible early on in the test. I moved the `metadata.yaml` parsing out of `runSingleTest` and into `determineAllApps` (which got renamed to `fetchAppsAndMetadata`), which runs right at the beginning of the test, before any VMs are started up.


martijnvans:
all the kokoro failures are unrelated, going to go ahead and merge.

integration_test/third_party_apps_test.go
```diff
@@ -510,14 +478,40 @@ const (
...
 )
 
+// incompatibleOperatingSystem looks at the supported_operating_systems field
+// of metadata.yaml for this app and returns a nonempty skip reason if it
+// thinks this app doesn't support the given platform.
+// supported_operating_systems should only contain "linux", "windows", or
+// "linux_and_windows".
+func incompatibleOperatingSystem(t *testing.T, testCase test) string {
+	supported := testCase.metadata.SupportedOperatingSystems
+	if supported == "linux_and_windows" {
```
igorpeshansky, line 488: [Optional] Why not `switch`?

integration_test/third_party_apps_test.go
```diff
@@ -510,14 +478,40 @@ const (
...
+		return "" // This app supports both Linux and Windows.
+	}
+	if supported == "windows" {
+		if !gce.IsWindows(testCase.platform) {
+			return fmt.Sprintf("Skipping test for platform %v because app %v only supports Windows.", testCase.platform, testCase.app)
```
igorpeshansky, line 493: #719 introduced `gce.PlatformKind()` that could make this code simpler:
```
func incompatibleOperatingSystem(t *testing.T, testCase test) string {
	switch supported := testCase.metadata.SupportedOperatingSystems; supported {
	case "linux_and_windows":
		return "" // This app supports both Linux and Windows.
	case "windows", "linux":
		if supported != gce.PlatformKind(testCase.platform) {
			return fmt.Sprintf("Skipping test for platform %v because app %v only supports %v.", testCase.platform, testCase.app, supported)
		}
		return "" // We are testing on a supported platform for this app.
	default:
		t.Fatalf("Unrecognized value %q for supported_operating_systems.", supported)
		return ""
	}
}
```
or even (if we choose to ignore the unrecognized platform case and rely on the validation):
```
func incompatibleOperatingSystem(t *testing.T, testCase test) string {
	supported := testCase.metadata.SupportedOperatingSystems
	if ! strings.Contains(supported, gce.PlatformKind(testCase.platform)) {
		return fmt.Sprintf("Skipping test for platform %v because app %v only supports %v.", testCase.platform, testCase.app, supported)
	}
	return "" // We are testing on a supported platform for this app.
}
```

integration_test/third_party_apps_test.go
```diff
@@ -510,14 +478,40 @@ const (
...
+		}
+		return "" // We are testing Windows and this app supports Windows.
+	}
+	if supported == "linux" {
+		if gce.IsWindows(testCase.platform) {
+			return fmt.Sprintf("Skipping test for platform %v because app %v only supports Linux.", testCase.platform, testCase.app)
+		}
+		return "" // We are testing Linux and this app supports Linux.
+	}
+	t.Fatalf("Unrecognized value %q for supported_operating_systems.", supported)
```
igorpeshansky, line 503: This case is probably not necessary given the validation rules… You could make this a `panic` instead…

integration_test/common/common.go
```diff
@@ -92,6 +92,7 @@ type IntegrationMetadata struct {
 	ExpectedMetrics              []*ExpectedMetric            `yaml:"expected_metrics" validate:"onetrue=Representative,unique=Type,dive"`
 	MinimumSupportedAgentVersion MinimumSupportedAgentVersion `yaml:"minimum_supported_agent_version"`
 	SupportedAppVersion          []string                     `yaml:"supported_app_version" validate:"required,unique,min=1"`
+	SupportedOperatingSystems    string                       `yaml:"supported_operating_systems" validate:"required,oneof=linux windows linux_or_windows"`
```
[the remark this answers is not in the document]
martijnvans: fixed

integration_test/third_party_apps_test.go
```diff
@@ -432,22 +385,36 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+		err = yaml.UnmarshalStrict(testCaseBytes, &metadata)
+		if err != nil {
+			t.Fatalf("could not unmarshal contents of applications/%v/metadata.yaml: %v", app, err)
+		}
+		if err = validate.Struct(&metadata); err != nil {
+			t.Fatalf("could not validate contents of applications/%v/metadata.yaml: %v", app, err)
+		}
+		allApps[app] = metadata
+	}
+	log.Printf("found %v apps", len(allApps))
```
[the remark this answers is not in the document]
martijnvans: The impacted apps list is still logged, on line 460.

I think it's fine to just log the number of apps here, since at this point there's no realistic way for it to be wrong anyway. It has to have a valid `metadata.yaml` file for each app or it will return an error and never get to this point anyway.

integration_test/third_party_apps_test.go
```diff
@@ -510,14 +478,40 @@ const (
...
+// incompatibleOperatingSystem looks at the supported_operating_systems field
+// of metadata.yaml for this app and returns a nonempty skip reason if it
+// thinks this app doesn't support the given platform.
+// supported_operating_systems should only contain "linux", "windows", or
+// "linux_and_windows".
+func incompatibleOperatingSystem(t *testing.T, testCase test) string {
+	supported := testCase.metadata.SupportedOperatingSystems
+	if supported == "linux_and_windows" {
+		return "" // This app supports both Linux and Windows.
+	}
```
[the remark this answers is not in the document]
martijnvans: try as I might, I can't understand this suggestion.

integration_test/third_party_apps_test.go
martijnvans: obsolete

integration_test/third_party_apps_test.go
martijnvans: obsolete

integration_test/third_party_apps_test.go
martijnvans: done

integration_test/common/common.go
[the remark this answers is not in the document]
martijnvans: I'm not understanding this suggestion. Where would this key go, and how is it different from what i have besides the name? I just don't get it.

PR #748, opened 2022-07-20
Internal: boost the recommended CPUs for build VMs

martijnvans:
In my benchmarks (3 data points before and after), using 8 CPUs was 10% (6 minutes) faster for a clean build (around 50 minutes).


ridwanmsharif (commented):
LGTM! Lets get speeeeeeed 🚤 🏃 💨

PR #752, opened 2022-07-25
Testing: fix mysql test failure

martijnvans:
https://github.com/GoogleCloudPlatform/ops-agent/pull/719 changed the label matching to be a full match instead of a partial match, and that caused mysql tests to fail with `workload.googleapis.com/mysql.double_writes: error: label value does not match pattern. label=kind, pattern=writes|written, value=pages_written` errors.

Looking at metrics explorer, everything is either "writes" or "pages_written", so i deleted "written" from the list of possible values.


avilevy18 (approved):
LGTM

PR #760, opened 2022-07-27
Remove empty ExpectedMetrics.Labels in metadata

avilevy18:
Refactoring ExpectedMetrics.Labels validation.

http://b/242048487


integration_test/metadata/integration_metadata.go
```diff
@@ -38,7 +38,7 @@ type ExpectedMetric struct {
 	MonitoredResource string `yaml:"monitored_resource" validate:"required,oneof=gce_instance"`
 	// Mapping of expected label keys to value patterns.
 	// Patterns are RE2 regular expressions.
-	Labels map[string]string `yaml:"labels"`
+	Labels map[string]string `yaml:"labels" validate:"omitempty,gt=0"`
```
igorpeshansky, line 41: I think `omitempty` should also be part of the `yaml` annotation (https://pkg.go.dev/gopkg.in/yaml.v2#Marshal)…

sophieyfang (approved):

integration_test/metadata/integration_metadata.go
sophieyfang: golang validator also has omitempty. They serve different purpose.

the one yaml uses is useful only during marshalling which we don't have that use case

here is the one for golang validator omitempty: https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Omit_Empty

sophieyfang: Here is the context: https://github.com/GoogleCloudPlatform/ops-agent/pull/732#discussion_r929109156

And golang playground:  https://go.dev/play/p/K13XVjHgqkS

image

igorpeshansky: Right, I saw the validator one — that's why I said "also". :grin:

avilevy18: Changed in: [16b3b36](https://github.com/GoogleCloudPlatform/ops-agent/pull/760/commits/16b3b36dd2fe40360225c2eeff81a2dde52783ec)

integration_test/third_party_apps_data/applications/active_directory_ds/metadata.yaml
```diff
@@ -88,32 +88,26 @@ expected_metrics:
   value_type: INT64
   kind: GAUGE
   monitored_resource: gce_instance
-  labels: {}
```
sophieyfang, line 91: Forgot that fsm_lib.py relies on labels always exist. Thus modifying fsm_lib.py from `labels=sorted(metric['labels'])` to `labels=sorted(metric.get('labels'))`

PR #761, opened 2022-07-27
Added Multiline Exception Detection to Python and Golang

LujieDuan:


qingling128:
This one is obsolete I guess?

LujieDuan:
> 

Yes that is correct. Deleting this now.

PR #762, opened 2022-07-28
Testing: refer to image_lists.presubmit

martijnvans:
This is instead of hardcoding the list of distros to test on presubmit in a second place.


igorpeshansky (approved):
LGTM :shipit:

PR #766, opened 2022-07-29
Turn on -test.short for presubmits

martijnvans:
This option was mistakenly dropped when transitioning the build configs from google's code silo to GitHub. This PR turns it back on.


kokoro/config/test/third_party_apps/common.gcl
```diff
@@ -3,5 +3,11 @@ import '../common.gcl' as common
 template config third_party_apps_test = common.go_test {
   params {
     test_suite = 'third_party_apps_test'
+
+    environment {
+      // For presubmits, we want to pass -test.short, which will skip
+      // most app+distro pairs.
+      SHORT = 'true'
```
Subbarker, line 10: Is this new? I don't recall seeing this before.

martijnvans: Keith introduced this feature in https://github.com/GoogleCloudPlatform/ops-agent/pull/396.

Subbarker (approved):

PR #767, opened 2022-07-29
Internal: support git-on-borg in build_package.bat

martijnvans:
This is to fix errors like:

```
& : The term 'T:\src/github/unified_agents/kokoro/scripts/build/build_package.ps1' is not recognized as the name of a cmdlet, function, script file, or operable program.
```

when running release builds. This happens because it should be "git" instead of "github" when running from git-on-borg, which is what we do for release builds.

Using `%~dp0` instead makes this script agnostic to github vs git-on-borg.


qingling128 (approved):
LGTM

PR #768, opened 2022-07-29
Refactor the `common` package and `integration_test` tags

braydonk:
The `common` package with a bunch of unrelated code is an undesirable pattern long term. Since all the code in `common` is currently related to metadata, I have changed the package from `common` to `metadata`. While I was there, I also removed the requirement for `integration_test` build tags from everything other than the actual files that contain integration tests.


martijnvans (approved):

integration_test/gce/gce_testing.go
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build integration_test
```
martijnvans, line 15: probably want to put this back because it's causing unit tests to fail (more init()  stuff)

braydonk: Yep, I'll do that. I wish we could refactor the package to not use the `init()` but it's way too big to do in this PR

PR #774, opened 2022-08-02
Presubmit check of public url in metadata.yaml

avilevy18:
Created presubmit check to test `third_party_apps_data` metadata.yaml public urls exists

http://b/238471921

### DIFF
--- a/integration_test/validate_metadata_test.go
+++ b/integration_test/validate_metadata_test.go
@@ -2,7 +2,9 @@ package integration
 
 import (
 	"embed"
+	"fmt"
 	"io/fs"
+	"net/http"
 	"os"
 	"strings"
 	"testing"
@@ -18,7 +20,42 @@ var agentMetricsMetadata []byte
 var thirdPartyDataDir embed.FS
 
 func TestValidateMetadataOfThirdPartyApps(t *testing.T) {
-	err := fs.WalkDir(thirdPartyDataDir, ".", func(path string, info fs.DirEntry, err error) error {
+	err := walkThirdPartyApps(func(contents []byte) error {
+		return validateMetadata(contents, &metadata.IntegrationMetadata{})
+	})
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestThirdPartyPublicUrls(t *testing.T) {
+	err := walkThirdPartyApps(func(contents []byte) error {
+		integrationMetadata := &metadata.IntegrationMetadata{}
+		err := validateMetadata(contents, integrationMetadata)
+		if err != nil {
+			return err
+		}
+		t.Run(integrationMetadata.ShortName, func(t *testing.T) {
+			t.Parallel()
+			r, err := http.Get(integrationMetadata.PublicUrl)
+			if err != nil {
+				t.Error(err)
+			}
+			if r.StatusCode == 404 {
+				t.Error(fmt.Sprintf("Invalid public url: %s", integrationMetadata.PublicUrl))
+			}
+			fmt.Println(r)
+		})
+		return nil
+	})
+
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func walkThirdPartyApps(fn func(contents []byte) error) error {
+	return fs.WalkDir(thirdPartyDataDir, ".", func(path string, info fs.DirEntry, err error) error {
 		if info.Name() != "metadata.yaml" {
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


integration_test/validate_metadata_test.go
```diff
@@ -18,7 +20,42 @@ var agentMetricsMetadata []byte
...
+	})
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestThirdPartyPublicUrls(t *testing.T) {
+	err := walkThirdPartyApps(func(contents []byte) error {
+		integrationMetadata := &metadata.IntegrationMetadata{}
+		err := validateMetadata(contents, integrationMetadata)
```
igorpeshansky, line 34: You shouldn't need to validate it again.

integration_test/validate_metadata_test.go
```diff
@@ -18,7 +20,42 @@ var agentMetricsMetadata []byte
 var thirdPartyDataDir embed.FS
 
 func TestValidateMetadataOfThirdPartyApps(t *testing.T) {
```
qingling128, line 22: `TestThirdPartyPublicUrls` seems to be a super set of `TestValidateMetadataOfThirdPartyApps`. Do we choose to keep them separate so that it's easier to tell what type of problem it is when things fail? aka the test failure will clearly say it's because `TestThirdPartyPublicUrls` fails.

integration_test/validate_metadata_test.go
```diff
@@ -18,7 +20,42 @@ var agentMetricsMetadata []byte
...
+	})
+
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func walkThirdPartyApps(fn func(contents []byte) error) error {
+	return fs.WalkDir(thirdPartyDataDir, ".", func(path string, info fs.DirEntry, err error) error {
 		if info.Name() != "metadata.yaml" {
```
qingling128, line 59: [Not directly related to this PR] Does `if info.Name() != "metadata.yaml" {` mean the `metadata.yaml` file is not present for this application. Do we need to error out in that case?

integration_test/validate_metadata_test.go
```diff
@@ -18,7 +20,42 @@ var agentMetricsMetadata []byte
...
+		if err != nil {
+			return err
+		}
+		t.Run(integrationMetadata.ShortName, func(t *testing.T) {
+			t.Parallel()
+			r, err := http.Get(integrationMetadata.PublicUrl)
+			if err != nil {
+				t.Error(err)
+			}
+			if r.StatusCode == 404 {
```
igorpeshansky, line 44: What if you get a 500, for example? Do you need to retry?

integration_test/validate_metadata_test.go
avilevy18: Yes, that is the case.

integration_test/validate_metadata_test.go
igorpeshansky: The function will be called for every directory entry under the given path. This conditional just means we'd skip any file that isn't named "metadata.yaml".

integration_test/validate_metadata_test.go
qingling128: Thanks for confirming.

integration_test/validate_metadata_test.go
avilevy18: `fs.WalkDir()` goes through all directories in `third_party_apps_data` and will also be called on all files that exist within each directory. There is more than just `metadata.yaml` files present so `if info.Name() != "metadata.yaml" {` serves as a filter to only address `metadata.yaml` files. 

We could put a test case that checks `third_party_apps_data/applications/<application_name>/metadata.yaml` exists.

qingling128: Ah i see. Makes sense.

Will the test fail if no `metadata.yaml` file exists?

integration_test/validate_metadata_test.go
qingling128: What about `r.StatusCode != 200`?

integration_test/validate_metadata_test.go
avilevy18: The newly purposed test would fail, if a `metadata.yaml` file does not exist. 

Currently for `walkThirdPartyApps()` if there is a missing `metadata.yaml` in a particular directory, said directory would be skipped.

integration_test/validate_metadata_test.go
avilevy18: Changed in: [92b982e](https://github.com/GoogleCloudPlatform/ops-agent/pull/774/commits/92b982e27964acb2eda56ba3376d6661306efb55)

integration_test/validate_metadata_test.go
avilevy18: [92b982e](https://github.com/GoogleCloudPlatform/ops-agent/pull/774/commits/92b982e27964acb2eda56ba3376d6661306efb55)

integration_test/validate_metadata_test.go
qingling128: SG

qingling128 (approved):
LGTM

PR #782, opened 2022-08-04
integration_test: Create a config back up when enabling an integration

ridwanmsharif:
This change adds a back up config file in the `enable` scripts for every
application. It also adds a check in the integration test suite to
fail if any new application doesn't create a back up as well.


sophieyfang:
You still need to touch every enable file so that it has the copy command. The enable file is user visible, e.g. https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/apache#example-config
The goal is the commands when copied by users, it can save a backup config in their local

sophieyfang (approved):

PR #784, opened 2022-08-05
Internal: remove deprecated distro "stretch"

martijnvans:
Turns out this distro wasn't being tested by any presubmits, and stretch support for release builds was recently removed by a change in Piper. So we don't need to refer to stretch in image_lists.gcl anymore.


qingling128 (approved):
LGTM

PR #789, opened 2022-08-08
Add Makefile, override auto test command

braydonk:
This PR is another attempt to add a Makefile to help with some development processes, mainly quickly running specific test commands and scripts. It should hopefully reduce some development toil spent trying to manually run commands, especially for testing.

Previously this did not work because `dh_auto_test` tried to detect a `make test` target and run it when we don't want to. This PR also adds `override_dh_auto_test`, since this command was not actually doing anything for us when there was no Makefile previously.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,11 @@
+test:
+	go test ./...
+
+test_confgenerator:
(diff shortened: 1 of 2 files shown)


qingling128 (approved):
LGTM modulo it's verified that it's compatible with the regular tests.

PR #790, opened 2022-08-08
Split metadata.yaml into metadata.yaml and doc.yaml

avilevy18:
Split metadata.yaml into metadata.yaml and doc.yaml

Refactored third_party_apps_test.go to work with metadata.UnmarshalMetadata()

http://b/241795483


ridwanmsharif:
Marked as obsolete on the bug.

PR #791, opened 2022-08-08
flinkreceiver: update flink labels for otel v0.57.2

JonathanWamsley:
Flink receiver `operator_name` and `garbage_collector_name` got updated to `name`. 
Otel [pr](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/c3a87a1bc6d38cea801f0c91152cd121a3f6e839) change of occurrence.

Also, mongodb had a change in which TLS setting was getting ignored. This Otel [pr](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/12093) reflects these changes. If you want this in another pr I can do so.

### DIFF
--- a/integration_test/third_party_apps_data/applications/mongodb/enable
+++ b/integration_test/third_party_apps_data/applications/mongodb/enable
@@ -11,6 +11,7 @@ metrics:
   receivers:
     mongodb:
       type: mongodb
+      insecure: true
   service:
     pipelines:
       mongo:
--- a/apps/flink.go
+++ b/apps/flink.go
@@ -48,10 +48,10 @@ func (r MetricsReceiverFlink) Pipelines() []otel.Pipeline {
 			otel.NormalizeSums(),
 			otel.MetricsTransform(
 				otel.AddPrefix("workload.googleapis.com"),
-				otel.UpdateMetric("flink.jvm.gc.collections.count", otel.RenameLabel("garbage_collector_name", "name")),
-				otel.UpdateMetric("flink.jvm.gc.collections.time", otel.RenameLabel("garbage_collector_name", "name")),
-				otel.UpdateMetric("flink.operator.record.count", otel.RenameLabel("operator_name", "name")),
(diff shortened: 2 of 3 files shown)


ridwanmsharif:
This is a backwards incompatible change is it not? Personally I would rename the processor to maintain the original names instead of changing the existing metric labels for the users of the integration.

cc @qingling128 for thoughts

qingling128:
Looks like the backwards incompatible change has been reverted. So we should be good with the label names.

integration_test/third_party_apps_data/applications/mongodb/enable
```diff
@@ -11,6 +11,7 @@ metrics:
   receivers:
     mongodb:
       type: mongodb
+      insecure: true
```
qingling128, line 14: What is the implication / reason behind this change?

schmikei: Since i authored the PR, I'll chime in here. 

There was an initial bug in the mongodb receiver where TLS settings were getting ignored in some scenarios, more details can be found in this PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/12093.

This meant that all usages of the mongodb receiver previously were using insecure client side connections which most likely would have gotten rejected by the mongodb server and preventing metrics from being collected. 

With the fix everything is updated to use proper TLS that the user supplies.

JonathanWamsley:
Hello, @ridwanmsharif  @qingling128 I believe the relevant integration tests for this pr have passed. The failed integration tests include a windows integration on the first attempt but not the second. Then also a SLES 12 integration test passed on the first attempt but failed on the second. Together all tests have passed so I think this is good but let me know if you want me to keep running kokoro tests. Thanks!

integration_test/third_party_apps_data/applications/mongodb/enable
qingling128: Got it. Thanks for the clarification!

ridwanmsharif (approved):

qingling128 (approved):
LGTM

PR #798, opened 2022-08-11
Deprecate windows-20h2-core

sophieyfang:
Deprecate windows-20h2-core as it reached EOL


sophieyfang:
ops-agent test on windows failed due to flakiness i guess? 
image

PR #802, opened 2022-08-17
Send metrics from Ops Agent main service

franciscovalentecastro:
## Description
Proof of concept on sending the `agent/ops_agent/enabled_receivers` metric from the Ops Agent main service (`google-cloud-ops-agent`). This implementation considers the future possibility of collecting and sending additional metrics obtained from the Ops Agent configuration.

The features implemented in this PR consist :
- Send metrics collected from the Ops Agent configuration to the Cloud Monitoring API.
- Register metrics to be sent every `n` minutes.
- Register metrics to be sent at different time intervals and couple Monitoring API requests when possible.

## Authors
@avilevy18 @franciscovalentecastro 

## Related issue
b/232815588

## How has this been tested?
Verified the correct collection of the `agent/ops_agent/enabled_receivers` metric in the Metrics explorer. After corroborating the feasibility of this approach, further testing and implementation of integration tests will be completed.

enabled_receivers

Screen Shot 2022-08-17 at 11 51 11 AM

## Checklist:
- Unit tests
  - [x] Unit tests will be implemented after design is approved. 
- Integration tests
  - [x] Integration tests will be implemented after design is approved.
- Documentation
- Minor version bump


cmd/ops_agent_windows/run_windows.go
```diff
@@ -163,6 +162,51 @@ func (s *service) startSubagents() error {
...
+	for _, m := range metrics {
+		tickers = append(tickers, time.NewTicker(time.Duration(m.Interval)*time.Minute))
+	}
+
+	for idx, m := range metrics {
+		go self_metrics.RegisterMetric(m, bufferChannel, tickers[idx])
+	}
+	s.log.Info(1, "registered metric tickers")
+
+	defer func() {
```
igorpeshansky, line 180: Why did you need to pull this into the function? You could create an independent goroutine that watches the ticker (and probably another channel for the exit signal).

internal/self_metrics/monitoring_api.go
```diff
@@ -0,0 +1,143 @@
...
+	// Wait for full buffer for 30 seconds
+	time.Sleep(time.Duration(30) * time.Second)
+
+	SendMetricsRequest(*buffer)
+
+	// Clear buffer
+	*buffer = make([]Metric, 0)
+}
+
+func SendMetricsEveryIntervalLinux(metrics []IntervalMetrics) error {
```
igorpeshansky, line 111: There is no reason to have this function be different on Linux and Windows… If you're worried about inconsistent death behavior, you can pass the `death` channel in as a parameter and use a different one on the two platforms.

cmd/google_cloud_ops_agent_engine/main.go
```diff
@@ -56,8 +58,21 @@ func run() error {
...
+	err = confgenerator.GenerateFilesFromConfig(&uc, *service, *logsDir, *stateDir, *outDir)
+	if err != nil {
+		return err
+	}
+	metrics, err := self_metrics.CollectOpsAgentSelfMetrics(&uc)
 	if err != nil {
 		return err
 	}
-	return confgenerator.GenerateFilesFromConfig(&uc, *service, *logsDir, *stateDir, *outDir)
+	err = self_metrics.SendMetricsEveryIntervalLinux(metrics)
```
igorpeshansky, line 72: We are still going to call this binary to generate configs, aren't we? If so, shouldn't we do this conditionally?

Sadly, because we still have this dual behavior on Linux of generating configs asynchronously when a sub-agent service starts up, the logic becomes more complex.

igorpeshansky (commented):
Some initial questions/comments.

internal/self_metrics/monitoring_api.go
franciscovalentecastro: Ok! I did it this way since I wasn't sure how to that with the different type of channels, for Linux being `os.Signal` and for Windows `svc.ChangeRequest` . Will work on unifying it in one function.

cmd/google_cloud_ops_agent_engine/main.go
franciscovalentecastro: Yes, you're right, it should go under an `if` depending on the `service` type, not on `fluent-bit` or `opentelemetry` config generation.

cmd/ops_agent_windows/run_windows.go
franciscovalentecastro: Yes, if I only pass the channel `changes` as an argument this is not needed to be inside the function. The real reason was mirroring what was already implemented in the current `Execute` implementation. This is probably linked to make a unified `SendMetricsEveryInterval` implementation for both platforms.

franciscovalentecastro:
> Some initial questions/comments.

Adressed comments.

cmd/ops_agent_windows/run_windows.go
```diff
@@ -109,21 +128,21 @@ func (s *service) checkForStandaloneAgents(unified *confgenerator.UnifiedConfig)
 	return nil
 }
 
-func (s *service) generateConfigs() error {
+func (s *service) generateConfigs() (confgenerator.UnifiedConfig, error) {
```
braydonk, line 131: It would probably be better if this returned a pointer to `confgenerator.UnifiedConfig`. `CollectOpsAgentSelfMetrics` already takes a pointed to it anyway, and you get the added benefit of returning `nil` instead of `confgenerator.UnifiedConfig{}`

internal/self_metrics/metrics.go
```diff
@@ -0,0 +1,167 @@
...
+			timestamp: time.Now().Unix(),
+		}
+
+		metricList = append(metricList, m)
+	}
+
+	return metricList
+}
+
+func GetEnabledReceivers(uc *confgenerator.UnifiedConfig) (enabledReceivers, error) {
```
avilevy18, line 122: This can be made private as it is only called in this file

systemd/google-cloud-ops-agent.service
```diff
@@ -17,11 +17,10 @@ Description=Google Cloud Ops Agent
 Wants=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opentelemetry-collector.service
 
 [Service]
-Type=oneshot
+Type=simple
```
quentinmit, line 20: Why are we doing this instead of relying on the OT Collector to write the metrics?

internal/self_metrics/metrics.go
```diff
@@ -0,0 +1,167 @@
...
+// limitations under the License.
+
+package self_metrics
+
+import (
+	"log"
+	"time"
+
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	timestamp "github.com/golang/protobuf/ptypes/timestamp"
```
braydonk, line 22: This doesn't seem necessary; the package is already `timestamp`

internal/self_metrics/metrics.go
```diff
@@ -0,0 +1,167 @@
...
+	metricpb "google.golang.org/genproto/googleapis/api/metric"
+	monitoredres "google.golang.org/genproto/googleapis/api/monitoredres"
```
braydonk, line 24: Is this package not already going to be `monitoredres`

internal/self_metrics/metrics.go
```diff
@@ -0,0 +1,167 @@
...
+	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
+)
+
+type TimeSeries interface {
+	ToTimeSeries(instance_id, zone string) []*monitoringpb.TimeSeries
+}
+
+type Metric struct {
+	metric_type   string
```
braydonk, line 33: We use `camelCase` rather than `snake_case` for struct property and variable names usually.

internal/self_metrics/metrics.go
```diff
@@ -0,0 +1,167 @@
+// Copyright 2020 Google LLC
```
braydonk, line 1: This should be copyright 2022

internal/self_metrics/monitoring_api.go
```diff
@@ -0,0 +1,137 @@
+// Copyright 2020, Google Inc.
```
braydonk, line 1: This should be copyright 2022

braydonk (commented):
Two general things:

The `self_metrics` package, at least the `metrics.go` file, should be unit tested if we want this to be merged.

Secondly, I'd be curious to see why this approach with writing directly to the monitoring API was chosen over using the Otel collector to do this. Was there a design doc?

systemd/google-cloud-ops-agent.service
franciscovalentecastro: The decision was taken due to the following arguments :
- If the OTel collector goes down, we can keep sending this metrics (also if there other health related metrics).
- A push to have the main service have more responsibility.

Still, it is a big change in the architecture, so there are a couple drawbacks.

I agree with the `simple` argument, `systemd` needs someway to know about the validation to correctly manage the subagents.

sophieyfang: Re:`Why are we doing this instead of relying on the OT Collector to write the metrics?`
### There are two parts for implementing this metrics.
### Part 1 - parsing and composing metrics
Parsing the config.yaml and composing the enabled_receivers timeseries shall really be done by the ops-agent binary rather the otel-ops-collector i think? `otel-ops-collector` as of today, it's a separate upstream repo that's not bundled with ops-agent(thought ops-agent seems the only one using it).  Thus there must be a binary running whenever ops-agent gets restarted that parses and composes this metric. Also this enabled_receiver metric is a gauge, thus it needs to be sent out every x interval. Therefore, it needs to maintain that running and ships this timeseries every x interval.

### Part two - send metrics
This PR is sending metrics directly into Monitoring API. One of the reason is this diagnostic metrics is tied to the config itself rather than any subagents, so whether the sub agents run or not shall not affect this metric being sent to the monitoring API.

Also grab @braydonk here who also mentioned using OT collector on writing the metrics. I always wonder what's the implied implementation for this option. How to achieve it.

quentinmit: I'm suggesting that you calculate what the timeseries should be as part of config generation and bake it into the generated collector config, relying on the collector to do what it's designed to do and report the metrics upstream. e.g. it looks like you could use `otlpjsonfilereceiver` to accomplish this.

This PR reimplements a bunch of what the collector does, which means you're potentially introducing new and different bugs, and now you need to keep it in sync with whatever the collector is doing.

braydonk: re: `what's the implied implementation for this option`

There's two options I see:

**Option 1: Ops Agent Receiver**

We'd make a configurable `opsagentreceiver` in `opentelemetry-operations-collector` that reads the Ops Agent config on the system and does similar calculations as here.

**Option 2: Static Value Receiver**

We could make a `staticvaluereceiver` that is configured to emit a set of values every collection interval. This would allow us to calculate the values and embed them in the otel config.

Personally I'm partial to option 1, since it's more in line with the way otel receivers are usually designed, but I'm open to dicussion on that (GitHub comments is probably not the best medium for those discussions though, we can summarize on this thread when we discuss more holistically elsewhere).

quentinmit: What's wrong with the existiing `otlpjsonfilereceiver` in lieu of option 2?

We should not do option 1 as-is because it's important that the configuration only gets read and parsed once. Otherwise what the metrics report as the running config might not match what's actually running (e.g. if the user edits the config files). So whatever we report from the collector, it needs to come from information that was assembled by the config generation. Whether we use `otlpjsonfilereceiver` or build our own generic receiver or build our own specific receiver I don't think matters much, but minimizing the code complexity is a valuable goal.

franciscovalentecastro: From all the options using the opentelemetry collector, the one using `otlpjsonfilereceiver` is probably the simplest, straightforward and with minimum code complexity. The only hypothetical downside is that this sets values statically, but for our current "new metrics" (enabled_receivers, feature_tracking) this is not an issue. 

I've already synced with @braydonk about it and now I have a clear idea on how to implement it. I documented all this opinions and options in the design document. Still looking if it's possible to use the Opentelemetry SDK.

PR #811, opened 2022-08-24
confgenerator: use golden package in unit tests

braydonk:
## Description
The [gotesttools package for golden tests](https://pkg.go.dev/gotest.tools/v3@v3.3.0/golden) lines up perfectly with the
golden tests that we currently do, and using it can greatly simplify our
unit test code. This PR reimagines the confgenerator_test.go tests using
this package. A few other minor changes were made:

* The invalid type for sqlserver test did not have an input.yaml (this
  new way of doing tests actually caught that; it's been that way
  forever)
* It never made much sense for `golden_otel` to have an extension of
  `conf`, this PR switches that to `yaml`
* Since we install `gotesttools` in this PR for `golden`, just opted to use the `assert` package as well

## Related issue
None yet, since I wanted to post this as RFC first. :slightly_smiling_face: 
EDIT: Created a bug for this. b/243791505

## How has this been tested?
Verbose test run: https://pastebin.com/SahVB1Kg
The `-update` flag worked as expected, as evidenced by the updated files in the PR.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


sophieyfang (approved):

qingling128:
Can we add some details (e.g. command output, transcripts, could be a file) for the `How has this been tested?` section for the records? BTW, have we gotten a chance to test the `-update_golden` flag with this change?

confgenerator/confgenerator_test.go
```diff
@@ -114,6 +83,52 @@ func testPlatformGenerateConfTests(t *testing.T, platform platformConfig) {
...
+	platform platformConfig,
+	testTypeDir string,
+	errAssertion func(*testing.T, error, string),
+) {
+	platformTestDir := filepath.Join(testTypeDir, platform.OS)
+	testNames := getTestsInDir(t, platformTestDir)
+
+	for _, testName := range testNames {
+		t.Run(testName, func(t *testing.T) {
+			testDir := filepath.Join(platformTestDir, testName)
```
avilevy18, line 127: Maybe `t.Parallel()`?

braydonk: Added in [e3d7607](https://github.com/GoogleCloudPlatform/ops-agent/pull/811/commits/e3d7607aea8abf17cea4eb14c04ff3d2a8fcd85b)

quentinmit:
I'm trying to handle a merge conflict on this PR - I'm wondering why this refactoring added so much duplicate code? The `golden` package is nice but it seems like it was coupled with unrolling a bunch of loops?

braydonk:
@quentinmit The unrolling of loops was to add some robustness to the tests.  
In the previous version, `valid` and `invalid` tests weren't technically treated any differently. If an invalid config was placed in the `valid` folder or vice versa the previous tests didn't care. There were some actual test cases before this PR that were afflicted by this. Now they are properly treated differently, i.e. if an invalid config is put in a valid test case folder or vice versa the test will fail.
When Lua files changed in the previous version there was no cleanup. Now, the case where they are changed includes cleanup of the old Lua files.  
The previous version of the tests checked for a test name that contained "builtin" which I thought would make more sense to split into its own test run rather than needing to check and do something different on that case. A lot of folks also use that builtin test case a lot to check what the default config generation is, so separating it into its own folder makes it more convenient to find (pretty minor thing compared to everything else).

PR #816, opened 2022-08-26
Shorten and fix incorrect capitalization in Vault integration description.

jkschulz:


integration_test/third_party_apps_data/applications/vault/metadata.yaml
```diff
@@ -2,7 +2,7 @@ app_url: "https://www.vaultproject.io/"
 short_name: Vault
 long_name: Hashicorp Vault
 description: |-
-  Vault is an identity-based secrets and encryption management system. This integration Collects Vault's audit logs. The integration also collects token, memory, and storage metrics.
+  Vault is an identity-based secrets and encryption management system. This integration collects Vault's audit logs, and token, memory, and storage metrics.
```
ridwanmsharif, line 5: extra `and` can also be removed?

ridwanmsharif (approved):

integration_test/third_party_apps_data/applications/vault/metadata.yaml
jkschulz: if they're in one sentence they shouldn't be, since they're nested 'and's: ((audit) logs) and ((token, memory, and storage) metrics)

if you don't think it reads well, i can split back out into two sentences though!

ridwanmsharif: The explanation makes sense to me. LGTM 🚀

igorpeshansky:
This was never merged. The capitalization bit seems to have since been (silently) taken care of by #817. The rest of the text is consistent with other integrations, so closing.

PR #818, opened 2022-08-26
Grab labels of pull request

avilevy18:
## Description
Added a script that gets the  up to date labels of a PR. 

## Related issue
b/243975669

## How has this been tested?
Tested through git commits and pushes to this PR. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


.github/workflows/pr_presubmit.yaml
```diff
@@ -18,6 +18,7 @@ jobs:
       with:
         ref: ${{ github.base_ref }}
         path: base_checkout
-      # `github.event.pull_request.labels.*.name` gives us the names of all labels for the PR
-    - run: go test ./ci -added="${{ steps.files.outputs.added }}" -modified="${{ steps.files.outputs.modified }}" -labels=${{ join(github.event.pull_request.labels.*.name, ' ') }} -base-dir="base_checkout" -tags=ci
-
+      # `steps.get_labels.outputs.*` gives us the names of all labels for the PR
+    - run: echo "::set-output name=labels::$(go run ci/get_labels/main.go -url=${{ github.event.pull_request.url }})"
```
braydonk, line 22: You can run the whole package here.
```suggestion
    - run: echo "::set-output name=labels::$(go run ci/get_labels -url=${{ github.event.pull_request.url }})"
```

ci/get_labels/main.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+var prUrl = flag.String("url", "", "Url of pull request")
+
+func main() {
+	flag.Parse()
+	labelsUrl := strings.Replace(*prUrl, "/pulls/", "/issues/", 1)
```
braydonk, line 29: A comment here explaining why this replace is necessary would be nice

ci/get_labels/main.go
```diff
@@ -0,0 +1,53 @@
...
+	labelsUrl += "/labels"
+	resp, err := http.Get(labelsUrl)
+	if err != nil {
+		panic(err)
+	}
+	defer resp.Body.Close()
+
+	type Labels struct {
```
braydonk, line 37: Would prefer this type to be defined at the top or for this to be an anonymous struct in the declaration on line 41. Probably also doesn't need to be capitalized

ci/get_labels/main.go
```diff
@@ -0,0 +1,53 @@
...
+
+	var responseLabels []Labels
+	err = json.NewDecoder(resp.Body).Decode(&responseLabels)
+	if err != nil {
+		panic(err)
+	}
+
+	var output string
+	for _, label := range responseLabels {
+		output += label.Name + " "
```
braydonk, line 49: Up to you, but it might be nicer here instead to just `fmt.Printf("%s ", label.Name)` instead of using an output variable and printing at the end.

avilevy18: Changed in: [8d1c90b](https://github.com/GoogleCloudPlatform/ops-agent/pull/818/commits/8d1c90b269b5dd22a9b4ee640a6a51609a31c97f)

ci/get_labels/main.go
```diff
@@ -0,0 +1,64 @@
...
+
+	labelCollector := LabelCollector{Client: &http.Client{}}
+	fmt.Println(labelCollector.GetLabels(*prUrl))
+}
+
+type LabelCollector struct {
+	Client *http.Client
+}
+
+func (l *LabelCollector) GetLabels(prUrl string) string {
```
braydonk, line 42: Optional: If I were writing this, I would have this function return an error, then `log.Fatal(err)` in main if it's not nil. Two reasons:

* I'm in favour of `log.Fatal` over a panic for a program like this that's more of a script, the panic rarely provides useful info
* I think it's nice to have the `main` function be the only one in charge of exiting the program

But what you have here is fine, I think you can leave it as is if you'd like. (I did already technically approve this part before so it's obviously fine enough lol)

ci/get_labels/main.go
```diff
@@ -0,0 +1,64 @@
...
+	"encoding/json"
+	"flag"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+var prUrl = flag.String("url", "", "Url of pull request")
+
+type respLabels struct {
```
braydonk, line 27: Optional: What might be good is making a `type RespLabelCollection []RespLabel` then make that type implement the `Stringer` interface (i.e. make a `func (lc RespLabelCollection) String() string`). Then you can return a `RespLabelCollection` here and simply `fmt.Println(collection)`.

ci/get_labels/main.go
```diff
@@ -0,0 +1,64 @@
...
+	Name string
```
braydonk, line 28: Although this is working right now because of the default JSON Unmarshalling behaviour, I'd suggest putting the `json` struct tag anyway.
```suggestion
	Name string `json:"name"`
```

ci/get_labels/main_test.go
```diff
@@ -0,0 +1,38 @@
...
+			t.Fatal(err)
+		}
+	}))
+
+	defer server.Close()
+	labelCollector := main.LabelCollector{
+		Client: server.Client(),
+	}
+
+	actual := labelCollector.GetLabels(server.URL + "/GoogleCloudPlatform/ops-agent/pulls/818")
```
braydonk, line 33: If you switch to returning error from `GetLabels` then you can check for errors here instead of dealing with a panic.

braydonk (approved):

ci/get_labels/main_test.go
```diff
@@ -0,0 +1,89 @@
...
+				},
+				{
+					Name: "foo",
+				},
+			},
+		},
+	}
+
+	for _, test := range testTable {
+		test := test
```
braydonk, line 36: This might not be necessary because this immediate `t.Run` isn't marked parallel.

ci/get_labels/main_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	actual, err := labelCollector.GetLabels(server.URL + "/GoogleCloudPlatform/ops-agent/pulls/818")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assertRespLabelCollection(t, expected, actual)
+}
+
+func assertRespLabelCollection(t *testing.T, expected main.RespLabelCollection, actual main.RespLabelCollection) {
```
braydonk, line 59: Maybe `assertRespLabelCollectionEqual` just to convey that you're asserting equality

braydonk (approved):
Super nitpick comments, otherwise looks great!

PR #821, opened 2022-08-30
Send metrics from Ops Agent main service with OTel SDK

franciscovalentecastro:
## Description
Proof of concept on sending the `agent/ops_agent/enabled_receivers` metric from the Ops Agent main service (`google-cloud-ops-agent`) using the Opentelemetry SDK library.

Note : This POC is a continuation of the work in #802.

## Related issue
b/232815588

## How has this been tested?
Verified the correct collection of the `agent/ops_agent/enabled_receivers` metric in the Metrics explorer. After corroborating the feasibility of this approach, further testing and implementation of integration tests will be completed.

Screen Shot 2022-08-30 at 2 04 31 PM
Screen Shot 2022-08-30 at 2 04 36 PM

## Checklist:
- Unit tests
  - [x] Unit tests will be implemented after design is approved. 
- Integration tests
  - [x] Integration tests will be implemented after design is approved.
- Documentation
- Minor version bump


internal/self_metrics/metrics.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return fmt.Errorf("Failed to establish pipeline: %v", err)
+	}
+	defer pusher.Stop(ctx)
+
+	// Start meter
+	meter := pusher.Meter("ops_agent/self_metrics")
+
+	// Collect GAUGE metric
+	gaugeObserver, err := meter.AsyncInt64().Gauge("agent/ops_agent/enabled_receivers")
```
sophieyfang, line 103: For our case i think we could just use SyncInt64()? 🤔 @dashpole to confirm

franciscovalentecastro: I tried using `SyncInt64()`, but it doesn't have support for `GAUGE` metrics ([link](https://github.com/open-telemetry/opentelemetry-go/blob/18f4cb85ece82b12cb9bd9af02efe2a47bd8f76e/metric/instrument/syncint64/syncint64.go) ) but `AsyncInt64()` does have it ([link](https://github.com/open-telemetry/opentelemetry-go/blob/18f4cb85ece82b12cb9bd9af02efe2a47bd8f76e/metric/instrument/asyncint64/asyncint64.go))

dashpole: Ah, thats right.  AsyncInt64 is probably what you want here.  If you don't want to check something too often, you should be able to implement something inside the callback that returns the previous value if it hasn't been long enough.  But its probably better to start with fetching it each time the metric is sent to start...

franciscovalentecastro: Ok! That makes sense. I think the way it is implemented right now, there is no further way to do more precalculations for the metric, it would stay this way (looping through every value and labels combination to call the Observe() method). Thanks for clarifying this.

PR #837, opened 2022-09-08
Fetch metadata from GCE VM Metadata Server

LujieDuan:
## Description
This change introduces a new package in confgenerator called `resourcedetector` to detect metadata of the environment Ops Agent is running in. The detected metadata can then be used by receivers/processors. Right now we only have a detector for GCE environment; the detector is not currently used by any components; this will be used by the Prometheus receivers.

## Related issue
b/244434005

## How has this been tested?
This has been tested with unit tests and manual tests on GCE. 
Integration tests for this is TBD. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/confgenerator/resourcedetector/detector.go
+++ b/confgenerator/resourcedetector/detector.go
@@ -4,12 +4,16 @@ import (
 	gcp_metadata "cloud.google.com/go/compute/metadata"
 )
 
+// An implementation of the Resource interface will have fields represent
+// available attributes about the current monitoring resource.
 type Resource interface {
 	GetType() string
 }
 
-// Get a resource instance for the current environment, which can be used
-// to get all the attributes
+// Get a resource instance for the current environment;
+// In order to access the attributes of a specific type of resource,
+// needs to cast the returned Resource instance to its underlying type:
+// actual, ok := resource.(GCEResource)
 func GetResource() (Resource, error) {
 	switch {
 	case gcp_metadata.OnGCE():
--- a/confgenerator/resourcedetector/detector.go
+++ b/confgenerator/resourcedetector/detector.go
@@ -4,30 +4,30 @@ import (
 	gcp_metadata "cloud.google.com/go/compute/metadata"
 )
 
-type Detector interface {
+type Resource interface {
 	GetType() string
 }
 
-// Get a Detector for the current environment, which can be used
+// Get a resource instance for the current environment, which can be used
 // to get all the attributes
-func GetDetector() (Detector, error) {
+func GetResource() (Resource, error) {
 	switch {
 	case gcp_metadata.OnGCE():
-		return GetGCEDetector()
+		return GetGCEResource()
 	default:
-		return GetUnrecognizedPlatformDetector()
+		return GetUnrecognizedPlatformResource()
 	}
 }
 
-// UnrecognizedPlatformDetector that returns an empty detector without any attributes
+// UnrecognizedPlatformResource that returns an empty resource instance without any attributes
 // for unrecognized environments
-type UnrecognizedPlatformDetector struct {
+type UnrecognizedPlatformResource struct {
 }
 
-func (UnrecognizedPlatformDetector) GetType() string {
+func (UnrecognizedPlatformResource) GetType() string {
 	return "unrecognized platform"
 }
 
-func GetUnrecognizedPlatformDetector() (Detector, error) {
-	return UnrecognizedPlatformDetector{}, nil
+func GetUnrecognizedPlatformResource() (Resource, error) {
+	return UnrecognizedPlatformResource{}, nil
 }
--- a/confgenerator/resourcedetector/detector.go
+++ b/confgenerator/resourcedetector/detector.go
@@ -1,0 +1,33 @@
+package resourcedetector
+
+import (
+	gcp_metadata "cloud.google.com/go/compute/metadata"
+)
+
+type Detector interface {
+	GetType() string
+}
+
+// Get a Detector for the current environment, which can be used
+// to get all the attributes
+func GetDetector() (Detector, error) {
+	switch {
+	case gcp_metadata.OnGCE():
+		return GetGCEDetector()
+	default:
+		return GetUnrecognizedPlatformDetector()
+	}
+}
+
+// UnrecognizedPlatformDetector that returns an empty detector without any attributes
+// for unrecognized environments
+type UnrecognizedPlatformDetector struct {
+}
+
+func (UnrecognizedPlatformDetector) GetType() string {
+	return "unrecognized platform"
+}
+
+func GetUnrecognizedPlatformDetector() (Detector, error) {
+	return UnrecognizedPlatformDetector{}, nil
+}
--- a/confgenerator/resourcedetector/gce_detector.go
+++ b/confgenerator/resourcedetector/gce_detector.go
@@ -1,0 +1,128 @@
+package resourcedetector
+
+type gceAttribute int
+
+const (
+	project gceAttribute = iota
+	zone
+	network
+	subnetwork
+	publicIP
+	privateIP
+	instanceID
+	instanceName
+	tags
+	machineType
+	metadata
+	label
+	interfaceIPv4
+)
+
+func GetGCEDetector() (Detector, error) {
+	provider := NewGCEMetadataProvider()
+	dt := GCEDetectorBuilder{provider: provider}
+	return dt.GetDetector()
+}
+
+// The data provider interface for GCE environment
+// Implementation of this provider can use either the metadata server on VM,
+// or the cloud API
+type gceDataProvider interface {
+	getProject() (string, error)
+	getZone() (string, error)
+	getNetwork() (string, error)
+	getSubnetwork() (string, error)
+	getPublicIP() (string, error)
+	getPrivateIP() (string, error)
+	getInstanceID() (string, error)
+	getInstanceName() (string, error)
+	getTags() (string, error)
+	getMachineType() (string, error)
+	getLabels() (map[string]string, error)
+	getMetadata() (map[string]string, error)
+	getInterfaceIPv4s() (map[string]string, error)
+}
+
+// List of single-valued attributes (non-nested)
+var singleAttributeSpec = map[gceAttribute]func(gceDataProvider) (string, error){
+	project:      gceDataProvider.getProject,
+	zone:         gceDataProvider.getZone,
+	network:      gceDataProvider.getNetwork,
+	subnetwork:   gceDataProvider.getSubnetwork,
+	publicIP:     gceDataProvider.getPublicIP,
+	privateIP:    gceDataProvider.getPrivateIP,
+	instanceID:   gceDataProvider.getInstanceID,
+	instanceName: gceDataProvider.getInstanceName,
+	tags:         gceDataProvider.getTags,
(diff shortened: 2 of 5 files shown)


confgenerator/resourcedetector/gce_detector.go
```diff
@@ -0,0 +1,125 @@
+package resourcedetector
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Keep synced with the Prometheus repo
+const (
```
ridwanmsharif, line 9: I would suggest keeping this detector generic and leave all the translation to the prometheus style labels to the `prometheus` receiver. This detector is useful for other components too. One such example is: [b/219991181](http://b/219991181) | P3 | Allow LogName to be used in processors

confgenerator/resourcedetector/gce_detector.go
```diff
@@ -0,0 +1,125 @@
...
+
+// Nested property takes two steps: first to get the list of keys,
+// and then use the key to get individual values.
+// Note that the key are usually used to construct the final label
+type nestedProperties struct {
+	first func(GCEDataProvider) (string, error)
+	then  func(GCEDataProvider, string) (string, string, error)
+}
+
+// List of single value labels (non-nested)
```
ridwanmsharif, line 56: `labels` feels like a prometheus specific thing to me. It can get confusing especially since the metadata has its own `labels` too. Naming it instead to something like `metadata` or `attributes` might make it clearer.

confgenerator/resourcedetector/gce_detector.go
```diff
@@ -0,0 +1,125 @@
...
+	gceLabelProject:      GCEDataProvider.getProject,
+	gceLabelZone:         GCEDataProvider.getZone,
+	gceLabelNetwork:      GCEDataProvider.getNetwork,
+	gceLabelSubnetwork:   GCEDataProvider.getSubnetwork,
+	gceLabelPublicIP:     GCEDataProvider.getPublicIP,
+	gceLabelPrivateIP:    GCEDataProvider.getPrivateIP,
+	gceLabelInstanceID:   GCEDataProvider.getInstanceID,
+	gceLabelInstanceName: GCEDataProvider.getInstanceName,
+	gceLabelTags:         GCEDataProvider.getTags,
+	gceLabelLabel:        GCEDataProvider.getLabel,
```
ridwanmsharif, line 67: Is this a single label or nested? I thought it was a nested one but could be wrong.

confgenerator/resourcedetector/gce_detector.go
```diff
@@ -0,0 +1,125 @@
...
+	gceLabelMetadata      = gceLabel + "metadata_"
+	gceLabelLabel         = gceLabel + "label_"
+	gceLabelMachineType   = gceLabel + "machine_type"
+	gceLabelInterfaceIPv4 = gceLabel + "interface_ipv4_"
+)
+
+// The data provider interface for GCE environment
+// Implementation of this provider can use either the metadata server on VM,
+// or the cloud API
+type GCEDataProvider interface {
```
ridwanmsharif, line 30: We should add some comments here around what is available in the detector now and what is not

confgenerator/resourcedetector/detector.go
```diff
@@ -0,0 +1,45 @@
...
+	default:
+		return EmptyDetector{}
+	}
+}
+
+type LabelName string
+
+type LabelValue string
+
+type Metalabels map[LabelName]LabelValue
```
ridwanmsharif, line 28: The contract we expose should probably be more clear - instead of having `map[string]string` it could be a struct that has fields for all the attributes. This way it is part of the public interface what fields are available as part of the detector, instead of maintaining another set of key labels.

confgenerator/resourcedetector/gce_metadata_provider.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (gmp *GCEMetadataProvider) getInterfaceNames() (string, error) {
+	return gmp.client.Get("instance/network-interfaces/")
+}
+
+// The metadata server would return interfaces as "0/", needs to trim the "/"
+// Also add `nic` to the name following the Prometheus's SD convention
+func (gmp *GCEMetadataProvider) getInterfaceIPv4(name string) (string, string, error) {
+	interface_name := strings.TrimSuffix(name, "/")
+	ipv4, err := gmp.client.Get(fmt.Sprintf("instance/network-interfaces/%s/ip", interface_name))
```
ridwanmsharif, line 90: These are very interesting mutations we have to do. I'd leave it out of the detector package and add it when the `prometheus` receiver uses it. Only it cares about the Prometheus SD's conventions.

ridwanmsharif (commented):
Overall this looks pretty solid. There are a few transformations and nuances here around differences from how the gce detector in prometheus collects things (Compute engine API) and how we do (metadata server). We should add an integration test in addition to the unit tests in this PR to make sure the metadata attributes look like what we expect for the `prometheus` use case.

confgenerator/resourcedetector/detector.go
```diff
@@ -0,0 +1,33 @@
...
+	case gcp_metadata.OnGCE():
+		return GetGCEDetector()
+	default:
+		return GetEmptyDetector()
+	}
+}
+
+// EmptyDetector that returns an empty detector without any attributes
+// for non-recognized environments
+type EmptyDetector struct {
```
LujieDuan, line 24: Not sure if it would be better to call this `UnrecognizedPlatformDetector` or something similar.

ridwanmsharif: +1 for `UnrecognizedPlatformDetector`.

confgenerator/resourcedetector/gce_detector.go
```diff
@@ -0,0 +1,159 @@
...
+	getLabel(key string) (string, error)
+	getMetadataKeys() ([]string, error)
+	getMetadata(key string) (string, error)
+	getInterfaceNames() ([]string, error)
+	getInterfaceIPv4(name string) (string, error)
+}
+
+// Nested attribute takes two steps: first to get the list of keys,
+// and then use the key to get individual values.
+type nestedAttribute struct {
```
ridwanmsharif, line 55: This struct and interface seems pretty confusing to me. It feels simpler to have the provider do both the fetching of the keys and then aggregate the values rather than having the detector work with this interface.

In other words, if the provider provided `getMetadata`, getLabels` and getLabels` methods that each returned a `map[string]string`, the GCE detector wouldn't need to do any of this nesting and unnesting logic right? That would be contained in the implementation of the `getMetadata`, getLabels` and getLabels` methods. Let me know what you think of this approach.

confgenerator/resourcedetector/gce_detector_test.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	if err.Error() != "some error" {
+		t.Errorf("should have return the correct error message")
+	}
+	if actual != nil {
+		t.Errorf("should return nil detector when having error")
+	}
+}
+
+func TestGettingDetectorOnGCE(t *testing.T) {
```
ridwanmsharif, line 46: This feels like it belongs in the `integration_test` folder right?


Also you will be able to test a lot more of the fields there since we have the VM information when the integration test sets it up:
https://github.com/GoogleCloudPlatform/ops-agent/blob/e7aa72d2b077a131440bdc999e6b7f36aba548be/integration_test/gce/gce_testing.go#L267-L282

See this as an example of what is used when setting up the environment for an integration test:
https://github.com/GoogleCloudPlatform/ops-agent/blob/e7aa72d2b077a131440bdc999e6b7f36aba548be/integration_test/ops_agent_test.go#L346

ridwanmsharif (commented):
This is looking very good! Left a couple comments

confgenerator/resourcedetector/gce_detector_test.go
ridwanmsharif: I don't know what the best way to integration test this. Maybe the easiest way is to have a small go command line runner that hits your interface and dumps the metadata into a json file. The integration test can then read that from the VM and compare it to what it expects. Just a thought, figured its worth sharing. Looking forward to whatever you end up going with.

confgenerator/resourcedetector/gce_metadata_provider.go
```diff
@@ -0,0 +1,112 @@
...
+		metadata[key] = val
+	}
+	return metadata, nil
+}
+
+func (gmp *GCEMetadataProvider) getMetadataValue(key string) (string, error) {
+	return gmp.client.Get(fmt.Sprintf("instance/attributes/%s", key))
+}
+
+// GCE VM labels are currently not available in the metadata server
```
ridwanmsharif, line 91: Can we create a bug for this and link the bug ID here as a TODO?

confgenerator/resourcedetector/gce_detector.go
```diff
@@ -0,0 +1,128 @@
...
+
+// List of nested attributes
+var nestedAttributeSpec = map[gceAttribute]func(gceDataProvider) (map[string]string, error){
+	metadata:      gceDataProvider.getMetadata,
+	interfaceIPv4: gceDataProvider.getInterfaceIPv4s,
+	label:         gceDataProvider.getLabels,
+}
+
+// GCEDetector implements the Detector interface and provide attributes of the VM when on GCE
+type GCEDetector struct {
```
ridwanmsharif, line 68: [nit] naming feels a bit off to me. `GCEDetector.Project` sounds like the project of the detector and not the metadata of the VM that is being detected. I'm struggling to come up with a better name/interface that works for both GCE and non-GCE resource detection however. Maybe we can just do the simple thing and rename this and `Detector` to `Resource` and the method `GetDetector` to `GetResource`. And we can add a comment that in GCE VM land the resource is the VM metadata.

It is a nit pick tho. Feel free to ignore.

ridwanmsharif (approved):
LGTM. I'd get another reviewer however since we were pretty involved in this

confgenerator/resourcedetector/detector.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	gcp_metadata "cloud.google.com/go/compute/metadata"
+)
+
+type Detector interface {
+	GetType() string
+}
+
+// Get a Detector for the current environment, which can be used
+// to get all the attributes
```
ridwanmsharif, line 12: We should also add what it needs to be casted to (for example `GCEDetector`)

confgenerator/resourcedetector/gce_metadata_provider.go
LujieDuan: I have created two bugs for the missing labels and subnetwork attributes from the metadata server.

confgenerator/resourcedetector/gce_detector.go
LujieDuan: I think this is a good idea - I have changed the name to `Resource`.

confgenerator/resourcedetector/detector.go
```diff
@@ -0,0 +1,33 @@
+package resourcedetector
+
+import (
+	gcp_metadata "cloud.google.com/go/compute/metadata"
+)
+
+type Detector interface {
```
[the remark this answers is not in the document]
LujieDuan: Added comments to what the interface means and how to cast to get an instance of the underlying type such as GCEResource.

ridwanmsharif (approved):
Wooo LGTM!

PR #853, opened 2022-09-14
Feature extraction by struct tags

avilevy18:
## Description
Feature extraction for feature tracking implemented by struct tags.

## Related issue
b/246809450

## How has this been tested?
Unit Tests

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+
+	var features []Feature
+	// Iterate over all available fields and read the tag value
+	for i := 0; i < t.NumField(); i++ {
+		// Get the field, returns https://golang.org/pkg/reflect/#StructField
+		field := t.Field(i)
+
+		// Grab actual value instead of pointer
+		v := reflect.Indirect(reflect.Indirect(c).FieldByName(field.Name))
+		if v.IsZero() {
```
LujieDuan, line 78: I am not sure about `v.IsZero()` here - my understanding is `false` is the zero value for bool? If the value we want to track is bool and is false, we would skip it incorrectly, right?

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+
+	return features, nil
+}
+
+func asString(v interface{}) (string, error) {
+	var s string
+	switch x := v.(type) {
+	case bool:
+		s = fmt.Sprintf("%t", x)
+	case int:
```
LujieDuan, line 139: I think we have a couple config taking int64 - would it make sense to have this one here as `case Int, Int8, Int16, Int32, Int64:`?

LujieDuan (commented):
This looks very good! The tests are nice and the feature extracting is nice and clean. I left a couple of questions and small suggestions.

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+var ErrTrackingInlineStruct = errors.New("cannot have tracking on inline struct")
+
+type Feature struct {
```
quentinmit, line 12: Please document the exported items in this package, because I can't tell how any of this is expected to be used just from the field and function names.

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+	Module, Kind, Type, Key, Value string
+}
+
+func ExtractFeatures(uc *UnifiedConfig) ([]Feature, error) {
```
quentinmit, line 16: If you're already doing the work to generically use reflection, why not use reflection for `ExtractFeatures` itself? You can just walk the entire config struct and expose any matching field generically with a key like

`Logging.Receivers.2[apache_access].WildcardRefreshInterval`

(i.e. use the full struct path as the key)

Then this function goes away, and you don't need any names in the struct tags. Since this is only for our own internal use, it's not really a problem if the names are less friendly.

(I would also suggest identifying fields to export based on the type of the field, i.e. int and bool fields are always (?) not PII, so don't need any tags.)

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+		tags := strings.Split(trackingTag, ",")
+		trackingName := tags[0]
+
+		// Check to see if value needs to be overridden
+		var overridden string
+		if len(tags) > 1 {
+			overridden = tags[1]
+		}
+		yamlTag := field.Tag.Get("yaml")
+		isInline := strings.Contains(yamlTag, ",inline")
```
quentinmit, line 93: Why bother handling some inline fields specially? Why not just expose them all? That will still produce a predictable name, which is all that matters for our debugging.

(I'm worried about trying to parse a different package's struct tag like this, because we might not have the same parsing behavior as they do.)

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+
+	if uc.Metrics != nil {
+		for _, r := range uc.Metrics.Receivers {
+			features, err := trackingFeatures(r, MetricsReceiverTypes.Subagent, MetricsReceiverTypes.Kind, r.Type())
+			if err != nil {
+				return nil, err
+			}
+			allFeatures = append(allFeatures, features...)
+		}
+		for _, p := range uc.Metrics.Processors {
```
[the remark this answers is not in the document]
avilevy18: We do not want to track any pipelines, we only would like to track if the `default_pipeline` has been modified

confgenerator/feature.go
avilevy18: The zero value of a feature can be inferred and therefore it does not need to be tracked.

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+var ErrTrackingInlineStruct = errors.New("cannot have tracking on inline struct")
+
+type Feature struct {
+	Module, Kind, Type, Key, Value string
```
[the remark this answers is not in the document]
avilevy18: Changed in: [0f93735](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/0f937356b0edf460a8aa421e4e0398dfc4c018f8)

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+				return nil, err
+			}
+			allFeatures = append(allFeatures, features...)
+		}
+	}
+
+	return allFeatures, nil
+}
+
+func trackingFeatures(component interface{}, module, kind, typ string) ([]Feature, error) {
```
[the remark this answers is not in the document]
avilevy18: Changed in: [0f93735](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/0f937356b0edf460a8aa421e4e0398dfc4c018f8)

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+				Key:    key,
+				Value:  featureTrackingString,
+			})
+		}
+	}
+
+	return features, nil
+}
+
+func asString(v interface{}) (string, error) {
```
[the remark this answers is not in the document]
avilevy18: Function was removed in: [0f93735](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/0f937356b0edf460a8aa421e4e0398dfc4c018f8)

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+			if trackingTag == "" {
+				continue
+			}
+			featureTrackingString, err := asString(v.Interface())
+			if err != nil {
+				return nil, err
+			}
+			if overridden != "" {
+				featureTrackingString = overridden
+			}
```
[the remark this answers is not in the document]
avilevy18: Changed in: [0f93735](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/0f937356b0edf460a8aa421e4e0398dfc4c018f8)

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+		var overridden string
+		if len(tags) > 1 {
+			overridden = tags[1]
+		}
+		yamlTag := field.Tag.Get("yaml")
+		isInline := strings.Contains(yamlTag, ",inline")
+
+		// Append prefix with dot notation for child features key
+		key := fmt.Sprintf("%s.%s", prefix, trackingName)
+		if prefix == "" || isInline {
```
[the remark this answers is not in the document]
avilevy18: No longer applicable in: [0f93735](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/0f937356b0edf460a8aa421e4e0398dfc4c018f8)

confgenerator/feature.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+			allFeatures = append(allFeatures, features...)
+		}
+	}
+
+	return allFeatures, nil
+}
+
+func trackingFeatures(component interface{}, module, kind, typ string) ([]Feature, error) {
+	return trackingFeaturesHelper(component, module, kind, typ, "")
```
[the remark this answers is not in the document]
avilevy18: `Feature.Key` has been refactored as `[]string` and will build a feature object recursively

confgenerator/feature.go
avilevy18: Changed in: [0f93735](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/0f937356b0edf460a8aa421e4e0398dfc4c018f8)

confgenerator/feature.go
qingling128: Discussed this in the design doc under the "Flatten the <key, value> pairs" section. 

Pros:
* Fewer labels.
Cons:
* More complex to query, filter by and aggregate.
* Needs heavy regex matching.

The cons outweighs the pros in this case, so keeping the `module` and `feature` is more convenient. Added a note for "alternative considered" in the design doc for the records.

quentinmit: I'm not sure I follow. The querying and filtering is going to happen on the Concord table, right? So why do the metric fields matter? We can have `module` and `feature` columns in the Concord table but keep the OA-side implementation simple.

qingling128: There are some internal-only details. Let's move to the internal chats.

confgenerator/feature.go
avilevy18: Inline structs are omitted because no name is included.

The approach for parsing the yaml tag is: https://github.com/goccy/go-yaml/blob/ecececdd257a265811b1646bc8e96b85cc2d2b0d/struct.go#L42-L79

confgenerator/feature.go
avilevy18: Changed in: [0f93735](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/0f937356b0edf460a8aa421e4e0398dfc4c018f8)

confgenerator/feature.go
quentinmit: Sure, but since this is for internal consumption, you don't need to care about the YAML tag. Embedded struct fields still have names from a reflection standpoint.

qingling128: Sync'd offline to understand the context and tradeoffs of the options. To summarize them:

A) Completely generic code walks the config struct and exposes every tagged variable with its struct path and value
* Pros
  * Simplest to implement
* Cons
  * The feature tracking metrics are not organized by features (Features generally map to receivers or processors). The path is flattened. More painful to query.

B) Non-generic code explicitly looks at top-level receiver and processor settings, and exposes the tagged ones with their struct field name and value
* Pros
  * Medium for implementation complexity
  * The feature tracking metrics are organized by features (Features generally map to receivers or processors).
* Cons
  * If the internal config parameter struct hierarchy or naming changes in the future, the corresponding feature tracking metrics labels will change, breaking existing feature tracking charts / dashboards unless someone notices and modify the charts / dashboards accordingly. The mitigation to prevent such changes is to stay locked in with the current structure naming and hierarchy.

C) Non-generic code explicitly looks at top-level receiver and processor settings, and exposes the tagged ones with an inferred YAML field name and value
* Pros
  * Internal config parameter struct hierarchy or naming can change freely. Because there is an extra layer to map to the yaml field names (aka user visible config parameter names that won't change without a breaking major release), the ingested feature tracking metrics can stay the same.
  * The feature tracking metrics are organized by features (Features generally map to receivers or processors).
* Cons
  * Most complex to implement

Given the tradeoffs, the decision is to stick with option C, aka what is implemented currently.

confgenerator/feature.go
qingling128: Summarized the offline chat in https://github.com/GoogleCloudPlatform/ops-agent/pull/853#discussion_r986287724.

confgenerator/feature.go
```diff
@@ -0,0 +1,210 @@
...
+			index := fmt.Sprintf("%d", i)
+
+			if ftr.Type == "" {
+				comp, ok := obj.Interface().(Component)
+				if !ok {
+					return nil, ErrInvalidType
+				}
+				ftr.Type = comp.Type()
+				ftr.Key = append(keys, index)
+			} else {
```
ridwanmsharif, line 112: What is an example of this case?

confgenerator/config.go
```diff
@@ -365,6 +365,17 @@ type LoggingReceiver interface {
 	Components(tag string) []fluentbit.Component
 }
 
+type LoggingReceiverMixin struct {
```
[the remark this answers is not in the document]
avilevy18: Whoops, forgot to remove this. Thanks

confgenerator/feature.go
```diff
@@ -0,0 +1,210 @@
...
+	v := reflect.Indirect(c)
+	if v.Kind() == reflect.Invalid || v.IsZero() {
+		return nil, nil
+	}
+
+	var features []Feature
+
+	switch kind := t.Kind(); {
+	case kind == reflect.Struct:
+		if m.IsInline && m.HasTracking {
```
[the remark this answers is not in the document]
avilevy18: Sync'd offline. Added a comment.

confgenerator/feature.go
```diff
@@ -0,0 +1,210 @@
...
+		// Else both the name of the field and the index of the item is appended to
+		// the key respectively.
+		for i, key := range v.MapKeys() {
+			obj := v.MapIndex(key)
+			md := m
+			ftr := feature
+			index := fmt.Sprintf("%d", i)
+
+			if ftr.Type == "" {
+				comp, ok := obj.Interface().(Component)
```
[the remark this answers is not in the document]
avilevy18: As we walk  through the `UnifiedConfig` we reach the `Receivers` and `Processors` maps which are all of type `Component`. Added comments to explain execution steps of `trackingFeatures()`

confgenerator/feature.go
```diff
@@ -0,0 +1,210 @@
...
+
+	switch kind := t.Kind(); {
+	case kind == reflect.Struct:
+		if m.IsInline && m.HasTracking {
+			return nil, ErrTrackingInlineStruct
+		}
+		if feature.Module == "" {
+			feature.Module = m.PathName
+		} else if feature.Kind == "" {
+			feature.Kind = m.PathName
```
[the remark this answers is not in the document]
avilevy18: Sync'd offline. Added comments for explanation.

confgenerator/feature.go
```diff
@@ -0,0 +1,246 @@
...
+	// Value defined from fields of UnifiedConfig.
+	Value string
+}
+
+// ExtractFeatures fields that containing a tracking tag will be tracked.
+// Automatic collection of bool or int fields. Any value that exists on tracking
+// tag will be used instead of value from UnifiedConfig.
+func ExtractFeatures(uc *UnifiedConfig) ([]Feature, error) {
+	allFeatures := getOverriddenDefaultPipelines(uc)
+	f, err := trackingFeatures(reflect.ValueOf(uc), metadata{}, Feature{})
```
ridwanmsharif, line 38: Can we unwrap and do some of the tracking of atleast the first 2 layers of `UnifiedConfig` here? instead of in `trackingFeatures`?

This is because `trackingFeatures` feels error prone to me as it relies on the depth of the recursion to build out `Feature.Kind` and `Feature.Kind`. Instead if we build those out here `trackingFeatures` could be a pure function with no assumptions being made about the depth of recursion.

Feel free to ignore this suggestion. I don't know yet which is more readable. My suggestion would add some duplication for sure in favor of readability but feel free to make the call yourself.

confgenerator/feature.go
```diff
@@ -0,0 +1,246 @@
...
+		// gets populated as we iterate through each level of the struct
+		if feature.Module == "" {
+			feature.Module = m.FieldTag
+		} else if feature.Kind == "" {
+			feature.Kind = m.FieldTag
+		} else if !m.IsInline {
+			feature.Key = append(feature.Key, m.FieldTag)
+		}
+
+		// For structs that in a Component
```
ridwanmsharif, line 86: Do you mean `For structs that are in a Component`?

See my other comment about assuming depth of recursion. Are we using `feature.Module != "" && feature.Kind != "" && feature.Type != "" && feature.Key != nil` to determine whether we are in a struct that is defined inside a receiver/processor component? 
Also what `Type` will these structs have? None?

confgenerator/feature.go
```diff
@@ -0,0 +1,210 @@
...
+		} else if feature.Kind == "" {
+			feature.Kind = m.PathName
+		} else if !m.IsInline {
+			feature.Key = append(feature.Key, m.PathName)
+		}
+
+		// Iterate over all available fields and read the tag value
+		for i := 0; i < t.NumField(); i++ {
+			// Get the field, returns https://golang.org/pkg/reflect/#StructField
+			field := t.Field(i)
```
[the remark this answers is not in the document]
ridwanmsharif: Could you add the contract of the `ExtractFeatures` interface somewhere as a comment for future readers? We should document that the above example is going to be behaving differently. The
```
a nested `yaml:"a", tracking:""`
```
Line above won't be allowed since tracking a field means giving it a key so it can't be empty. The above example produces 5 metrics:

- a
- a.str
- a.in
- b.str
- b.in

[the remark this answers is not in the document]
ridwanmsharif: Also this makes for a good unit test

ridwanmsharif (commented):
Looks pretty close to me

confgenerator/feature.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func getMetadata(field reflect.StructField) metadata {
+	trackingTag, hasTracking := field.Tag.Lookup("tracking")
+	hasOverride := false
+	if trackingTag != "" {
+		hasOverride = true
+	}
+
+	yamlTag, _ := field.Tag.Lookup("yaml")
```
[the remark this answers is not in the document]
avilevy18: Added panic

confgenerator/feature.go
avilevy18: Changed in: [237e331](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/237e331081f573c60c5da6d731121d7f4cd56a46)

confgenerator/feature.go
avilevy18: Changed in: [237e331](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/237e331081f573c60c5da6d731121d7f4cd56a46)

confgenerator/feature.go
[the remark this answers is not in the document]
avilevy18: Changed in, unit test added: [237e331](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/237e331081f573c60c5da6d731121d7f4cd56a46)

confgenerator/feature.go
```diff
@@ -0,0 +1,273 @@
...
+				ftr.Value = m.OverrideValue
+				features = append(features, ftr)
+			}
+		}
+
+		// Iterate over all available fields and read the tag value
+		for i := 0; i < t.NumField(); i++ {
+			// Get the field, returns https://golang.org/pkg/reflect/#StructField
+			field := t.Field(i)
+			if field.Name == "Type" {
```
ridwanmsharif, line 152: We should document here what this is. Basically the `Type` field name is part of the `ConfigComponent` definition. And all user visible component inlines that component. That means this field can help us assert that a certain component is enabled.

confgenerator/feature.go
```diff
@@ -0,0 +1,273 @@
...
+				// Capture special metrics for enabled receiver or processor
```
ridwanmsharif, line 153: Move this comment before the `if` statement with the siggestion above.

confgenerator/feature.go
```diff
@@ -0,0 +1,273 @@
...
+				continue
+			}
+
+			tf, err := trackingFeatures(v.Field(i), getMetadata(field), feature)
+			if err != nil {
+				return nil, err
+			}
+			features = append(features, tf...)
+		}
+	case kind == reflect.Map:
```
ridwanmsharif, line 167: Are we not supporting feature tracking with maps at all for v1?

ridwanmsharif (commented):
Left a couple questions and suggestions but looks good to me. Will approve once some of these are addressed

confgenerator/feature.go
avilevy18: Added `TODO`: [e4b0486](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/e4b0486e87c25b5f9f09f38906f8078ee015ae6b)

confgenerator/feature.go
```diff
@@ -0,0 +1,273 @@
...
+		tempTrackedFeatures, err = trackedMappedComponents("logging", "processors", uc.Logging.Processors)
+		if err != nil {
+			return nil, err
+		}
+		allFeatures = append(allFeatures, tempTrackedFeatures...)
+	}
+	return allFeatures, nil
+}
+
+func getSortedKeys[C Component](m map[string]C) []string {
```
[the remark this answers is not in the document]
avilevy18: Changed in: [e4b0486](https://github.com/GoogleCloudPlatform/ops-agent/pull/853/commits/e4b0486e87c25b5f9f09f38906f8078ee015ae6b)

PR #867, opened 2022-09-21
Decrease fluent-bit buffer_max_size to 2M for smaller outgoing requests.

igorpeshansky:
## Description
Some outgoing requests end up being over the API limit. This PR reduces the chance of that happening by lowering the maximum size of the accumulated buffer.

## Related issue
http://b/244498953

## How has this been tested?
Only automatic unit tests.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


qingling128 (approved):
LGTM

igorpeshansky:
Thanks. The Windows integration test failure was a flake. Merging.

igorpeshansky:
@sophieyfang:
> Shall we also add a test case with a log entry really big and see it didn't get the API denied response? 🤔

I think the issue here is that the accumulated size of the log entries exceeds the API request limit. A single entry will still be rejected, as it'd be over the entry size limit (256k). While we can try to simulate the "multiple entries add up to a request that's too large" scenario, it seems very fragile and hard to reproduce, so the test will likely be flaky.

PR #869, opened 2022-09-22
Prometheus receiver: add integration test with JSON exporter

LujieDuan:
## Description
Add integration test for Prometheus receiver using [JSON exporter](https://github.com/prometheus-community/json_exporter). 

To use the JSON exporter, we need to start a Python http server to serve a static JSON file on port 8000. We then config Ops Agent to scrape Prometheus style metrics from JSON exporter listening on port 7979. The JSON exporter will go request the JSON from the Python server and convert the JSON to metrics as we configured. 

The JSON exporter will only export metric types gauge and counter. 

## Related issue
b/244434541

## How has this been tested?
- Manual trigger the integration test to make sure it would fail with unexpected metrics from GCM. 
- Make sure all integration tests pass. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/testdata/setup_json_exporter.sh
```diff
@@ -0,0 +1,18 @@
+set -e
+
+wget -O json_exporter.tar.gz \
```
ridwanmsharif, line 3: Does this only work on certain distros? Should we exclude the others in the test?

integration_test/ops_agent_test.go
```diff
@@ -1893,6 +1901,153 @@ func TestPrometheusMetrics(t *testing.T) {
...
+      type: prometheus
+      config:
+        scrape_configs:
+        - job_name: json
+          metrics_path: /probe
+          params:
+            module: [default]
+          static_configs:
+            - targets:
+              - http://localhost:8000/data.json
```
ridwanmsharif, line 1981: What is the target the Ops Agent tries to listen on?

`http://localhost:8000/data.json/probe`?

Will it also work if the target is 
`localhost:8000` and `metrics_path` is `/data.json/probe`? 

I ask because this feels like a validation error that is a bug and not a feature lol

integration_test/ops_agent_test.go
```diff
@@ -1893,6 +1901,153 @@ func TestPrometheusMetrics(t *testing.T) {
...
+		}
+
+		packages := []string{"wget", "python3"}
+		err := agents.InstallPackages(ctx, logger.ToMainLog(), vm, packages)
+		if err != nil {
+			t.Fatalf("failed to install %v with err: %s", packages, err)
+		}
+
+		// Install either tmux or screen to run the JSON exporter and the http server in the background
+		tmux_err := agents.InstallPackages(ctx, logger.ToMainLog(), vm, []string{"tmux"})
```
ridwanmsharif, line 1947: Did we need `tmux` to run things in the background? I was fairly certain there is a bash way to run things in the background too no? Like appending `&` at the end or something?

Worth looking into at the very least to simplify the dependencies of this test.

ridwanmsharif (commented):
Looks very good!

integration_test/testdata/setup_json_exporter.sh
LujieDuan: Right now this should work on all our linux distros. I was having trouble with tmux and screen since tmux is missing from SLES and screen is missing from Rocky. 
So in the test I try to install wget/python and either tmux or screen, and now it seems ok for all our linux distros.

integration_test/ops_agent_test.go
LujieDuan: The final target after the relabeling is: `http://localhost:7979/probe?module=default&target=http://localhost:8000/data.json`. 
Here is what I think how this got built: 
- Initially, `__address__` is the target which is `http://localhost:8000/data.json`;
- The first relabeling move `__address__` to `__param_target`, so that we now have `target=http://localhost:8000/data.json` of the final url;
- The `params: module: [default]` add the `module=default` so we have `module=default&target=http://localhost:8000/data.json`
- The `metric_path: /probe` add that to url and get `probe?module=default&target=http://localhost:8000/data.json`;
- And the last relabeling change the `__address__` to `localhost:7979` and get `http://localhost:7979/probe?module=default&target=http://localhost:8000/data.json`. 

It should also work if target is `localhost:7979` (the JSON exporter's port) and `metric_path` is `/probe`, and params has `module: [default]` and `target: [http://localhost:8000/data.json]`.

integration_test/ops_agent_test.go
LujieDuan: Good point! I was using tmux/screen when developing the test because I can easily verify the two processes are running. Let me try `nohup` with `&` now.

integration_test/ops_agent_test.go
```diff
@@ -1893,6 +1901,181 @@ func TestPrometheusMetrics(t *testing.T) {
...
+
+		// Wait long enough for the data to percolate through the backends
+		// under normal circumstances. Based on some experiments, 2 minutes
+		// is normal; wait a bit longer to be on the safe side.
+		time.Sleep(3 * time.Minute)
+		window := time.Minute
+
+		var multiErr error
+		// Gauge type
+		gaugeMetric := "prometheus.googleapis.com/test_gauge_value/gauge"
```
ridwanmsharif, line 2017: A lot of the tests look like they can be templated. Something like:

```
type promMetricTest struct {
  metricName string
  metricKind   MetricKind
  valueType     ValueType
  expectedValue float64
}

testData = []promMetricTest{
{"prometheus.googleapis.com/test_gauge_value/gauge", metric.MetricDescriptor_GAUGE, metric.MetricDescriptor_DOUBLE, 789},
{"prometheus.googleapis.com/test_counter_value/counter", metric.MetricDescriptor_CUMULATIVE, metric.MetricDescriptor_DOUBLE, 0},
{"prometheus.googleapis.com/test_untyped_metric_value/gauge", metric.MetricDescriptor_GAUGE, metric.MetricDescriptor_DOUBLE, 56},
}

for test := range testData {
  ...
}
```

LujieDuan: Great suggestion! I have made the change to reuse code for all the metric types.

integration_test/ops_agent_test.go
```diff
@@ -1893,6 +1901,161 @@ func TestPrometheusMetrics(t *testing.T) {
...
+			liveCheckOut, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", `curl "http://localhost:7979/probe?module=default&target=http://localhost:8000/data.json"`)
+			// We will retry when:
+			// 1. JSON exporter is not started: in this case the stderr will have: "curl: (7) Failed to connect to localhost port 7979 after 1 ms: Connection refused"
+			// 2. The Python HTTP server is not started: in this case the stdout will not have the expected Prometheus style metrics
+			// If neither cases - break the retrying loop
+			if err == nil && !strings.Contains(liveCheckOut.Stderr, "Connection refused") && strings.Contains(liveCheckOut.Stdout, `test_counter_value{test_label="counter_label"} 1234`) {
+				break
+			}
+
+		}
```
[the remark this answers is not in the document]
LujieDuan: Fixed! Now will display the last error message when out of attempts.

integration_test/ops_agent_test.go
```diff
@@ -1893,6 +1901,161 @@ func TestPrometheusMetrics(t *testing.T) {
...
+
+		var multiErr error
+		for _, test := range testData {
+			if pts, err := gce.WaitForMetric(ctx, logger.ToMainLog(), vm, test.metricName, window, nil, true); err != nil {
+				multiErr = multierr.Append(multiErr, err)
+			} else {
+				if pts.MetricKind != test.expectedMetricKind {
+					multiErr = multierr.Append(multiErr, fmt.Errorf("Metric %s has metric kind %s; expected kind %s", test.metricName, pts.MetricKind, test.expectedMetricKind))
+				}
+				if pts.ValueType != metric.MetricDescriptor_DOUBLE {
```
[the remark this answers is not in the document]
LujieDuan: Fixed! Nice catch

PR #886, opened 2022-10-03
prometheus: use prometheus styled regex instead of OTel

ridwanmsharif:
This is mainly focussed on the `replacement` field and us not using the otel styled `$` syntax for the user visible prom config.

## Checklist:
- Unit tests
- Integration tests
  - [] Integration tests do not apply.
- Documentation
- Minor version bump


confgenerator/prometheus.go
```diff
@@ -84,15 +86,62 @@ func (r PrometheusMetrics) Pipelines() []otel.Pipeline {
...
+	copyPromConfig, err := deepCopy(promConfig)
+	if err != nil {
+		// This should never happen since we already validated the prometheus config.
+		panic(fmt.Errorf("failed to deep copy prometheus config: %w", err))
+	}
+
+	for i := range promConfig.ScrapeConfigs {
+		for j := range copyPromConfig.ScrapeConfigs[i].RelabelConfigs {
+			// Escape the $ characters in the regexes.
+			replacement := copyPromConfig.ScrapeConfigs[i].RelabelConfigs[j].Replacement
```
LujieDuan, line 109: Can we operate on the copy entirely here, and since the ScapeConfigs and RelabelConfigs are list of pointers to the actual config, we may simplify it here:
```
	for _, sc := range copyPromConfig.ScrapeConfigs {
		for _, rc := range sc.RelabelConfigs {
			// Escape the $ characters in the regexes.
			rc.Replacement = strings.ReplaceAll(rc.Replacement, "$", "$$")
```

confgenerator/prometheus.go
```diff
@@ -84,15 +86,62 @@ func (r PrometheusMetrics) Pipelines() []otel.Pipeline {
...
+		Type:   "prometheus",
+		Config: map[string]interface{}{"config": copyPromConfig},
+	}
+}
+
+func deepCopy(config promconfig.Config) (promconfig.Config, error) {
+	var out any
+	if err := mapstructure.Decode(config, &out); err != nil {
+		return promconfig.Config{}, err
+	}
```
LujieDuan, line 131: Question: why do we need to convert the config to any here?

LujieDuan (commented):
The change looks good! Left a comment and a question.

confgenerator/prometheus.go
ridwanmsharif: Made a change to clean up some of this. Interesting note is we do need to use the index values though to deal with the references/pointers and not a copy. That was why I had done it the first way. But I cleaned it up a bit more for readability.

Also see: https://stackoverflow.com/questions/20185511/range-references-instead-values

confgenerator/prometheus.go
ridwanmsharif: Good catch. We don't need the `Decode` at all it looks like, this was from when I copied over the yaml marshalling that we do in the `ModularConfig.Generate()` but we can make this simpler. Done in latest commit.

LujieDuan (approved):
LGTM!

PR #897, opened 2022-10-06
Add test for non-default time zones on Windows

jefferbrecht:
## Description
Add test for non-default time zones on Windows, which is currently broken. When fluent-bit is fixed upstream, merge the fix here and re-run the tests.

## Related issues
b/218888265
https://github.com/fluent/fluent-bit/issues/2941

## How has this been tested?
Ran tests locally; will let the standard tests run in the PR.

## Checklist:
- Unit tests
- Integration tests
  - [x] Integration tests have been added/modified.
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -1611,6 +1611,41 @@ func TestWindowsEventLog(t *testing.T) {
 	})
 }
 
+func TestWindowsEventLogWithNonDefaultTimeZone(t *testing.T) {
+	t.Parallel()
+	gce.RunForEachPlatform(t, func(t *testing.T, platform string) {
+		t.Parallel()
+		if !gce.IsWindows(platform) {
+			t.SkipNow()
+		}
+		ctx, logger, vm := agents.CommonSetup(t, platform)
+		if _, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", `Set-TimeZone -Id "Eastern Standard Time"`); err != nil {
+			t.Fatal(err)
+		}
+		if err := setupOpsAgent(ctx, logger, vm, ""); err != nil {
+			t.Fatal(err)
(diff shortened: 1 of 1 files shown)


integration_test/ops_agent_test.go
```diff
@@ -1611,6 +1611,41 @@ func TestWindowsEventLog(t *testing.T) {
...
+func TestWindowsEventLogWithNonDefaultTimeZone(t *testing.T) {
+	t.Parallel()
+	gce.RunForEachPlatform(t, func(t *testing.T, platform string) {
+		t.Parallel()
+		if !gce.IsWindows(platform) {
+			t.SkipNow()
+		}
+		ctx, logger, vm := agents.CommonSetup(t, platform)
+		if _, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", `Set-TimeZone -Id "Eastern Standard Time"`); err != nil {
+			t.Fatal(err)
```
sophieyfang, line 1623: What if the new windows VM default time zone is just the same as Eastern Time.  Does the VM created with zone us-east1 default to have Eastern Time?

jefferbrecht: If the default time zone is already Eastern Time then this command is a no-op and the test should still fail as expected.

In my testing the VM time zone was always UTC regardless of which Cloud zone it was created in.

franciscovalentecastro:
Should we update also `TestCustomLogFormat` in this PR too ? Since it uses `UTC` to avoid hitting the timezone bug.

https://github.com/GoogleCloudPlatform/ops-agent/blob/3ec1e254feddb1277aeb742f1f340937c83aec9c/integration_test/ops_agent_test.go#L837-L844

jefferbrecht:
Closing because this has been moved into #1078.

PR #901, opened 2022-10-13
confgenerator: use new generic set

braydonk:
## Description
This PR creates a new `Set[T comparable]` type that we can use to replace some confusing `map` usage. The new `Set` type may end up being useful in other places.

There were a couple other spots nearby that I thought would benefit from using generics instead which I also adjusted.

I also ran `make addlicense` which found a couple unrelated files missing license headers.

## How has this been tested?
`make test_confgenerator`, since the unit tests already covered this code.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/config.go
```diff
@@ -793,83 +795,46 @@ var (
...
-func sortedKeys(m interface{}) []string {
-	var r []string
-	for k := range mapKeys(m) {
-		r = append(r, k)
-	}
-	sort.Strings(r)
-	return r
-}
-
 // findInvalid returns all strings from a slice that are not in allowed.
```
avilevy18, line 810: This only returns 1 invalid string, not all. The comment might need to be updated

braydonk: While trying to write a better descriptive comment I realized that this function is pretty unnecessary and decided to inline it.

qingling128:
Sync'd offline @martijnvans will take this one as the primary reviewer.

internal/set/set_test.go
```diff
@@ -0,0 +1,80 @@
...
+	assert.Assert(t, testSet.Contains(newKey))
+}
+
+func TestRemove(t *testing.T) {
+	key := "key"
+	testSet := set.Set[string]{
+		key: struct{}{},
+	}
+	testSet.Remove(key)
+	assert.Equal(t, len(testSet), 0)
```
martijnvans, line 60: I would throw in an assert.Assert(t, !testSet.Contains(key)) here.

tasks.mak
```diff
@@ -73,6 +73,9 @@ test_metadata_update:
 test_metadata_validate:
 	go test ./integration_test/validate_metadata_test.go
 
+test_set:
+	go test ./internal/set
```
martijnvans, line 77: this seems like an oddly specific thing to have in the makefile. How often is this going to get used after the PR is merged? I'd say likely never.

internal/set/set.go
```diff
@@ -0,0 +1,56 @@
...
+
+func SliceToSet[T comparable](sl []T) Set[T] {
+	s := Set[T]{}
+	for _, v := range sl {
+		s.Add(v)
+	}
+	return s
+}
+
+func MapToSet[K comparable, V any](m map[K]V) Set[K] {
```
martijnvans, line 27: This name kind of bugs me, because I wasn't sure when looking at the callsite whether it was a set of the keys in the map or a set of the values. How about set.KeySet ? I know that this loses parallelism with SliceToSet, but I feel like that's a small price to pay.

Weak suggestion, feel free to ignore.

martijnvans (approved):

internal/set/set.go
braydonk: I probably wouldn't change the name of the type to `set.KeySet` since the `Set` could be made of anything not just map keys. I could change this function name to `MapKeysToSet`?

tasks.mak
braydonk: I just kind of got in the habit of adding a target for each package, but I don't have to keep doing that.

internal/set/set.go
martijnvans: huh? if you called this function on a map, the resulting set can't be made of anything, it is specifically made of the keys of the given map. 

oh you think I wass suggesting changing the name of the type? no, I Iwas just suggesting changing the name of ths function.

MapKeysToSet is all rght, i still prefer KeySet, but it's up to you.

tasks.mak
martijnvans: ok

quentinmit: Just use a pattern rule to define tests on arbitrary packages.

internal/set/set.go
braydonk: Yeah that works, and maps with the Effective Go practice of not using the package name in exported members from the package. Will do that.

tasks.mak
braydonk: Yeah I suppose that would work too. I'll remove this target from this PR and do that in a future follow up

braydonk: Done in [8c8978f](https://github.com/GoogleCloudPlatform/ops-agent/pull/901/commits/8c8978ffc91e2e1e2dcc118c3017717a6d266f39)

internal/set/set.go
braydonk: Done in [8c8978f](https://github.com/GoogleCloudPlatform/ops-agent/pull/901/commits/8c8978ffc91e2e1e2dcc118c3017717a6d266f39)

internal/set/set_test.go
braydonk: Done in [8c8978f](https://github.com/GoogleCloudPlatform/ops-agent/pull/901/commits/8c8978ffc91e2e1e2dcc118c3017717a6d266f39)

martijnvans (approved):

PR #909, opened 2022-10-17
Integration test for Prometheus receiver of metric types histogram and summary

LujieDuan:
## Description
Add integration test case for the Prometheus receiver, with histogram and summary types of metrics. 

## Related issue
b/253060243

## How has this been tested?
Run the test manually and use unexpected values to make sure the test can catch that. 
Run the test in the PR. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -2134,7 +2134,10 @@ func TestPrometheusMetricsHistogramAndSummary(t *testing.T) {
 	t.Parallel()
 	gce.RunForEachPlatform(t, func(t *testing.T, platform string) {
 		t.Parallel()
-		if gce.IsWindows(platform) {
+		// TODO: Enable this test for all distros once the prometheus receiver is GA.
+		// For some reason, the featuregate, when set in the default systemd environment
+		// file, is not being picked up on centOS distros. This is a temporary workaround.
+		if gce.IsWindows(platform) || gce.IsCentOS(platform) || gce.IsRHEL(platform) {
 			t.SkipNow()
 		}
 		ctx, logger, vm := agents.CommonSetup(t, platform)
@@ -2213,7 +2216,10 @@ func TestPrometheusMetricsHistogramAndSummary(t *testing.T) {
       prom_pipeline:
         receivers: [prom_app]
 `
-
+		// Turn on the prometheus feature gate.
+		if err := gce.SetEnvironmentVariables(ctx, logger.ToMainLog(), vm, map[string]string{"UNSUPPORTED_BETA_PROMETHEUS_RECEIVER": "enabled"}); err != nil {
+			t.Fatal(err)
+		}
 		if err := setupOpsAgent(ctx, logger, vm, config); err != nil {
 			t.Fatal(err)
 		}
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -60,15 +60,10 @@ import (
 	"github.com/GoogleCloudPlatform/ops-agent/integration_test/logging"
 	"github.com/GoogleCloudPlatform/ops-agent/integration_test/metadata"
 	"github.com/GoogleCloudPlatform/ops-agent/integration_test/util"
-	"golang.org/x/exp/slices"
-	"google.golang.org/genproto/googleapis/api/metric"
-	"google.golang.org/genproto/googleapis/monitoring/v3"
-	"gopkg.in/yaml.v2"
-
-	cloudlogging "cloud.google.com/go/logging"
 	"github.com/google/uuid"
 	"go.uber.org/multierr"
-	distribution "google.golang.org/genproto/googleapis/api/distribution"
+	"golang.org/x/exp/slices"
+	"google.golang.org/genproto/googleapis/api/distribution"
 	"google.golang.org/genproto/googleapis/api/metric"
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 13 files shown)


integration_test/ops_agent_test.go
```diff
@@ -2091,6 +2093,386 @@ func TestPrometheusMetricsWithJSONExporter(t *testing.T) {
...
+		// buckets w/ le=[0, 20, 40, 60, 80] plus the +inf final bucket
+		// For step 1, we observe points [0, 10, 20, 30, 40, 50, 60, 70, 80, 90]
+		// And get:
+		// Bounds (less than or equal) |0  |20     |40     |60     |80     |+inf
+		// Points                      |[0]|[10,20]|[30,40]|[50,60]|[70,80]|[90]
+		// Count                       |1  |2      |2      |2      |2      |1
+		// And histogram metrics are stored as cumulative type metrics, so for
+		// this initial step, Count/Mean/SumOfSquaredDeviation are all zeros,
+		// and the BucketCounts is nil
+		stepOneExpectedHistogram := &distribution.Distribution{
```
ridwanmsharif, line 2213: I don't completely understand why these are all 0 values? They all do have some values right (in your example above, they have a total of 10 points). Shouldn't this have the first values? I see in step 2 they get the delta between the second and the first set of values, but in this case the delta between the first and 0 is dropped?

ridwanmsharif: If this is expected, please let me know why. I may be missing something

integration_test/ops_agent_test.go
```diff
@@ -2091,6 +2093,386 @@ func TestPrometheusMetricsWithJSONExporter(t *testing.T) {
...
+		// Note: We could wait for 3 mins here again and assume the new points
+		// are available, but in most case it should need less than 3 mins; we
+		// also want to make sure the new points are indeed available before we
+		// check them against the step 2 expected values, otherwise the error
+		// message won't be useful
+		time.Sleep(1 * time.Minute)
+		// Since test_summary_count/summary is a cumulative type metric, as long
+		// as it is greater than 0, we know the step 2 points have arrived
+		testMetricName := "prometheus.googleapis.com/test_summary_count/summary"
+		var previousTotal float64 = 0
```
ridwanmsharif, line 2267: Is this always zero? Should it not be set to what was found before step 4?

ridwanmsharif: Also looks like we are already testing for Step 2 summary points below. Is this needed?

integration_test/ops_agent_test.go
```diff
@@ -2091,6 +2093,386 @@ func TestPrometheusMetricsWithJSONExporter(t *testing.T) {
...
+		multiErr = multierr.Append(multiErr, assertPrometheusHistogramMetric(ctx, logger, vm, "test_histogram", window, stepOneExpectedHistogram))
+		// For Summary: We use Objectives [0.5, 0.9, 0.99]
+		// For step 1, we observe points [0, 10, 20, 30, 40, 50, 60, 70, 80, 90]
+		// And get:
+		// Objectives |0.5            |0.9               |0.99
+		// Points     |[0,10,20,30,40]|[..., 50,60,70,80]|[...,90]
+		// Quantile   |40             |80                |90
+		// And summary metrics' quantiles are stored as gauge type metrics, so
+		// for this initial step, quantiles have the actual values.
+		// But count and sum are stored as cumulative values, so those two are 0
```
ridwanmsharif, line 2234: Same question for these values like for the histogram. It feels like the first points are basically just dropped?

integration_test/ops_agent_test.go
```diff
@@ -2091,6 +2093,386 @@ func TestPrometheusMetricsWithJSONExporter(t *testing.T) {
 	})
 }
 
+// Test the Histogram and Summary metric types using static testing files
+// The files will contain metrics in the right format and
+// hosted by a simple Python HTTP server so that the agent can scrape the metrics
+// The test will send two sets of metric points, to verify the cumulative metrics
+// are correctly received and processed
+func TestPrometheusMetricsHistogramAndSummary(t *testing.T) {
```
ridwanmsharif, line 2101: [super ultra nit]: This is a pretty massive test. I wonder if we can simplify some of this into a separate method for the set up and then have two tests: 1 for histogram and 1 for summary. That way if one of them breaks, debugging and fixing may be easier

ridwanmsharif (commented):
Overall looks pretty good to me but I had a few questions. Will quick scan once more after

integration_test/ops_agent_test.go
```diff
@@ -2091,6 +2093,386 @@ func TestPrometheusMetricsWithJSONExporter(t *testing.T) {
...
+		testMetricName := "prometheus.googleapis.com/test_summary_count/summary"
+		var previousTotal float64 = 0
+		totalGreaterThanPreviousTotal := func(series *monitoring.TimeSeries) (bool, error) {
+			if series.ValueType == metric.MetricDescriptor_DOUBLE {
+				lastPoint := series.Points[len(series.Points)-1]
+				return lastPoint.Value.GetDoubleValue() > previousTotal, nil
+			}
+			return false, fmt.Errorf("wrong metric value type")
+		}
+		if _, err := gce.WaitForMetricWithCondition(ctx, logger.ToMainLog(), vm, testMetricName, window, nil, true, totalGreaterThanPreviousTotal); err != nil {
```
ridwanmsharif, line 2275: Was it not possible to use the `extraFilters` param instead of adding a condition after we wait for the metric?
Asking to see if we can push the query for this condition to the server and not do it client side.

integration_test/ops_agent_test.go
```diff
@@ -2091,6 +2093,386 @@ func TestPrometheusMetricsWithJSONExporter(t *testing.T) {
...
+			setupOut, err := gce.RunScriptRemotely(ctx, logger, vm, string(setupScript), nil, env)
+			// Since the script will start the processes in the background, the
+			// script should finish without error
+			if err != nil {
+				t.Fatalf("failed to run the http server in VM with err: %v, stderr: %s", err, setupOut.Stderr)
+			}
+			// Wait until the http server is ready
+			time.Sleep(30 * time.Second)
+			liveCheckOut, liveCheckErr := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", `curl "http://localhost:8000/data.json"`)
+			// We will retry when the Python HTTP server is not started:
```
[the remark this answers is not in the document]
LujieDuan: The retry was needed because the Python http server with nohup would randomly fail. The Python http server also different across distros and Python versions and make this test unstable. I have changed to use a go http server (since we can already set up go env in the vm) with systemd (to make sure the server starts, and logs to `journalctl_output.txt` for better debugging). From my testing it looks this is much more reliable than the previous approach.

integration_test/ops_agent_test.go
```diff
@@ -2091,6 +2093,386 @@ func TestPrometheusMetricsWithJSONExporter(t *testing.T) {
...
+		ctx, logger, vm := agents.CommonSetup(t, platform)
+
+		// 1. Upload the step one metric points
+		workDir := path.Join(workDirForPlatform(vm.Platform), "sample_data")
+		stepOneMetrics := "testdata/sample_histogram_summary_1.txt"
+		f, err := os.Open(stepOneMetrics)
+		if err != nil {
+			t.Fatal(err)
+		}
+		err = gce.UploadContent(ctx, logger, vm, f, path.Join(workDir, "data.json"))
```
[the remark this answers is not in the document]
LujieDuan: Good catch! Renamed to just `data`.

integration_test/ops_agent_test.go
LujieDuan: Sync'ed offline, the cumulative points (histogram, count & sum of summary) get normalized when exporting to GCM: the first point of a cumulative time series is not sent and used to subtract from all following points. That's why even though the `first step` point has values, the agent scrape this multiple times and they are all received as 0 values. Added comments to point to the normalization of the cumulative points.

integration_test/ops_agent_test.go
LujieDuan: Together with `WaitForMetricWithCondition`, this check here has been removed and replaced with a sleep(3m).

integration_test/ops_agent_test.go
LujieDuan: Summary's count and sum are cumulative and get normalized as well.

integration_test/ops_agent_test.go
LujieDuan: Good suggestion! I have broken this into two separate tests and look cleaner this way.

integration_test/ops_agent_test.go
LujieDuan: `WaitForMetricWithCondition` was used to make sure `step two` points have been received before checking their values. Now I have found that those `step two` points normally get received pretty quick (with 10s scrape interval for the agent, it usually takes another couple seconds before those become available) so I have replaced this with a sleep(3m) and we don't have to keep retrying here if something wrong with the `step two` points.

integration_test/ops_agent_test.go
ridwanmsharif: This makes sense to me. I think we should atleast manually verify that this is actually what promql users expect once they migrate off of prometheus. We should be doing what GMP does on GKE, or what prometheus does by default

ridwanmsharif (commented):
The new http server is very cool. Can we use it for `TestPrometheusMetricsWithJSONExporter` too?

Overall looks pretty solid, just not sure yet about whether we want the normalization that the exporter enables by default or whether we want to toggle that off.

integration_test/testdata/prometheus/http_server.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+var (
+	dir  = flag.String("dir", "./", "directory to serve")
+	port = flag.String("port", "8000", "port to listen to")
+)
+
+func serve(dir, port string) error {
+	dirHandler := http.FileServer(http.Dir(dir))
+	http.Handle("/", dirHandler)
```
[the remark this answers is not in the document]
LujieDuan: The http server will make all files in a folder available, in case we need to host/query multiple files. Currently the test curls the `{dir}/data` file to get the metrics.

integration_test/ops_agent_test.go
```diff
@@ -2118,6 +2120,477 @@ func TestPrometheusMetricsWithJSONExporter(t *testing.T) {
...
+		err := uploadFiles(ctx, logger, vm, testdataDir, filesToUpload["step_one"])
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// 2. Setup the golang and start the go http server
+		if err := installGolang(ctx, logger, vm); err != nil {
+			t.Fatal(err)
+		}
+		setupScript := `sudo systemctl daemon-reload
```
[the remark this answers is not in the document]
LujieDuan: Will use this same method for the `TestPrometheusMetricsWithJSONExporter` as well!

integration_test/ops_agent_test.go
```diff
@@ -2118,6 +2120,477 @@ func TestPrometheusMetricsWithJSONExporter(t *testing.T) {
...
+      prom_pipeline:
+        receivers: [prom_app]
+`
+		// Turn on the prometheus feature gate.
+		if err := gce.SetEnvironmentVariables(ctx, logger.ToMainLog(), vm, map[string]string{"EXPERIMENTAL_FEATURES": "prometheus_receiver"}); err != nil {
+			t.Fatal(err)
+		}
+		if err := setupOpsAgent(ctx, logger, vm, config); err != nil {
+			t.Fatal(err)
+		}
```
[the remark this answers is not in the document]
LujieDuan: Done! Right now the top level test functions will only specify step one and two's files & checks against expected values, and the helper function will have all the common steps.

ridwanmsharif (approved):

PR #913, opened 2022-10-19
Kill service processes if Restart-Service fails

jefferbrecht:
Lots of flakes lately on Windows 2012 due to timeouts on Restart-Service

## Description
The `Restart-Service` command can timeout and fail the test if the sub-agents don't respond. We're trying to find a root cause for such situations, but for now we'd like to work around test flakes which depend on service restarts.

## Related issue
b/240564518

## How has this been tested?
Will let the PR presubmits run the tests

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -80,7 +80,18 @@ func configPathForPlatform(platform string) string {
 
 func restartCommandForPlatform(platform string) string {
 	if gce.IsWindows(platform) {
-		return "Restart-Service google-cloud-ops-agent -Force"
+		return `
+Restart-Service google-cloud-ops-agent -Force
+if (!$?) {
+	Write-Output 'Could not restart services gracefully. Killing processes directly...'
+	Get-WmiObject -Class Win32_Service -Filter "Name LIKE 'google-cloud-ops-agent%'" | ForEach-Object {
+		Stop-Process -Force $_.ProcessId
+		if (!$?) {
+			Write-Output "Could not stop process $($_.ProcessId); proceeding anyway"
+		}
+	}
+	# The SCM will automatically restart services after the processes are killed
+}`
 	}
 	// Return a command that works for both < 2.0.0 and >= 2.0.0 agents.
 	return "sudo service google-cloud-ops-agent restart || sudo systemctl restart google-cloud-ops-agent.target"
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -82,6 +82,7 @@ func restartCommandForPlatform(platform string) string {
 	if gce.IsWindows(platform) {
 		return `
 Restart-Service google-cloud-ops-agent -Force
+#TODO(b/240564518): remove process-killing once bug is fixed
 if (!$?) {
 	Write-Output 'Could not restart services gracefully. Killing processes directly...'
 	Get-Service -Name 'google-cloud-ops-agent*' | Set-Service -StartupType Disabled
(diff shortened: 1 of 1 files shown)


martijnvans:
I believe this PR will not do anything to alleviate the restart-service timeouts. The timeout that is being hit is actually 30 seconds and is not imposed by the test. The Restart-Service command spends 30 seconds waiting for, I believe, fluent-bit to shut down, before it gives up and returns an error.

These flakes are b/240564518. I tried to figure out what was going on but gave up after a while and handed off the bug. However, the flakes are really bad and I would love it if someone could prioritize b/240564518.

jefferbrecht:
Ah, I see. My theory was that the service was plugged up waiting for background stuff to complete and couldn't respond to the shutdown request in time, so I wanted to give that background stuff more time to complete, but it makes more sense that it's actually deadlocked or something and the delay won't help.

As a workaround then we could force-kill the process associated with the service, until we figure out b/240564518.

martijnvans:
Sure, we could force-kill the relevant processes, probably just fluent-bit. I'm OK with that as a temporary workaround.

jefferbrecht:
```
=== CONT  TestInvalidConfig/windows-2019
    ops_agent_test.go:986: Expected agent to reject bad config.
```

I didn't realize we're relying on `Restart-Service` failing as the signal that the bad config was rejected. This change makes the service-restart implicit through the recovery mechanism.

Two options:

1. Temporarily disable automatic service startup to prevent the recovery mechanism from running, kill the processes, re-enable automatic service startup, then explicitly start service via `Start-Service` which would return any errors that `Restart-Service` would have returned had it succeeded in the first place. This should bring it to parity with the original tests as written. But...
2. IMO the test design for `TestInvalidConfig` could be improved. Currently any error from `Restart-Service` will be treated as "bad config rejected" but it's possible for other errors to occur and still pass the test. We can improve this by explicitly looking for an event log entry starting with `failed to generate config files`. That way, if we get an error from `Restart-Service` and we _don't_ see a matching event log with that error message, we can conclude that the test didn't run properly. But either way I would still need to do the `Start-Service` thing in (1). WDYT? cc @martijnvans

jefferbrecht:
Ahh, I guess I'm playing dice with the recovery mechanism now, because it could kick in during the time between killing the first process and calling `Start-Service`. The recovery mechanism waits 1000ms before restarting the first time, 2000ms the second time, etc. I guess I'll see how the kokoro run goes -- it worked locally for me at least. I added a comment to the code with an idea on how to make this even more complicated in case the recovery mechanism interferes and causes flakes.

integration_test/ops_agent_test.go
```diff
@@ -215,6 +229,18 @@ func setupOpsAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
 	return setupOpsAgentFrom(ctx, logger, vm, config, locationFromEnvVars())
 }
 
+// restartOpsAgent restarts the Ops Agent and waits for it to become available.
+func restartOpsAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM) error {
+	if output, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", restartCommandForPlatform(vm.Platform)); err != nil {
+		logger.ToMainLog().Printf("stdout=%s\nstderr=%s", output.Stdout, output.Stderr)
```
martijnvans, line 235: RunREmotely already prints stdout/stderr to the log, no need to do it manually. stdout/stderr is also included in err if there is an error.

integration_test/ops_agent_test.go
```diff
@@ -80,7 +80,21 @@ func configPathForPlatform(platform string) string {
 
 func restartCommandForPlatform(platform string) string {
 	if gce.IsWindows(platform) {
-		return "Restart-Service google-cloud-ops-agent -Force"
+		return `
+Restart-Service google-cloud-ops-agent -Force
+if (!$?) {
```
martijnvans, line 85: consider a TODO to remove this once the bug  is fixed. optional.

martijnvans (approved):

integration_test/ops_agent_test.go
jefferbrecht: Oops, this was a leftover debugging thing, removed.

integration_test/ops_agent_test.go
jefferbrecht: Added

jefferbrecht:
In [this build](https://source.cloud.google.com/results/invocations/aa3ec0f0-738e-4f26-9e7e-14ae48cbec46/targets), no logs showed up after reinstalling. I presume fluent-bit got stuck during the googet reinstall, which strangely completed without error. We can work around that flake too but I'll leave it for a separate PR if needed -- the existing retries might be enough now to unblock the nightlies. At any rate I re-ran that test locally and it passed so I'm convinced it's just a flake and we're no worse off (this change doesn't affect that specific test). Everything else passed so I'm merging now.

PR #915, opened 2022-10-19
Revert "Testing: use ssh to connect to Windows instances"

martijnvans:
Reverts GoogleCloudPlatform/ops-agent#873. It breaks active_directory_ds tests. :(

PR #873, opened 2022-09-26
Testing: use ssh to connect to Windows instances

martijnvans:
## Description
This has several benefits, such as allowing ObservIQ members to be able to run Windows tests locally, and deletes our dependency on winrm.par. (No more need to specify WINRM_PAR_PATH!). It is also just really nice that we are doing things more or less the same for Linux and for Windows.

## Related issue
b/209616302

## How has this been tested?
Integration tests

## Checklist:
- Unit tests
  - [X] Unit tests do not apply.
  - [ ] Unit tests have been added/modified and passed for this PR.
- Integration tests
  - [ ] Integration tests do not apply.
  - [X] Integration tests have been added/modified and passed for this PR.
- Documentation
  - [X] This PR introduces no user visible changes.
  - [ ] This PR introduces user visible changes and the corresponding documentation change has been made.
- Minor version bump
  - [X] This PR introduces no new features.
  - [ ] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.
  - [ ] This PR bumps the version.

PR #916, opened 2022-10-20
Restore "Testing: use ssh to connect to Windows instances"

martijnvans:
Reverts GoogleCloudPlatform/ops-agent#915, restoring GoogleCloudPlatform/ops-agent#873, with additional fixes for active directory.

Tested locally with all applications, on the following non-comprehensive list of platforms: debian-10,rocky-linux-8,sql-std-2019-win-2019,windows-2019,windows-2016,windows-2012-r2


jefferbrecht (approved):
Test failures look like the typical 2012 flakes, LGTM

PR #944, opened 2022-11-04
Testing: Add workaround for windows-2012 flakes

jefferbrecht:
## Description
Applies the processor affinity workaround we discovered when debugging the fluent-bit startup lock issue. This should be reverted once b/240564518 is fixed in fluent-bit.

## Related issue
b/240564518

## How has this been tested?
Tested a sample of previously-flaky nightly integration tests locally.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -253,9 +253,27 @@ func setupOpsAgentFrom(ctx context.Context, logger *logging.DirectoryLogger, vm 
 			// Sleep to avoid some flaky errors when restarting the agent because the
 			// services have not fully started up yet.
 			time.Sleep(startupDelay)
+			// TODO(b/240564518): Workaround for fluent-bit startup hang on 2012. This requires
+			// a restart of the agents, so perform this step before restartOpsAgent below.
+			if strings.Contains(vm.Platform, "2012") {
+				if _, err := gce.RunScriptRemotely(
+					ctx,
+					logger,
+					vm,
+					`$ErrorActionPreference = "Stop"
+					$key = "HKLM:\SYSTEM\CurrentControlSet\Services\google-cloud-ops-agent-fluent-bit"
+					$old_path = (Get-ItemProperty -Path $key -Name "ImagePath").ImagePath
+					$new_path = "cmd.exe /k start /AFFINITY FF /WAIT `+"`\"`\""+` $old_path"
+					Set-ItemProperty -Path $key -Name "ImagePath" $new_path`,
+					nil,
+					nil,
+				); err != nil {
+					return fmt.Errorf("setupOpsAgentFrom() windows-2012 workaround failed: %v", err)
+				}
+			}
 		}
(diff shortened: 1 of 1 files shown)


integration_test/ops_agent_test.go
```diff
@@ -253,9 +253,27 @@ func setupOpsAgentFrom(ctx context.Context, logger *logging.DirectoryLogger, vm
 			// Sleep to avoid some flaky errors when restarting the agent because the
 			// services have not fully started up yet.
 			time.Sleep(startupDelay)
+			// TODO(b/240564518): Workaround for fluent-bit startup hang on 2012. This requires
+			// a restart of the agents, so perform this step before restartOpsAgent below.
+			if strings.Contains(vm.Platform, "2012") {
```
martijnvans, line 258: what's the relationshiip of this workaround to the one inside `restartCommandForPlatform()`? can we remove the one inside `restartCommandForPlatform()` or do we need both? :/

jefferbrecht: We need both unfortunately, because ops agent starts up immediately after we install it which doesn't give us an opportunity to install the workaround -- it will have already locked up.

Another approach would be to add this workaround to our production service definition for Windows builds, which should eliminate the need for both workarounds in the tests, but then customers have to see it too...

martijnvans (approved):

integration_test/ops_agent_test.go
jefferbrecht: As an alternative, we could at least move the restart workaround into here, since this should be the only spot where we actually need to force-kill fluent-bit.

e.g.:

```go
			// TODO(b/240564518): Workaround for fluent-bit startup hang on 2012. This requires
			// a restart of the agents, so perform this step before restartOpsAgent below.
			if strings.Contains(vm.Platform, "2012") {
				if _, err := gce.RunScriptRemotely(
					ctx,
					logger,
					vm,
					`$ErrorActionPreference = "Stop"
					Get-WmiObject -Class Win32_Service -Filter "Name LIKE 'google-cloud-ops-agent%'" | ForEach-Object {		
						Stop-Process -Force $_.ProcessId -ErrorAction SilentlyContinue
						if (!$?) {
							Write-Output "Could not stop process $($_.ProcessId); proceeding anyway"
						}
					}
					$key = "HKLM:\SYSTEM\CurrentControlSet\Services\google-cloud-ops-agent-fluent-bit"
					$old_path = (Get-ItemProperty -Path $key -Name "ImagePath").ImagePath
					$new_path = "cmd.exe /k start /AFFINITY FF /WAIT `+"`\"`\""+` $old_path"
					Set-ItemProperty -Path $key -Name "ImagePath" $new_path`,
					nil,
					nil,
				); err != nil {
					return fmt.Errorf("setupOpsAgentFrom() windows-2012 workaround failed: %v", err)
				}
			}
```

and then we can pull the workaround out of the restart command.

martijnvans: either way is fine with me

PR #948, opened 2022-11-07
Put Windows 2012 workaround in correct spot

jefferbrecht:
## Description
I accidentally put the workaround from #944 in the section that only runs when there's a non-empty Ops Agent config. I want it to run unconditionally so this moves the code outside that section.

## Related issue
b/240564518

## How has this been tested?
Ran `TestSystemLogByDefault/windows-2012-r2` locally, which uses a blank config and which failed on the nightlies even after #944; it passes now.

## Checklist:
- Unit tests
  - [x] Unit tests do not apply.
  - [ ] Unit tests have been added/modified and passed for this PR.
- Integration tests
  - [ ] Integration tests do not apply.
  - [x] Integration tests have been added/modified and passed for this PR.
- Documentation
  - [x] This PR introduces no user visible changes.
  - [ ] This PR introduces user visible changes and the corresponding documentation change has been made.
- Minor version bump
  - [ ] This PR introduces no new features.
  - [ ] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.
  - [ ] This PR bumps the version.

PR #952, opened 2022-11-08
Add workaround for Windows 2012 fluent-bit lockups

jefferbrecht:
## Description
This is the user-facing counterpart to #944 and #948. The intent is to unblock users while simultaneously simplifying the workarounds away from our test code.

While this is a user-facing change, I don't expect there to be any negative impact on users, and if something unforeseen _does_ occur then the user still has a workaround which is to manually change the service startup parameters back to where they were before.

## Related issue
b/240564518

## How has this been tested?
Tested locally + presubmits

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/cmd/ops_agent_windows/main_windows.go
+++ b/cmd/ops_agent_windows/main_windows.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 
 	"github.com/kardianos/osext"
+	"golang.org/x/sys/windows/registry"
 	"golang.org/x/sys/windows/svc"
 )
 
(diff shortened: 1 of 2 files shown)


jefferbrecht:
`TestExcludeMetrics` is still failing for some reason, and it appears to be the lockup again, so I'm adding a debug log to look into that...

`TestUpgradeOpsAgent` may still fail on 2012 even with this fix since that tests operates on an already-released Ops Agent without the fix.

Options:

1. Add a workaround in the tests for that specific test configuration (I don't want this)
2. Do nothing and hopefully there are enough retries for the nightly to get a passing build (I'm ok to try this first)
3. Disable this test on 2012 temporarily until we get the next release out (I'd also be ok with this, provided we run the test locally until it passes which should be easy)

braydonk (approved):
I love it and I hate it LGTM

PR #953, opened 2022-11-08
Adding feature tracking into diagnostic service

avilevy18:
## Description

Sending feature tracking through diagnostic service

## Related issue

b/246812085

## How has this been tested?

Integration tests

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/covprofile
```diff
@@ -0,0 +1,9433 @@
+mode: set
```
ridwanmsharif, line 1: This file doesn't need to be checked in does it?

integration_test/gce/gce_testing.go
```diff
@@ -476,6 +476,30 @@ func WaitForMetric(ctx context.Context, logger *log.Logger, vm *VM, metric strin
 	return nil, fmt.Errorf("WaitForMetric(metric=%s, extraFilters=%v) failed: %s", metric, extraFilters, exhaustedRetriesSuffix)
 }
 
+func WaitForMetricSeries(ctx context.Context, logger *log.Logger, vm *VM, metric string, window time.Duration, extraFilters []string, isPrometheus bool) ([]*monitoringpb.TimeSeries, error) {
```
jefferbrecht, line 479: Please add a function header documenting this function.

integration_test/ops_agent_test.go
```diff
@@ -1688,21 +1690,22 @@ func testDefaultMetrics(ctx context.Context, t *testing.T, logger *logging.Direc
...
 	}
-
 	err = yaml.UnmarshalStrict(bytes, &agentMetrics)
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	expectedMetrics := agentMetrics.ExpectedMetrics
 
+	logger.ToMainLog().Println("Testing Logger: finding representative metrics")
```
jefferbrecht, line 1700: Are these still needed?

integration_test/ops_agent_test.go
```diff
@@ -1761,12 +1766,40 @@ func testDefaultMetrics(ctx context.Context, t *testing.T, logger *logging.Direc
...
+
+	var fc feature_tracking.FeatureTrackingContainer
+	logger.ToMainLog().Printf("Read features.yaml successful\n")
+
+	err = yaml.UnmarshalStrict(featureBytes, &fc)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	series, err := gce.WaitForMetricSeries(ctx, logger.ToMainLog(), vm, "agent.googleapis.com/agent/internal/ops/feature_tracking", window, nil, false)
```
jefferbrecht, line 1785: If I understand this right, we're using one feature per time series? It seems like it would be easier to combine all features into one time series. Doing it this way makes it difficult to know whether you've gotten them all, because metrics are eventually consistent.

In this test, we know ahead of time how many features to expect since they're read from features.yaml, so I would think `WaitForMetricSeries` should take that number as a parameter and return either an array of time series of that length or an error.

integration_test/third_party_apps_test.go
```diff
@@ -460,7 +461,14 @@ func runLoggingTestCases(ctx context.Context, logger *logging.DirectoryLogger, v
...
 
-func runMetricsTestCases(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM, metrics []*metadata.ExpectedMetric) error {
+type feature struct {
+	Module  string
+	Feature string
+	Key     string
+	Value   string
+}
+
+func runMetricsTestCases(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM, metrics []*metadata.ExpectedMetric, fc feature_tracking.FeatureTrackingContainer, skipFeatureTracking bool) error {
```
jefferbrecht, line 471: Do we really need two new parameters here? Why not just make `fc` nillable and use its nil-ness as a flag to skip?

internal/self_metrics/self_metrics.go
```diff
@@ -113,9 +117,40 @@ func InstrumentEnabledReceiversMetric(uc *confgenerator.UnifiedConfig, meter met
...
+		return fmt.Errorf("failed to register callback: %w", err)
+	}
+	return nil
+}
+
+func CollectOpsAgentSelfMetrics(userUc, mergedUc *confgenerator.UnifiedConfig, death chan bool) error {
 	// Resource for GCP and SDK detectors
 	ctx := context.Background()
+
+	//ENABLED RECEIVERS
```
jefferbrecht, line 153: Remove?

cmd/google_cloud_ops_agent_diagnostics/main.go
```diff
@@ -21,18 +21,23 @@ import (
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
 )
 
-func getUnifiedConfigAndValidate(userConfPath, platform string) (confgenerator.UnifiedConfig, error) {
+func getUnifiedConfigAndValidate(userConfPath, platform string) (confgenerator.UnifiedConfig, confgenerator.UnifiedConfig, error) {
```
jefferbrecht, line 24: * Please document the return values -- currently there is no distinction between the two `UnifiedConfig`s
* Why do we need to separate the user config from the merged config?

confgenerator/covprofile
avilevy18: I have no idea how it got in there. Removing it now.

integration_test/ops_agent_test.go
avilevy18: Removed

internal/self_metrics/self_metrics.go
avilevy18: Removed.

integration_test/ops_agent_test.go
```diff
@@ -1761,12 +1766,40 @@ func testDefaultMetrics(ctx context.Context, t *testing.T, logger *logging.Direc
...
+
+	logger.ToMainLog().Printf("Attempting to read features.yaml\n")
+	featureBytes, err := os.ReadFile(path.Join("agent_metrics", "features.yaml"))
+	if err != nil {
+		t.Fatalf("Could not find features.yaml\n")
+		return
+	}
+
+	var fc feature_tracking.FeatureTrackingContainer
+	logger.ToMainLog().Printf("Read features.yaml successful\n")
```
jefferbrecht, line 1778: You generally don't need `Printf` or `Fatalf` if you're not formatting any text. Just use `Println` and `Fatal` without the `\n`.

integration_test/ops_agent_test.go
jefferbrecht: The test doesn't even seem to reach this point actually; it stops abruptly at `Read features.yaml successful`. (See https://source.cloud.google.com/results/invocations/d38c3603-5544-4f8d-a3bd-9002179e188b, then look at `main_log.txt` in the test logs for `TestDefaultMetricsNoProxy/debian-10`.)

jefferbrecht: Nevermind, it looks like it's being run and it simply doesn't print anything.

cmd/google_cloud_ops_agent_diagnostics/main.go
```diff
@@ -21,18 +21,25 @@ import (
...
+func getUnifiedConfigAndValidate(userConfPath, platform string) (userUc, mergedUc confgenerator.UnifiedConfig, err error) {
+	userUc, err = confgenerator.ReadUnifiedConfigFromFile(userConfPath, platform)
+	if err != nil {
+		return confgenerator.UnifiedConfig{}, confgenerator.UnifiedConfig{}, err
+	}
 	_, mergedConfig, err := confgenerator.MergeConfFiles(userConfPath, platform, apps.BuiltInConfStructs)
 	if err != nil {
-		return confgenerator.UnifiedConfig{}, err
+		return confgenerator.UnifiedConfig{}, confgenerator.UnifiedConfig{}, err
+
```
jefferbrecht, line 34:
```suggestion
```

integration_test/feature_tracking/feature_tracking.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/ops-agent/internal/set"
+	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
+)
+
+var errMissingFeatureTrackingMetrics = errors.New("missing feature tracking metrics")
+
+type MissingExpectedFeatureTrackingMetricsError struct {
+	Err                     error
```
jefferbrecht, line 14: This field doesn't get used anywhere outside this package and can be removed. The string used to create `errMissingFeatureTrackingMetrics` can just be inlined in the `Error()` function.

integration_test/feature_tracking/feature_tracking.go
```diff
@@ -0,0 +1,57 @@
...
+	Feature string
+	Key     string
+	Value   string
+}
+
+type FeatureTrackingContainer struct {
+	Features []*FeatureTracking `yaml:"features"`
+}
+
+func AssertFeatureTrackingMetrics(series []*monitoringpb.TimeSeries, features []*FeatureTracking) error {
```
jefferbrecht, line 33: This function (and the error type) should be moved to test code. I don't see a use case for it in the `feature_tracking` package.

integration_test/gce/gce_testing.go
```diff
@@ -415,14 +415,15 @@ func lookupTrace(ctx context.Context, logger *log.Logger, vm *VM, window time.Du
 // time series. An error is returned if the evaluation fails.
 // A return value of (nil, nil) indicates that the evaluation succeeded
 // but returned no data.
-func nonEmptySeries(logger *log.Logger, it *monitoring.TimeSeriesIterator) (*monitoringpb.TimeSeries, error) {
+func nonEmptySeries(logger *log.Logger, it *monitoring.TimeSeriesIterator) ([]*monitoringpb.TimeSeries, error) {
```
jefferbrecht, line 418: * Please update the function name and header so it matches the new behaviour
* This will no longer terminate early in case the caller only needs a single time series. I would add a `minimumRequiredSeries int` parameter to facilitate that.

integration_test/gce/gce_testing.go
```diff
@@ -415,14 +415,15 @@ func lookupTrace(ctx context.Context, logger *log.Logger, vm *VM, window time.Du
...
 	// Loop through the iterator, looking for at least one non-empty time series.
+	tsList := make([]*monitoringpb.TimeSeries, 0)
 	for {
 		series, err := it.Next()
 		logger.Printf("nonEmptySeries() iterator supplied err %v and series %v", err, series)
 		if err == iterator.Done {
-			// Either there were no data series in the iterator or all of them were empty.
-			return nil, nil
+			// Success, we found a time series with len(series.Points) > 0.
```
jefferbrecht, line 425: What if the iterator is empty? It is awkward for a function named `nonEmptySeries` to produce an empty series without an error.

integration_test/ops_agent_test.go
```diff
@@ -1680,6 +1680,8 @@ func testDefaultMetrics(ctx context.Context, t *testing.T, logger *logging.Direc
 		}
 	}
 
+	logger.ToMainLog().Println("Testing Logger: testDefaultMetrics")
+
```
jefferbrecht, line 1684:
```suggestion
```

internal/set/set.go
```diff
@@ -54,3 +54,7 @@ func (s Set[T]) Keys() []T {
 	}
 	return result
 }
+
+func (s Set[T]) IsEmpty() bool {
```
jefferbrecht, line 58: This isn't used and can be removed.

integration_test/third_party_apps_test.go
```diff
@@ -627,14 +647,35 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
 
 	if metadata.ExpectedMetrics != nil {
 		logger.ToMainLog().Println("found expectedMetrics, running metrics test cases...")
-		if err = runMetricsTestCases(ctx, logger, vm, metadata.ExpectedMetrics); err != nil {
+
+		fc, err := getExpectedFeatures(app)
+		//if err ==
+
```
jefferbrecht, line 653:
```suggestion
```

integration_test/third_party_apps_test.go
```diff
@@ -460,7 +461,14 @@ func runLoggingTestCases(ctx context.Context, logger *logging.DirectoryLogger, v
 	return err
 }
 
-func runMetricsTestCases(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM, metrics []*metadata.ExpectedMetric) error {
+type feature struct {
+	Module  string
+	Feature string
+	Key     string
+	Value   string
+}
```
jefferbrecht, line 469: This doesn't appear to be used and can be removed.

integration_test/feature_tracking/feature_tracking.go
avilevy18: Changed file name to `feature_tracking_metadata`

integration_test/feature_tracking/feature_tracking.go
jefferbrecht: This field can still be removed, and then `Error()` can be simplified to:

```go
func (e *MissingExpectedFeatureTrackingMetricsError) Error() string {
	return fmt.Sprintf("missing feature tracking metrics: %v", e.MissingExpectedFeatures)
}
```

integration_test/gce/gce_testing.go
```diff
@@ -411,18 +416,28 @@ func lookupTrace(ctx context.Context, logger *log.Logger, vm *VM, window time.Du
...
 
-// nonEmptySeries evaluates the given iterator, returning its first non-empty
-// time series. An error is returned if the evaluation fails.
-// A return value of (nil, nil) indicates that the evaluation succeeded
-// but returned no data.
-func nonEmptySeries(logger *log.Logger, it *monitoring.TimeSeriesIterator) (*monitoringpb.TimeSeries, error) {
+// nonEmptySeriesList evaluates the given iterator, returning a non-empty array of
+// time series, the length of the array is guaranteed to be of size minimumRequiredSeries or greater.
+// A panic is issued if minimumRequiredSeries is zero or negative.
+// An error is returned if the iterator provides an array less than minimumRequiredSeries.
```
jefferbrecht, line 422: We're still returning `(nil, nil)` in the case of a successful empty iterator, so I would reword this as:

```
An error is returned if the evaluation fails or produces a non-empty slice with length less than minimumRequiredSeries.
A return value of (nil, nil) indicates that the evaluation succeeded but returned no data.
```

integration_test/gce/gce_testing.go
```diff
@@ -451,29 +465,45 @@ func firstTrace(it *trace.TraceIterator) (*cloudtrace.Trace, error) {
...
+	return series[0], nil
+}
+
+// WaitForMetricSeries looks for the given metrics in the backend and returns it if it
 // exists. An error is returned otherwise. This function will retry "no data"
 // errors a fixed number of times. This is useful because it takes time for
 // monitoring data to become visible after it has been uploaded.
-func WaitForMetric(ctx context.Context, logger *log.Logger, vm *VM, metric string, window time.Duration, extraFilters []string, isPrometheus bool) (*monitoringpb.TimeSeries, error) {
+// A return value of (nil, nil) indicates that the evaluation succeeded
+// but returned no data.
```
jefferbrecht, line 483:
```suggestion
```

integration_test/gce/gce_testing.go
```diff
@@ -451,29 +465,45 @@ func firstTrace(it *trace.TraceIterator) (*cloudtrace.Trace, error) {
...
+func WaitForMetricSeries(ctx context.Context, logger *log.Logger, vm *VM, metric string, window time.Duration, extraFilters []string, isPrometheus bool, minimumRequiredSeries int) ([]*monitoringpb.TimeSeries, error) {
```
jefferbrecht, line 484: Optional: it is unfortunate that the plural of `Series` is `Series`. I wonder if there's a better name for this function or if it's already obvious enough how `WaitForMetric` and `WaitForMetricSeries` differ based on the function signatures.

integration_test/gce/gce_testing.go
```diff
@@ -451,29 +465,45 @@ func firstTrace(it *trace.TraceIterator) (*cloudtrace.Trace, error) {
 	return trace, nil
 }
 
-// WaitForMetric looks for the given metric in the backend and returns it if it
+// WaitForMetric uses WaitForMetricSeries to return a single metric or appropriate error
```
jefferbrecht, line 468: Function headers should be self-contained IMO. I'd preserve the existing header for this function, and then the header for `WaitForMetricSeries` can pretty much be a duplicate of it, with the wording slightly changed to indicate that it returns a slice instead of a single time series.

jefferbrecht (approved):

PR #954, opened 2022-11-09
Get bison package from team vendor repo

sophieyfang:
## Description
15.1 and 15.2 needs bison > 3.4 < 3.8. The public suse repo kept getting wiped out. Thus we will hold the pkg ourselves.

## Related issue
b/256835289

## How has this been tested?
Create a new sles 15.1 then add repo then install bison(b/256835289#comment8)

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -231,8 +231,8 @@ enabled=1
 autorefresh=0
 repo_gpgcheck=0
 gpgcheck=0
-EOF && \
-    zypper addrepo agent-vendor.repo && \
+EOF
+RUN set -x; zypper addrepo agent-vendor.repo && \
     zypper -n --gpg-auto-import-keys refresh && \
     zypper -n update && \
     zypper -n install bison>3.4 && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -223,7 +223,16 @@ FROM opensuse/leap:15.1 AS sles15-build
 
 RUN set -x; zypper -n install git systemd autoconf automake flex libtool libcurl-devel libopenssl-devel libyajl-devel gcc gcc-c++ zlib-devel rpm-build expect cmake systemd-devel systemd-rpm-macros java-11-openjdk-devel unzip zip && \
     # Add home:d4vid:co22 repo to install >3.4 bison
-    zypper addrepo https://packages.cloud.google.com/yum/repos/google-cloud-monitoring-sles15-x86_64-test-20221109-1 && \
+    cat > agent-vendor.repo << EOF
+[google-cloud-monitoring-sles15-x86_64-test]
+name=google-cloud-monitoring-sles15-x86_64-test
+baseurl=https://packages.cloud.google.com/yum/repos/google-cloud-monitoring-sles15-x86_64-test-20221109-1
+enabled=1
+autorefresh=0
+repo_gpgcheck=0
+gpgcheck=0
+EOF && \
+    zypper addrepo agent-vendor.repo && \
     zypper -n --gpg-auto-import-keys refresh && \
     zypper -n update && \
     zypper -n install bison>3.4 && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -219,11 +219,18 @@ COPY . /work
...
-FROM opensuse/leap:15.3 AS sles15-build
-
-RUN set -x; zypper -n install git systemd autoconf automake flex libtool libcurl-devel libopenssl-devel libyajl-devel gcc gcc-c++ zlib-devel rpm-build expect cmake systemd-devel systemd-rpm-macros java-11-openjdk-devel unzip zip && \
-    # Add home:d4vid:co22 repo to install >3.4 bison
-    zypper addrepo https://download.opensuse.org/repositories/home:/d4vid:/co22/15.3/home:d4vid:co22.repo && \
+FROM opensuse/leap:15.1 AS sles15-build
+SHELL ["/bin/bash", "-c"]
+RUN set -x; zypper -n install git systemd autoconf automake flex libtool libcurl-devel libopenssl-devel libyajl-devel gcc gcc-c++ zlib-devel rpm-build expect cmake systemd-devel systemd-rpm-macros java-11-openjdk-devel unzip zip
+# Add agent-vendor.repo to install >3.4 bison
+RUN echo $'[google-cloud-monitoring-sles15-x86_64-test] \n\
```
jefferbrecht, line 226: Can we host this repo file somewhere on Google's side as well? Not a blocker, just curious.

Dockerfile
```diff
@@ -219,11 +219,18 @@ COPY . /work
 WORKDIR /work
 RUN ./pkg/rpm/build.sh
 
-FROM opensuse/leap:15.3 AS sles15-build
-
-RUN set -x; zypper -n install git systemd autoconf automake flex libtool libcurl-devel libopenssl-devel libyajl-devel gcc gcc-c++ zlib-devel rpm-build expect cmake systemd-devel systemd-rpm-macros java-11-openjdk-devel unzip zip && \
-    # Add home:d4vid:co22 repo to install >3.4 bison
-    zypper addrepo https://download.opensuse.org/repositories/home:/d4vid:/co22/15.3/home:d4vid:co22.repo && \
+FROM opensuse/leap:15.1 AS sles15-build
+SHELL ["/bin/bash", "-c"]
```
jefferbrecht, line 223: Why do we need bash?

Dockerfile
```diff
@@ -219,11 +219,18 @@ COPY . /work
...
+# Add agent-vendor.repo to install >3.4 bison
+RUN echo $'[google-cloud-monitoring-sles15-x86_64-test] \n\
+name=google-cloud-monitoring-sles15-x86_64-test \n\
+baseurl=https://packages.cloud.google.com/yum/repos/google-cloud-monitoring-sles15-x86_64-test-20221109-1 \n\
+enabled         = 1 \n\
+autorefresh     = 0 \n\
+repo_gpgcheck   = 0 \n\
+gpgcheck        = 0' > agent-vendor.repo
+RUN set -x; zypper addrepo agent-vendor.repo && \
     zypper -n --gpg-auto-import-keys refresh && \
```
jefferbrecht, line 234: Do we still need this if `gpgcheck` is disabled in our repo?

Dockerfile
sophieyfang: docker by default why not specified, it uses `/bin/sh`. I got the solution from https://stackoverflow.com/a/33439625/8514230 which uses `/bin/bash`

Dockerfile
sophieyfang: Rapture repo is google-owned and also public. Though it will soon be deprecated.

jefferbrecht: Sorry I meant just the text that we're echoing to `agent-vendor.repo`, rather than the package repo itself. If we host `agent-vendor.repo` then we don't have to echo to the file locally.

Dockerfile
jefferbrecht: This should be doable on `/bin/sh`, e.g. https://stackoverflow.com/a/59245613/1188632

Dockerfile
sophieyfang: Creating any file under /etc/yum.repo.d will need extra permission and steps. I found it easy to do it this way.

Dockerfile
sophieyfang: Not sure, i haven't tried that yet.

jefferbrecht (approved):
LGTM, my comments are minor and can be addressed in a later cleanup if warranted. @sophieyfang could you update the branch to the latest master? After it passes I think we can merge this.

PR #980, opened 2022-11-16
Jefferbrecht otlp

jefferbrecht:
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


jefferbrecht:
I'll try to review @quentinmit's work from `quentin-otlp` in this PR. The diff between this branch and that one is a rebase on master + integration tests (OTLP and a tiny Prometheus refactor). @quentinmit LMK if you'd rather proceed with review on your branch instead, though we'd have to rebase it anyway before it can be merged.

confgenerator/config.go
```diff
@@ -743,30 +723,53 @@ func (l *Logging) Validate(platform string) error {
...
-func (m *Metrics) Validate(platform string) error {
+func (uc *UnifiedConfig) MetricsReceivers() (map[string]MetricsReceiver, error) {
+	validReceivers := map[string]MetricsReceiver{}
+	for k, v := range uc.Metrics.Receivers {
+		validReceivers[k] = v
+	}
+	if uc.Combined != nil {
+		for k, v := range uc.Combined.Receivers {
+			if _, ok := uc.Metrics.Receivers[k]; ok {
+				return nil, fmt.Errorf("metrics receiver %q has the same name as generic receiver %q", k, k)
```
jefferbrecht, line 734: @quentinmit: Can we move this to validation? Then we could remove the returned `error` and not have to check the result every time `MetricsReceivers` is called.

apps/otlp.go
```diff
@@ -0,0 +1,62 @@
...
+		Receiver: otel.Component{
+			Type: "otlp",
+			Config: map[string]interface{}{
+				"protocols": map[string]interface{}{
+					"grpc": map[string]interface{}{
+						"endpoint": endpoint,
+					},
+				},
+			},
+		},
```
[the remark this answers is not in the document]
jefferbrecht: Synced offline; this would be in scope for the upcoming refactor when adding trace support.

confgenerator/experimental.go
```diff
@@ -14,3 +22,23 @@ func IsExperimentalFeatureEnabled(receiver string) bool {
...
+func componentValidator(sl validator.StructLevel) {
+	comp, ok := sl.Current().Interface().(ConfigComponent)
+	if !ok {
+		return
+	}
+	feature, ok := requiredFeatureForType[comp.Type]
+	if !ok || IsExperimentalFeatureEnabled(feature) {
+		return
+	}
+	sl.ReportError(nil, "", "", "experimental", fmt.Sprintf("Component of type %q cannot be used with the current version of the Ops Agent", comp.Type))
```
franciscovalentecastro, line 39: I'm confused about this message (maybe I'm misunderstanding the code). Wouldn't this show only if the feature is disabled ( via "EXPERIMENTAL_FEATURES" ) and not related to the Ops Agent version ? Maybe something like  `"Component of type %q is disabled in the Ops Agent."`

integration_test/ops_agent_test.go
```diff
@@ -2475,14 +2485,13 @@ func runResourceDetectorCli(ctx context.Context, logger *logging.DirectoryLogger
...
-	goPath, err := InstallGolang(ctx, logger, vm)
-	if err != nil {
+	if err := installGolang(ctx, logger, vm); err != nil {
 		return nil, err
 	}
 	cmd := fmt.Sprintf(`
 		%s
 		cd %s
-		go run run_resource_detector.go`, goPath, workDir)
+		go run run_resource_detector.go`, goPathCommandForPlatform(vm.Platform), workDir)
```
franciscovalentecastro, line 2494: It seems possible to refactor the code here to use the new `runGoCode` to test the `resourcedetector`, but probably it is too much for this PR. I could be done in a subsequent PR.

integration_test/testdata/otlp/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+			log.Fatal(err)
+		}
+	}()
+
+	meter := global.MeterProvider().Meter("foo")
+	gauge, err := meter.AsyncFloat64().Gauge("otlp.test")
+	if err != nil {
+		log.Fatal(err)
+	}
+	err = meter.RegisterCallback([]instrument.Asynchronous{gauge}, func(c context.Context) {
```
franciscovalentecastro, line 46: Is it enough to test with a `GAUGE` metric ? Maybe the flake was due to the timestamps in a `CUMULUATIVE` are different to a `GAUGE` (start and end are the same).

franciscovalentecastro (commented):
Adding some comments :)

integration_test/ops_agent_test.go
```diff
@@ -2518,17 +2527,130 @@ func InstallGolang(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+			t.Fatal(err)
+		}
+
+		if _, err = gce.WaitForMetric(ctx, logger.ToMainLog(), vm, "workload.googleapis.com/otlp.test", time.Hour, nil, false); err != nil {
+			t.Error(err)
+		}
+	})
+}
+
+func testOtlpTraces(t *testing.T) {
```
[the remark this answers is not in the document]
jefferbrecht: This test is intentionally disabled until traces support is added, which is coming in a future PR.

confgenerator/experimental.go
jefferbrecht: From the customer's point of view, the features enabled via this flag are eventually going to be made GA in a future version -- the error message would go away simply by upgrading the Ops Agent version once the feature is GA.

integration_test/ops_agent_test.go
jefferbrecht: Yeah, I have some other ideas too for simplifying that test which could avoid needing to upload Go code at all for it -- we can definitely do that in a later PR.

integration_test/ops_agent_test.go
```diff
@@ -2518,17 +2527,130 @@ func InstallGolang(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
+		%s
+		cd %s
+		go mod init main
+		go get ./...
+		go run main.go`, goPathCommandForPlatform(vm.Platform), workDir)
+	_, err := gce.RunScriptRemotely(ctx, logger, vm, goInitAndRun, nil, nil)
+	return err
+}
+
+func TestOtlpMetrics(t *testing.T) {
```
franciscovalentecastro, line 2566: NIT : Maybe `TestOTLPMetrics`

integration_test/testdata/otlp/metrics.go
jefferbrecht: The flakes were really weird; for some reason it only happened on Windows, and only sometimes, and I could make it pass more often by adding a simple delay at the end before tearing it down. When I dug into the backend I noticed that GCM was trying to connect two time series points together which were sent on totally separate runs. So my guess is that it needs to either send two points in quick succession to work, or the program needs to live long enough to create its own delta, or something like that.

I figure GAUGE is more reliable at any rate since it doesn't require any delta shenanigans. It is a good point though that we could be testing all supported metric types, but I'd have to root cause the flake issue first.

integration_test/ops_agent_test.go
jefferbrecht: Thanks, will fix

integration_test/testdata/otlp/metrics.go
jefferbrecht: Looks like the CUMULATIVE issue was caused by only sending one point -- if I send a second point with a short delay in between it seems to work reliably.

There also doesn't seem to be a way to send DELTA metrics so I'll leave it at GAUGE and CUMULATIVE for now.

franciscovalentecastro: Awesome!

confgenerator/config.go
```diff
@@ -99,6 +99,8 @@ func (ve validationError) Error() string {
 		return fmt.Sprintf("%q must be a duration of at least %s", ve.Field(), ve.Param())
 	case "endswith":
 		return fmt.Sprintf("%q must end with %q", ve.Field(), ve.Param())
+	case "experimental":
+		return experimentalValidationErrorString(ve)
```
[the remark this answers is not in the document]
jefferbrecht: `"experimental"` comes from the `tag` argument of [`sl.ReportError`](https://github.com/GoogleCloudPlatform/ops-agent/blob/jefferbrecht-otlp/confgenerator/experimental.go#L39) in this case.

I chose to encode the whole error message inside of `ve.Param` because it felt simpler, but I agree maybe it's not great usage of the API. I'll refactor this.

integration_test/gce/gce_testing.go
```diff
@@ -413,6 +436,16 @@ func nonEmptySeries(logger *log.Logger, it *monitoring.TimeSeriesIterator) (*mon
 	}
 }
 
+// firstTrace evaluates the given iterator, returning its first trace,
+// or nil if the iterator is empty.
+func firstTrace(it *trace.TraceIterator) *cloudtrace.Trace {
+	trace, err := it.Next()
+	if err == iterator.Done {
```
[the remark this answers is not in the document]
jefferbrecht: The only [documented](https://pkg.go.dev/cloud.google.com/go/trace@v1.3.0/apiv1#TraceIterator.Next) error here is `iterator.Done`, but I can add that check in just in case.

LujieDuan (commented):
The integration tests looks good to me! Thanks for refactoring the Prometheus test. Left a couple of comments:

confgenerator/feature_tracking_test.go
```diff
@@ -464,10 +464,10 @@ func TestOverrideDefaultPipeline(t *testing.T) {
...
 func TestGolden(t *testing.T) {
 	_ = apps.BuiltInConfStructs
-	components := confgenerator.GetComponentsFromRegistry(confgenerator.LoggingReceiverTypes)
-	components = append(components, confgenerator.GetComponentsFromRegistry(confgenerator.LoggingProcessorTypes)...)
-	components = append(components, confgenerator.GetComponentsFromRegistry(confgenerator.MetricsReceiverTypes)...)
-	components = append(components, confgenerator.GetComponentsFromRegistry(confgenerator.MetricsProcessorTypes)...)
+	components := confgenerator.LoggingReceiverTypes.GetComponentsFromRegistry()
+	components = append(components, confgenerator.LoggingProcessorTypes.GetComponentsFromRegistry()...)
+	components = append(components, confgenerator.MetricsReceiverTypes.GetComponentsFromRegistry()...)
+	components = append(components, confgenerator.MetricsProcessorTypes.GetComponentsFromRegistry()...)
```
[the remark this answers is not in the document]
jefferbrecht: We were going to leave it for later since it doesn't need to block this PR, but if it's just a one-liner then I may as well add it in. Thanks for pointing this out!

confgenerator/experimental.go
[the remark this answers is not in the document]
jefferbrecht: Neat, I didn't realize it would add that in, thanks. I'll change this.

confgenerator/feature_tracking_test.go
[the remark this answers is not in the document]
jefferbrecht: Actually upon second glance this looks like a non-trivial change, so I'd prefer to leave it for a future PR.

confgenerator/experimental.go
```diff
@@ -1,10 +1,18 @@
...
+
+	"github.com/go-playground/validator/v10"
 )
 
+var requiredFeatureForType = map[string]string{
+	"prometheus": "prometheus_receiver",
+	"otlp":       "otlp_receiver",
+}
+
 func IsExperimentalFeatureEnabled(receiver string) bool {
```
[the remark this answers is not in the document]
jefferbrecht: Fixed!

LujieDuan (approved):
LGTM!

PR #986, opened 2022-11-17
Try getting sudo access before running tests

jefferbrecht:
## Description
sles-12 failed in the nightlies for the sudo issue, for which we had previously added a retry policy (5 times with 5 seconds in between). This change moves the retry to a more general area before tests are run, and tries a general sudo command instead: if it succeeds then later testing code should not have issues running sudo commands anymore. The retry count is also greatly increased, from 5 to 60.

## Related issue
b/259122953

## How has this been tested?
Tested sles-12-sp2-sap locally, though this is a flake to begin with so we'll have to assume it's working until it fails again.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/gce/gce_testing.go
```diff
@@ -1497,10 +1492,19 @@ func waitForStartLinux(ctx context.Context, logger *log.Logger, vm *VM) error {
...
-	// TODO(b/259122953): SUSE needs additional startup time. Remove once we have more
-	// sensible/deterministic workarounds for each of the individual problems.
 	if IsSUSE(vm.Platform) {
+		// TODO(b/259122953): SUSE needs additional startup time. Remove once we have more
+		// sensible/deterministic workarounds for each of the individual problems.
 		time.Sleep(slesStartupDelay)
+		// TODO(b/259122953): wait until sudo is ready
+		backoffPolicy := backoff.WithContext(backoff.WithMaxRetries(backoff.NewConstantBackOff(slesStartupSudoDelay), slesStartupSudoMaxAttempts), ctx)
+		err := backoff.Retry(func() error {
+			_, err := RunRemotely(ctx, logger, vm, "", "sudo ls /root")
```
franciscovalentecastro, line 1502: Is there a specific reason to choose `sudo ls /root` as the sudo command ? I like it and I think it's simple enough. My only concern is that something weird would happen like e.g. The `/root` folder takes a while to be created even after `is-system-running` is a success.

jefferbrecht: I wanted a built-in non-destructive command that's guaranteed to only be runnable with `sudo`, and `sudo ls /root` was just the first thing that came to mind. I think `/root` gets created pretty early during initialization, but even if there's a delay after `is-system-running` passes I think that's actually a good thing -- the goal is to block on as much startup initialization as possible.

I suppose I could change it to `sudo stat /etc/sudoers.d/google_sudoers`, which would more directly accomplish the goal of "make sure sudo works", but I'm not sure it matters much either way.

franciscovalentecastro: Ok! Understood, I'm fine with either of the commands you mentioned (`sudo ls /root` or `sudo stat /etc/sudoers.d/google_sudoers`). LGTM!

franciscovalentecastro (approved):

PR #993, opened 2022-11-18
[Draft] Health Check POC

franciscovalentecastro:
## Description
This a very rough and early draft of the Health Checks functionality. I mostly want to get initial feedback on the current implemented ideas and consider ideas (or design options) I might have not considered.

## Related issue
b/258672047

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


internal/health_checks/health_checks.go
```diff
@@ -0,0 +1,248 @@
...
+        // Not on GCE
+        projectId = "Not-on-GCE"
+    }
+    fmt.Println(fmt.Sprintf("==> projectId : %s \n \n", projectId))
+
+    var multiErr error
+    fmt.Println("Health_Checks \n \n")
+
+    if err := NetworkCheck(); err != nil {
+        fmt.Println(fmt.Sprintf("==> NetworkCheckErr : %s \n \n", err))
```
hsmatulisgoogle, line 80: You can use Printf directly instead doing println(sprintf("...."))

franciscovalentecastro:
Reworked some of the structure of the structure of the `health_checks` package by adding a `HealthCheck` interface and registering each check.

internal/health_checks/network_check.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (c NetworkCheck) RunCheck(uc *confgenerator.UnifiedConfig) error {
+
+    // Request to logging API
+    status, _, err := runGetHTTPRequest(loggingAPIUrl)
+    c.LogMessage("==>" + status)
+    if err != nil {
+        fmt.Println(err)
+        return err
+    }
```
jinghan-ma, line 58: According to [the docs for http.Get](https://pkg.go.dev/net/http#Get), "An error is returned if there were too many redirects or if there was an HTTP protocol error. "

In particular, if the network we are on has issues such as improper proxies, I can see an infinite redirect situation happening. 

Should this situation be handled as an error or a test failure?

internal/health_checks/health_checks.go
```diff
@@ -0,0 +1,129 @@
...
+        }    
+
+        result = append(result, fmt.Sprintf("Check: %s, Status: %s", name, c.GetResult()))
+        result = append(result, fmt.Sprintf("Failure: %s", c.GetFailureMessage()))
+        result = append(result, fmt.Sprintf("Solution: %s \n", c.GetSolutionMessage()))
+        // result = append(result, "Log : " + c.GetLogMessage())
+    }
+    result = append(result, "===========================================================")
+
+    return strings.Join(result, "\n"), multiErr
```
jinghan-ma, line 128: Related to my comment on the `http.GET` section, we will have to think about what we do with errors vs failures.

internal/health_checks/api_check.go
```diff
@@ -0,0 +1,82 @@
...
+    c.LogMessage(fmt.Sprintf("==> project : %s \n \n", project))
+
+    c.LogMessage("\n> APICheck \n \n")
+    ctx := context.Background()
+
+    // New Logging Client
+    c.LogMessage("==> New Logging Client \n")
+    logClient, err := logging.NewClient(ctx, project)
+    if err != nil {
+        c.Fail("Logging Client didn't create successfully.", "Check the Logging API is enabled.")
```
jinghan-ma, line 54: I think it might be a good idea to start creating a set of error codes for failure cases. A (probably not very well thought out) example would be something like a struct with an error code and it's associated human readable message, like:

```
{
    code: logging-api-disabled,
    message: "Logging API disabled"
    action: "Check the Logging API is enabled"
    isFatal: true
}
```


Having them all defined and listed out in a separate file allows us to:
1. Easily find and modify errors as the Ops Agent evolves
2. Quickly identify all the tests that could generate a particular error
3. **Have a list of all the possible errors that the Health Checks will produce**. This will come in handy when we need to integrate the results of the health checks to other systems (such as a Health Oracle, or a Health Dashboard)

jinghan-ma (commented):
This is a great start to the Health Checks project! I left a few comments as things to think about moving forward :)

internal/health_checks/network_check.go
franciscovalentecastro: Yes, I think an overall thing to address (see the last commits I added) is how differentiate clearly between `check failure` and `error`.
- How to report an error ?
- Should we handle the errors ?
- How to differentiate this in very detailed cases, e.g. `http.Get` ?

internal/health_checks/health_checks.go
franciscovalentecastro: Responded in previous comment.

internal/health_checks/api_check.go
franciscovalentecastro: Great! I really like this solution and it's very in tune with the next things to add to the current code. I will start working on this. Just a couple of questions and details regarding this solution :

1. How should we differentiate between `check failures` and `errors` ? Is the value `isFatal` would show this difference ? Should we add another value to your proposed struct ?
2. This actually solves the problem **How to do write integration test that work reliably ?** since we only a have a limited amount of error codes to check. This also will make the testing in multiple environments reliable and easier to understand the results.

jinghan-ma:
Another thing to consider once we have the individual tests worked out is having a retry system. We should be able to specify how many times a test should be retried, and in what interval.

This will also update our hierarchy to something like: Health check - (contains multiple) -> Tests - (contain multiple) -> Attempts

franciscovalentecastro:
@jinghan-ma Good! That's a great to add `retires` to final version of the Health Checks. I'm documenting this suggestion in the roadmap for the last milestone.

internal/health_checks/failure_codes.go
```diff
@@ -0,0 +1,100 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package health_checks
+
+import (
+    "fmt"
```
jefferbrecht, line 18: I'd check your IDE settings -- the indentation in this file is all spaces instead of tabs.

internal/health_checks/health_checks.go
```diff
@@ -0,0 +1,154 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package health_checks
+
+import (
+	"fmt"
+	"log"
+    "time"
```
jefferbrecht, line 20: Likewise, there's a mix of spaces and tabs in this file.

internal/health_checks/failure_codes.go
franciscovalentecastro: Thank you for mentioning this ! I'm going to check the settings.

internal/health_checks/ports_check.go
```diff
@@ -58,21 +58,24 @@ func (c PortsCheck) RunCheck() error {
 	// Check for fluent-bit self metrics port
 	err := c.check_port(self_metrics_host, strconv.Itoa(fluentbit.MetricsPort))
 	if err != nil {
+		log.Printf("failed to connect: %s", net.JoinHostPort(self_metrics_host, strconv.Itoa(fluentbit.MetricsPort)))
```
JuddRogers, line 61: Please include the purpose of the port as most people don't have this memorized. 
Fluent-bit metrics port,  OTEL metrics port, ...

franciscovalentecastro: Great suggestion! I'm thinking to create 3 (or 4) separate error codes to clearly differentiate the errors (on `health_checks_error.go`) :
- `FB_METRICS_PORT_ERR` 
- `OTEL_METRICS_PORT_ERR` 
- `FB_RECEIVER_PORT_ERR`
- `OTEL_RECEIVER_PORT_ERR`

JuddRogers: Nice idea.

If these messages will end up in logs Cx can see then the text should also clarify the purpose of the port.

franciscovalentecastro: I added the following error codes that will have all the clarifying information about the failure and which port (and how it is used) is affected : 

https://github.com/GoogleCloudPlatform/ops-agent/blob/4126ab17858a7f648c8ff712dd7986e468f949b4/internal/health_checks/health_check_error.go#L30-L66

They are going to reported from `network_check.go` :

https://github.com/GoogleCloudPlatform/ops-agent/blob/4126ab17858a7f648c8ff712dd7986e468f949b4/internal/health_checks/ports_check.go#L51-L83

The logs are only for further debugging, the error codes are the ones intended for the user.

PR #996, opened 2022-11-21
Change receiver feature gate to a list

jefferbrecht:
## Description
This is in preparation for the OTLP receiver; we'd like to avoid having to define a new env variable for each feature gate.

## Related issue
b/239598894

## How has this been tested?
Unit tests pass, will let integration tests run in PR.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/confgenerator_test.go
```diff
@@ -292,3 +272,27 @@ func TestMain(m *testing.M) {
...
+		Metadata:      map[string]string{"test-key": "test-value"},
+		Label:         map[string]string{"test-label-key": "test-label-value"},
+		InterfaceIPv4: map[string]string{"test-interface": "test-interface-ipv4"},
+	}
+
+	// Set up the test environment with mocked data.
+	confgenerator.MetadataResource = testResource
+
+	// Enable experimental receivers.
+	os.Setenv("EXPERIMENTAL_RECEIVERS", "prometheus")
```
ridwanmsharif, line 297: If we're expanding this to be a more generic featuregate, we should probably not stop at receivers.

Maybe we rename it to `EXPERIMENTAL_FEATURES` and use gates like `prometheus_receiver` and `XYZ_processor` to toggle those components on and off.

confgenerator/prometheus.go
```diff
@@ -221,7 +218,7 @@ func checkTLSConfig(tlsConfig commonconfig.TLSConfig) error {
 
 // validatePrometheus checks the receiver configuration is valid.
 func validatePrometheus(promConfig promconfig.Config) (string, error) {
-	if PrometheusFeatureGate != "enabled" {
+	if !IsExperimentalReceiverEnabled("prometheus") {
```
ridwanmsharif, line 221: Can we define the list of featuregates (like `prometheus`) somewhere central? Perhaps a separate file?

ridwanmsharif (approved):

confgenerator/prometheus.go
jefferbrecht: I'm not sure what the goal of a predefined list would be -- are you looking for validating the list in case the user supplies an invalid feature gate?

confgenerator/confgenerator_test.go
jefferbrecht: Done!

confgenerator/prometheus.go
ridwanmsharif: I was thinking the purpose was twofold:

- If we have a file that lists all the feature flags we support, we have a defined naming convention of what the feature flags should look like and have one place to find them all.
- We could validate that user supplies a valid feature flag that the agent supports.

jefferbrecht: I tried this and I ended up with quite a bit of additional complexity. IMO it's not worth doing this considering experimental features are unsupported, not user-facing, and transient in nature -- we'll be GA-ing both prometheus and OTLP in the coming months, which means all of this can be reverted at that point anyway.

Nonetheless, I pushed it to [`jefferbrecht-refactor-receiver-gate-alternate`](https://github.com/GoogleCloudPlatform/ops-agent/compare/jefferbrecht-refactor-receiver-gate...jefferbrecht-refactor-receiver-gate-alternate) if you'd like to take a look. LMK if you see any opportunities to simplify it.

ridwanmsharif: I think your explanation is pretty sound. lets ship this as is and revisit this if the needs grow in the future. Normally we go to GA directly, so I imagine this is not a problem for a while.

jefferbrecht:
All tests passed in a prior run except for Bionic which was an unrelated failure; merging.

PR #997, opened 2022-11-21
Testing: improve Windows test logs

martijnvans:
## Description

The problem this PR is addressing is that powershell wraps the stderr we want in unnecessary XML, even when we specify -OutputFormat Text as an argument. That bug was fixed in pwsh, but not in powershell unfortunately. The result is that our test logs get clogged up with weird XML garbage, see "Before" example below:

The idea is to parse the XML returned by powershell over stderr, and extract the human readable part and print that.

Before: https://storage.googleapis.com/ops-agents-public-buckets-test-logs/prod/stackdriver_agents/testing/consumer/ops_agent/test/windows/4974/20230622-071546/logs/TestLoggingAgentCrashRestart_windows-2019/main_log.txt

After: https://storage.googleapis.com/ops-agents-public-buckets-test-logs/prod/stackdriver_agents/testing/consumer/ops_agent/test/windows/4977/20230622-082246/logs/TestLoggingAgentCrashRestart_windows-2019/main_log.txt

To do this, I had to separate out the part of runCommand() that logs the command and the raw stdout/stderr into a function runAndLog(), and for runRemotelyWindows(), get it to call runCommand() directly, skipping runAndLog() in order to do its own logging.

## Related issue
b/190860079

## How has this been tested?

I also checked that when there is something useful on stderr, like a "missing cmdlet" error, it does show up right in the logs and in the ResultStore web page, and I checked that if the XML fails to parse, that the raw XML is dumped to the log, as desired.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/gce/gce_testing.go
```diff
@@ -803,7 +822,54 @@ func RunRemotely(ctx context.Context, logger *log.Logger, vm *VM, stdin string,
...
+		}
+		logger.Printf("exit code: %v", exitCode)
+		logger.Printf("stdout+stderr: %s", output.Interleaved)
+	}
+
+	return output, err
+}
+
+// stderrFromXML extracts the human-readable part of the XML returned by
+// powershell over stderr.
```
ridwanmsharif, line 852: Maybe we should clarify here, that if its not XML already (in case the powershell bug is fixed later), this returns the raw stderr

PR #998, opened 2022-11-21
Fix go installation in integration tests

LujieDuan:
## Description
The previous method of extract go and link to `/usr/bin` is not a reliable solution and it caused nightly builds to fail. Borrow the method from PR #980 by @jefferbrecht to extract go and return the command of setting up the PATH, so that we can first run the command and then invoke go. 

Also enable the `TestResourceDetectorOnGCE` test on Windows. 

## Related issue
b/246615284

## How has this been tested?
Tested locally covering nightly distros.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -2437,13 +2437,11 @@ func TestResourceDetectorOnGCE(t *testing.T) {
 // env in the VM. Then run the runner to print out the JSON formatted
 // GCEResource and finally unmarshal it back to an instance of GCEResource
 func runResourceDetectorCli(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM) (*resourcedetector.GCEResource, error) {
-	type fileToUpload struct {
-		local, remote string
-	}
-
 	// Update the resourcedetector package and the go.mod and go.sum
 	// So that the main function can locate the package from the work directory
-	filesToUpload := []fileToUpload{
+	filesToUpload := []struct {
+		local, remote string
+	}{
 		{local: "cmd/run_resource_detector/run_resource_detector.go",
 			remote: "run_resource_detector.go"},
 		{local: "../confgenerator/resourcedetector/detector.go",
@@ -2461,7 +2459,12 @@ func runResourceDetectorCli(ctx context.Context, logger *logging.DirectoryLogger
 	// Create the folder structure on the VM
 	workDir := path.Join(workDirForPlatform(vm.Platform), "run_resource_detector")
 	packageDir := path.Join(workDir, "confgenerator", "resourcedetector")
-	createFolderCmd := fmt.Sprintf("mkdir -p %s", packageDir)
+	var createFolderCmd string
+	if gce.IsWindows(vm.Platform) {
+		createFolderCmd = fmt.Sprintf("New-Item -ItemType Directory -Path %s", packageDir)
+	} else {
+		createFolderCmd = fmt.Sprintf("mkdir -p %s", packageDir)
+	}
 	out, err := gce.RunScriptRemotely(ctx, logger, vm, createFolderCmd, nil, nil)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create folder with %s in VM: %s", createFolderCmd, out.Stderr)
@@ -2473,6 +2476,7 @@ func runResourceDetectorCli(ctx context.Context, logger *logging.DirectoryLogger
 		if err != nil {
 			return nil, err
 		}
+		defer f.Close()
 		err = gce.UploadContent(ctx, logger, vm, f, path.Join(workDir, file.remote))
 		if err != nil {
 			return nil, err
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -2433,7 +2433,7 @@ func TestResourceDetectorOnGCE(t *testing.T) {
 	})
 }
 
-// runResourceDetectorCli uploads the resource detector runner and setup up the
+// runResourceDetectorCli uploads the resource detector runner and sets up the
 // env in the VM. Then run the runner to print out the JSON formatted
 // GCEResource and finally unmarshal it back to an instance of GCEResource
 func runResourceDetectorCli(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM) (*resourcedetector.GCEResource, error) {
@@ -2461,12 +2461,7 @@ func runResourceDetectorCli(ctx context.Context, logger *logging.DirectoryLogger
 	// Create the folder structure on the VM
 	workDir := path.Join(workDirForPlatform(vm.Platform), "run_resource_detector")
 	packageDir := path.Join(workDir, "confgenerator", "resourcedetector")
-	var createFolderCmd string
-	if gce.IsWindows(vm.Platform) {
-		createFolderCmd = fmt.Sprintf("New-Item -ItemType Directory -Path %s", strings.ReplaceAll(packageDir, "/", "\\"))
-	} else {
-		createFolderCmd = fmt.Sprintf("mkdir -p %s", packageDir)
-	}
+	createFolderCmd := fmt.Sprintf("mkdir -p %s", packageDir)
 	out, err := gce.RunScriptRemotely(ctx, logger, vm, createFolderCmd, nil, nil)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create folder with %s in VM: %s", createFolderCmd, out.Stderr)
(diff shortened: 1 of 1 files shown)


integration_test/ops_agent_test.go
```diff
@@ -2436,18 +2433,18 @@ func TestResourceDetectorOnGCE(t *testing.T) {
 	})
 }
 
-// runResourceDetectorCli uploads the cli and setup up the env in the VM.
-// Then run the cli to print out the JSON formatted GCEResource and finally
-// unmarshal it back to an instance of GCEResource
+// runResourceDetectorCli uploads the resource detector runner and setup up the
```
jefferbrecht, line 2436:
```suggestion
// runResourceDetectorCli uploads the resource detector runner and sets up the
```

integration_test/ops_agent_test.go
```diff
@@ -2461,29 +2458,54 @@ func runResourceDetectorCli(ctx context.Context, logger *logging.DirectoryLogger
...
 			return nil, err
 		}
-		err = gce.UploadContent(ctx, logger, vm, f, path.Join(workDir, file.remote))
+		var remotePath string
+		if gce.IsWindows(vm.Platform) {
+			remotePath = strings.ReplaceAll(path.Join(workDir, file.remote), "/", "\\")
+		} else {
+			remotePath = path.Join(workDir, file.remote)
+		}
+		err = gce.UploadContent(ctx, logger, vm, f, remotePath)
```
jefferbrecht, line 2487:
```suggestion
		remotePath := path.Join(workDir, file.remote)
		if gce.IsWindows(vm.Platform) {
			remotePath = strings.ReplaceAll(remotePath, "/", "\\")
		}
		err = gce.UploadContent(ctx, logger, vm, f, remotePath)
```

integration_test/ops_agent_test.go
```diff
@@ -2461,29 +2458,54 @@ func runResourceDetectorCli(ctx context.Context, logger *logging.DirectoryLogger
...
+	if err != nil {
 		return nil, err
 	}
-
-	cmd := fmt.Sprintf("cd %s && go run run_resource_detector.go", workDir)
-	test_out, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", cmd)
+	cmd := fmt.Sprintf(`
+		%s
+		cd %s
+		go run run_resource_detector.go`, goPath, workDir)
```
jefferbrecht, line 2501: `workDir` hasn't had its path separators normalized yet. If that's needed then it should be added, or if it's not needed then we can probably also remove the `ReplaceAll(remotePath, "/", "\\")` from earlier to simplify.

integration_test/ops_agent_test.go
LujieDuan: Thanks! :)

integration_test/ops_agent_test.go
LujieDuan: Good suggestion! Together with the next suggestion, I have removed the different commands of Windows/Non-Windows and simplify them to the same command.

integration_test/ops_agent_test.go
LujieDuan: Removed `ReplaceAll(remotePath, "/", "\\")` as suggested.

integration_test/ops_agent_test.go
```diff
@@ -2461,7 +2458,16 @@ func runResourceDetectorCli(ctx context.Context, logger *logging.DirectoryLogger
 			remote: "go.sum"},
 	}
 
+	// Create the folder structure on the VM
 	workDir := path.Join(workDirForPlatform(vm.Platform), "run_resource_detector")
+	packageDir := path.Join(workDir, "confgenerator", "resourcedetector")
+	createFolderCmd := fmt.Sprintf("mkdir -p %s", packageDir)
```
jefferbrecht, line 2464: Nit: since `runResourceDetectorCli` now runs on Windows targets then `mkdir -p s` isn't _technically_ correct, but I guess we got lucky since in PowerShell `mkdir %s` achieves the desired behaviour and the `-p` is ignored instead of producing an error. We should either provide an alternative command for Windows, or if not then at least add a comment here noting that `mkdir -p %s` happens to work on both platforms.

integration_test/ops_agent_test.go
```diff
@@ -2388,9 +2388,6 @@ func TestResourceDetectorOnGCE(t *testing.T) {
 	t.Parallel()
```
jefferbrecht, line 2388: Maybe not in scope for this PR, but why do we even need this test to begin with? Is it not feasible to test the resource detector via a prometheus pipeline? It might be simpler that way; there's quite a bit of machinery here for this test.

It's also important to call out that the version of `resourcedetector` under test here does not necessarily match the version under test in the rest of the integration tests -- those tests are done against the installed Ops Agent package (which could be the latest public release, could be from REPO_SUFFIX, or could be from AGENT_PACKAGE_IN_GCS), while this test is done against HEAD. Not much of an issue for the automated test pipelines but this could affect anyone testing locally.

If it's not feasible to test `resourcedetector` via prometheus, there are some ways we can still simplify this:

* We can just clone the Ops Agent repo on the test VM instead of needing to specify each individual file to upload
* If that's not preferable, then you can just upload the whole `cmd/run_resource_detector` directory and generate `go.mod` automatically -- then we only have to define one thing to upload instead of six
* The regexes don't add value IMO and can be removed
* We don't need to set up the Ops Agent for this test, we just need a VM
* The assertions can each be condensed into e.g. `assert.Equal(t, actual.InstanceName, vm.Name)` (you can import the `assert` package that we use in `confgenerator_test`)

integration_test/ops_agent_test.go
```diff
@@ -2436,18 +2433,18 @@ func TestResourceDetectorOnGCE(t *testing.T) {
...
 }
 
-// runResourceDetectorCli uploads the cli and setup up the env in the VM.
-// Then run the cli to print out the JSON formatted GCEResource and finally
-// unmarshal it back to an instance of GCEResource
+// runResourceDetectorCli uploads the resource detector runner and sets up the
+// env in the VM. Then run the runner to print out the JSON formatted
+// GCEResource and finally unmarshal it back to an instance of GCEResource
 func runResourceDetectorCli(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM) (*resourcedetector.GCEResource, error) {
 	type fileToUpload struct {
```
jefferbrecht, line 2440: Why not use an anonymous struct:

```
	filesToUpload := []struct {
		local, remote string
	}{
		{
			local:  "cmd/run_resource_detector/run_resource_detector.go",
			remote: "run_resource_detector.go",
		},
		{
			local:  "../confgenerator/resourcedetector/detector.go",
			remote: "confgenerator/resourcedetector/detector.go",
		},
		...
```

integration_test/ops_agent_test.go
LujieDuan: This is good suggestion. I didn't know that PS would ignore `-p`. I agree I think it is better to have the alternative command for Windows here.

integration_test/ops_agent_test.go
LujieDuan: Done!

integration_test/ops_agent_test.go
LujieDuan: Those are good points! I think I will have a separate PR to address those suggestions. 
- Why this test is needed: we have another test `TestPrometheusMetrics` which checks the detected metadata via a Prometheus pipeline. it covers many similar checks as this `resourcedetector` test. This standalone `resourcedetector` was added because: 1. The Prometheus receiver would transform the metadata to the Prometheus format, so with this test we can check the raw values, as a way to make sure the data source (metadata server on GCE) always returns expected values. 2. The `resourcedetector` and some of its fields might be used by other components other than the Prometheus receiver. Though we probably can find ways to cover those two without this test. Let me know what you think. 
- Regarding the issue that this test always test against local HEAD: this is a good point - indeed this is a different behaviour here. My thought was by doing it this way, we can test local changes when running this test locally. The options of cloning the entire repo from a branch or generating `go.mod` and resolving the package automatically would not allow us to test local changes. Again let me know what you think here. I should at least add comments to the test to highlight that this always tests against HEAD and ignore the provided Ops Agent. 
- Yeah I think the regex for extract the json from stdout is not needed. Will remove. 
- Yes - we do not need to set up the Ops Agent for this test. Do you mean I can skip `agents.CommonSetup(t, platform)` and instead just create a VM? 
- Good suggestion - I will replace those with `assert.Equal`!

Again thanks for the questions! I could have this PR ready first to unblock the nightlies and have a separate PR to improve this test. Or I can revert the previous PR to remove this test for now from master while working on those.

PR #1000, opened 2022-11-23
Makefile: Add `go vet` support

braydonk:
## Description
Added go vet support to the makefile. It ignores the package `confgenerator/filter/internal/generated` because there's nothing we can do about that. This PR also fixes a vet issue in
`generate_expected_metrics.go` where a sync.Mutex was being copied by value.

Signed-off-by: braydonk <braydonk@google.com>

## How has this been tested?
Ran the new `go_vet` target and fixed the errors it found.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/cmd/generate_expected_metrics/generate_expected_metrics.go
+++ b/integration_test/cmd/generate_expected_metrics/generate_expected_metrics.go
@@ -60,7 +60,7 @@ var (
 	}
 )
 
-type expectedMetricsMap map[string]metadata.ExpectedMetric
+type expectedMetricsMap map[string]*metadata.ExpectedMetric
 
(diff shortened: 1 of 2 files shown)


avilevy18 (approved):

braydonk:
btw the `go vet` error was a `sync.Mutex` copied by value. The Metric Descriptor struct has a MessageState proto, which has a sync.Mutex. Because of that, we need to change to passing metric descriptors around by reference and not by value so we're not copying the sync.Mutex.

braydonk:
The tests passed on previous run (before force push). Will merge after unit tests pass.

PR #1004, opened 2022-11-24
files: use DB.locking config option to improve input performance

ridwanmsharif:
Fluent Bit has this option available in the `tail` input plugin. From looking at the upstream code, it looks like should help to reduce the number of syscalls on every commit to the DB but at the price of locking the access to the database file to third party programs. But for us, that trade-off seems okay?

We should soak test/performance test this to see if we get any visible benefit.


ridwanmsharif:
Looks like we get some better memory utilization:
image

Marginal/negligible improvement of CPU utilization:
image

And virtually no impact on throughput:
image

braydonk (approved):

PR #1009, opened 2022-12-01
Only prefix metrics with unknown domain

jefferbrecht:
## Description
Existing behaviour was to prefix all metrics with `workload.googleapis.com`. It would be better to match the behaviour of the `googlecloudexporter` which only prefixes metrics that don't match a [list of known domains](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/config.go#L158).

## Related issue
b/259696576

## How has this been tested?
Will run tests in PR.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/apps/otlp.go
+++ b/apps/otlp.go
@@ -15,12 +15,20 @@
 package apps
 
 import (
+	"fmt"
+	"regexp"
+	"strings"
+
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
 )
 
 // TODO: The collector defaults to this, but should we default to 127.0.0.1 or ::1 instead?
 const defaultGRPCEndpoint = "0.0.0.0:4317"
+
+// Keep these in sync:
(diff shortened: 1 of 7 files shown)


apps/otlp.go
```diff
@@ -49,8 +61,34 @@ func (r ReceiverOTLP) Pipelines() []otel.Pipeline {
 			},
 		},
 		Processors: []otel.Component{
+			// The intent here is to add the workload.googleapis.com prefix to any metrics
+			// that do not match the list of known domains in the exporter [1].
+			//
+			// This would ordinarily be accomplished using a metricstransform processor
+			// with a negative-lookahead regexp, but RE2 (the regexp engine used by
```
quentinmit, line 68: Nit: I think it's the Go regex package, not RE2, though they support almost the same set of functionality.

quentinmit (approved):

apps/otlp.go
jefferbrecht: Updated

PR #1017, opened 2022-12-06
Add trace support for OTLP

quentinmit:
## Description
Adds trace support for OTLP.

Creating this PR now so that the integration tests can run, but it's not ready for submission yet.

## Related issue

b/239598894 b/207217898

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


jefferbrecht:
The design doc mentions that both `metrics` and `traces` need to be defined when using `combined`, but the following config passes:

```yaml
combined:
  receivers:
    otlp:
      type: otlp
traces:
  service:
    pipelines:
      otlp:
        receivers:
          - otlp
```

Is this a bug, or was the intent of the design doc that `metrics` and `traces` need to be defined in _the merged config_, not _the user config_? I interpreted the design doc as referring to the user config.

quentinmit:
> Is this a bug, or was the intent of the design doc that `metrics` and `traces` need to be defined in _the merged config_, not _the user config_? I interpreted the design doc as referring to the user config.

You're correct that this is the behavior as-implemented.

We don't have any mechanism to apply validations to the user config. I think I'm assuming that users are unlikely to unintentionally try to enable traces without metrics, so it's not worth trying to hack around that.

jefferbrecht:
Another config that is allowed but I think ought not to be:

```yaml
combined:
  receivers:
    otlp:
      type: otlp
    otlp2:
      type: otlp
metrics:
  service:
    pipelines:
traces:
  service:
    pipelines:
      otlp:
        receivers:
          - otlp
          - otlp2
```

quentinmit:
> Another config that is allowed but I think ought not to be:
> 
> ```yaml
> combined:
>   receivers:
>     otlp:
>       type: otlp
>     otlp2:
>       type: otlp
> metrics:
>   service:
>     pipelines:
> traces:
>   service:
>     pipelines:
>       otlp:
>         receivers:
>           - otlp
>           - otlp2
> ```

I agree, that's only broken because two receivers are trying to bind the same (default, in this case) port. Checking that at validation time seems very difficult to do right (https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use#Preventing_TOCTOU) but I suppose we could at least do a string comparison to see if two receivers use the exact same string as a listen address.

jefferbrecht (approved):
LGTM, we can continue looking through edge cases in a follow-up PR.

confgenerator/testdata/invalid/linux/combined-receiver_otlp_missing_traces/golden/error
```diff
@@ -0,0 +1 @@
+combined receiver "otlp" found with no traces section; did you forget to define a traces pipeline?
```
qingling128, line 1: Can we make the action item more explicit in this error message?

e.g. `when an "otlp" type receiver is specified in the "combined" section, two pipelines are required separately in the metrics and trace sections to include this receiver.`

confgenerator/testdata/invalid/linux/traces-wrong_receiver_type/golden/error
```diff
@@ -0,0 +1 @@
+traces receiver "otlp" from pipeline "otlp" is not defined.
```
qingling128, line 1: Can we make this consistent with logging and metrics sides? Would that require a lot of efforts? If so, ignore this.

aka adjust the folder name to `traces-receiver_unsupported_type` and adjust the error message to:

```
traces receiver with type "apache" is not supported. Supported logging receiver types: [otlp].
```


Reference:
https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/testdata/invalid/linux/logging-receiver_unsupported_type/golden/error

has

```
logging receiver with type "unsupported_type" is not supported. Supported logging receiver types: [apache_access, apache_error, cassandra_debug, cassandra_gc, cassandra_system, couchbase_general, couchbase_goxdcr, couchbase_http_access, couchdb, elasticsearch_gc, elasticsearch_json, files, flink, fluent_forward, hadoop, hbase_system, jetty_access, kafka, mongodb, mysql_error, mysql_general, mysql_slow, nginx_access, nginx_error, oracledb_alert, oracledb_audit, postgresql_general, rabbitmq, redis, saphana, solr_system, syslog, systemd_journald, tcp, tomcat_access, tomcat_system, varnish, vault_audit, wildfly_system, zookeeper_general].
```

confgenerator/config.go
```diff
@@ -791,6 +856,35 @@ func (uc *UnifiedConfig) ValidateMetrics(platform string) error {
...
+	if err != nil {
+		return err
+	}
+	for _, id := range sortedKeys(t.Service.Pipelines) {
+		p := t.Service.Pipelines[id]
+		if err := validateComponentKeys(receivers, p.ReceiverIDs, subagent, "receiver", id); err != nil {
+			return err
+		}
+		if len(p.ProcessorIDs) > 0 {
+			return fmt.Errorf("Traces pipelines do not support processors.")
```
qingling128, line 875: Can we add an invalid unit test case for this `Traces pipelines do not support processors.` error?

confgenerator/config.go
```diff
@@ -737,7 +790,7 @@ func (uc *UnifiedConfig) MetricsReceivers() (map[string]MetricsReceiver, error)
 	if uc.Combined != nil {
 		for k, v := range uc.Combined.Receivers {
 			if _, ok := uc.Metrics.Receivers[k]; ok {
-				return nil, fmt.Errorf("metrics receiver %q has the same name as generic receiver %q", k, k)
+				return nil, fmt.Errorf("metrics receiver %q has the same name as combined receiver %q", k, k)
```
qingling128, line 793: Can we add an invalid unit test case for this `metrics receiver %q has the same name as combined receiver` error?

qingling128 (approved):
LGTM with a few non-blocking comments. Follow up PRs are totally fine.

confgenerator/otel/modular.go
```diff
@@ -24,10 +24,22 @@ import (
 
 const MetricsPort = 20201
 
-// Pipeline represents a single OT receiver and zero or more processors that must be chained after that receiver.
+// ReceiverPipeline represents a single OT receiver and zero or more processors that must be chained after that receiver.
+type ReceiverPipeline struct {
```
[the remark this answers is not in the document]
quentinmit: Remember that this is in package `otel`, so the type is already named `otel.ReceiverPipeline` (basically what your third bullet is); the idea is that it was an OTel receiver + a pipeline processing that receiver.

confgenerator/testdata/invalid/linux/traces-wrong_receiver_type/golden/error
quentinmit: It's not testing the same thing as `receiver_unsupported_type`; this is testing trying to reference a metrics receiver in a trace pipeline, not trying to use a metrics receiver type in a trace receiver definition. I just renamed the receiver in the test to make the golden error a little clearer.

confgenerator/testdata/invalid/linux/combined-receiver_otlp_missing_traces/golden/error
quentinmit: Improved, PTAL.

confgenerator/config.go
quentinmit: Done

confgenerator/config.go
quentinmit: Done

qingling128 (approved):
LGTM

PR #1029, opened 2022-12-22
Update Prometheus JSON Exporter Integration Test

LujieDuan:
## Description
`TestPrometheusMetricsWithJSONExporter` tests Prometheus's counter/gauge/untyped metrics with a JSON exporter, to make sure Ops Agent's Prometheus receiver is compatible with the exporter and those types get collected correctly. 

When developing the integration test for histogram/summary types #909, couple of improvements have been made to improve the stability of Prometheus related tests. Bring those to `TestPrometheusMetricsWithJSONExporter`:
- Replace the Python HTTP server with the Golang server (which also fixes b/260616780)
- Use common helper functions to upload files and assert the expected values of the metrics;
- Use systemd to launch the JSON exporter and the http server. 

## Related issue
b/260616780

## How has this been tested?
Tested to make sure the test passes on all distros and can report errors when set to unexpected values. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/ops_agent_test.go
```diff
@@ -1957,84 +1957,77 @@ func TestPrometheusMetrics(t *testing.T) {
...
-			f, err := testdataDir.Open(file.local)
-			if err != nil {
-				t.Fatal(err)
-			}
-			defer f.Close()
-			err = gce.UploadContent(ctx, logger, vm, f, path.Join(workDir, file.remote))
-			if err != nil {
-				t.Fatal(err)
-			}
+			{local: path.Join(prometheusTestdata, "http_server.go"),
```
braydonk, line 1972: Up to you, but I think this format is a bit more readable:
```go
{
    local:  path.Join("testdata", "prometheus", "http_server.go"),
    remote: path.Join("/opt", "go-http-server", "http_server.go"),
},
```

integration_test/ops_agent_test.go
```diff
@@ -1957,84 +1957,77 @@ func TestPrometheusMetrics(t *testing.T) {
...
-			// Out of attempts
-			if attempt == maxAttampts-1 {
-				errString := fmt.Sprintf("last stdout: %s last stderr: %s", liveCheckOut.Stdout, liveCheckOut.Stderr)
-				if liveCheckErr != nil {
-					errString = fmt.Sprintf("last err: %v %s", liveCheckErr, errString)
-				}
-				t.Fatalf("failed to start the HTTP server or JSON exporter - exhausted retries. %s", errString)
-			}
-		}
+		setupScript := `curl -L -o json_exporter.tar.gz \
```
braydonk, line 1993: I see there was originally a script file for setting up the json exporter, and that it was removed in favour of this. What's the goal with that? I kind of like the script file more.

braydonk (commented):
A minor suggestion and a question

integration_test/ops_agent_test.go
LujieDuan: Good question - so the script was doing two things: 
1. download the exporter, and 
2. run it and the http server in the background. 

2 was replaced by the systemd services because of better debug-ability: e.g., `systemctl restart json-exporter-for-prometheus-test` would immediately fail if the service, the binary or the exporter's config is not set up correctly in the VM, and logs of the services get uploaded as part of the test's output.  
For 1 and those `systemctl` commands - there is no particular reason to have them in there, and I agree having a separate script file would make the test cleaner - let me move them to a script file!

braydonk (approved):
LGTM!

PR #1030, opened 2023-01-01
Expose NVML and DCGM receivers as Ops Agent apps

suffiank:
## Description
Expose basic GPU metrics by adding an NVML app (as default Linux configuration) and advanced GPU metrics by adding an DCGM app.

## Checklist:
- Unit tests
  - [x] Unit tests have been added/modified and passed for this PR in submodule opentelemetry-operations-collector.
Note: Currently we are running unit tests manually. We plan to automate them in the submodule opentelemetry-operations-collector later.
- Integration tests
- Documentation
Note: The doc pages will be released in parallel with public preview Ops Agent + GPU metrics feature binary. As we are not merging the GPU metrics into master, it needn't be a blocker for this PR.
- Minor version bump
Note: We will bump the Ops Agent version when we decide to go GA and merge into the master.


suffiank:
cc @LujieDuan The integration tests finally passed. For now it's only runs on the Ubuntu 20.04 platform. Martijn suggested is was OK since we have to run various GPU types rather expect differences across Linux distributions.

suffiank:
@LujieDuan 
1. I would be interested to disable GPU utilization and GPU used memory in the DCGM app.
2. I would be interested to relabel the metrics so as to drop the "profiling" prefix.

apps/dcgm.go
```diff
@@ -0,0 +1,61 @@
...
+			Config: map[string]interface{}{
+				"collection_interval": r.CollectionIntervalString(),
+            "endpoint": r.Endpoint,
+			},
+		},
+		Processors: []otel.Component{
+			otel.NormalizeSums(),
+			otel.MetricsTransform(
+				otel.AddPrefix("workload.googleapis.com"),
+			),
```
LujieDuan, line 52: For dropping the `profiling` prefix from the metrics, please give this a try:
```suggestion
			otel.MetricsTransform(
				otel.ChangePrefix("dcgm\\.gpu\\.profiling\\.", "dcgm.gpu."),
				otel.AddPrefix("workload.googleapis.com"),
			),
```

apps/dcgm.go
```diff
@@ -0,0 +1,61 @@
...
+      r.Endpoint = defaultDcgmEndpoint
+   }
+
+	return []otel.Pipeline{{
+		Receiver: otel.Component{
+			Type: "dcgm",
+			Config: map[string]interface{}{
+				"collection_interval": r.CollectionIntervalString(),
+            "endpoint": r.Endpoint,
+			},
```
LujieDuan, line 46: Try this to disable the receiver from scraping of those metrics:
```suggestion
			Config: map[string]interface{}{
				"collection_interval": r.CollectionIntervalString(),
				"endpoint":            r.Endpoint,
				"metrics": map[string]interface{}{
					"dcgm.gpu.utilization": map[string]bool{
						"enabled": false,
					},
					"dcgm.gpu.memory.bytes_used": map[string]bool{
						"enabled": false,
					},
				},
			},
```

confgenerator/testdata/valid/linux/metrics-receiver_solr/golden/otel.yaml
```diff
@@ -28,6 +28,11 @@ processors:
       exclude:
         match_type: regexp
         metric_names: []
+  filter/default__pipeline_nvml_2:
```
LujieDuan, line 31: We can copy this test `confgenerator/testdata/valid/linux/metrics-receiver_solr/` and use it to create one for dcgm, so that we can use this unit test to easily verify that we are disabling metrics & dropping prefix.

apps/dcgm.go
suffiank: Works ✔️

apps/dcgm.go
suffiank: Is it possible to enable these in the Ops Agent configuration just for integration test? If not, I'm OK to omit it from the integration test since there are scrape tests for the receiver itself.

suffiank:
> > 
> 
> Those two SGTM - regarding disabling the GPU utilization and GPU used memory - do you think it is necessary to make them configurable individually?

I think it's OK to not expose them unless we find a reason they would add any value in addition to the NVML metrics, which will always be accessible whenever the DCGM ones are.

apps/dcgm.go
LujieDuan: We can add configuration fields to optionally enable this, if we believe there are users who would love to include those two kinds of metrics. Otherwise I don't think we can just enable those for the integration tests. 
For example: 
```
type MetricsReceiverDcgm struct {
	confgenerator.ConfigComponent       `yaml:",inline"`
	confgenerator.MetricsReceiverShared `yaml:",inline"`

	Endpoint              string `yaml:"endpoint" validate:"omitempty"`
	CollectGpuUtilization bool   `yaml:"collect_gpu_utilization"`
	CollectMemory         bool   `yaml:"collect_memory"`
}
```

confgenerator/testdata/valid/linux/metrics-receiver_solr/golden/otel.yaml
suffiank: Happy to have that. Can you clarify more how to do so? I'm somewhat confused because I thought these are auto-generated files.

apps/dcgm.go
suffiank: Hm. OK, in that case I'd suggest let's skip them from the integration test.

integration_test/third_party_apps_data/applications/nvml/enable
```diff
@@ -0,0 +1,13 @@
+# The built-in Ops Agent already includes the nvml receiver in the default pipeline.
```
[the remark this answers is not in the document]
suffiank: Unfortunately the process metrics will not exist if no load is generated.

integration_test/gce/gce_testing.go
```diff
@@ -146,7 +146,7 @@ const (
 	// hangs, the test still will be cancelled eventually and all its VMs will be cleaned up.
 	// This amount needs to be less than 4 hours, which is the limit on how long a Kokoro build can
 	// take before it is forcibly killed.
-	SuggestedTimeout = 2 * time.Hour
+	SuggestedTimeout = 60 * time.Hour
```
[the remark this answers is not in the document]
suffiank: Good catch, removed.

integration_test/third_party_apps_data/applications/nvml/enable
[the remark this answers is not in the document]
LujieDuan: That's right - sorry I forgot about that! 
Let's still keep only the comment at the top (as we may potentially use this `enable` file in the public doc `example configuration` section), and move the steps that generates the load to `integration_test/third_party_apps_data/applications/nvml/exercise`. 
Also, instead of use wget to download the `test_cuda_kernel.cc` file, we may want to hard code the source code in the `exercise` file, something like this: 
```
tee /etc/google-cloud-ops-agent/config.yaml > /dev/null << EOF
...the content of test_cuda_kernel.cc...
EOF
g++ -I/usr/local/cuda/include test_cuda_kernel.cc -L/usr/local/cuda/lib64 -lcublas -lcudart -o loopcudakernel
nohup ./loopcudakernel </dev/null >loopcuda.out 2>&1 &
echo "Submitted ./loopcudakernel as pid $!"
```
in the case when we need to modify this, we can run the integration test locally to verify the new change. WDYT?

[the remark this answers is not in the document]
suffiank: Oh, perfect.

LujieDuan:
SLES 12 & 15 tests failure seems unrelated - force running the tests now

LujieDuan (approved):
LGTM! We can merge this to the feature branch now!
Could you update the `submodules/opentelemetry-operations-collector` to the head of the master?

suffiank:
> LGTM! We can merge this to the feature branch now! Could you update the `submodules/opentelemetry-operations-collector` to the head of the master?

PR #1031, opened 2023-01-06
Introduce log rotation by default for self log files

hsmatulisgoogle:
## Description
Fluent-bit self logs can grow quickly and fill up the disk. Currently ops agent doesn't have any built-in log rotation, and users have to configure this themselves. This PR adds log rotation by default, along with some configuration options

## Related issue
b/203811937
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/config.go
```diff
@@ -43,6 +43,7 @@ type UnifiedConfig struct {
 	Metrics  *Metrics  `yaml:"metrics"`
 	// FIXME: OTel uses metrics/logs/traces but we appear to be using metrics/logging/traces
 	Traces *Traces `yaml:"traces,omitempty"`
+	Global *Global `yaml:"global,omitempty"`
```
sophieyfang, line 46: Shall we add some testcases into https://github.com/GoogleCloudPlatform/ops-agent/tree/master/confgenerator/testdata/valid

cmd/agent_wrapper/main.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+	config := getLogFileRotation(ucConfig)
+
+	cmd := exec.Command(command[0], command[1:]...)
+	cmd.Stderr = os.Stderr
+	if logFilename != "" && config.GetEnabled() {
+		logger := lumberjack.Logger{
+			Filename:   logFilename,
+			MaxSize:    config.GetMaxFileSize(), // megabytes
```
jefferbrecht, line 56: The `megabytes` comment might be more appropriate as part of a function header for `GetMaxFileSize`, so that all callers know about it, not just this callsite.

jefferbrecht (changes requested):
Left some comments. Looks pretty good overall!

I really think this would benefit from integration tests to cover at least the following cases:
* Do self logs still make it into Cloud Logging (testing that the basic wrapping functionality works as expected)? We don't have any self-log integration tests currently.
* After a rotation, do new logs still make it to Cloud Logging (testing that the rotation mechanism in fluent-bit works as expected)?
* The by-design purpose of this feature is about preventing disk explosions under abnormally high self-log load. Have we run any performance tests to confirm that this works as expected under such loads?

confgenerator/config.go
sophieyfang: I am a bit confused maybe. Are we exposing this log rotation for customer to config or not? 🤔 Will there be a public doc explaining how to config it? If so, it's user visible then some testcases inside testcases will be good to have as it's straightforward in look and clear to understand right away what config works and what not.

cmd/agent_wrapper/main_test.go
```diff
@@ -0,0 +1,240 @@
...
+		},
+		{
+			beforeLogSizes:   []int64{400 * megabyte},
+			expectedLogSizes: []int64{10, 400 * megabyte},
+			bytesWritten:     10,
+			config:           "",
+		},
+		{
+			beforeLogSizes:   []int64{399 * megabyte},
+			expectedLogSizes: []int64{399*megabyte + 10},
```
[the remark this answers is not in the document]
hsmatulisgoogle: This is enforced by gofmt. If I make this change, gofmt suggests the following diff

```
gofmt -d main_test.go
diff main_test.go.orig main_test.go
--- main_test.go.orig
+++ main_test.go
@@ -169,7 +169,7 @@
 		},
 		{
 			beforeLogSizes:   []int64{399 * megabyte},
-			expectedLogSizes: []int64{399 * megabyte + 10},
+			expectedLogSizes: []int64{399*megabyte + 10},
 			bytesWritten:     10,
 			config:           "",
 		},
```

confgenerator/config.go
sophieyfang: Ugh.. my bad. This one never gets translated into either fluentbit or otel configs. It's remaining as its service

hsmatulisgoogle:
I have replied to every comment AFAIK:
> * Do self logs still make it into Cloud Logging (testing that the basic wrapping functionality works as expected)? We don't have any self-log integration tests currently.

I added a test, but it looks like we do have one already so I removed it: https://github.com/GoogleCloudPlatform/ops-agent/blob/c42b36ec95a70577f8cdf186594b0272e40db945/integration_test/ops_agent_test.go#L2691
> * After a rotation, do new logs still make it to Cloud Logging (testing that the rotation mechanism in fluent-bit works as expected)?

We do not have a test specifically for self log file rotation, but our benchmark tests have periodic rotations (every 10 minutes or so) so any issues with log rotation being picked up should be covered there

> * The by-design purpose of this feature is about preventing disk explosions under abnormally high self-log load. Have we run any performance tests to confirm that this works as expected under such loads?

## log generator + agent wrapper
I ran the `agent_wrapper` tool overnight, and it was able to maintain a throughput of 16MB/s [link](https://screenshot.googleplex.com/9H5FiMmMmRdd4Ve)
image

## VM w/o networking
I also tested it on a VM by disabling networking and having fluent bit spam

```
"Request had insufficient authentication scopes.
```
errors. There were not that many errors spammed though, only a rate of 500bytes/second which we can more than keep up with. I ran this overnight and it was steady as well.

## log generator + agent wrapper directed at `logging-module.log`
 This ended up sort of working, until the two different agent wrappers step on each other by rotating the same file. At this point I feel that the benchmarks we currently have adequately cover this case and its not worth it to spend more time benchmarking

cmd/agent_wrapper/main_test.go
```diff
@@ -0,0 +1,238 @@
...
+const (
+	timeFormat  = "2006-01-02T15-04-05.000"
+	logFileName = "log_output_file"
+	megabyte    = 1024 * 1024
+)
+
+type testCase struct {
+	beforeLogSizes   []int
+	expectedLogSizes []int
+	bytesWritten     int8
```
jefferbrecht, line 22: Nit:
```suggestion
	bytesWritten     int
```

jefferbrecht (approved):
Thanks for the in-depth analysis -- LGTM 👍

cmd/agent_wrapper/main_windows_test.go
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+
+//go:build windows
+// +build windows
+
+package main
+
+import "strings"
+
+func (tc *testCase) getCommand() []string {
```
rafaelwestphal, line 22: optional: Maybe explain a bit what this command is doing, seems like witchery to me.

rafaelwestphal (approved):
LGTM

quentinmit:
Hey, I just saw this get merged and looked at the code. This appears to introduce an extra parent process in the tree, which will cause problems on both Linux and Windows without additional changes that I don't see in this PR. On Windows fluent-bit won't be able to connect to the service manager, and on Linux systemd will not know which process is the main process (see `GuessMainPID` and `Type` in the `systemd.service` manpage.

PR #1038, opened 2023-01-13
Improve linux Dockerfile caching

hsmatulisgoogle:
## Description
Improve Dockerfile caching to speed up subsequent builds on the same machine. It now takes ~8 minutes on my machine to recompile the agent after changing the source code

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


Subbarker (approved):

build.sh
```diff
@@ -17,10 +17,6 @@
 # Package build scripts can use this by setting $DESTDIR before launching the
```
jefferbrecht, line 17: Should we add `set -x` here too like in the other scripts? How about adding `-e` as well?

dockerfiles/compile_test.go
```diff
@@ -0,0 +1,21 @@
...
+		t.Error(err)
+	}
+	expectedDockerfile, err := getDockerfile()
+	if err != nil {
+		t.Error(err)
+	}
+	if expectedDockerfile != string(gotDockerfile) {
+		t.Fatalf("Dockerfile was not generated by ./dockerfiles/compile. Please run `go run ./dockerfiles`")
+	}
+
```
jefferbrecht, line 20:
```suggestion
```

jefferbrecht (approved):

build.sh
hsmatulisgoogle: Done. Added -x to the other scripts as well

build.sh
```diff
@@ -42,94 +38,20 @@ if [ -z "$DESTDIR" ]; then
...
-  # flags to disable them. Prune after build.
-  rm "${DESTDIR}/lib/systemd/system/fluent-bit.service"
-  rm -r "${DESTDIR}${subagentdir}/fluent-bit/etc"
-}
-
 function build_opsagentengine() {
-  mkdir -p "$DESTDIR$prefix/libexec"
-  go build -buildvcs=false -o "$DESTDIR$prefix/libexec/google_cloud_ops_agent_engine" \
+  if [[ ! -f /work/google_cloud_ops_agent_engine ]]; then
+  go build -buildvcs=false -o "/work/google_cloud_ops_agent_engine" \
```
shuLhan, line 44: @hsmatulisgoogle This changes break the build.sh if its run directly. For example,

```
$ BUILD_DISTRO=arch CODE_VERSION="${pkgver}" DESTDIR="$_destdir"  ./build.sh
```

See build output here: https://build.kilabit.info/karajo/job/log?id=aur_google-cloud-ops-agent-git&counter=51

shuLhan: > build.sh is an implementation detail of Dockerfile and is not intended to be run standalone.

I see. I thought it was for building the binaries (the usual `make` flow installed on `$DESTDIR`). I remember reading the rpm spec script and it execute the `build.sh` directly.

> Would you mind sharing the reason why you're building it that way instead of using Dockerfile -- is there something about Dockerfile that doesn't work for you?

I use it to build the ops-agent package for Arch Linux [1] which is not available in Dockerfile.
In that case I will patch the build script manually, and run each .sh script manually.

```
@@ -72,6 +74,13 @@ build() {
        echo "destdir: $_destdir"

        cd "${pkgname}"
+
+       CGO_ENABLED=1 ./builds/otel.sh "$_destdir"
+       ./builds/fluent_bit.sh "$_destdir"
+       ./builds/systemd.sh "$_destdir"
+       ./builds/ops_agent_diagnostics.sh "$_destdir"
+
+       git apply "${srcdir}/0001-build-sh.patch"
        BUILD_DISTRO=arch CODE_VERSION="${pkgver}" DESTDIR="$_destdir" \
                ./build.sh
 }
```
It succesfully build last time I run, but need to test it on live image.

Sorry for the noise.

[1] https://aur.archlinux.org/packages/google-cloud-ops-agent-git

PR #1039, opened 2023-01-16
Add functionality to add and remove tags from VMs.

rafaelwestphal:
## Description
Add functionality to add and remove tags from VMs.

## How has this been tested?
This code will be used to test other functionalities.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


franciscovalentecastro (approved):
LGTM

integration_test/gce/gce_testing.go
```diff
@@ -1731,3 +1733,33 @@ func RunForEachPlatform(t *testing.T, f func(t *testing.T, platform string)) {
 func ArbitraryPlatform() string {
 	return strings.Split(os.Getenv("PLATFORMS"), ",")[0]
 }
+
+func AddTagToVm(ctx context.Context, logger *log.Logger, vm *VM, tag string) (CommandOutput, error) {
```
martijnvans, line 1737: Since the underlying command `add-tags` supports multiple tags, what do you think about making the `tag` argument a []string? We could always make that change later, but I think there's a certain appeal to making this function map onto `add-tags` directly, without adding new restrictions like "only 1 tag at a time".

optional suggestion.

martijnvans, line 1737: I think we should return an error if the input tag contains a comma, because that will confuse gcloud.

martijnvans (commented):
It's more normal to include the code that needs to call these functions in the same PR as the PR adding the functions themselves. That allows reviewers to have more context when reviewing the change, and they might make suggestions like "Oh, it's just this one test that needs these functions? Why not just add them as local functions in ops_agent_test.go?" In this case, I think it's fine to have them as functions in gce_testing.go, but anyway, in the future I'd suggest not splitting apart the caller and the callee into different PRs.

PR #1040, opened 2023-01-16
Parse errors into severity

avilevy18:
## Description

Parse severity from Fluent Bit self logs

## Related issue

b/265979940

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/confgenerator.go
```diff
@@ -32,6 +32,8 @@ import (
 	"github.com/shirou/gopsutil/host"
 )
 
+const FluentBitSelfLogTag = "ops-agent-fluent-bit"
```
jefferbrecht, line 35: This should be unexported

confgenerator/logging_processors.go
```diff
@@ -135,6 +135,7 @@ type LoggingProcessorParseRegex struct {
 	ConfigComponent `yaml:",inline"`
 	ParserShared    `yaml:",inline"`
 	Field           string `yaml:"field,omitempty"`
+	PreserveKey     bool   `yaml:"preserve_key,omitempty"`
```
jefferbrecht, line 138: Was it intentional to include this as an exposed config in the YAML? If so, are we going to leave it undocumented for now?

avilevy18: Changed in: [8145132](https://github.com/GoogleCloudPlatform/ops-agent/pull/1040/commits/81451323dc4ba31fa2296dcc8c76b75838800967)

confgenerator/confgenerator.go
avilevy18: Changed in: [8145132](https://github.com/GoogleCloudPlatform/ops-agent/pull/1040/commits/81451323dc4ba31fa2296dcc8c76b75838800967)

confgenerator/confgenerator.go
```diff
@@ -395,12 +396,16 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
 
 func generateSeveritySelfLogsParser() []fluentbit.Component {
 	out := make([]fluentbit.Component, 0)
+
+	_, timeOffset := time.Now().Zone()
```
jefferbrecht, line 400: `timeOffset` is in seconds here, but `Time_Offset` in fluent-bit is expecting a `+-hhmm`-style offset. Leading zero is also required. Something like this should work:

```go
timeOffset := time.Now().Format("-0700")
```

confgenerator/confgenerator.go
```diff
@@ -395,12 +396,16 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
...
+
 	parser := LoggingProcessorParseRegex{
 		Regex:       `(?<message>\[[ ]*(?<time>\d+\/\d+\/\d+ \d+:\d+:\d+)] \[[ ]*(?<severity>[a-z]+)\].*)`,
 		PreserveKey: true,
 		ParserShared: ParserShared{
 			TimeKey:    "time",
 			TimeFormat: "%Y/%m/%d %H:%M:%S",
+			TimeOffset: fmt.Sprintf("%d", timeOffset),
```
jefferbrecht, line 408: In addition to the above, I also think we should omit this if it's zero to simplify the diff.

confgenerator/fluentbit/processors.go
```diff
@@ -110,6 +111,9 @@ func ParserComponentBase(TimeFormat string, TimeKey string, Types map[string]str
 	if TimeKey != "" {
 		parser.Config["Time_Key"] = TimeKey
 	}
+	if TimeOffset != "" {
```
jefferbrecht, line 114: Actually, the omit-if-zero idea might be more appropriate here instead:

```suggestion
	if TimeOffset != "" && TimeOffset != "+0000" && TimeOffset != "-0000" {
```

confgenerator/confgenerator.go
avilevy18: Great catch, thanks!

confgenerator/confgenerator.go
```diff
@@ -383,12 +386,55 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
...
+	out = append(out, fluentbit.Component{
+		Kind: "FILTER",
+		Config: map[string]string{
+			"Key_Name": "message",
+			"Name":     "parser",
+			"Parser":   "logging-severity-parser.self-logs-severity",
+			"Match":    fluentBitSelfLogTag,
+		},
+	})
+
```
LujieDuan, line 426: Question: Do we need the extra `FILTER` here? Can we let `LoggingProcessorParseRegex` match on `fluentBitSelfLogTag` directly? 
```suggestion
	}.Components(fluentBitSelfLogTag, "self-logs-severity")

	out = append(out, parser...)

```

LujieDuan (commented):
Suggestion: we could add an integration test case to verify the severity got extracted correctly. I believe Fluent Bit always has those `info` log lines on start up, so maybe we could just ```gce.WaitForLog(..., "ops-agent-fluent-bit", ..., `severity="INFO"`)```.

integration_test/ops_agent_test.go
```diff
@@ -2672,6 +2672,27 @@ func TestLoggingAgentCrashRestart(t *testing.T) {
...
+		ctx, logger, vm := agents.CommonSetup(t, platform)
+
+		if err := setupOpsAgent(ctx, logger, vm, ""); err != nil {
+			t.Fatal(err)
+		}
+
+		// Wait long enough for the data to percolate through the backends
+		// under normal circumstances. Based on some experiments, 2 minutes
+		// is normal; wait a bit longer to be on the safe side.
+		time.Sleep(3 * time.Minute)
```
jefferbrecht, line 2688: Is this actually necessary in practice? `gce.WaitForLog` already waits for 400 seconds...

confgenerator/confgenerator.go
avilevy18: Removed extra filter, and adding integration tests

integration_test/ops_agent_test.go
avilevy18: Good catch! Will remove

integration_test/ops_agent_test.go
```diff
@@ -2688,6 +2688,30 @@ func TestLoggingAgentCrashRestart(t *testing.T) {
...
+
+		if err := setupOpsAgent(ctx, logger, vm, ""); err != nil {
+			t.Fatal(err)
+		}
+
+		cmd := `sudo timedatectl set-timezone America/Toronto`
+		if gce.IsWindows(platform) {
+			cmd = `Set-TimeZone -Name "Eastern Standard Time"`
+		}
+		if _, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", cmd); err != nil {
```
jefferbrecht, line 2705: This is happening too late into the test: this changes the time zone at runtime which isn't what we want, and fluent-bit has already finished flushing its startup self-logs by this point anyway so you wouldn't be asserting on logs with the adjusted timestamp. The `setupOpsAgent` command should be run after changing the time zone.

jefferbrecht (approved):

LujieDuan (approved):
LGTM!

quentinmit:
Won't these tests fail if the developer's workstation is not in EST? (Also... does it handle DST correctly?)

jefferbrecht:
I've been using `TZ=UTC+0 go test ./...` to get around it, maybe we can just set that in the unit tests?

quentinmit:
> I've been using `TZ=UTC+0 go test ./...` to get around it, maybe we can just set that in the unit tests?

I don't think that's reasonable to tell anyone, and besides, won't this still break in production during DST due to the hardcoded offset?

jefferbrecht:
> > I've been using `TZ=UTC+0 go test ./...` to get around it, maybe we can just set that in the unit tests?
> 
> I don't think that's reasonable to tell anyone, and besides, won't this still break in production during DST due to the hardcoded offset?

I meant that we can just set the TZ variable in TestMain or something so that the tests run under UTC; the user doesn't have to do anything different.

I'm assuming that the DST offset is rolled into the offset that we emit to the config so I think it'd be fine, but it should still be checked.

quentinmit:
> I'm assuming that the DST offset is rolled into the offset that we emit to the config so I think it'd be fine, but it should still be checked.

Nothing is going to restart fluent-bit every time the DST offset changes, so it will be wrong until the next time fluent-bit happens to restart.

Issue #1035, opened 2023-01-11
Vault prometheus-metrics token exposure


jjtroberts:
**Describe the bug**
The [ops-agent-engine prints its config.yaml](https://github.com/GoogleCloudPlatform/ops-agent/blob/47d43ec9bc7e1dd10a673a169940b089d2e00f23/cmd/google_cloud_ops_agent_engine/main.go#L51) which, when [configured with a vault token](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/vault#example-config), exposes the token in cloud logging.

**To Reproduce**
1. Start a GCE VM with image 'debian-11'
2. Install Ops Agent version '2.24.0' following instructions [here](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/installation#install-latest-version)
3. Customize the agent with configuration from [example here](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/vault#example-config) and restart the agent
4. See error (redacted)
```
google_cloud_ops_agent_engine[4509]: 2023/01/10 19:32:16 Built-in config:
...
google_cloud_ops_agent_engine[4476]:       token: xxxxxxxxxxxxxxxxx”
```

**Expected behavior**
I expect the token to be redacted, or documentation updated to note how to avoid logging the secrets used in the config.yaml.

**Environment (please complete the following information):**
 - Project ID dev-octo-origin
 - VM ID 5802857856404292122
 - VM distro / OS: Debian
 - Ops Agent version 2.24.0
 - Ops Agent configuration
 
Terraform template_file as part of an instance group template startup script:
```
tee /etc/google-cloud-ops-agent/config.yaml > /dev/null << EOF
metrics:
  receivers:
    vault:
      type: vault
      insecure_skip_verify: true
      token: $VAULT_TOKEN
      endpoint: 127.0.0.1:8200
  service:
    pipelines:
      vault:
        receivers:
          - vault
logging:
  receivers:
    vault_audit:
      type: vault_audit
      include_paths: [/var/log/vault/audit.log]
  service:
    pipelines:
      vault:
        receivers:
          - vault_audit
EOF
```

jefferbrecht:
Thanks for raising this; this is definitely something we will try to address for the next release.

Until that release is ready, you can use `exclude_logs` as a workaround to filter token logs (and any other logs with sensitive information by adding filters to `match_any` as appropriate) with this configuration:

```yaml
logging:
[...]

PR #1060, opened 2023-01-24
secret: added secret string type, use for vault

braydonk:
## Description
Added a secret.Data interface and a secret.String type that will allow for custom secret data which will be redacted on output (yaml or print).

Showing example usage on vault Token, future PRs should address other sensitive data.

## Related issue
#1035
b/266204153

## How has this been tested?
Unit tests and manual runs in ops agent engine.

## Checklist:
- Unit tests
  - [] Unit tests have been added/modified and passed for this PR.
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/internal/secret/string.go
+++ b/internal/secret/string.go
@@ -20,6 +20,8 @@ import (
 	"github.com/goccy/go-yaml"
 )
 
+var RedactedValue string = "__redacted__"
+
 type Secret[T any] interface {
 	fmt.Stringer
 	yaml.BytesMarshaler
@@ -30,7 +32,7 @@ type String string
 
 // fmt.Stringer
 func (s String) String() string {
-	return "xxxxx"
+	return RedactedValue
 }
 
 // yaml.BytesMarshaler
--- a/internal/secret/string.go
+++ b/internal/secret/string.go
@@ -28,12 +28,12 @@ type Secret[T any] interface {
 
 type String string
 
+// fmt.Stringer
 func (s String) String() string {
 	return "xxxxx"
 }
 
-// From github.com/goccy/go-yaml. See:
-// https://github.com/goccy/go-yaml/blob/master/yaml.go
+// yaml.BytesMarshaler
 func (s String) MarshalYAML() ([]byte, error) {
 	return []byte(s.String()), nil
 }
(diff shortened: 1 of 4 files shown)


internal/secret/string.go
```diff
@@ -0,0 +1,43 @@
...
+
+package secret
+
+import (
+	"fmt"
+
+	"github.com/goccy/go-yaml"
+)
+
+type Secret[T any] interface {
```
braydonk, line 23: I did not use this interface currently, but I figure if ever we are going to need another type of secret data I decided to make a generic interface. We may decide to remove it later. Currently, it's not causing any harm.

internal/secret/string.go
```diff
@@ -0,0 +1,43 @@
...
+
+type String string
+
+// fmt.Stringer
+func (s String) String() string {
+	return "xxxxx"
+}
+
+// yaml.BytesMarshaler
+func (s String) MarshalYAML() ([]byte, error) {
```
ridwanmsharif, line 37: 🤦 this is so much nicer that what we were discussing. Great find!

ridwanmsharif (approved):
Can we also apply this to all the `password` fields we have in all the apps? Also I'm assuming that this doesn't affect any of the unmarshalling since it falls back to the string unmarshaller, but having that as a test case would be nice!

braydonk:
> Can we also apply this to all the password fields we have in all the apps? 

We can, but I thought that should be done in a followup PR.

> I'm assuming that this doesn't affect any of the unmarshalling since it falls back to the string unmarshaller, but having that as a test case would be nice!

I am pretty sure that's correct, but I will add a test case anyway.

internal/secret/string.go
```diff
@@ -0,0 +1,43 @@
...
+	fmt.Stringer
+	yaml.BytesMarshaler
+	SecretValue() T
+}
+
+type String string
+
+// fmt.Stringer
+func (s String) String() string {
+	return "xxxxx"
```
jefferbrecht, line 33:
```suggestion
	return "redacted"
```

jefferbrecht (approved):

braydonk:
Pretty sure the failures are flakes (windows2012-r2; I'm guessing fluent bit lockups)

jjtroberts:
Thank you so much!

PR #1062, opened 2023-01-25
testing: Disable `rhui-rhel*-7-*` flaky repositories.

franciscovalentecastro:
## Description
b/266093260 | Disabling the following list of rhel 7 repositories that are flaky when testing platforms `rhel-7-7-sap-ha` and `rhel-7-9-sap-ha` :

```
rhui-rhel-7-server-e4s-optional-rhui-debug-rpms
rhui-rhel-7-server-rhui-debug-rpms
rhui-rhel-7-server-rhui-extras-debug-rpms
rhui-rhel-7-server-rhui-extras-source-rpms
rhui-rhel-7-server-rhui-optional-debug-rpms
rhui-rhel-7-server-rhui-optional-rpms
rhui-rhel-7-server-rhui-rh-common-debug-rpms
rhui-rhel-7-server-rhui-rh-common-rpms
rhui-rhel-ha-for-rhel-7-server-e4s-rhui-debug-rpms 
rhui-rhel-server-rhui-rhscl-7-source-rpms
rhui-rhel-server-rhui-rhscl-7-debug-rpms
```

The disabling consists on installing `yum-utils` and then using the command `yum-config-manager` to disable the repositories.

## Checklist:
Tested the fix locally on platforms `rhel-7-7-sap-ha` and `rhel-7-9-sap-ha`.

- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/gce/gce_testing.go
+++ b/integration_test/gce/gce_testing.go
@@ -1161,7 +1161,7 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
 	// Removing flaky rhel-7 repositories due to b/265341502
 	if isRHEL7SAPHA(vm.Platform) {
 		if _, err := RunRemotely(ctx,
-			logger, vm, "", `sudo yum -y --disablerepo=rhui-rhel*-7-server-* install yum-utils && sudo yum-config-manager --disable "rhui-rhel*-7-server-*"`); err != nil {
+			logger, vm, "", `sudo yum -y --disablerepo=rhui-rhel*-7-* install yum-utils && sudo yum-config-manager --disable "rhui-rhel*-7-*"`); err != nil {
 			return nil, fmt.Errorf("disabling flaky repos failed : %w", err)
 		}
 	}
@@ -1182,7 +1182,7 @@ func IsRHEL(platform string) bool {
 }
 
 func isRHEL7SAPHA(platform string) bool {
-	return platform == "rhel-7-7-sap-ha" || platform == "rhel-7-9-sap-ha"
+	return strings.HasPrefix(platform, "rhel-7") && strings.HasSuffix(platform, "-sap-ha")
 }
 
 // CreateInstance launches a new VM instance based on the given options.
--- a/integration_test/gce/gce_testing.go
+++ b/integration_test/gce/gce_testing.go
@@ -1161,7 +1161,7 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
 	// Removing flaky rhel-7 repositories due to b/265341502
 	if isRHEL7(vm.Platform) {
 		if _, err := RunRemotely(ctx,
-			logger, vm, "", `sudo yum -y --disablerepo=rhui-rhel-*-rpms install yum-utils && sudo yum-config-manager --disable "rhui-rhel-*-rpms"`); err != nil {
+			logger, vm, "", `sudo yum -y --disablerepo=rhui-rhel*-7-* install yum-utils && sudo yum-config-manager --disable "rhui-rhel*-7-*"`); err != nil {
 			return nil, fmt.Errorf("disabling flaky repos failed : %w", err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


integration_test/ops_agent_test.go
```diff
@@ -222,6 +222,13 @@ func installOpsAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *g
 		}
 		return nil
 	}
+	// Removing flaky rhel-7 repositories due to b/265341502
+	if gce.IsRHEL7(vm.Platform) {
+		if _, err := gce.RunRemotely(ctx,
+			logger.ToMainLog(), vm, "", `sudo yum -y --disablerepo=rhui-rhel*-7-* install yum-utils && sudo yum-config-manager --disable "rhui-rhel*-7-*"`); err != nil {
+			return fmt.Errorf("yum check-update failed : %v", err)
```
martijnvans, line 229: This error message doesn't match the actual yum commands. Since the error will already have the commands embedded in it, no need to repeat that, maybe just:

`return fmt.Errorf("disabling flaky repos failed : %w", err)`

Also, for new code prefer %w when wrapping an inner error with a new error message because it preserves the type of the underlying error, sorta. https://go.dev/blog/go1.13-errors

integration_test/ops_agent_test.go
```diff
@@ -222,6 +222,13 @@ func installOpsAgent(ctx context.Context, logger *logging.DirectoryLogger, vm *g
 		}
 		return nil
 	}
+	// Removing flaky rhel-7 repositories due to b/265341502
+	if gce.IsRHEL7(vm.Platform) {
```
martijnvans, line 226: we should do this somewhere more general. I suggest here: https://github.com/GoogleCloudPlatform/ops-agent/blob/a200ece61d0bbba1d7e21ae6c54004a69e4b9799/integration_test/gce/gce_testing.go#L1159

Then the repos will be disabled exactly once, at VM startup, before we start using it.

martijnvans: also if you do move it there, let's make isRhel7() private (lowercase) because I don't think we're going to want to use it outside of this file. We can always make it public later.

if you want to keep it as public (uppercase), that's OK with me but in general I prefer to only make things public if we have to.

martijnvans (changes requested):

integration_test/ops_agent_test.go
franciscovalentecastro: Addressed in https://github.com/GoogleCloudPlatform/ops-agent/pull/1062/commits/585ed6fe32bd2c38005fd7f344bcae7993f4d980 , made method private.

integration_test/ops_agent_test.go
franciscovalentecastro: Addressed in https://github.com/GoogleCloudPlatform/ops-agent/pull/1062/commits/585ed6fe32bd2c38005fd7f344bcae7993f4d980 .

martijnvans (approved):

integration_test/gce/gce_testing.go
```diff
@@ -1173,6 +1181,10 @@ func IsRHEL(platform string) bool {
 	return strings.HasPrefix(platform, "rhel-")
 }
 
+func isRHEL7SAPHA(platform string) bool {
+	return platform == "rhel-7-7-sap-ha" || platform == "rhel-7-9-sap-ha"
```
martijnvans, line 1185: we should do something like "if it starts with 'rhel-' and ends with '-sap-ha'" instead of hardcoding 7-7 and 7-9.

martijnvans (changes requested):

integration_test/gce/gce_testing.go
franciscovalentecastro: Ok! That makes more sense. Will update.

franciscovalentecastro: Addressed in https://github.com/GoogleCloudPlatform/ops-agent/pull/1062/commits/62a78b43dd19318c2ffa2b884786833b72098a55 .

franciscovalentecastro: Addressed in https://github.com/GoogleCloudPlatform/ops-agent/pull/1062/commits/774e41cf475c450b283d40dea91401d06d2c5fdb .

martijnvans (approved):

franciscovalentecastro:
Ran locally all integration tests for `rhel-7`, `rhel-7-7-sap-ha`, `rhel-7-9-sap-ha` platforms and they passed with the current state of the fix. Presumbit tests mostly passed, aside from flakes.

PR #1082, opened 2023-02-03
healthchecks: Fix `Ports Check` flake on restart (systemd)

franciscovalentecastro:
## Description
The [Ports Health Check](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/internal/healthchecks/ports_check.go) sometimes shows a false positive in Linux (systemd) since there is no restriction of execution for the subagent services (units)  so they don't overlap during startup with the `google-cloud-ops-agent.service`. This causes the port `20202` still being used by fluent-bit during the startup checks. This doesn't happen in Windows version.

This updates the `systemd` configuration so `google-cloud-ops-agent-fluent-bit.service` &`google-cloud-ops-agent-opentelemetry-collector.service` unit start after the `google-cloud-ops-agent.service` unit. This using the unit `After` systemd configuration.

The false positive error is the following :
```
Ports Check - Result: FAIL, Error code: FbMetricsPortErr, Failure: Port 20202 needed for Ops Agent self metrics is unavailable., Solution: Verify that port 20202 is open., Resource: https://cloud.google.com/logging/docs/agent/ops-agent/tr
oubleshooting
```

## Related issue
b/268046122

## How has this been tested?
- Verified that after adding the `After=` configuration, `fluent-bit` and `otelopscol` are still independent in following sense : 
  - Killing `fluent-bit` or `otelopscol` process don't restart or change the other process and viceversa (same PID).
  - Restarting a subagent service (by `sudo service * restart` ) doesn't change the other process `PID`.
- Integration and unit tests pass.
- Can't reproduce the check flake anymore with 20 repetitions of `restart` in different scenarios :
  - `fluent-bit` is busy (simulating log load)
  -  `fluent-bit` not busy (normal Ops Agent execution).

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/cmd/google_cloud_ops_agent_engine/main.go
+++ b/cmd/google_cloud_ops_agent_engine/main.go
@@ -34,9 +34,7 @@ var (
 )
 
 func runStartupChecks(service string) {
-	switch service {
-	// Run checks in main service
-	case "":
+	if service == "" {
 		time.Sleep(time.Second)
 		gceHealthChecks := healthchecks.HealthCheckRegistryFactory()
 		logger, closer := healthchecks.CreateHealthChecksLogger(*logsDir)
@@ -47,12 +45,6 @@ func runStartupChecks(service string) {
 			log.Printf(result.Message)
 		}
 		log.Println("Startup checks finished")
-	// Adding sleep to reduce flakyness in Ports Checks
-	// when restarting ops agent service
-	case "fluentbit":
-		time.Sleep(2 * time.Second)
-	case "otel":
-		time.Sleep(2 * time.Second)
 	}
 }
 
--- a/cmd/google_cloud_ops_agent_engine/main.go
+++ b/cmd/google_cloud_ops_agent_engine/main.go
@@ -18,7 +18,6 @@ import (
 	"flag"
 	"log"
 	"os"
-	"time"
 
 	"github.com/GoogleCloudPlatform/ops-agent/apps"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
@@ -35,7 +34,6 @@ var (
 
 func runStartupChecks(service string) {
 	if service == "" {
-		time.Sleep(time.Second)
 		gceHealthChecks := healthchecks.HealthCheckRegistryFactory()
 		logger, closer := healthchecks.CreateHealthChecksLogger(*logsDir)
 		defer closer()
--- a/internal/healthchecks/healthchecks.go
+++ b/internal/healthchecks/healthchecks.go
@@ -19,6 +19,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"time"
 )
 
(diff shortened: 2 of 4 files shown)


quentinmit:
`Requires` will cause shared fate (if the dependency dies for any reason, so does the dependent service). We didn't want that for any of the subagents because we want metrics to keep working if the logging agent dies, or any other combination like that. (As the documentation says, you almost never want a `Requires` dependency - `Wants` is generally the right dependency, but in this case we don't want a dependency at all.)

These checks really should be happening as part of each subagent's `PreStart`, since that's the only time the subagent is guaranteed to not be running.

franciscovalentecastro:
Summary discussion with @quentinmit , @braydonk :

- `After` in systemd fixes the ordering when starting the main unit (this fixes the current Ports Check flakiness, current fix). It doesn't help when a subagent restarts by itself. Its aligned with the current `Health Checks` design.
 
- Running the check in `ExecPre` of the subagent is the only way to reliable say that the subagent is not running. This would require to redirect the result journald logs to the main unit. The check would run every time a subagent is restarted, not only when the main service is.

- Doing a `Port Available Check` in Unix by `Listening` to the socket can be unreliable if `SO_REUSEADDR`/ `SO_REUSEPORT` are set.

CC @jinghan-ma

KengoTODA:
Thanks everyone working for this issue and the ops-agent!

This is a note for who is searching solution regarding the `FbMetricsPortErr` error:

```
Ports Check - Result: FAIL, Error code: FbMetricsPortErr, Failure: Port 20202 needed for Ops Agent self metrics is unavailable., Solution: Verify that port 20202 is open., Resource: https://cloud.google.com/logging/docs/agent/ops-agent/tr
oubleshooting
```

In my understanding, no mitigation we have for now, better to use 2.25.0 instead of the latest release 2.25.1.
Hope that this post helps someone working for service relaibility and observability! 👋

franciscovalentecastro:
- Verified (manually in debian-10) that after adding the `After=` configuration, `fluent-bit` and `otelopscol` are still independent in following sense : 
  - Killing `fluent-bit` or `otelopscol` process don't restart or change the other process and viceversa (same PID).
  - Restarting a subagent service (by `sudo service * restart` ) doesn't change the other process `PID`.
- Integration and unit tests pass.
- Can't reproduce the check flake anymore with 20 repetitions of `restart` in different scenarios :
  - `fluent-bit` is busy (simulating log load)
  -  `fluent-bit` not busy (normal Ops Agent execution).

braydonk (approved):

franciscovalentecastro:
All tests passed, aside from unrelated `Debian Buster 3P app` failures and Windows expected flakiness.

jefferbrecht:
@KengoTODA: This fix has now been released: https://github.com/GoogleCloudPlatform/ops-agent/releases/tag/2.26.0

PR #1083, opened 2023-02-03
Enable instrumentation scope labels for workload.googleapis.com

quentinmit:
## Description
Enable the `instrumentation_library_labels` exporter flag for everything except `agent.googleapis.com/*`.  Unfortunately, this means doubling the number of qps a typical Ops Agent sends to `CreateTimeSeries`, because the flag is exporter-wide.

## Related issue
b/267350044

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/otel/modular.go
```diff
@@ -24,14 +24,36 @@ import (
 
 const MetricsPort = 20201
 
+type ExporterType int
+
+const (
+	// N.B. Every ExporterType increases the QPS and thus quota
```
qingling128, line 30: Just curious: Why would new exporter type increase QPS?

confgenerator/otel/modular.go
```diff
@@ -24,14 +24,36 @@ import (
...
+	// consumption in consumer projects; think hard before adding
+	// another exporter type.
+	OTel ExporterType = iota
+	System
+	GMP
+)
+
+func (t ExporterType) Name() string {
+	if t == System || t == GMP {
+		// OTel and GMP have different exporter types so can share the empty string.
```
qingling128, line 40: Does this mean to be `System and GMP have different exporter types`?

integration_test/third_party_apps_test.go
```diff
@@ -641,6 +641,19 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
 	if metadata.ExpectedMetrics != nil {
 		logger.ToMainLog().Println("found expectedMetrics, running metrics test cases...")
 
+		// All integrations are expected to set the instrumentation_source label.
```
qingling128, line 644: Is it possible to add an integration test for the non-3rd party app receivers to verify the behavior of the instrumentation scope labels there?

confgenerator/confgenerator.go
```diff
@@ -52,7 +52,7 @@ func googleCloudExporter(userAgent string) otel.Component {
 				// descriptors to be created implicitly with new time series.
 				"skip_create_descriptor": true,
 				// Omit instrumentation labels, which break agent metrics.
-				"instrumentation_library_labels": false,
+				"instrumentation_library_labels": instrumentationLabels,
 				// Omit service labels, which break agent metrics.
 				"service_resource_labels": false,
```
ridwanmsharif, line 57: From the offline discussions it sounded like the service_resource_labels were also supposed to be turned on for the `OTEL` exporter type?

https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/150#issuecomment-1416170876 for some context

Is that happening at another time or not needed at all for 3P metrics?

confgenerator/confgenerator.go
```diff
@@ -124,12 +124,15 @@ func (uc *UnifiedConfig) GenerateOtelConfig(hostInfo *host.InfoStat) (string, er
...
-		GlobalProcessors:                []otel.Component{gcpResourceDetector()},
-		GoogleCloudExporter:             googleCloudExporter(userAgent),
-		GoogleManagedPrometheusExporter: googleManagedPrometheusExporter(userAgent),
+		LogLevel:          uc.Metrics.Service.LogLevel,
+		ReceiverPipelines: receiverPipelines,
+		Pipelines:         pipelines,
+		GlobalProcessors:  []otel.Component{gcpResourceDetector()},
+		Exporters: map[otel.ExporterType]otel.Component{
+			otel.System: googleCloudExporter(userAgent, false),
+			otel.OTel:   googleCloudExporter(userAgent, true),
```
ridwanmsharif, line 133: quick question: For existing 3P metrics, when users upgrade to Ops Agent with this change in, they will see a discontinuity in their time series at the time of upgrade right? From before and after the label was added.

because this is a static valued label, it won't disaggregate the existing metrics so won't "break" any alerts or dashboards. But are there other unintended affects to a user that we might care about? Do we need some public docs about this change?

ridwanmsharif (commented):
made a pass. Things look good overall, just wanted to ask a couple clarifying questions.

confgenerator/otel/modular.go
quentinmit: Yeah, it's only the case if we're not filling up batches. But the batch size is 200 points, so I think in most applications we're usually sending partial batches.

confgenerator/otel/modular.go
quentinmit: No, it's referring to the collector's exporter type and not the `ExporterType` enum... I rephrased it.

integration_test/third_party_apps_test.go
quentinmit: What are you including in "non-3rd party app receivers"? The existing integration test should be checking the agent metrics, and GMP is separate from this, so isn't everything else being covered by `third_party_apps_test`?

confgenerator/confgenerator.go
quentinmit: It needs to be turned on separately because we need to do more work on making the data correct first (and it's not on the critical path for `instrumentation_scope`).

confgenerator/confgenerator.go
quentinmit: Yeah, they'll see a discontinuity. This is in the set of acceptable changes because it won't change the behavior of any alerts or dashboards. We have b/268237708 to document the new fields.

ridwanmsharif (approved):

integration_test/third_party_apps_test.go
ridwanmsharif: @qingling128 do you mean to check that the instrumentation scope labels are not present in the system metrics? To make sure consumers of those metrics see no discontinuity with their timeseries once this lands?

qingling128: Yes. Basically how the labels behave for the OTLP receiver and host metrics receivers. Ok N the documentation we probably wanna call out their difference as well.

quentinmit: Since system metrics' descriptors are pre-defined, there is no way for Ops Agent to report additional metric labels, so we don't have to worry that users will see any changes. (At worst, the points would just be rejected - which the existing integration tests should catch.)

PR #1086, opened 2023-02-07
Add windows_event_log receiver v2

jefferbrecht:
## Description
This adds a new `receiver_version: 2` for `windows_event_log` receivers, which uses the newer Windows Event Log API. In particular, this will allow for collecting channels that were previously incompatible with the old API, such as `Microsoft-Windows-PrintService/Operational`.

Somewhat orthogonally, this also now registers the Event Log as a sink for internal `log.Print` messages, which were previously discarded. (I made this change to give visibility to a new warning I've added which occurs when you attempt to use an incompatible channel with the old v1 receiver.)

## Related issue
b/262037807

## How has this been tested?
* Added & passed unit tests
* Added & passed integration tests
  * Caveat: I originally called out a few more tests in the design doc to cover all reported-as-not-working channels from customers, but some limitations of the scripting API (see code comments for details) made this very cumbersome, so I am only testing one representative channel for now. This should still provide adequate coverage.

## Checklist:
- Unit tests
- Integration tests
- Documentation
  - [x] Documentation is WIP
- Minor version bump

### DIFF
--- a/cmd/ops_agent_windows/run_windows.go
+++ b/cmd/ops_agent_windows/run_windows.go
@@ -17,6 +17,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 
@@ -31,6 +32,7 @@ import (
 
 const (
 	EngineEventID uint32 = 1
+	StdoutEventID uint32 = 2
 )
 
 func containsString(all []string, s string) bool {
@@ -198,6 +200,19 @@ func (s *service) startSubagents() error {
 	return nil
 }
 
+type eventLogWriter struct {
+	EventID  uint32
+	EventLog *eventlog.Log
+}
+
+func (w *eventLogWriter) Write(p []byte) (int, error) {
+	err := w.EventLog.Info(w.EventID, string(p))
+	if err != nil {
+		return 0, err
+	}
+	return len(p), nil
+}
+
 func run(name string) error {
 	elog, err := eventlog.Open(name)
 	if err != nil {
@@ -205,6 +220,12 @@ func run(name string) error {
 		return err
 	}
 	defer elog.Close()
+
+	// Redirect stdout to the event log to capture internal messages
+	log.SetOutput(&eventLogWriter{
+		EventID:  StdoutEventID,
+		EventLog: elog,
+	})
 
 	elog.Info(1, fmt.Sprintf("starting %s service", name))
 	err = svc.Run(name, &service{log: elog})
--- a/confgenerator/config.go
+++ b/confgenerator/config.go
@@ -29,10 +29,12 @@ import (
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
 	"github.com/GoogleCloudPlatform/ops-agent/internal/set"
+	"github.com/GoogleCloudPlatform/ops-agent/internal/windows"
 	"github.com/go-playground/validator/v10"
 	yaml "github.com/goccy/go-yaml"
 	"github.com/kardianos/osext"
 	promconfig "github.com/prometheus/prometheus/config"
+	"go.uber.org/multierr"
 	"golang.org/x/exp/constraints"
 )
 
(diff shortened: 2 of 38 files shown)


cmd/ops_agent_windows/run_windows.go
```diff
@@ -207,6 +222,12 @@ func run(name string) error {
 	}
 	defer elog.Close()
 
+	// Redirect stdout to the event log to capture internal messages
+	log.SetOutput(&eventLogWriter{
+		EventID:  StdoutEventID,
+		EventLog: elog,
```
sophieyfang, line 228: In this setup, what is the os.stdout being binded to io.Writer of log.SetOutput? 🤔 

Is there test added for verifying the warning message that was previously lost now is tracked in eventlog?

integration_test/ops_agent_test.go
```diff
@@ -1625,6 +1632,128 @@ func TestWindowsEventLog(t *testing.T) {
...
+
+		// Quote-and-escape the query string so that Cloud Logging accepts it
+		expectedWarning := fmt.Sprintf(`"\"logging.receivers.%s.channels\" contains a channel, \"%s\", which may not work properly on version 1 of windows_event_log"`, log, channel)
+		if err := gce.WaitForLog(ctx, logger.ToMainLog(), vm, log, time.Hour, logMessageQueryForPlatform(vm.Platform, expectedWarning)); err != nil {
+			t.Fatal(err)
+		}
+	})
+}
+
+func TestWindowsEventLogV2(t *testing.T) {
```
sophieyfang, line 1670: What will the error message be like when winlog is dealing with a non-exstent channel, e.g. a typo when they input the channel name.

cmd/ops_agent_windows/run_windows.go
jefferbrecht: This test verifies the warning message which was previously lost: https://github.com/GoogleCloudPlatform/ops-agent/pull/1086/files#diff-4c9b98fddf589bb837ae19fbe885c647cdb7d97f048fc2e742e84c0d80554152R1664

integration_test/ops_agent_test.go
jefferbrecht: For v1, fluent-bit silently accepts the mis-typed channel but treats it as `Application` so you end up with duplicate logs.

For v2, fluent-bit emits an error and then terminates. These errors aren't surfaced to Cloud Logging but would be included in the diagnostics bundle.

```
[2023/02/15 10:27:52] [error] [in_winevtlog] cannot subscribe 'Applicatio' (0)
[2023/02/15 10:27:52] [error] [lib] backend failed
[2023/02/15 10:27:52] [error] [input:winevtlog:winevtlog.1] failed to open channels
[2023/02/15 10:27:52] [error] Failed initialize input winevtlog.1
```

confgenerator/config.go
```diff
@@ -273,11 +279,59 @@ func newValidator() *validator.Validate {
 		}
 		return t%tfactor == 0
 	})
+	// winlogchannels wraps validateWinlogchannels when receiver_version > 1
```
sophieyfang, line 282: Why can't we match exactly 2 as the expected winlog version?

confgenerator/testdata/invalid/windows/logging-receiver_windows_event_log_type_receiver_version_not1_unsupported_channel_characters/golden/error
```diff
@@ -0,0 +1,9 @@
+[21:7] "channels[0]" (a,b) contains an invalid character: ,
```
sophieyfang, line 1: I suppose client who wrongly input channels like this was going to do below stuff: 
`channels: [a,b]` 
or 
```
channels:
- a
- b
```
Can we modify the top error message from `"channels[0]" (a,b) contains an invalid character: ,` to  something obvious so that they know they need to put treat channels as a yaml list.

confgenerator/config.go
sophieyfang: Only in the comment that mentions >1, why can't we just say == 2, i saw the validator uses "oneof 1 2"

sophieyfang (approved):

confgenerator/testdata/invalid/windows/logging-receiver_windows_event_log_type_receiver_version_not1_unsupported_channel_characters/golden/error
jefferbrecht: The yaml library already enforces that it's a list and produces a different error if a non-list is supplied. This error is attempting to catch someone adding a literal `,` in the channel name, which isn't allowed.

PR #1087, opened 2023-02-08
Adding disk buffer limit size configuration to ops-agent

rafaelwestphal:
## Description
Adding fluent-bit configuration (total_limit_size) to prevent disk buffer from growing uncontrollably.

## Related issue
b/268344747

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump
  - [X] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.


confgenerator/logging.go
```diff
@@ -44,32 +44,38 @@ func setLogNameComponents(tag, logName, receiverType string, hostName string) []
...
-func stackdriverOutputComponent(match, userAgent string) fluentbit.Component {
-	return fluentbit.Component{
-		Kind: "OUTPUT",
-		Config: map[string]string{
-			// https://docs.fluentbit.io/manual/pipeline/outputs/stackdriver
-			"Name":              "stackdriver",
-			"Match_Regex":       fmt.Sprintf("^(%s)$", match),
-			"resource":          "gce_instance",
-			"stackdriver_agent": userAgent,
+func stackdriverOutputComponent(match, userAgent string, storage_limit_size string) fluentbit.Component {
```
jefferbrecht, line 47:
```suggestion
func stackdriverOutputComponent(match, userAgent string, storageLimitSize string) fluentbit.Component {
```

confgenerator/confgenerator.go
```diff
@@ -373,7 +373,7 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
 			out = append(out, s.components...)
 		}
 		if len(tags) > 0 {
-			out = append(out, stackdriverOutputComponent(strings.Join(tags, "|"), userAgent))
+			out = append(out, stackdriverOutputComponent(strings.Join(tags, "|"), userAgent, "20MB"))
```
jefferbrecht, line 376: Formatting-wise, I think this should be `20M` according to the docs.

But that aside, if I understand this correctly, a limit of 20 MB means that if a customer loses API connectivity for even one second when logging at a rate of 20 MB/s, they'd start to drop logs? Why are we choosing such a low limit?

confgenerator/confgenerator.go
```diff
@@ -383,7 +383,7 @@ func (l *Logging) generateFluentbitComponents(userAgent string, hostInfo *host.I
 		BufferInMemory: true,
 	}.Components("ops-agent-fluent-bit")...)
 
-	out = append(out, stackdriverOutputComponent("ops-agent-fluent-bit", userAgent))
+	out = append(out, stackdriverOutputComponent("ops-agent-fluent-bit", userAgent, ""))
```
jefferbrecht, line 386: Why aren't we also setting a limit for self-logs?

rafaelwestphal: Self logs don't buffer on disk, so this configuration does not have effect.

confgenerator/confgenerator.go
rafaelwestphal: The docs goes a bit into more details, but this number is number of chunks, not effectively MB. Which is weird because chunks can have different sizes... but that's how fluent-bit implemented it.

jefferbrecht: The code seems to indicate that it's bytes, not chunks: https://github.com/fluent/fluent-bit/blob/caf63dd59dd3eecb66d96b6f969cb30143ad447e/include/fluent-bit/flb_output.h#L370, https://github.com/fluent/fluent-bit/blob/caf63dd59dd3eecb66d96b6f969cb30143ad447e/tests/internal/input_chunk.c#L230

sophieyfang (commented):
According to the fluentbit doc, this total_limit_size set up is used when one log stream is used to be output into multiple destinations. If one destination is slower than the other, don't hold on the chunks for too long meaning if the limit is hit, drop the chunk. 

If the doc is speaking truth, this set up might not be for our case as all the streams are going to one destination which is Google cloud logging API. I can see it's a good way on throttling the stream on output plugin but still clients often don't wanna drop logs. This set up is dumping chunks which is equal to dropping logs. We use buffer to disk to prevent dropping logs which is in the opposite on what this set up is doing.

rafaelwestphal:
> 

This is very similar to the issue that Jeff brought up. I went through this a little bit on the design doc, but I'll add here for visibility. 

This setup is not necessarily dropping chunks. In a normal execution scenario (network connectivity is OK) nothing will be dropped and Ops-Agent will behave as it does nowadays.

In the case that there is back pressure, we will still buffer the logs in disk but we are *limiting the amount of disk* we are using. We have to keep in mind that, as it is right now, we are still going to drop those logs. Either because the machine will run out of disk space to buffer (due to not having this configuration) or, in the case of a network outage, Ops-Agent will retry 3 times and then completely drop the log. So the log loss is there either way, we are just gracefully preventing ops-agent from trashing the machine.

jefferbrecht:
> > 
> 
> This is very similar to the issue that Jeff brought up. I went through this a little bit on the design doc, but I'll add here for visibility.
> 
> This setup is not necessarily dropping chunks. In a normal execution scenario (network connectivity is OK) nothing will be dropped and Ops-Agent will behave as it does nowadays.
> 
> In the case that there is back pressure, we will still buffer the logs in disk but we are _limiting the amount of disk_ we are using. We have to keep in mind that, as it is right now, we are still going to drop those logs. Either because the machine will run out of disk space to buffer (due to not having this configuration) or, in the case of a network outage, Ops-Agent will retry 3 times and then completely drop the log. So the log loss is there either way, we are just gracefully preventing ops-agent from trashing the machine.

Thanks for the explanation and offline discussion; I'll copy here that I'm fine with the number for the limit as long as it's sufficient to avoid dropping logs before the 3-retries window expires. I'll add some comments to the design doc to learn more about the details.

integration_test/ops_agent_test.go
```diff
@@ -3213,6 +3214,88 @@ func TestNetworkHealthCheck(t *testing.T) {
...
+		_, err := gce.RunScriptRemotely(ctx, logger, vm, generateLogsScript, nil, nil)
+
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		scripts := fmt.Sprintf(`
+		cat  /run/google-cloud-ops-agent-fluent-bit/fluent_bit_main.conf 
+		`)
+		gce.RunScriptRemotely(ctx, logger, vm, scripts, nil, nil)
```
jefferbrecht, line 3274: I don't think this is needed

jefferbrecht (approved):
LGTM overall, we can fine tune the number for `total_limit_size` later if needed, doesn't need to be a blocker for this PR -- most real-world use cases will be far below this number anyway I imagine.

sophieyfang:
> This setup is not necessarily dropping chunks. In a normal execution scenario (network connectivity is OK) nothing will be dropped and Ops-Agent will behave as it does nowadays.

Another scenario is when client has big throughput and it's been through quite some backpressure then the buffer disk size is now over 20M. Suppose each log is around 1000 bytes then 20M is around 20,000 log entries buffered in disk. 20,000 backlogs seems not a big number?

> as it is right now, we are still going to drop those logs. Either because the machine will run out of disk space to buffer (due to not having this configuration) 
20M is really not a big enough number. If the partial goal is to prevent disk growing huge, i will suggest a bigger number than 20M.

> in the case of a network outage, Ops-Agent will retry 3 times and then completely drop the log. 
I think this statement only holds true for UP state chunk, only chunk in up state is being attempted to be sent off with three tries. For chunk that is in down state and stored in the disk but not load into memory just yet, this statement does not apply.

"storage.total_limit_size which limits the number of Chunks that exist in the filesystem"  this will wipe out the content in disk once it hits the limit.

hsmatulisgoogle:
> Another scenario is when client has big throughput and it's been through quite some backpressure then the buffer disk size is now over 20M. Suppose each log is around 1000 bytes then 20M is around 20,000 log entries buffered in disk. 20,000 backlogs seems not a big number?

If a `storage.total_limit_size` of 20M is too small, a longer benchmarks over say a day should be able to catch that. However I agree that 20M seems a bit small for a filesystem size limit. For context, for self log rotation the plan is to add a default limit of 400M. If we are not sure, it might be better to leave it as a larger value and reduce it later

rafaelwestphal:
I'm ok with fine tunning `storage.total_limit_size` number. It is still unclear for me after some experiments, and even after looking at the documentation, if the unit for this is in bytes of Number of chunks. It might be safer to assume it's in bytes and then I agree that the number is low.

franciscovalentecastro (approved):

PR #1101, opened 2023-02-13
Revert "Introduce log rotation by default for self log files" for further investigation

hsmatulisgoogle:
After some internal discussion we decided GoogleCloudPlatform/ops-agent#1031 could have some consequences we need to evaluate. For now the PR will be reverted.

## Windows
Currently the PR results in a process other than the one launched by the SCM to handle signals, resulting in the following warning:

```
A service process other than the one launched by the Service Control Manager connected when starting the Google Cloud Ops Agent - Logging Agent service.  The Service Control Manager launched process 2164 and process 8144 connected instead.

  Note that if this service is configured to start under a debugger, this behavior is expected.
```

## Linux

It is possible that during a shutdown event, the agent wrapper is shut down by systemd before fluent bit completes its shutdown process at which point the agent wrapper will send a SIGKILL to fluent bit, ending its shut down.


qingling128 (approved):
LGTM

PR #1102, opened 2023-02-14
Introduce log rotation by default for self log files

hsmatulisgoogle:
Reverts GoogleCloudPlatform/ops-agent#1101

See [b/203811937](http://b/203811937)


hsmatulisgoogle:
For fixing the windows side of things, there are multiple approaches:

1. Connect to Windows Service Control manager through the agent_wrapper and prevent fluent bit from doing so by
  a.  Making subcommands console processes by attaching a Pseudo Console to them
  b. Change fluent bit so we can stop it from connecting to the SCM 
2. Use a fork to switch PIDs with the child process (through cygwin)

jefferbrecht:
> For fixing the windows side of things, there are multiple approaches:
> 
> 1. Connect to Windows Service Control manager through the agent_wrapper and prevent fluent bit from doing so by
>    a.  Making subcommands console processes by attaching a Pseudo Console to them
>    b. Change fluent bit so we can stop it from connecting to the SCM
> 2. Use a fork to switch PIDs with the child process (through cygwin)

Can you elaborate on (1a)? Are you referring to [this](https://devblogs.microsoft.com/commandline/windows-command-line-introducing-the-windows-pseudo-console-conpty/)? If so, it seems overly complicated, not to mention it only seems to be available on Windows 10 and later.

(2) seems similarly complicated. I'm not keen on introducing a dependency on cygwin just for this.

Why can't we just spawn fluent-bit as an ordinary process from the agent wrapper? fluent-bit won't think it's a service when run that way, so it'll implicitly not even attempt to talk to the SCM.

quentinmit:
> Why can't we just spawn fluent-bit as an ordinary process from the agent wrapper? fluent-bit won't think it's a service when run that way, so it'll implicitly not even attempt to talk to the SCM.

I believe fluent-bit will still think it's a service when run that way. Also, without talking to the SCM I think we don't know how to signal fluent-bit to cleanly shutdown.

quentinmit:
There's nothing changed here from the original, right? Can we mark this as a draft PR until it's ready to review?

jefferbrecht:
> > Why can't we just spawn fluent-bit as an ordinary process from the agent wrapper? fluent-bit won't think it's a service when run that way, so it'll implicitly not even attempt to talk to the SCM.
> 
> I believe fluent-bit will still think it's a service when run that way. Also, without talking to the SCM I think we don't know how to signal fluent-bit to cleanly shutdown.

I have no reason to believe it'd think it's a service when run that way, and some testing I've done locally confirms that. Only processes started directly by the service manager will succeed in calling `StartServiceCtrlDispatcher`, which is the mechanism fluent-bit uses to detect that scenario.

"Signals" on Windows is chaotic-evil territory, but we might be able to simply send CTRL+C to the "console" window of the spawned process: https://stackoverflow.com/a/15281070/1188632

hsmatulisgoogle:
> There's nothing changed here from the original, right? Can we mark this as a draft PR until it's ready to review?

Marked as draft. However, there are differences. The issue of signal & exit status propagation on the linux side should be fixed & tested. However, what to do about the windows issue is unclear

hsmatulisgoogle:
> Why can't we just spawn fluent-bit as an ordinary process from the agent wrapper? fluent-bit won't think it's a service when run that way, so it'll implicitly not even attempt to talk to the SCM.

Currently I believe we are spawning fluent-bit as a regular process and it is still talking to the SCM. Unless I misunderstood what you meant by regular process as we are spawning fluent bit here through `cmd.Run()`

https://github.com/GoogleCloudPlatform/ops-agent/blob/622e96975a5a561bbb1eb30bad7bb8c9f91c3bcb/cmd/agent_wrapper/main_windows.go#L70

Currently on windows, fluent bit seems to always attempt to contact SCM but gives up and runs normally if it gets an error code for `ERROR_FAILED_SERVICE_CONTROLLER_CONNECT`

[int win32_main(int argc, char **argv)](https://github.com/fluent/fluent-bit/blob/f58e4ceff0344801c389cfd7ac14cb5c4a2b9905/src/win32/winsvc.c#L130-L148)

```
{
    win32_argc = argc;
    win32_argv = argv;

    if (StartServiceCtrlDispatcher(svc_table)) {
        return 0;
    }

    if (GetLastError() != ERROR_FAILED_SERVICE_CONTROLLER_CONNECT) {
        return -1;
    }

    /*
     * If we cannot connect to SCM, we assume that "fluent-bit.exe"
     * was invoked from the command line.
     */
    return flb_main(argc, argv);
}
```

[`ERROR_FAILED_SERVICE_CONTROLLER_CONNECT`](https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-startservicectrldispatchera#return-value) is only returned when `the program is being run as a console application rather than as a service`. 

Documentation on what this means is unclear, but I suggested the Pseudo Console as it seems to be related to running console applications from a service applications, which should sever fluent bit's connection to the SCM. This would refer to 1a).

Alternatively we could submit a PR to fluent bit to change how it decides to connect to the SCM, to allow embedding it in other services. This would refer to 1b)

jefferbrecht:
> Currently I believe we are spawning fluent-bit as a regular process and it is still talking to the SCM. 
> ...
> Currently on windows, fluent bit seems to always attempt to contact SCM but gives up and runs normally if it gets an error code for `ERROR_FAILED_SERVICE_CONTROLLER_CONNECT`

This is only the case currently because the agent wrapper _doesn't_ register itself as the service process. The proposal was to turn the agent wrapper into the service process; I'm saying that if we do that, then the spawned fluent-bit will no longer act as a service process.

hsmatulisgoogle:
> This isn't true, because currently fluent-bit _is_ properly responding to the service signals (the integration test logs from the PR prove this), which it can only do if `StartServiceCtrlDispatcher` succeeds. If it was giving up then we'd be seeing tons of service errors in the logs instead.

Sorry, that was badly worded. I updated the comment to see if I can explain what I was trying to convey better. I was trying to describe how fluent bit behaves depending on the different error codes it receives from `StartServiceCtrlDispatcher`. Specifically I was trying to point out how it will abort when receiving a `ERROR_SERVICE_ALREADY_RUNNING` error which I think is what will happen if our process talks to the SCM without changing fluent bit. 

This should be easy to test though, I can set up a PR that tries to talk to the SCM and see if fluent bit breaks or not

jefferbrecht:
> Specifically I was trying to point out how it will abort when receiving a `ERROR_SERVICE_ALREADY_RUNNING` error which I think is what will happen if our process talks to the SCM without changing fluent bit.
> 
> This should be easy to test though, I can set up a PR that tries to talk to the SCM and see if fluent bit breaks or not

I've tried this, and it doesn't receive `ERROR_SERVICE_ALREADY_RUNNING`, because that's only returned if the calling process has already successfully called the dispatcher, which it hasn't. Instead, it receives `ERROR_FAILED_SERVICE_CONTROLLER_CONNECT`. But in case my setup was wrong, I'm interested to know if you can replicate that result on your end.

jefferbrecht:
I've been toying with using `AttachConsole`/`SetConsoleCtrlHandler(0, true)`/`GenerateConsoleCtrlEvent(0, 0)` to cleanly terminate the spawned fluent-bit, but I can only make it work when the spawned fluent-bit is started from a console, and not from a service process. Not sure why yet.

My next theory is that we can spawn the fluent-bit process with `CREATE_NEW_PROCESS_GROUP`, which give it its own console, and then call `GenerateConsoleCtrlEvent(0, pid)` instead, to terminate it cleanly. Will test tomorrow, my laptop is about to run out of battery and I didn't bring the charger -_-

jefferbrecht:
Update: `GenerateConsoleCtrlEvent(1, pid)` won't work either when we're running from a service (it always fails with `INVALID_HANDLE`).

I'm running out of ideas for simple ways to accomplish this. Maybe we can use `CreateRemoteThread` to inject some code into the running fluent-bit process which calls `raise(SIGINT)` 🤷 We can discuss more during the sync.

cmd/agent_wrapper/main_linux.go
```diff
@@ -0,0 +1,57 @@
...
+	"os/exec"
+	"os/signal"
+	"syscall"
+	"time"
+)
+
+func handleSignals(cmd *exec.Cmd) {
+	sigs := make(chan os.Signal, 1)
+	for {
+		signal.Notify(sigs, syscall.SIGTERM, syscall.SIGTERM)
```
jefferbrecht, line 31: Was the second `SIGTERM` meant to be something different?

cmd/agent_wrapper/main_linux.go
```diff
@@ -0,0 +1,57 @@
...
+		sig := <-sigs
+		start := time.Now()
+		for {
+			if cmd.Process != nil {
+				cmd.Process.Signal(sig)
+				break
+
```
jefferbrecht, line 38: Nit:

```suggestion
```

cmd/agent_wrapper/main_linux.go
```diff
@@ -0,0 +1,56 @@
...
+			} else if 80*time.Second+time.Until(start) <= 0 {
+				break
+			}
+			time.Sleep(50 * time.Millisecond)
+		}
+		if sig == syscall.SIGTERM {
+			time.Sleep(80*time.Second + time.Until(start))
+			log.Fatalf("Wrapped %v did not terminate in 80 seconds\n", cmd.Path)
+		}
+	}
```
jefferbrecht, line 47: I think I'm confused about a few things:

* What's the purpose of the outer `for` loop? `sig := <-sigs` will block until a signal is received, and all of the signals we subscribe to are terminating.
* Why is `signal.Notify` inside the `for` loop? I can't think of a reason why we'd need to subscribe the same signals to the same channel more than once.
* The `did not terminate` error doesn't make sense to me -- it seems we're actually trying to check whether the process was _started_, not terminated.
  * In that case, why is this a fatal error? We can still handle this situation gracefully in the wrapper.
  * Why are we checking for `SIGTERM` on line 43? The condition for "process was/wasn't started" should be the result of the `cmd.Process != nil` evaluation on line 35 and should have nothing to do with whether the signal was `SIGTERM` or `SIGINT`.
* What's the reason for the `time.Sleep` before `log.Fatalf`?

hsmatulisgoogle: 1. I was missing the SIGCONT signal that fluent-bit uses to dump information, updated it
2. Good catch, thanks!
3. After thinking about it, I think that the `did not terminate` block is not required.  It was meant to make sure the sub-process doesn't take too long to stop after receiving a SIGTERM.
  3.2 I added some comments to explain why we check for `cmd.Process != nil`. It is unrelated to what I was trying to do with the `process did not terminate` block
5. It was meant to give the subprocess time to stop before being killed, as it needs time to do cleanup

cmd/agent_wrapper/main_linux.go
```diff
@@ -0,0 +1,60 @@
...
+	"os/signal"
+	"syscall"
+	"time"
+)
+
+const MaximumWaitForProcessStart = 5 * time.Second
+
+func handleSignals(cmd *exec.Cmd) {
+	/* Relay signals that should be passed down to the subprocess we are wrapping.
+	 */
```
jefferbrecht, line 32:
```suggestion
	// Relay signals that should be passed down to the subprocess we are wrapping.
```

jefferbrecht (approved):

cmd/agent_wrapper/main_linux.go
jefferbrecht: Makes sense, thanks for the clarification!

PR #1139, opened 2023-03-09
healthchecks: Handle multiple Health Check failures

franciscovalentecastro:
## Description
- Handle multiple Health Check failures using `errros.Join()` and `Unwrap() []errors` from `Go 1.20`.
- Implemented `String()` for `HealthCheckResult` to generate result message on demand.
- Improve message when API responds with `UNAUTHENTICATED`.
- Report `Context Deadline Exceeded` from API as connection failure.
- Save resulting Health Checks log as `health-checks.txt` in Windows.

## Related issue
b/272305209 , b/272501350

## Checklist:
- Unit tests
  - [ ] Unit tests do not apply.
  - [x] Unit tests have been added/modified and passed for this PR.
- Integration tests
  - [ ] Integration tests do not apply.
  - [x] Integration tests have been added/modified and passed for this PR.
- Documentation
  - [ ] This PR introduces no user visible changes.
  - [ ] This PR introduces user visible changes and the corresponding documentation change has been made.
- Minor version bump
  - [ ] This PR introduces no new features.
  - [ ] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.
  - [ ] This PR bumps the version.

PR #1148, opened 2023-03-14
Add additional network healthchecks

avilevy18:
## Related issue
b/228050119

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


internal/healthchecks/error.go
```diff
@@ -69,6 +90,33 @@ var (
 		ResourceLink: "https://cloud.google.com/monitoring/agent/ops-agent/troubleshooting",
 		IsFatal:      true,
 	}
+	PacApiConnErr = HealthCheckError{
+		Code:    "PacApiConnErr",
+		Class:   Connection,
+		Message: "Request to Packages API failed.",
+		Action:  "Check your internet connection and firewall rules.",
+		// TODO: Don't know what to put for the resource link
```
rafaelwestphal, line 98: Maybe 
https://cloud.google.com/logging/docs/agent/ops-agent/troubleshoot-find-info
is a better landing page for these connectivity troubleshooting errors.

cmd/google_cloud_ops_agent_engine/main.go
```diff
@@ -25,24 +25,22 @@ import (
...
-	outDir   = flag.String("out", os.Getenv("RUNTIME_DIRECTORY"), "directory to write configuration files to")
-	input    = flag.String("in", "/etc/google-cloud-ops-agent/config.yaml", "path to the user specified agent config")
-	logsDir  = flag.String("logs", "/var/log/google-cloud-ops-agent", "path to store agent logs")
-	stateDir = flag.String("state", "/var/lib/google-cloud-ops-agent", "path to store agent state like buffers")
+	service      = flag.String("service", "", "service to generate config for")
+	outDir       = flag.String("out", os.Getenv("RUNTIME_DIRECTORY"), "directory to write configuration files to")
+	input        = flag.String("in", "/etc/google-cloud-ops-agent/config.yaml", "path to the user specified agent config")
+	logsDir      = flag.String("logs", "/var/log/google-cloud-ops-agent", "path to store agent logs")
+	stateDir     = flag.String("state", "/var/lib/google-cloud-ops-agent", "path to store agent state like buffers")
+	healthChecks = flag.Bool("healthchecks", false, "run health checks")
```
rafaelwestphal, line 33: "only run health checks" or "run health checks and exit". This can be a bit ambiguous as to "We are enabling/disabling health checks when running the service", so a different wording might be better.

internal/healthchecks/network_check.go
```diff
@@ -20,9 +20,38 @@ import (
...
+		},
+		{
+			name:             "Monitoring API",
+			url:              "https://monitoring.googleapis.com/$discovery/rest",
+			successMessage:   "Request to the Monitoring API was successful.",
+			healthCheckError: MonApiConnErr,
+		},
+		{
+			name:             "Packages API",
+			url:              "https://packages.googleapi.com/$discovery/rest",
```
rafaelwestphal, line 38: Although I don't mind that much adding these urls to the health checks, they are more valuable at the diagnostics script and don't add much value after the agent is already installed. Since they refer to the endpoints that will enable the installation to succeed, without access to packages and/or dl.google you can't download the Ops-Agent and run the health checks.

rafaelwestphal:
Thanks! I'll give @franciscovalentecastro a chance to look at these changes, since he is working closely on this.

internal/healthchecks/network_check.go
franciscovalentecastro: This is a great catch that I haven't realized before. It doesn't hurt to check them here, but there is more value in the diagnostics script if the user can't even install the Ops Agent.

cmd/google_cloud_ops_agent_engine/main.go
```diff
@@ -52,7 +50,14 @@ func main() {
 		log.Fatalf("The agent config file is not valid. Detailed error: %s", err)
 	}
 }
+
 func run() error {
+	if *healthChecks {
+		runHealthChecks()
```
franciscovalentecastro, line 56: I have the concern that running with `--healthchecks` flag is going to run all of the Health Checks (Ports, Network, API), but the `Ports Check` is going to fail if the Ops Agent is running, causing confusion in the results. 

We can disable that `Ports Check` (not registering that check for this case) to avoid this.

cmd/google_cloud_ops_agent_engine/main.go
```diff
@@ -65,7 +70,10 @@ func run() error {
 	log.Printf("Built-in config:\n%s", apps.BuiltInConfStructs["linux"])
 	log.Printf("Merged config:\n%s", uc)
 
-	runStartupChecks(*service)
+	if *service == "" {
```
franciscovalentecastro, line 73: Is there a way to only have one entry point for the `runHealthChecks` that works for either the Ops Agent startup or just the `--healthchecks` flag.

internal/healthchecks/network_check.go
franciscovalentecastro: Checking the `metadata` server here still has a lot of value.

cmd/google_cloud_ops_agent_engine/main.go
jefferbrecht: What if we just make the ports check short-circuit as a `PASS` if the Ops Agent (and subagents) are already running?

franciscovalentecastro: Yeah, that's a good idea. I like it better than disabling the `Ports Check`. What is the best way to detect the Ops Agent is already running ?

internal/healthchecks/network_check.go
```diff
@@ -20,9 +20,38 @@ import (
 )
 
 var (
-	// API urls
-	loggingAPIUrl    = "https://logging.googleapis.com/$discovery/rest"
-	monitoringAPIUrl = "https://monitoring.googleapis.com/$discovery/rest"
+	requests = [...]networkCheckRequest{
```
jefferbrecht, line 23: Why not just use a slice here?

internal/healthchecks/error.go
```diff
@@ -36,6 +38,25 @@ func (e HealthCheckError) Error() string {
...
+type MutliHealthCheckError []HealthCheckError
+
+func (e MutliHealthCheckError) Error() {
+
+}
+
+func (e MutliHealthCheckError) FullMessage() {
+
+}
+
```
jefferbrecht, line 59: Is any of this used?

internal/healthchecks/error.go
```diff
@@ -69,6 +90,30 @@ var (
...
+		Class:        Connection,
+		Message:      "Request to dl.google.com failed",
+		Action:       "Check your internet connection and firewall rules.",
+		ResourceLink: "https://cloud.google.com/logging/docs/agent/ops-agent/troubleshoot-find-info",
+		IsFatal:      false,
+	}
+	MetaApiConnErr = HealthCheckError{
+		Code:         "MetaApiConnErr",
+		Class:        Connection,
+		Message:      "Request to GCE Metadata server",
```
jefferbrecht, line 112: Message looks incomplete

internal/healthchecks/network_check.go
jefferbrecht, line 23: Optional: since `requests` is only used once, you could define both the struct and its elements inline within `RunCheck` and avoid needing to define a named struct. (This is a common pattern used in unit tests too: https://gobyexample.com/testing)

cmd/google_cloud_ops_agent_engine/main.go
jefferbrecht: I guess there are at least two ways:

1. Check that the subagent services are running. If they're successfully running then necessarily they must be holding onto the ports, so it's a pass.
2. Check which processes (if any) own ports 20201 and 20202; if they match what we expect (e.g. `/opt/whatever/fluent-bit` and `/opt/whatever/otelopscol` for Linux, likewise for Windows) then the subagents are already running with those ports, so it's a pass.

I am probably missing some gotchas here so I'm curious to know what others think.

igorpeshansky: 2 might be somewhat problematic, since checking which process owns a port is often a privileged operation, and the health check invocation will likely be in user space… 1 could be sufficient.

franciscovalentecastro: Yes, I'm also curious about what others think. This came up at the last discussion relating the `Ports Check Flakiness`. 

I would go with `1.` (but still not sure how to do it). `2.` (as mentioned by @igorpeshansky) would be problematic since probably running a `netstat` command or something similar is going to be platform/implementation dependent.

cmd/google_cloud_ops_agent_engine/main.go
```diff
@@ -25,25 +25,32 @@ import (
...
+	stateDir     = flag.String("state", "/var/lib/google-cloud-ops-agent", "path to store agent state like buffers")
+	healthChecks = flag.Bool("healthchecks", false, "run health checks and exit")
 )
 
-func runStartupChecks(service string) {
-	if service == "" {
-		gceHealthChecks := healthchecks.HealthCheckRegistryFactory()
-		logger, closer := healthchecks.CreateHealthChecksLogger(*logsDir)
-		defer closer()
+func runHealthChecks(registry healthchecks.HealthCheckRegistry) {
```
franciscovalentecastro, line 36: Also, we should also add this in the Windows service too :
https://github.com/GoogleCloudPlatform/ops-agent/blob/master/cmd/ops_agent_windows/run_windows.go#L136-L151

If this feels to big of a change, maybe we can only add the new `Network Checks` here and defer the `--healthchecks` fucntion to a followup `PR`.

cmd/google_cloud_ops_agent_engine/main.go
jefferbrecht: On Linux: check the exit code of `systemctl is-active --quiet google-cloud-ops-agent-fluent-bit` (likewise for the metrics service)

On Windows: check the output of `sc query google-cloud-ops-agent-fluent-bit | findstr "STATE"` for `RUNNING` (likewise for the metrics service)

cmd/google_cloud_ops_agent_engine/main.go
franciscovalentecastro: I think this was addressed.

cmd/google_cloud_ops_agent_engine/main.go
```diff
@@ -65,7 +63,13 @@ func run() error {
 	log.Printf("Built-in config:\n%s", apps.BuiltInConfStructs["linux"])
 	log.Printf("Merged config:\n%s", uc)
 
-	runStartupChecks(*service)
+	if *service == "" {
+		runHealthChecks(healthchecks.HealthCheckRegistryFactory())
+		log.Println("Startup checks finished")
```
franciscovalentecastro, line 68: Some suggestions for this : 
- Moving the execution of the `healthchecks` to the beginning of `run()` so it's the first thing. The reasoning for this is if `--healthchecks` is set, then it will only run the `healthchecks` and no other part of the confgenerator.
- I think there's no need anymore to pass as a paramter `healthchecks.HealthCheckRegistryFactory()`.
- Maybe add a comment to the `return nil` explaining why it's set this way.

internal/healthchecks/linux.go
```diff
@@ -20,9 +20,23 @@ package healthchecks
...
 	"errors"
 	"net"
+	"os/exec"
 	"syscall"
 )
 
+func isSubagentActive(subagent string) (bool, error) {
+	_, err := exec.Command("systemctl", "is-active", "--quiet", subagent).Output()
+	if err != nil {
+		// if the service doesn't exist return false with no error
```
franciscovalentecastro, line 30: What happens in the case the `service exists`, but it is not running.

cmd/ops_agent_windows/main_windows.go
```diff
@@ -57,6 +58,16 @@ func main() {
 				log.Fatal(err)
 			}
 			infoLog.Printf("uninstalled services")
+		} else if *healthChecks {
+			healthCheckResults := getHealthCheckResults()
+			for _, result := range healthCheckResults {
+				if result.Err != nil {
+					log.Printf("Error: %s", result.Message)
+				} else {
+					log.Println(result.Message)
```
franciscovalentecastro, line 67: Should we use `infoLog` here ?

internal/healthchecks/ports_check.go
```diff
@@ -45,6 +45,18 @@ func checkIfPortAvailable(host string, port string, network string) (bool, error
 }
 
 func (c PortsCheck) RunCheck(logger *log.Logger) error {
+	fbActive, err := isSubagentActive("google-cloud-ops-agent-fluent-bit")
+	if err != nil {
+		return err
+	}
+	ocActive, err := isSubagentActive("google-cloud-ops-agent-opentelemetry-collector")
```
franciscovalentecastro, line 52: I think the name of the `opentelemetry-collector` subagent in windows is different to the one in linux. See https://github.com/GoogleCloudPlatform/ops-agent/blob/master/integration_test/ops_agent_test.go#L144-L150

jefferbrecht: That's the process name -- we're checking the service name here, which looks correct: https://github.com/GoogleCloudPlatform/ops-agent/blob/master/cmd/ops_agent_windows/main_windows.go#L120

internal/healthchecks/linux.go
avilevy18: The command exits with status code 3 so the code will run as if the service doesn't exists.

internal/healthchecks/ports_check.go
franciscovalentecastro: Ok! Thanks for the clarification.

franciscovalentecastro (approved):
LGTM! Approved modulo addressing the last comments regarding refactors.

internal/healthchecks/ports_check.go
```diff
@@ -45,6 +45,18 @@ func checkIfPortAvailable(host string, port string, network string) (bool, error
 }
 
 func (c PortsCheck) RunCheck(logger *log.Logger) error {
+	fbActive, err := isSubagentActive("google-cloud-ops-agent-fluent-bit")
```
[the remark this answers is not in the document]
jefferbrecht: I don't think we should merge them, because there exist intermediate states (especially on Windows) where only one subagent is running. Each subagent uses a different port so the checks divide pretty cleanly anyway.

cmd/ops_agent_windows/main_windows.go
```diff
@@ -57,6 +58,16 @@ func main() {
 				log.Fatal(err)
 			}
 			infoLog.Printf("uninstalled services")
+		} else if *healthChecks {
+			healthCheckResults := getHealthCheckResults()
+			for _, result := range healthCheckResults {
+				if result.Err != nil {
+					infoLog.Printf("Error: %s", result.Message)
```
jefferbrecht, line 65: Does `result.Err` contain anything relevant that we might want to include in the printed text?

internal/healthchecks/network_check.go
```diff
@@ -32,36 +66,21 @@ func (c NetworkCheck) Name() string {
 }
 
 func (c NetworkCheck) RunCheck(logger *log.Logger) error {
-	// Request to logging API
-	response, err := http.Get(loggingAPIUrl)
-	if err != nil {
-		if isTimeoutError(err) || isConnectionRefusedError(err) {
-			return LogApiConnErr
+	for _, r := range requests {
```
jefferbrecht, line 69: Currently `RunCheck` will fail fast on the first encountered failure. Would it make more sense to run all of the checks independently and aggregate the result with `multierr` instead of failing fast?

internal/healthchecks/linux.go
jefferbrecht, line 30: Exit code 3 means "unit is not active", not "service doesn't exist".

internal/healthchecks/network_check.go
franciscovalentecastro: Yes, it should report an aggregate result for the failures, not just fail fast. We (@avilevy18, @braydonk) discussed this at beginning of this task. 

I will address this in the followup PR https://github.com/GoogleCloudPlatform/ops-agent/pull/1139 , since the `API Check` (and now it seems also the `Ports Check`) will benefit from reporting all of the fails/errors. Initially (the current state of #1139) I was only going to separate in different checks (e.g. Log API and Mon API checks), but after discussing the solution was to create a `type MultiHealthCheckError []HealthCheckError` and report each of the relevant failures.

internal/healthchecks/network_check.go
```diff
@@ -20,9 +20,43 @@ import (
...
+			successMessage:   "Request to dl.google.com was successful.",
+			healthCheckError: DLApiConnErr,
+		},
+		{
+			name:             "GCE Metadata Server",
+			url:              "http://metadata.google.internal",
+			successMessage:   "Request to the GCE Metadata server was successful.",
+			healthCheckError: MetaApiConnErr,
+		},
+	}
```
jefferbrecht, line 59: Would it make sense (maybe as a later refactor after this PR) to merge this struct with the one from `error.go`? Everything looks like it's one-to-one so it might make more sense to just have everything related to a check be all in one place...

avilevy18: Lets refactor in another another PR.

cmd/ops_agent_windows/main_windows.go
jefferbrecht: Discussed offline; `result.Message` already embeds the contents of `result.Err`.

I would suggest a small refactor of `HealthCheckResult`:
* Get rid of the `Message` field and change it into a `String()` function that generates the formatted message on-demand. This makes it so that we don't have to manually keep `Err` and `Message` in sync, with the added benefit of implementing `Stringer`.
* Add a `Name` field to support the above function since it's needed for generating the message.

WDYT?

franciscovalentecastro: I really like this idea of generating the formatted message on demand. I was trying to figure out where to put this and `result` seems to be great (it has all the info, with the `Err` and name it should be enough).

internal/healthchecks/error.go
```diff
@@ -69,6 +69,30 @@ var (
...
+		IsFatal:      false,
+	}
+	MetaApiConnErr = HealthCheckError{
+		Code:         "MetaApiConnErr",
+		Class:        Connection,
+		Message:      "Request to GCE Metadata server failed",
+		Action:       "Check your internet connection and firewall rules.",
+		ResourceLink: "https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/troubleshoot-run-ingest#network-issues",
+		IsFatal:      true,
+	}
```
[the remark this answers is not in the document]
avilevy18: Spoke offline, more robust tests will be available in https://github.com/GoogleCloudPlatform/ops-agent/pull/1139

cmd/ops_agent_windows/main_windows.go
avilevy18: Spoke offline, will be part of https://github.com/GoogleCloudPlatform/ops-agent/pull/1139

cmd/ops_agent_windows/main_windows.go
```diff
@@ -57,10 +59,14 @@ func main() {
...
 			infoLog.Printf("uninstalled services")
+		} else if *healthChecks {
+			healthCheckResults := getHealthCheckResults()
+			healthchecks.LogHealthCheckResults(healthCheckResults, infoLog)
+			infoLog.Println("Health checks finished")
 		} else {
 			// TODO: add an interactive GUI box with the Install, Uninstall, and Cancel buttons.
 			fmt.Println("Invoked as a standalone program with no flags. Nothing to do.")
-			fmt.Println("Use either --install or --uninstall to take action.")
+			fmt.Println("Use either --healtchecks, --install, --uninstall to take action.")
```
jefferbrecht, line 69:
```suggestion
			fmt.Println("Use either --healthchecks, --install, --uninstall to take action.")
```

internal/healthchecks/ports_check.go
```diff
@@ -45,38 +50,65 @@ func checkIfPortAvailable(host string, port string, network string) (bool, error
...
+func runOtelCollectorCheck(logger *log.Logger) error {
+	ocActive, err := isSubagentActive("google-cloud-ops-agent-opentelemetry-collector")
+	if err != nil {
+		return err
+	}
+	if ocActive {
+		return nil
+	}
+
+	//Opentelemetry-collector listens in both tcp4 and tcp6. Check for opentelemetry-collector self metrics port.
```
jefferbrecht, line 90: Nit:

```suggestion
	// Opentelemetry-collector listens in both tcp4 and tcp6. Check for opentelemetry-collector self metrics port.
```

jefferbrecht (approved):
LGTM, just a couple typos remaining.

internal/healthchecks/healthchecks.go
```diff
@@ -81,3 +81,13 @@ func (r HealthCheckRegistry) RunAllHealthChecks(logger *log.Logger) map[string]H
 
 	return result
 }
+
+func LogHealthCheckResults(healthCheckResults map[string]HealthCheckResult, logger *log.Logger) {
+	for _, result := range healthCheckResults {
+		if result.Err != nil {
+			logger.Printf("Error: %s", result.Message)
```
franciscovalentecastro, line 88: There is no need to add `Error: ` as prefix here, since the message already will be formatted as, e.g.`API Check - Result: ERROR, Detail: ****`

franciscovalentecastro (approved):

jefferbrecht:
If the helper refactor is causing issues then feel free to revert it. What I had in mind was something like this:

```go
func LogHealthCheckResults(healthCheckResults map[string]HealthCheckResult, infoLogger func(string), errorLogger func(string)) {
	for _, result := range healthCheckResults {
		if result.Err != nil {
			errorLogger(result.Message)
		} else {
			infoLogger(result.Message)
		}
	}
}
```

Which can be invoked e.g. on Linux as:

```go
LogHealthCheckResults(healthCheckResults, func(s string) { logger.Println(s) }, func(s string) { logger.Println(s) })
```

Or on Windows as:

```go
LogHealthCheckResults(healthCheckResults, func(s string) { s.Log.Info(EngineEventID, s) }, func(s string) { s.Log.Error(EngineEventID, s) })
```

Or if you need stdout:

```go
LogHealthCheckResults(healthCheckResults, func(s string) { log.Println(s) }, func(s string) { log.Println(s) })
```

PR #1161, opened 2023-03-16
Move `test_config.yaml` settings into per-app `metadata.yaml` files

martijnvans:
## Description

This change splits up `test_config.yaml`, putting its contents into each app's `metadata.yaml` file. This is nice from a tidiness perspective, because there are fewer files to worry about, but it's also nice because now when you adjust the test skips for an app, it **doesn't run all tests for all apps**, which is what it had to do when there was a central `test_config.yaml` file. This is extra nice because running all tests for all apps is "heavy" in that it takes longer and is more likely to overload the apache servers, run us out of various quotas, etc.

## Related issue
offshoot of b/270147175.

## How has this been tested?
In addition to automated tests, I ran the tests before and after this change and verified that the PASS/SKIP log messages were identical before and after (after sorting). They were, after fixing some bugs.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


sophieyfang (approved):

martijnvans:
some things are still running, but I had a clean run previously so I'm going to call this good.

PR #1164, opened 2023-03-17
Make Windows testing hermetic

quentinmit:
## Description
Centralize platform detection such that tests can be hermetic and not use any real platform information

## Related issue
b/271422538

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


quentinmit:
I suspect the remaining integration test failures are flakes, but I'm still trying to prove that. PTAL.

cmd/ops_agent_windows/run_windows.go
```diff
@@ -88,6 +91,8 @@ func (s *service) Execute(args []string, r <-chan svc.ChangeRequest, changes cha
 			default:
 				s.log.Error(EngineEventID, fmt.Sprintf("unexpected control request #%d", c))
 			}
+		case <-ctx.Done():
+			return
```
jefferbrecht, line 95: Do we need to call `cancel()` here too? If not, do we need `WithCancel` to begin with?

quentinmit: `cancel()` will be called by the defer on line 56 after the return executes.

jefferbrecht: Oops, I left my comment on the wrong line; I meant to refer to line 89,

```
			case svc.Stop, svc.Shutdown:
				return
```

In the [diagnostics service](https://github.com/GoogleCloudPlatform/ops-agent/blob/quentin-config-hermetic/cmd/google_cloud_ops_agent_diagnostics/main_windows.go#L101-L103), we call `cancel()` on `svc.Stop`/`svc.Shutdown()`. I guess I'm asking why we're not doing that here too? And if it turns out that the services are just different and we don't need to call it here for whatever reason, then can't we just get rid of the whole `WithCancel()` and `defer cancel()` from here -- what is it that we're expecting to still be running with that context which needs cancellation when `Execute()` returns?

Likewise, does the diagnostics service also need a `case <-ctx.Done():` like we've done here? I'm generally just wondering why the two service goroutines are different w.r.t. `ctx`.

jefferbrecht (approved):
My thread isn't a blocker so LGTM 👍 Thanks for working on this!

braydonk (approved):

cmd/ops_agent_windows/run_windows.go
quentinmit: In the diagnostics service, the service event loop is running in a separate goroutine, which means that `return` wouldn't actually trigger the `defer cancel()`. I guess technically both should have a `case <-ctx.Done()` to break out of the loop, but I was thinking that it doesn't matter if the extra goroutine has an unclean death. These could probably benefit from some refactoring to share common code, but I didn't want to bite that off in this PR.

internal/platform/platform.go
```diff
@@ -0,0 +1,90 @@
...
+
+type Type int
+
+const (
+	Linux Type = 1 << iota
+	Windows
+	All = Linux | Windows
+)
+
+func (p Platform) Is2012() bool {
```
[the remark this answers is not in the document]
quentinmit: Not a bad idea, but I'll leave that for the next person doing cleanup 🙃

PR #1166, opened 2023-03-20
Fix and enable apache test on SLES and CentOS 7

martijnvans:
## Description

This PR adds two new features in service of removing the last few `platforms_to_skip` for apache:

1. Certain metrics can be marked as unavailable on certain platforms in `metadata.yaml`.
2. Same for certain log fields.

This is better than just marking the logs/metrics as optional, because they can be important on some platforms (so we do want coverage).

## Related issue
https://github.com/GoogleCloudPlatform/ops-agent/issues/1173

## How has this been tested?
automated tests only.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/metadata/integration_metadata.go
```diff
@@ -49,6 +49,9 @@ type ExpectedMetric struct {
 	Representative bool `yaml:"representative,omitempty" validate:"excluded_with=Optional,excluded_with=Platform"`
 	// Exclusive metric to a particular kind of platform.
 	Platform string `yaml:"platform,omitempty" validate:"excluded_with=Representative,omitempty,oneof=linux windows"`
+	// A list of platforms that this metric is not available on.
+	// Examples: centos-7,debian-10. Not valid are linux,windows.
+	UnavailableOn []string `yaml:"unavailable_on,omitempty" validate:"excluded_with=Representative"`
```
igorpeshansky, line 54: I wonder if this information also needs to be propagated into the generated documentation… @sophieyfang WDYT?

martijnvans: sophie and i discussed this a bit, and we agreed that since we don't surface "optional" for log fields or metrics in the docs, we don't need to surface "unavailable_on" either.

If we decided to surface that information later, we could do it by saying something like "this metric/log field may be unavailable on some platforms" if optional or unavailable_on are specified.

regarding sophie's suggestion to use "optional' instead of 'unavailable_on" for log fields, we haven't decided yet. I'm OK with that approach but I think unavailable_on is going to produce slightly better test coverage. The difference is probably minimal though.

integration_test/third_party_apps_test.go
```diff
@@ -407,16 +407,40 @@ func verifyLog(actualLog *cloudlogging.Entry, expectedLog *metadata.ExpectedLog)
...
+	var result []*metadata.LogFields
+	for _, field := range fields {
+		if !metadata.SliceContains(field.UnavailableOn, platform) {
+			result = append(result, field)
+		}
+	}
+	return result
+}
+
 func runLoggingTestCases(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM, logs []*metadata.ExpectedLog) error {
```
rafaelwestphal, line 422: Since this is related, can we remove the hack related to b/239240173 on line 602 (bug is marked as fixed)?
If we can not, change it to this new format.

martijnvans: I updated the bug number. I think the bug is still not fixed. I'm not sure what you mean about changing it to the new format; the feature in this PR cannot be used for aerospike + SLES because this PR only includes a feature for missing log FIELDS, not logs that are entirely missing. It wouldn't be hard to add that feature though, do you want me to do that in this PR or a separate one?

rafaelwestphal: ah ok, got it. Thanks for updating it.

rafaelwestphal (approved):

PR #1187, opened 2023-04-05
integration_test: add feature tracking test for prom receiver

ridwanmsharif:
## Description
Add integration test for feature tracking metrics that come from the prometheus receiver.

## Related issue
[b/260123506](http://b/260123506) | P2 | Add integration tests for Prometheus feature tracking metrics

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -28,13 +28,16 @@ The following variables are optional:
 
 REPO_SUFFIX: If provided, what package repo suffix to install the ops agent from.
 ARTIFACT_REGISTRY_REGION: If provided, signals to the install scripts that the
-    above REPO_SUFFIX is an artifact registry repo and specifies what region it
+
+	above REPO_SUFFIX is an artifact registry repo and specifies what region it
 	is in. If not provided, that means that the packages are accessed through
 	packages.cloud.google.com instead, which may point to Cloud Rapture or
 	Artifact Registry under the hood.
+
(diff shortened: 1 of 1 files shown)


integration_test/ops_agent_test.go
```diff
@@ -2300,6 +2303,51 @@ func TestPrometheusMetrics(t *testing.T) {
 		if multiErr != nil {
 			t.Error(multiErr)
 		}
+
+		expectedFeatures := []*feature_tracking_metadata.FeatureTracking{
+			{
+				Module:  "logging",
```
LujieDuan, line 2309: Question: for prom we extract other features as well: https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/testdata/valid/linux/metrics-receiver_prometheus_relabel/golden/features.yaml
Do we normally include all features in the integration tests, or just check the representative ones is good?

ridwanmsharif: I figured we would just test what matter to us here since the rest are covered by the unit tests but alas I was wrong. The tests fail because even though all the metrics come up correctly (http://shortn/_9Yh3qHC1mT) but our tests require that all metrics are listed here so it knows how many times to call `gce.WaitForMetric` since all metrics write too the same metric type.

I fixed this now

LujieDuan (approved):
LGTM!

PR #1196, opened 2023-04-13
Add support for replication metrics in mysql

dehaansa:
## Description
Add support for replication metrics in MySQL, and ensure they are supported in 5.7.

## Related issue

[b/267141861](http://b/267141861)

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


dehaansa:
There appears to be a Cassandra integration test issue at the moment affecting debian third party apps tests, but the MySQL tests are all passing and validating replication support for both 5.7 and 8.0+

apps/sqlqueryreceiver_common.go
igorpeshansky, line 1: [Optional] I wonder if it's time to add a "common" subdirectory under "apps"?

apps/mysql.go
```diff
@@ -34,6 +36,8 @@ type MetricsReceiverMySql struct {
 
 	Password string `yaml:"password" validate:"omitempty"`
 	Username string `yaml:"username" validate:"omitempty"`
+
+	ApplicationVersion string `yaml:"application_version" validate:"omitempty"`
```
igorpeshansky, line 40: [Optional] Have we explored ways of doing this fallback automatically, without user input? E.g., modifying the upstream mysql receiver to detect whether new-style metrics are unavailable and issue the old-style queries?

igorpeshansky (commented):
A couple of optional comments.

apps/mysql.go
dehaansa: We have explored it, but when I discussed it with some opentelemetry community members there was some resistance to adding the archaic language that MySQL chose to replace (`slave status`) for a MySQL version that's nearly EOL, so we chose to avoid the uphill battle and implement in a way independent of the MySQL receiver.

apps/sqlqueryreceiver_common.go
dehaansa: Certainly seems reasonable. It's a bit of a refactor to touch all of the files using common_logging_processors so I would propose this refactor be done in a separate PR.

integration_test/third_party_apps_data/applications/mysql/sles/install
```diff
@@ -20,6 +20,9 @@ else
     mysql_repo_pkg_name=mysql80-community-release-sl15-1.noarch.rpm
   fi
 
+
+
+
```
martijnvans, line 25: extra whitespace here

martijnvans (approved):
From the testing side, looks good. I didn't review the changes to the core functionality.

integration_test/third_party_apps_data/applications/mysql/sles/install
```diff
@@ -50,3 +50,5 @@ EOF
 fi
 
 sudo service mysql start
+
+# TODO - when reenabling sles support, replication will need to be configured here as in debian & centos
```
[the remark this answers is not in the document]
martijnvans: old comment, ignore

integration_test/third_party_apps_data/applications/mysql5.7/metadata.yaml
```diff
@@ -0,0 +1,410 @@
...
+      - name: jsonPayload.rowsSent
+        type: number
+        description: 'The number of rows sent to the client'
+      - name: jsonPayload.rowsExamined
+        type: number
+        description: 'The number of rows examined by the server layer'
+      - name: jsonPayload.errorNumber
+        type: number
+        description: 'The statement error number, or 0 if no error occurred'
+        optional: true # Not present in 5.7
```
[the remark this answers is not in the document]
dehaansa: Yeah, will remove them.

integration_test/third_party_apps_data/applications/mysql/centos_rhel/install
```diff
@@ -25,7 +25,35 @@ EOF
...
+sudo mkdir /var/log/mysql
+sudo chown mysql:mysql /var/log/mysql
+
+# Required to allow mysql to accept connections on 3307
+if [ ${ID} == rocky ]; then
+  sudo yum install -y policycoreutils-python-utils
+else
+  sudo yum install -y policycoreutils-python
+fi
+sudo semanage port -a -t mysqld_port_t -p tcp 3307
```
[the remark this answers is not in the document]
dehaansa: Makes sense, updated the semanage calls. I'll remember this guideline and apply it moving forward, maybe we should just have a refactor PR to apply it to all uses of yum and tee and similar commands?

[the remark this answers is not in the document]
martijnvans: well I was trying to say that I don't think it's even necessarily better to have yum --assumeyes, it's a lot of extra text and I think yum -y may be obvious enough to people that they don't need to look it up.
This is not an important guideline, it's just a personal preference and I don't think we need to commit to some kind of big cleanup. I also think that if you forget this guideline going forward, it's not a big deal and I won't be mad. ;)

apps/mysql.go
igorpeshansky: What if we just issued old-style SQL queries unconditionally from the Ops Agent-generated configs? Would that cause a lot of log spam on MySQL 8.x?

(I'm just trying to avoid adding new configuration options if we can help it…)

dehaansa: Due to the way it's implemented, it wouldn't log spam but it would query both ways on MySQL 8. We wouldn't have duplicate metrics because the filtering happens post-query in the mysql receiver so it wouldn't be problematic.

This new configuration option was the decision the ops team made, so I'll defer to @jefferbrecht and team, do we want to just _always_ configure a secondary receiver to get replication metrics?

jefferbrecht: I might've had an exaggerated sense of badness in my mind re: allowing both queries on version 8 -- if it turns out to not be a big deal then we can reconsider that design choice.

If I understand right, are you proposing that we just disable the replica metrics in the MySQL OTel receiver and use a second receiver (`sqlqueryreceiver`) with `slave status` instead, which supposedly works on both versions? That'd avoid duplication even before filtering, right?

dehaansa: That's correct, that should work. The query would still be executed on version 8, but the metrics would never be created from that query.

jefferbrecht: Sounds good -- We're fine on our end to give this a try, hopefully it simplifies things.

dehaansa: Re-implemented in the described way (no more configuration, always collect replication metrics with `slave status` regardless of version), awaiting tests in kokoro.

igorpeshansky (approved):
LGTM :shipit:

igorpeshansky (commented):
A couple of cleanup issues I missed earlier…

apps/mysql.go
```diff
@@ -55,70 +57,137 @@ func (r MetricsReceiverMySql) Pipelines() []otel.ReceiverPipeline {
...
-						"enabled": false,
-					},
-					"mysql.table.io.wait.count": map[string]interface{}{
-						"enabled": false,
-					},
-					"mysql.table.io.wait.time": map[string]interface{}{
-						"enabled": false,
-					},
+	// MySQL replication metrics are implemented separate to the main metrics pipeline so that 5.7 and 8.0 are both supported,
+	// this configuration is used for the sqlqueryreceiver
```
igorpeshansky, line 61: Nit: the term `sqlqueryreceiver` no longer appears anywhere, right? You could add a space to turn it into "sqlquery receiver" (appearing below), or avoid the need for this comment altogether by renaming this variable, e.g., to `sqlqueryReceiverDriverConf` or even `sqlReceiverDriverConfig` (to match `sqlReceiverQueriesConfig`).

dehaansa: Resolved.

apps/mysql.go
```diff
@@ -57,9 +57,7 @@ func (r MetricsReceiverMySql) Pipelines() []otel.ReceiverPipeline {
 		r.Username = "root"
 	}
 
-	// MySQL replication metrics are implemented separate to the main metrics pipeline so that 5.7 and 8.0 are both supported,
```
igorpeshansky: Sorry to be unclear, I was referring to the second part of this comment ("this configuration is used for…") earlier. This part was still useful.

PR #1208, opened 2023-04-20
Support Dockerfile on ARM

jefferbrecht:
## Description
This abstracts away the hard-coded `amd64`, etc. strings from our Dockerfile so that it can be run on any platform.

Currently amd64 and arm64 are supported, but it can be extended easily in the future if needed, though I can't imagine why.

## Related issue
[b/274914224](http://b/274914224)

## How has this been tested?
Tested the build offline for the ARM platforms we care about, which passed. Though until we add CI support, the goal of this PR is simply to not break the existing `amd64` builds.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


Dockerfile
```diff
@@ -31,20 +31,18 @@ RUN set -x; yum -y update && \
 		gcc gcc-c++ make bison flex file systemd-devel zlib-devel gtest-devel rpm-build java-11-openjdk-devel \
 		expect rpm-sign zip wget && \
 		yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
-		yum install -y cmake3 && \
-		ln -fs cmake3 /usr/bin/cmake && \
-		# Download and install CMake manually (zypper only gives us CMake 3.5, but we need >= 3.12)
-		wget https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-x86_64.sh -O cmake.sh && \
-		(echo '4d98de8d605da676e71a889dd94f80c76abb377fade2f21e3510e62ece1e1ada  cmake.sh' | sha256sum --check) && \
-		bash cmake.sh --skip-license --prefix=/usr/local
+		wget https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-$(uname -m).sh -O cmake.sh && \
```
igorpeshansky, line 34: This is `$basearch`, which, as I recall, is `$(uname -i)` (see, e.g., https://www.linuxquestions.org/questions/linux-software-2/yum-variables-%24distribution-%24releasever-%24basearch-how-to-check-648050/). Also, you need to quote the command substitution.

But this would also prevent any sort of cross-build in the future, so maybe parameterize the architecture as a build argument instead?

Dockerfile
```diff
@@ -537,24 +535,23 @@ RUN set -x; \
...
-	
+		wget https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-$(uname -m).sh -O cmake.sh && \
+	((echo '4d98de8d605da676e71a889dd94f80c76abb377fade2f21e3510e62ece1e1ada  cmake.sh' | sha256sum --check) || \
+	(echo '73a35cab2174a3eb8f35083d55c80871185dc3808f3dae3558cd5fbdb29a4614  cmake.sh' | sha256sum --check)) && \
+	bash cmake.sh --skip-license --prefix=/usr/local
 		ADD https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.13%2B8/OpenJDK11U-jdk_x64_linux_hotspot_11.0.13_8.tar.gz /tmp/OpenJDK11U-jdk_x64_linux_hotspot_11.0.13_8.tar.gz
 		RUN set -xe; \
-			mkdir -p /usr/local/java-11-openjdk && \
-			tar -xf /tmp/OpenJDK11U-jdk_x64_linux_hotspot_11.0.13_8.tar.gz -C /usr/local/java-11-openjdk --strip-components=1
+		mkdir -p /usr/local/java-11-openjdk && \
```
rafaelwestphal, line 544: Should the indentation here be the previous?

rafaelwestphal (commented):
Just nit that the indentations seem to be off for some of the changes you made. Other than that LGTM

Dockerfile
```diff
@@ -31,20 +31,18 @@ RUN set -x; yum -y update && \
...
 		yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
-		yum install -y cmake3 && \
-		ln -fs cmake3 /usr/bin/cmake && \
-		# Download and install CMake manually (zypper only gives us CMake 3.5, but we need >= 3.12)
-		wget https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-x86_64.sh -O cmake.sh && \
-		(echo '4d98de8d605da676e71a889dd94f80c76abb377fade2f21e3510e62ece1e1ada  cmake.sh' | sha256sum --check) && \
-		bash cmake.sh --skip-license --prefix=/usr/local
+		wget https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-$(uname -m).sh -O cmake.sh && \
+	((echo '4d98de8d605da676e71a889dd94f80c76abb377fade2f21e3510e62ece1e1ada  cmake.sh' | sha256sum --check) || \
+	(echo '73a35cab2174a3eb8f35083d55c80871185dc3808f3dae3558cd5fbdb29a4614  cmake.sh' | sha256sum --check)) && \
```
igorpeshansky, line 36: Is this because the ARM package has a different checksum? Would it be possible to make that explicit (e.g., define a map from architecture to checksum and key it by the parameter) instead?

Dockerfile
```diff
@@ -135,14 +133,14 @@ RUN set -x; yum -y update && \
 		yum -y install git systemd \
 		autoconf libtool libcurl-devel libtool-ltdl-devel openssl-devel yajl-devel \
 		gcc gcc-c++ make cmake bison flex file systemd-devel zlib-devel gtest-devel rpm-build systemd-rpm-macros java-11-openjdk-devel \
-		expect rpm-sign zip
+		expect rpm-sign zip wget
```
igorpeshansky, line 136: `curl` is usually pre-installed. Do we also need `wget`?

igorpeshansky (commented):
Some initial comments.

Dockerfile
jefferbrecht: `uname -i` returns `unknown` on my x86_64 instances :(

Parameterizing with a build argument would also resolve the `ADD` -> `RUN` issue; I wanted to avoid that complexity but if it helps here too then maybe there's enough upsides to go through with it. I'll give it a try.

Dockerfile
jefferbrecht: Yeah, ARM has a different checksum. I'll try making it a map.

Dockerfile
jefferbrecht: There were one or two images that didn't have `curl`. I'm fine just adding `curl` to those if that's preferable over adding `wget` everywhere.

Dockerfile
jefferbrecht: I'm not too concerned with the indentation in the Dockerfile since it's auto-generated; my preference is consistent indentation in `compile.go` since that file is the hand-written one. But I'm also fine to revert this to reduce the size of the diff.

Dockerfile
igorpeshansky: I usually prefer `curl` over `wget`…

Dockerfile
hsmatulisgoogle: I think it's worth trying to parametrize this as well, as the ADD -> RUN is a bit painful. If so, it might be worth it to add this to the template itself and make it so every build will build cmake in this way to make things consistent

jefferbrecht: Changed to `BUILDARCH` which automatically populates with `amd64`/`arm64` when using `DOCKER_BUILDKIT=1`.

Dockerfile
jefferbrecht: Changed to a map.

Dockerfile
jefferbrecht: Removed `wget` and added `curl` when needed.

hsmatulisgoogle (approved):

Dockerfile
```diff
@@ -639,7 +678,10 @@ COPY --from=sles12-build /google-cloud-ops-agent*.rpm /
 
 FROM opensuse/leap:15.1 AS sles15-build-base
 
-RUN set -x; zypper -n install git systemd autoconf automake flex libtool libcurl-devel libopenssl-devel libyajl-devel gcc gcc-c++ zlib-devel rpm-build expect cmake systemd-devel systemd-rpm-macros java-11-openjdk-devel unzip zip wget
+# Requires DOCKER_BUILDKIT=1
+ARG BUILDARCH
+
+RUN set -x; zypper -n install curl git systemd autoconf automake flex libtool libcurl-devel libopenssl-devel libyajl-devel gcc gcc-c++ zlib-devel rpm-build expect cmake systemd-devel systemd-rpm-macros java-11-openjdk-devel unzip zip wget
```
igorpeshansky, line 684: You no longer need `wget`, right?

Dockerfile
```diff
@@ -25,26 +25,39 @@
...
+                arch='aarch64' \
+                hash='73a35cab2174a3eb8f35083d55c80871185dc3808f3dae3558cd5fbdb29a4614' \
+                ;; \
+        *) \
+                echo 'unsupported' && false \
+                ;; \
+        esac && \
+	curl --location "https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-$arch.sh" --output cmake.sh && \
+	(echo "$hash  cmake.sh" | sha256sum --check) && \
+	bash cmake.sh --skip-license --prefix=/usr/local
```
igorpeshansky, line 52: There's actually a simpler way. Try adding the following at the top of the `Dockerfile`:
```
FROM alpine:latest AS cmake-amd64

ENV hash=4d98de8d605da676e71a889dd94f80c76abb377fade2f21e3510e62ece1e1ada
ADD https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-x86_64.sh \
    /cmake.sh

FROM alpine:latest AS cmake-arm64

ENV hash=73a35cab2174a3eb8f35083d55c80871185dc3808f3dae3558cd5fbdb29a4614
ADD https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-aarch64.sh \
    /cmake.sh

FROM cmake-${BUILDARCH} AS cmake-install
RUN set -xe; (echo "$hash  /cmake.sh" | sha256sum -c)
```
and then just use the following in each image:
```
COPY --from=cmake-install /cmake.sh /cmake.sh
RUN … && \
    bash /cmake.sh --skip-license --prefix=/usr/local
```

(aside: wish the [`--checksum` feature](https://docs.docker.com/engine/reference/builder/#verifying-a-remote-file-checksum-add---checksumchecksum-http-src-dest) were already available ­— then we'd be able to just do this:
```
FROM scratch AS cmake-amd64

ADD --checksum=sha256:4d98de8d605da676e71a889dd94f80c76abb377fade2f21e3510e62ece1e1ada \
    https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-x86_64.sh \
    /cmake.sh

FROM scratch AS cmake-arm64

ADD --checksum=sha256:73a35cab2174a3eb8f35083d55c80871185dc3808f3dae3558cd5fbdb29a4614 \
    https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-aarch64.sh \
    /cmake.sh

FROM cmake-${BUILDARCH} AS cmake-install
```
Oh, well…)

jefferbrecht: Thanks -- added!

Dockerfile
jefferbrecht: Right -- updated.

Dockerfile
```diff
@@ -25,26 +25,39 @@
...
 		ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk/
 
 SHELL ["/bin/bash", "-c"]
 
 # Install golang
-ADD https://golang.org/dl/go1.20.2.linux-amd64.tar.gz /tmp/go1.20.2.linux-amd64.tar.gz
+ADD https://golang.org/dl/go1.20.2.linux-$BUILDARCH.tar.gz /tmp/go.tar.gz
```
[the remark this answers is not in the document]
jefferbrecht: Updated.

Dockerfile
```diff
@@ -140,9 +160,10 @@ RUN set -x; yum -y update && \
 SHELL ["/bin/bash", "-c"]
 
 # Install golang
-ADD https://golang.org/dl/go1.20.2.linux-amd64.tar.gz /tmp/go1.20.2.linux-amd64.tar.gz
+ARG BUILDARCH
+ADD https://golang.org/dl/go1.20.2.linux-$BUILDARCH.tar.gz /tmp/go1.20.2.tar.gz
```
igorpeshansky, line 164: [Optional]
```suggestion
ADD https://golang.org/dl/go1.20.2.linux-${BUILDARCH}.tar.gz /tmp/go1.20.2.tar.gz
```

Dockerfile
```diff
@@ -19,32 +19,50 @@
...
 RUN set -x; yum -y update && \
 		yum -y install git systemd \
 		autoconf libtool libcurl-devel libtool-ltdl-devel openssl-devel yajl-devel \
 		gcc gcc-c++ make bison flex file systemd-devel zlib-devel gtest-devel rpm-build java-11-openjdk-devel \
-		expect rpm-sign zip wget && \
+		expect rpm-sign zip && \
 		yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
-		yum install -y cmake3 && \
-		ln -fs cmake3 /usr/bin/cmake && \
-		# Download and install CMake manually (zypper only gives us CMake 3.5, but we need >= 3.12)
```
igorpeshansky, line 36: Might be worth keeping at least part of this comment (well, with `s/zypper/yum/` or something)…

Dockerfile
```diff
@@ -129,6 +147,8 @@ COPY --from=centos7-build /google-cloud-ops-agent*.rpm /
 
 FROM rockylinux:8 AS centos8-build-base
 
+COPY --from=cmake-install /cmake.sh /cmake.sh
```
igorpeshansky, line 150: Only needed if you actually install `cmake`, right? You could pull `cmake` installation out into a couple of separate layers — that way you could add both:
```
COPY --from=cmake-install /cmake.sh /cmake.sh
RUN set -x; bash /cmake.sh --skip-license --prefix=/usr/local
```
at once when needed…

Dockerfile
```diff
@@ -520,11 +550,13 @@ COPY --from=buster-build /google-cloud-ops-agent*.deb /
...
 FROM opensuse/archive:42.3 AS sles12-build-base
 
+COPY --from=cmake-install /cmake.sh /cmake.sh
+
 RUN set -x; \
 		# The 'OSS Update' repo signature is no longer valid, so verify the checksum instead.
 		zypper --no-gpg-check refresh 'OSS Update' && \
 		(echo 'b889b4bba03074cd66ef9c0184768f4816d4ccb1fa9ec2721c5583304c5f23d0  /var/cache/zypp/raw/OSS Update/repodata/repomd.xml' | sha256sum --check) && \
-		zypper -n install git systemd autoconf automake flex libtool libcurl-devel libopenssl-devel libyajl-devel gcc gcc-c++ zlib-devel rpm-build expect cmake systemd-devel systemd-rpm-macros unzip zip wget && \
+		zypper -n install curl git systemd autoconf automake flex libtool libcurl-devel libopenssl-devel libyajl-devel gcc gcc-c++ zlib-devel rpm-build expect cmake systemd-devel systemd-rpm-macros unzip zip && \
```
igorpeshansky, line 559: Where is `curl` used now?

Dockerfile
jefferbrecht: I'm not sure I follow; are you suggesting that we leave only the following in the header:

```
# Manually prepare a new enough version of CMake.
# This should be used on platforms where the default package manager
# does not provide a new enough version (we require >= 3.12).
FROM alpine:latest AS cmake-amd64

ENV hash=4d98de8d605da676e71a889dd94f80c76abb377fade2f21e3510e62ece1e1ada
ADD https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-x86_64.sh \
    /cmake.sh

FROM alpine:latest AS cmake-arm64

ENV hash=73a35cab2174a3eb8f35083d55c80871185dc3808f3dae3558cd5fbdb29a4614
ADD https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-aarch64.sh \
    /cmake.sh
```

and then move the `COPY / RUN` lines to each image on an as-needed basis?

Dockerfile
jefferbrecht: Added a catchall comment to `installCMake`. LMK if that fits or if you'd rather see inline comments so they show up in the Dockerfile.

Also, WDYT about renaming the layer to `cmake-3.25.2-install` so that the version we're actually installing is clear from within compile.go?

Dockerfile
jefferbrecht: Thanks, makes sense -- updated.

Dockerfile
jefferbrecht: Oops, forgot to remove the lingering ones. Fixed.

Dockerfile
```diff
@@ -19,32 +19,50 @@
...
 
+COPY --from=cmake-install /cmake.sh /cmake.sh
+
 RUN set -x; yum -y update && \
 		yum -y install git systemd \
 		autoconf libtool libcurl-devel libtool-ltdl-devel openssl-devel yajl-devel \
 		gcc gcc-c++ make bison flex file systemd-devel zlib-devel gtest-devel rpm-build java-11-openjdk-devel \
-		expect rpm-sign zip wget && \
+		expect rpm-sign zip && \
 		yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
```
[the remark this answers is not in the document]
jefferbrecht: Seems to work fine without it 🤷 I'll remove it and we'll see if anything else breaks.

Dockerfile
igorpeshansky: Missed this one. I think adding the version to the rule would result in a larger change if we ever decide to update it (plus, from what I understand, we don't actually care that the version is 3.25.2, as long as it's recent enough). You *could* rename the rules `cmake-*-recent` (e.g., `cmake-install-recent`) and make the version a build argument (defaulting to 3.25.2)… Alternatively, rename the script to `/cmake-recent.sh` (though I'd still pull out the version into a build argument).

jefferbrecht: Renamed to `cmake-*-recent`.

Dockerfile
```diff
@@ -536,25 +561,24 @@ RUN set -x; \
...
 		zypper -n install bison>3.4 && \
 		# Allow fluent-bit to find systemd
-		ln -fs /usr/lib/systemd /lib/systemd && \
-		# Download and install CMake manually (zypper only gives us CMake 3.5, but we need >= 3.12)
-		wget https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2-linux-x86_64.sh -O cmake.sh && \
-		(echo '4d98de8d605da676e71a889dd94f80c76abb377fade2f21e3510e62ece1e1ada  cmake.sh' | sha256sum --check) && \
-		bash cmake.sh --skip-license --prefix=/usr/local
-	
+		ln -fs /usr/lib/systemd /lib/systemd
 		ADD https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.13%2B8/OpenJDK11U-jdk_x64_linux_hotspot_11.0.13_8.tar.gz /tmp/OpenJDK11U-jdk_x64_linux_hotspot_11.0.13_8.tar.gz
```
igorpeshansky, line 565: Just realized — don't you need to change this as well (`x64`)?

Dockerfile
```diff
@@ -639,7 +663,7 @@ COPY --from=sles12-build /google-cloud-ops-agent*.rpm /
 
 FROM opensuse/leap:15.1 AS sles15-build-base
 
-RUN set -x; zypper -n install git systemd autoconf automake flex libtool libcurl-devel libopenssl-devel libyajl-devel gcc gcc-c++ zlib-devel rpm-build expect cmake systemd-devel systemd-rpm-macros java-11-openjdk-devel unzip zip wget
+RUN set -x; zypper -n install git systemd autoconf automake flex libtool libcurl-devel libopenssl-devel libyajl-devel gcc gcc-c++ zlib-devel rpm-build expect cmake systemd-devel systemd-rpm-macros java-11-openjdk-devel unzip zip
 		# Add agent-vendor.repo to install >3.4 bison
 		RUN echo $'[google-cloud-monitoring-sles15-x86_64-test] \n\
```
igorpeshansky, line 668: This would need to change as well, right (`x86_64`)? I'm pretty sure we can rename this to use `vendor` and ignore the arch, and use `$basearch` in the `baseurl` instead.

igorpeshansky (commented):
Looks like we've missed some instances of hard-coded architecture…

Dockerfile
jefferbrecht: Blah, yet another arch syntax to handle. It's extra painful because we don't need to support sles12 yet anyway. Might have to use the same strategy here that we used for CMake.

Dockerfile
jefferbrecht: (Or, since we won't be supporting ARM sles-15 in the first cut, can we just fail the build in this specific scenario? I don't have a sense of how big an effort it is to add ARM support for whatever this package is.)

Dockerfile
jefferbrecht: Fixed.

Dockerfile
jefferbrecht: Decided to fail the build on ARM for sles-15 for now, we can add the repo later.

igorpeshansky: You could have still used `$basearch` in the repo definition — the build would fall if the repo is nonexistent.

jefferbrecht: Updated.

igorpeshansky (approved):
LGTM :shipit:

PR #1224, opened 2023-04-27
Testing: Spread test VMs across multiple zones

martijnvans:
## Description
Support a weighted round robin zone selection algorithm, which will allow us to spread our test VMs across multiple zones. For now the new feature is unused and we're still all in one region. But I anticipate getting more quota in different zones soon, and when that happens it will be good to be able to use it.

This will be particularly helpful for ARM testing, where we have only 16 CPUs per zone currently.

## Related issue
[b/280064697](http://b/280064697)

## How has this been tested?
I did a bit of manual testing on the zone picker to see that it was more or less doing its job. It did look to be doing a weighted round robin algorithm, albeit with a bit more randomness than I was expecting.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


kokoro/config/test/common.gcl
```diff
@@ -15,7 +15,15 @@ template config common = {
...
-      ZONE = 'us-central1-b'
+      ZONES =join([
+        'us-central1-a=3',
+        'us-central1-b=3',
+        'us-central1-c=3',
+        'us-central1-f=3',
+        'us-east1-b=2',
+        'us-east1-c=2',
+        'us-east1-d=2',
+      ], ',')
```
igorpeshansky, line 26: Any reason we don't add a function in `utils/functions.gcl`, like:

```
// Convert a map into a comma-separated list of key-value pairs.
//
// Example usage:
//  make_zones(['foo': 'bar', 'x': 'y'])
// Returns
//  'foo=bar,x=y'
make_zones = lambda zones_: {
  assert is_map(zones_)
  ret = join(map(lambda x_: "%x_[0]%=%x_[1]%", zones_.items()), ',')
}.ret
```
and use it here:
```suggestion
      ZONES = functions.make_zones([
        'us-central1-a': 3,
        'us-central1-b': 3,
        'us-central1-c': 3,
        'us-central1-f': 3,
        'us-east1-b': 2,
        'us-east1-c': 2,
        'us-east1-d': 2,
      ])
```
?

integration_test/gce/zone_picker.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// parseZoneEnvironmentVariables looks at `weightedZones` (and `zone` if
+// needed) and extracts the zones and weights into a a weighted round robin
+// zone picker.
+// TODO(martijnvans): Remove support for `zone`.
+func newZonePicker(weightedZones, zone string) (*weightedRoundRobin, error) {
+	if weightedZones == "" {
+		if zone == "" {
+			return nil, errors.New("either ZONE or ZONES must be specified")
```
igorpeshansky, line 39: Since we prefer `ZONES`, let's put it first?

```suggestion
			return nil, errors.New("either ZONES or ZONE must be specified")
```

integration_test/gce/zone_picker.go
```diff
@@ -0,0 +1,61 @@
...
+		weightedZones = zone + "=1"
+	}
+
+	sw := &weighted.SW{}
+	// Each zoneSpec should look like <string>=<integer>.
+	for _, zoneSpec := range strings.Split(weightedZones, ",") {
+		zoneAndWeight := strings.Split(zoneSpec, "=")
+		if len(zoneAndWeight) != 2 {
+			return nil, fmt.Errorf(`invalid zone specification %q from ZONES=%q; should be like "us-central1=5"`, zoneSpec, weightedZones)
+		}
```
igorpeshansky, line 50: [Optional] I suspect that in many cases, the weights will be equal, so it might make sense to support a simple comma-separated list of zones as well…

```suggestion
		if len(zoneAndWeight) == 1 {
			zoneAndWeight = append(zoneAndWeight, "1")
		} else if len(zoneAndWeight) != 2 {
			return nil, fmt.Errorf(`invalid zone specification %q from ZONES=%q; should be like "us-central1=5"`, zoneSpec, weightedZones)
		}
```

integration_test/gce/zone_picker.go
```diff
@@ -0,0 +1,61 @@
...
+		weightedZones = zone + "=1"
+	}
+
+	sw := &weighted.SW{}
+	// Each zoneSpec should look like <string>=<integer>.
+	for _, zoneSpec := range strings.Split(weightedZones, ",") {
+		zoneAndWeight := strings.Split(zoneSpec, "=")
+		if len(zoneAndWeight) != 2 {
+			return nil, fmt.Errorf(`invalid zone specification %q from ZONES=%q; should be like "us-central1=5"`, zoneSpec, weightedZones)
```
igorpeshansky, line 49: Nit: `us-central1` is a region. Should we use a real zone in the example usage?

integration_test/gce/gce_testing.go
```diff
@@ -44,7 +44,8 @@ AGENT_PACKAGES_IN_GCS, for details see README.md.
 
 This library needs the following environment variables to be defined:
 PROJECT: What GCP project to use.
-ZONE: What GCP zone to run in.
+ZONES: What GCP zones to run in, with integer weights attached to each zone,
+in the format: zone1=weight1,zone2=weight2
```
igorpeshansky, line 48: Nit: missing trailing '.'.

integration_test/gce/zone_picker.go
```diff
@@ -0,0 +1,61 @@
...
+// Next() selects the next zone to spawn a VM in.
+// It is thread safe.
+func (wrr *weightedRoundRobin) Next() string {
+	wrr.mutex.Lock()
+	defer wrr.mutex.Unlock()
+
+	return wrr.sw.Next().(string)
+}
+
+// parseZoneEnvironmentVariables looks at `weightedZones` (and `zone` if
```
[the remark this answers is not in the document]
martijnvans: yup, renamed.

kokoro/config/test/common.gcl
martijnvans: well I don't see the point in doing that. It's adding an extra layer of indirection and a helper function with some actual logic in it, and I don't think the result is any better.

integration_test/gce/zone_picker.go
martijnvans: sure

integration_test/gce/gce_testing.go
martijnvans: I did this on purpose so that the period wouldn't be misinterpreted as part of the ZONES variable format.

kokoro/config/test/common.gcl
igorpeshansky: Making it a map guarantees that the keys are unique, and removes the possibility of, e.g., stray spaces screwing up the flag. But up to you.

integration_test/gce/zone_picker.go
martijnvans: done, but I plan to get rid of ZONE anyway.

integration_test/gce/zone_picker.go
martijnvans: I think supporting more than one format makes things more complicated, not simpler. Like, what do you do with this:

zone1=50,zone2

the magical =1 insertion is a surprise, I'd rather just have one consistent behavior.

igorpeshansky: True, that one would be somewhat controversial. I was thinking more of `zone1,zone2`…

kokoro/config/test/common.gcl
martijnvans: I still think I'd rather not add that layer of indirection just for that.

integration_test/gce/zone_picker.go
martijnvans: implemented.

igorpeshansky (commented):
One typo and two optional comments.

integration_test/gce/gce_testing.go
```diff
@@ -1026,8 +1036,10 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
 		vm.Network = "default"
 	}
 	if vm.Zone == "" {
-		vm.Zone = os.Getenv("ZONE")
+		// Chooses the next zone according to the weights in WEIGHTED_ZONES.
```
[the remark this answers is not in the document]
martijnvans: done

integration_test/gce/zone_picker.go
```diff
@@ -0,0 +1,69 @@
...
+	return wrr.sw.Next().(string)
+}
+
+// newZonePicker looks at `zones` (and `zone` if needed) and extracts
+// the zones and weights into a a weighted round robin zone picker.
+// `zones` should be a comma-separated list of zone specs, where each
+// zone spec is either in the format <zone>=<integer weight> or just <zone>
+// (in which case the weight defaults to 1).
+// TODO(martijnvans): Remove support for the `zone` argument.
+func newZonePicker(zones, zone string) (*weightedRoundRobin, error) {
```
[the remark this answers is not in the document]
martijnvans: OK let's do that later if the current setup proves problematic in some way.

integration_test/gce/gce_testing.go
```diff
@@ -44,7 +44,10 @@ AGENT_PACKAGES_IN_GCS, for details see README.md.
 
 This library needs the following environment variables to be defined:
 PROJECT: What GCP project to use.
-ZONE: What GCP zone to run in.
+ZONES: What GCP zones to run in as a comma-separated list, with optional
+integer weights attached to each zone, in the format:
+zone1=weight1,zone2=weight2. Any zone with no weight is given a default weight
+of 1.
```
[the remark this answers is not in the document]
martijnvans: I don't want to do that because I plan on getting rid of ZONE imminently.

igorpeshansky (approved):
LGTM :shipit:

PR #1226, opened 2023-05-01
Removing logging of metadata query result

rafaelwestphal:
## Description
Removing the result from metadata query from being added to health-checks.log

## Related issue
b/279212943

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/internal/healthchecks/api_check.go
+++ b/internal/healthchecks/api_check.go
@@ -97,7 +97,7 @@ func runLoggingCheck(logger *log.Logger) error {
 	if err != nil {
 		return fmt.Errorf("can't get GCE metadata: %w", err)
 	}
-	logger.Printf("gce metadata: %+v", gceMetadata)
+	logger.Printf("GCE Metadata queried successfully")
 
 	// New Logging Client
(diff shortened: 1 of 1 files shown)


internal/healthchecks/api_check.go
```diff
@@ -97,7 +97,7 @@ func runLoggingCheck(logger *log.Logger) error {
 	if err != nil {
 		return fmt.Errorf("can't get GCE metadata: %w", err)
```
franciscovalentecastro, line 98: After thinking this after a couple of days. I think removing the `err` here is a good idea. We already have a `Network Check` that queries the metadata server and the error here might (in a super rare case) could expose PII details (e.g. can't find attribute `x`). The `API Check` responsibility is not to verify the metadata server.

Just leaving (or improving a little) the message should be enough. 
```
fmt.Errorf("can't get GCE metadata`)
``` 

Synced with @LujieDuan about this before. CC if he has more opinions about this.

rafaelwestphal: SGTM

franciscovalentecastro (approved):
LGTM

PR #1229, opened 2023-05-03
fluent-bit[coro_stack_size] make configurable

babakhanov:
Our team recently encountered issues with the Fluent Bit service failing on various GCP VMs, most of which were running MongoDB, although this was not always the case.

Upon investigating the issue, we found that it was related to the `coro_stack_size` Fluent Bit parameter (https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/yaml/configuration-file) which default value is too low in our case. 

I think that this parameter should be made configurable and this changes will be beneficial for the `ops-agent` service and the users who rely on it.


babakhanov:
reopened https://github.com/GoogleCloudPlatform/ops-agent/pull/1231

PR #1249, opened 2023-05-11
Testing: Move installation functions into `agents` package

martijnvans:
## Description
This change will allow me to reuse the SetupOpsAgent function from soak tests.

## Related issue
[b/278590309](http://b/278590309)

## How has this been tested?
automated tests only

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/agents/agents.go
+++ b/integration_test/agents/agents.go
@@ -29,6 +29,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 	"path"
 	"regexp"
 	"strings"
@@ -37,6 +38,7 @@ import (
 
 	"github.com/GoogleCloudPlatform/ops-agent/integration_test/gce"
 	"github.com/GoogleCloudPlatform/ops-agent/integration_test/logging"
+	"github.com/GoogleCloudPlatform/ops-agent/integration_test/util"
 
 	"github.com/blang/semver"
 	"github.com/cenkalti/backoff/v4"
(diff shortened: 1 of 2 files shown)


integration_test/agents/agents.go
```diff
@@ -664,6 +666,115 @@ func InstallStandaloneWindowsMonitoringAgent(ctx context.Context, logger *log.Lo
 	return err
 }
 
+// PackageLocation describes a location where packages
+// (currently, only the Ops Agent packages) live.
+type PackageLocation struct {
```
Subbarker, line 671: this is a lot of inferred behavior, can we make the struct be more strict perhaps? do do validation in it's creation?

martijnvans: PTAL, I added code to throw an error for conflicting install locations, not sure if that is sufficient. We could throw an error when creating the struct in the first place, but I think that wouldn't be any better than throwing an error when we try to use the struct (which is what I did) because that happens immediately afterr creating the struct anyways. In other words, the error would be returned at the same moment anyway, it would just cost a slight amount of additional code at each callsite for PackageLocationForEnvVars.

Subbarker (approved):

integration_test/agents/agents.go
```diff
@@ -664,6 +666,124 @@ func InstallStandaloneWindowsMonitoringAgent(ctx context.Context, logger *log.Lo
...
+	return PackageLocation{
+		packagesInGCS:          os.Getenv("AGENT_PACKAGES_IN_GCS"),
+		repoSuffix:             os.Getenv("REPO_SUFFIX"),
+		artifactRegistryRegion: os.Getenv("ARTIFACT_REGISTRY_REGION"),
+	}
+}
+
+// InstallOpsAgent installs the Ops Agent on the given VM. Consults the given
+// PackageLocation to determine where to install the agent from. For details
+// about PackageLocation, see the documentation for the PackageLocation struct.
```
igorpeshansky, line 703: Trailing whitespace?

martijnvans: fixed

PR #1250, opened 2023-05-15
Testing: Soak tests on new codebase, supports Windows and Linux

martijnvans:
## Description
Implements soak tests on Windows (new functionality) and replaces our current code for running soak tests on Ubuntu.

## Related issue
[b/278590309](http://b/278590309)

## How has this been tested?

see screenshots inline. Works on debian-11:

linux_soak

and windows-2019:

windows_soak

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/README.md
```diff
@@ -296,3 +296,22 @@ https://storage.googleapis.com/ops-agents-public-buckets-vendored-deps/mirrored-
...
+parameters as desired) and run:
+
+```
+PROJECT=my_project \
+  DISTRO=debian-11 \
+  ZONES=us-central1-b \
+  TTL=100m \
+  LOG_SIZE_IN_BYTES=1000 \
+  LOG_RATE=1000 \
+  go run -tags=integration_test soak_test/launcher.go
```
quentinmit, line 312: You should never use `go run` with a filename or filenames.

Put the launcher in its own directory and use

```
go run ./soak_test/cmd/launcher
```

(Note the leading `./` is required, making it a relative package path.)

integration_test/soak_test/launcher.go
```diff
@@ -0,0 +1,191 @@
...
+
+//go:build integration_test
+
+/*
+The launcher command launches a VM and begins a soak test on it.
+
+Specifically, it installs the Ops Agent and a Python program that
+logs to a specific file that the Ops Agent is watching.
+
+This command is configured by the following environment variables,
```
quentinmit, line 23: Why does this use environment variables and not command-line flags?

integration_test/soak_test/launcher.go
```diff
@@ -0,0 +1,191 @@
...
+    pipelines:
+      my_pipeline:
+        receivers: [mylog_source]
+        exporters: [google]
+`, logPath)
+	if err := agents.SetupOpsAgent(ctx, logger, vm, config); err != nil {
+		return err
+	}
+
+	// Install python.
```
quentinmit, line 134: Leave a TODO to explore options that avoid installing Python on Windows. (e.g. building a self-contained executable, or rewriting the log generator in Go)

kokoro/scripts/test/start_soak_test.sh
```diff
@@ -0,0 +1,24 @@
...
+    break
+  fi
+done
+
+LOG_RATE=${LOG_RATE-100000} \
+LOG_SIZE_IN_BYTES=${LOG_SIZE_IN_BYTES-1000} \
+VM_NAME="${VM_NAME:-soak-test-${KOKORO_BUILD_NUMBER}-$(date +%F)}" \
+DISTRO="${DISTRO:-ubuntu-2004-lts}" \
+TTL="${TTL:-5m}" \
+  go run "launcher.go" \
```
quentinmit, line 21: Same comment about `go run`.

quentinmit (approved):

integration_test/soak_test/launcher.go
martijnvans: Not sure I agree with this one. Also that makes me not sure what to put in the todo, specifically the "why" part of the TODO. Is it because we don't like installing Python on windows? (IMO, go would be no easier). Is it because we don't like Python's performance?

integration_test/soak_test/launcher.go
martijnvans: I dunno, a few reasons:

- env variables are easy
- certain env variables that we use are already supplied by kokoro, no need to explicitly copy them into flags in the .sh file
- the integration tests currently use env variables instead of flags
- env variables and flags have pretty much the same syntax anyway, not sure what the advantage of flags would be?

integration_test/README.md
martijnvans: > You should never use go run with a filename or filenames.

why?

also, because of the way embed works, I'd need to put log_generator.py inside cmd too, which seems a bit odd. But I'll do it if you have some reason for it.

quentinmit: `go run` will build the filenames in a subtly different way that can cause hard-to-debug problems. In particular it will compose a new main package from the passed filenames, which is completely different from the normal one-package-per-directory rule (that all other tools, like `go test`, assume). So the usual advice is simply "do not use go run with filename(s)". By convention, `main` packages should be defined in directories named `.../cmd/foo`.

integration_test/soak_test/launcher.go
quentinmit: It's just non-idiomatic for Go. If you're consuming Kokoro-defined variables then I guess you should do the locally normal thing.

integration_test/soak_test/launcher.go
quentinmit: The idea is to do the building of the tool not on the target machine, so that nothing (Python or Go) needs to be installed on it.

PR #1287, opened 2023-06-14
Integration Tests for TCP and Fluent Forward Receivers on Windows

shafinsiddique:
## Description

We don't currently test the TCP and fluent forward receivers on Windows. This PR adds those integration tests. 

There is currently a bug on TCP receivers on Windows so the integration test is set to be skipped. Once the bug is fixed, we can start running the integration test as part of our testing suite. 

Note : The Powershell script included in this PR is from [this](https://github.com/fleschutz/PowerShell) Github repo. I have checked the license and it is CCO which means we can freely use it.   
 
## Related Issue

b/285883136

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/ops_agent_test.go
```diff
@@ -1486,11 +1535,25 @@ func TestFluentForwardLog(t *testing.T) {
 		//
 		// The forwarding Fluent Bit uses the tag "forwarder_tag" when sending the
 		// log record. This will be preserved in the LogName.
-		if _, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", "echo '{\"msg\":\"test fluent forward log\"}' | /opt/google-cloud-ops-agent/subagents/fluent-bit/bin/fluent-bit -i stdin -o forward://127.0.0.1:24224 -t forwarder_tag"); err != nil {
-			t.Fatalf("Error writing dummy forward protocol log line: %v", err)
+
+		var log string
```
ridwanmsharif, line 1539: `log` shouldn't change based on the distro either.

integration_test/ops_agent_test.go
```diff
@@ -1486,11 +1535,25 @@ func TestFluentForwardLog(t *testing.T) {
...
+		var command string
+		wait := true
```
ridwanmsharif, line 1541: Why does `wait` depend on distro here?

integration_test/ops_agent_test.go
```diff
@@ -1441,9 +1443,48 @@ func TestTCPLog(t *testing.T) {
 			t.Fatal(err)
 		}
 
+		command := "echo '{\"msg\":\"test tcp log 1\"}{\"msg\":\"test tcp log 2\"}' > /dev/tcp/localhost/5170 && echo '{\"msg\":\"test tcp log 3\"}{\"msg\":\"test tcp log 4\"}' > /dev/tcp/localhost/5170"
+		// Installing Netcat or equivalent tools on Windows is tricky so we can instead
+		// use this Powershell Script that can send TCP messages.
+		if gce.IsWindows(platform) {
```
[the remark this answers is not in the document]
ridwanmsharif: Would also be nice to have the linux part do the same tbh

integration_test/ops_agent_test.go
```diff
@@ -1420,8 +1420,10 @@ func TestTCPLog(t *testing.T) {
 	gce.RunForEachPlatform(t, func(t *testing.T, platform string) {
 		t.Parallel()
 		if gce.IsWindows(platform) {
+			// DELETE THIS BLOCK WHEN TCP BUG IS FIXED. The code for testing on Windows is already present below.
```
ridwanmsharif, line 1423: [nit] Add a TODO here and reference the bug number tracking this.

ridwanmsharif (commented):
Looks pretty solid overall! Left a few comments but nothing major

integration_test/ops_agent_test.go
braydonk, line 1539: This can be one line `var log, command string`

integration_test/ops_agent_test.go
braydonk, line 1423: To follow the format of other comments like this, we can change this to:
```suggestion
			// TODO: Delete when b/285865631 is fixed.
```

integration_test/ops_agent_test.go
```diff
@@ -1441,9 +1443,48 @@ func TestTCPLog(t *testing.T) {
 			t.Fatal(err)
 		}
 
+		command := "echo '{\"msg\":\"test tcp log 1\"}{\"msg\":\"test tcp log 2\"}' > /dev/tcp/localhost/5170 && echo '{\"msg\":\"test tcp log 3\"}{\"msg\":\"test tcp log 4\"}' > /dev/tcp/localhost/5170"
```
jefferbrecht, line 1446: You can use backtick strings here to avoid backslash-soup:

```
command := `echo '{"msg":"test ...`
```

integration_test/gce/gce_testing.go
```diff
@@ -802,7 +811,32 @@ func RunRemotely(ctx context.Context, logger *log.Logger, vm *VM, stdin string,
...
+// StartRemotely starts a command on the provided VM.
+// The command should be a shell command if the VM is Linux, or powershell if the VM is Windows.
+// Returns the combined stdout+stderr as a string, plus an error if there was
+// a problem.
+// If the wait argument is true, the function waits for the command to finish running. Otherwise,
+// it returns immediately.
+// 'command' is what to run on the machine. Example: "cat /tmp/foo; echo hello"
+// 'stdin' is what to supply to the command on stdin. It is usually "".
+// TODO: Remove the stdin parameter, because it is hardly used.
+func StartRemotely(ctx context.Context, logger *log.Logger, vm *VM, stdin string, command string, wait bool) (_ CommandOutput, err error) {
```
jefferbrecht, line 838: Can we avoid the `StartRemotely` and `startCommand` stuff by just following what the Linux version of the fluentd-forward test does?

braydonk: Fluent Bit won't read from stdin on Windows, meaning we have to run it as a background process. @shafinsiddique that means `cmd.Run` blocks forever because Fluent Bit keeps running right?

jefferbrecht: Even if it doesn't read from stdin, there are simpler workarounds available, e.g. write a single line to a text file and tail it.

braydonk: That doesn't solve the problem I don't think. The reason reading from stdin worked before if I understand correctly is that it ran as a one off and then concluded when `stdin` was done. Whether it's `random` input or `tail` it can't just do the same one-off operation it does on Linux.

shafinsiddique: > Fluent Bit won't read from stdin on Windows, meaning we have to run it as a background process. @shafinsiddique that means `cmd.Run` blocks forever because Fluent Bit keeps running right?

Yes, cmd.Run is blocking

shafinsiddique: > Starting the process is already a one-off operation. It's not clear why `RunRemotely` isn't sufficient here...
> 
> `&` probably won't work since I think processes started that way are bound to the shell, but certainly the WMI way should work to keep it backgrounded (`Invoke-WmiMethod -Class Win32_Process -Name Create -ArgumentList 'fluent-bit.exe' ...`)

We CAN use RunRemotely if we change it's implementation to call cmd.Start() instead of cmd.Run(). Is that what you're suggesting?

shafinsiddique: > `&` probably won't work since I think processes started that way are bound to the shell, but certainly the WMI way should work to keep it backgrounded (`Invoke-WmiMethod -Class Win32_Process -Name Create -ArgumentList 'fluent-bit.exe' ...`)

OH i see what you mean, okay this is a good idea

integration_test/ops_agent_test.go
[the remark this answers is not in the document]
shafinsiddique: I do agree, fluent bit is better than the TCP script. I changed the Linux test to use fluent bit. 

For Windows, the only way we could use fluent bit to send TCP requests is if we use stdin or tail input plugins. Stdin isnt supported in Windows and tail, in my experience has been flaky or at least quite tricky to get right as it seems to take a few commands for fluent bit to actually pick it up. 

The reason why we can't use somethign like 'random' or 'dummy' like we did in fluent forward test is because we want to query with actual log id to test that data from multiple requests are going through. With 'dummy' the data is always the same and with random the data is random values so querying is hard.

integration_test/ops_agent_test.go
shafinsiddique: agreed, made the change to keep it consistent

integration_test/ops_agent_test.go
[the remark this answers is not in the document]
ridwanmsharif: https://docs.fluentbit.io/manual/pipeline/inputs/dummy#configuration-parameters

Oh I thought there was a `dummy` config field that would let you configure what message to send. We could have 2 receivers with different messages, and then output it all to the same tcp destination. I don't feel too strongly about this tho, so if its too much trouble maybe its not worth it. 

deferring to @braydonk and @shafinsiddique for what is best here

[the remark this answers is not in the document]
ridwanmsharif: For other receivers, we often test with 1 log. In this case, it sounds like that makes it much simpler too. Is there a specific reason we need to test several tcp payloads here in this test? Has it got something to do with the `in_tcp` bug we found?

ridwanmsharif (commented):
Looks much nicer!

integration_test/ops_agent_test.go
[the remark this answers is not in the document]
jefferbrecht: Having multiple messages will test fluent-bit's ability to separate independent JSON chunks into records when there's no newline in-between.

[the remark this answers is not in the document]
shafinsiddique: yep, the in_tcp bug arises only on second emitted logs. The first ones go through. That's why we're adding an additional log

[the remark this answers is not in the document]
shafinsiddique: > Having multiple messages will test fluent-bit's ability to separate independent JSON chunks into records when there's no newline in-between.

this i believe you're talking about why we send two jsons per message. Ridwan i believe was asking why we send 2 messages (using the && or -and)

integration_test/ops_agent_test.go
```diff
@@ -1440,19 +1442,54 @@ func TestTCPLog(t *testing.T) {
...
 
-		if err := gce.WaitForLog(ctx, logger.ToMainLog(), vm, "tcp_logs", time.Hour, "jsonPayload.msg:test tcp log 1"); err != nil {
-			t.Error(err)
+		if _, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", command); err != nil {
+			t.Fatalf("Error writing dummy TCP log line: %v", err)
 		}
 
-		if err := gce.WaitForLog(ctx, logger.ToMainLog(), vm, "tcp_logs", time.Hour, "jsonPayload.msg:test tcp log 2"); err != nil {
-			t.Error(err)
+		for i := 1; i <= 4; i++ {
```
martijnvans, line 1489: we should do these in parallel. Otherwise, it will wait ~7 minutes back to back, 4 times, stretching the test duration out to 28+ minutes needlessly. For an example of how to do this, see https://github.com/GoogleCloudPlatform/ops-agent/blob/69bacda8ce09e5d9a1d2e5e90f8f70fe8c539be0/integration_test/ops_agent_test.go#L1996

The pattern is:
```
var waitGroup sync.WaitGroup
for i := 1; i <= 4; i++ {
  i:= i // https://go.dev/doc/faq#closures_and_goroutines
 waitGroup.Add(1)
 go func () {
    defer waitGroup.Done()
    if err := gce.WaitForLog(...); err != nil {
      t.Error(err)
    }
  }()
} // end for loop
```

integration_test/ops_agent_test.go
```diff
@@ -1440,19 +1442,54 @@ func TestTCPLog(t *testing.T) {
...
+			} catch {
+				"?? Error in line $($_.InvocationInfo.ScriptLineNumber): $($Error[0])"
+				exit 1
+			}`
+
+			if err := gce.UploadContent(ctx, logger.ToMainLog(), vm, strings.NewReader(tcp_script), `C:\Program Files\Google\Cloud Operations\Ops Agent\send_tcp.ps1`); err != nil {
+				t.Fatalf("Error uploading TCP script: %v", err)
+			}
+
+			command = `(& "C:\Program Files\Google\Cloud Operations\Ops Agent\send_tcp.ps1" 127.0.0.1 5170 '{"msg":"test tcp log 1"}{"msg":"test tcp log 2"}') -and (& "C:\Program Files\Google\Cloud Operations\Ops Agent\send_tcp.ps1" 127.0.0.1 5170 '{"msg":"test tcp log 3"}{"msg":"test tcp log 4"}')`
```
jefferbrecht, line 1479: WDYT about this version:

```suggestion
			tcpScript := `$ErrorActionPreference = 'Stop'
			function Send-TcpString {
			  param(
			    [string]$TargetIP,
			    [int]$TargetPort,
			    [string]$Message
			  )
			  $Socket = New-Object System.Net.Sockets.TCPClient($TargetIP, $TargetPort) 
			  $Stream = $Socket.GetStream() 
			  $Writer = New-Object System.IO.StreamWriter($Stream)
			  $Message | % {
			    $Writer.WriteLine($_)
			    $Writer.Flush()
			  }
			  $Stream.Close()
			  $Socket.Close()
			}
			Send-TcpString 127.0.0.1 5170 '{"msg":"test tcp log 1"}{"msg":"test tcp log 2"}'
			Send-TcpString 127.0.0.1 5170 '{"msg":"test tcp log 3"}{"msg":"test tcp log 4"}'`

			if _, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", tcpScript); err != nil {
				t.Fatalf("Error writing dummy TCP log line: %v", err)
			}
```

Rationale, mostly simplifications:
* Params should either have valid defaults or be required; may as well make them required here
* No need for intermediate `$IP` and `$Address` variables; the `TCPClient` constructor has an overload that takes a host and port directly
* You can define and invoke functions inline so we can skip the intermediate file upload
* The `Read-Host` stuff isn't needed since we aren't running an interactive prompt
* We've recently been trying to push for `$ErrorActionPreference = 'Stop'` -- the default error format is already sufficiently descriptive so we don't need to reformat it in a catch
* (nit) Use camelCase for local Go vars

shafinsiddique: I fully agree that this is a much neater and better script in general! thank you for helping me with it!

integration_test/ops_agent_test.go
shafinsiddique: Yep, fully agreed. Made the change to wait for the logs in parallel.

braydonk (approved):
I think this looks good to me now. The skip for the TCP test can be added in a followup, it will be fixed imminently when the release is ready.

PR #1290, opened 2023-06-16
confgenerator: Structured Ops Agent Health Logs

franciscovalentecastro:
## Description
This change creates the `confgenerator/self_logs.go` as a first step to organize all self logs into the same structured format. This is PR only applies the format for `health-checks.log` and the followup https://github.com/GoogleCloudPlatform/ops-agent/pull/1300 will enforce the format to a sample of `logging-module.log` logs.

Every Ops Agent Health Logs should have the following fields : 

1. Inherited Fields
    - [MonitoredResource](https://cloud.google.com/logging/docs/reference/v2/rest/v2/MonitoredResource) : The resource from which telemetry is being collected.

2. Required Fields
    - Code : A code to uniquely identify signals.
    - Agent : Description of agent collecting telemetry.
    - Message : A full description of the signal.
    - [LogSeverity](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logseverity) : A description of how impactful the problem is.
    - SchemaVersion : The version of the current schema used for the Health Signal.
3. Optional Fields
    - [SourceLocation](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logentrysourcelocation) : Location in source code of agent where the signal originated.

Google Cloud Logging LogEntry Output Format : 

## Related issue
b/287980723

## How has this been tested?
Some details about the testing of this PR : 
- Health Checks unit tests passed.
- Improved integration test `TestLoggingSelfLogs`.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/cmd/ops_agent_windows/run_windows.go
+++ b/cmd/ops_agent_windows/run_windows.go
@@ -25,6 +25,7 @@ import (
 	"github.com/GoogleCloudPlatform/ops-agent/apps"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
 	"github.com/GoogleCloudPlatform/ops-agent/internal/healthchecks"
+	"github.com/GoogleCloudPlatform/ops-agent/internal/logs"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/debug"
 	"golang.org/x/sys/windows/svc/eventlog"
@@ -145,39 +146,9 @@ func getHealthCheckResults() []healthchecks.HealthCheckResult {
 	return gceHealthChecks.RunAllHealthChecks(logger)
 }
 
-type WindowsServiceLogger struct {
-	srv *service
-}
-
-func (wsl WindowsServiceLogger) Infof(format string, v ...any) {
-	wsl.srv.log.Info(EngineEventID, fmt.Sprintf(format, v...))
-}
-
-func (wsl WindowsServiceLogger) Warnf(format string, v ...any) {
-	wsl.srv.log.Warning(EngineEventID, fmt.Sprintf(format, v...))
-}
-
-func (wsl WindowsServiceLogger) Errorf(format string, v ...any) {
-	wsl.srv.log.Error(EngineEventID, fmt.Sprintf(format, v...))
-}
-
-func (wsl WindowsServiceLogger) Infow(msg string, keysAndValues ...any) {
-	wsl.srv.log.Info(EngineEventID, msg)
-}
-
-func (wsl WindowsServiceLogger) Warnw(msg string, keysAndValues ...any) {
-	wsl.srv.log.Warning(EngineEventID, msg)
-}
-
-func (wsl WindowsServiceLogger) Errorw(msg string, keysAndValues ...any) {
-	wsl.srv.log.Warning(EngineEventID, msg)
-}
-
-func (wsl WindowsServiceLogger) Println(v ...any) {}
-
 func (srv *service) runHealthChecks() {
 	healthCheckResults := getHealthCheckResults()
-	logger := WindowsServiceLogger{srv}
+	logger := logs.WindowsServiceLogger{EventID: EngineEventID, Logger: srv.log}
 	healthchecks.LogHealthCheckResults(healthCheckResults, logger)
 	srv.log.Info(EngineEventID, "Startup checks finished")
 }
--- a/cmd/ops_agent_windows/run_windows.go
+++ b/cmd/ops_agent_windows/run_windows.go
@@ -152,7 +152,7 @@ type WindowsServiceLogger struct {
 
 func (wsl WindowsServiceLogger) Infof(format string, v ...any) {
 	if len(v) > 0 {
-		wsl.srv.log.Info(EngineEventID, fmt.Sprintf(format, logs.DiscardZapFields(v ...)))
+		wsl.srv.log.Info(EngineEventID, fmt.Sprintf(format, logs.DiscardZapFields(v ...)...))
 	} else {
 		wsl.srv.log.Info(EngineEventID, format)
 	}
@@ -160,7 +160,7 @@ func (wsl WindowsServiceLogger) Infof(format string, v ...any) {
 
 func (wsl WindowsServiceLogger) Warnf(format string, v ...any) {
 	if len(v) > 0 {
-		wsl.srv.log.Warning(EngineEventID, fmt.Sprintf(format, logs.DiscardZapFields(v ...)))
+		wsl.srv.log.Warning(EngineEventID, fmt.Sprintf(format, logs.DiscardZapFields(v ...)...))
 	} else {
 		wsl.srv.log.Warning(EngineEventID, format)
 	}
@@ -168,7 +168,7 @@ func (wsl WindowsServiceLogger) Warnf(format string, v ...any) {
 
 func (wsl WindowsServiceLogger) Errorf(format string, v ...any) {
 	if len(v) > 0 {
-		wsl.srv.log.Error(EngineEventID, fmt.Sprintf(format, logs.DiscardZapFields(v ...)))
+		wsl.srv.log.Error(EngineEventID, fmt.Sprintf(format, logs.DiscardZapFields(v ...)...))
 	} else {
 		wsl.srv.log.Error(EngineEventID, format)
 	}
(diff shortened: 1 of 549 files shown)


cmd/ops_agent_windows/run_windows.go
```diff
@@ -150,28 +150,27 @@ type WindowsServiceLogger struct {
...
 func (wsl WindowsServiceLogger) Errorf(format string, v ...any) {
-	if len(v) > 0 {
-		wsl.srv.log.Error(EngineEventID, fmt.Sprintf(format, v...))
-	} else {
-		wsl.srv.log.Error(EngineEventID, format)
-	}
+	wsl.srv.log.Error(EngineEventID, fmt.Sprintf(format, v...))
+}
+
+func (wsl WindowsServiceLogger) Infow(format string, v ...any) {
```
jefferbrecht, line 164: Two things:

1) Why call these e.g. `Infow` and not just `Info`?
2) IIUC the purpose of using `wsl.srv.log.Info(EngineEventID, format)` before is to avoid crashes when the string contains formattable control sequences. The new code just forwards everything into `Infof` which formats everything wholesale, which is a dangerous breaking change.

jefferbrecht, line 164: Two things:

1) Why call these e.g. `Infow` and not just `Info`?
2) IIUC the purpose of using `wsl.srv.log.Info(EngineEventID, format)` before is to avoid crashes when the string contains formattable control sequences. The new code just forwards everything into `Infof` which formats everything wholesale, which is a dangerous breaking change.

All of the changes in this file seem unnecessary... What problem are these changes trying to solve?

jefferbrecht (changes requested):

cmd/ops_agent_windows/run_windows.go
franciscovalentecastro: > What problem are these changes trying to solve?
The main goal of adding `Infow` to the `StructuredLogger` interface is to being able to add `Code` as a separate field to the `health-checks.log` (see `healthchecks.go`). More details below : 

1. [Infow](https://pkg.go.dev/go.uber.org/zap#SugaredLogger.Infow) is a method from zap structured logger that adds additional fields ('w'ith metadata) directly by calling the logger, e.g. `Infow("message", "code", "MonApi...")`. Adding `Infow` here is to follow the interface and for loggers that that don't have the `adding Keys And Values` capability, we drop the parameters.

2. This `WindowsServiceLogger` logger is only used by the Health Checks methods, and needs to follow the `StructuredLogger` interface. At some point we can (or not) update all logs in the windows service to be using this structured logger.

> All of the changes in this file seem unnecessary
I updated the signature of the methods so it's clearer the goal of `Infow`, and avoided to redirect everything to `Infof` since that is dangerous. I also moved all the `WindowsServiceLogger` logic to `logs/windows.go`.

internal/healthchecks/api_check.go
```diff
@@ -108,7 +108,7 @@ func runLoggingCheck(logger logs.StructuredLogger) error {
 	logger.Infof("logging client was created successfully")
 
 	if err := logClient.Ping(ctx); err != nil {
-		logger.Println(err)
+		logger.Errorf(err.Error())
```
rafaelwestphal, line 111: Here we are producing an log with severity=ERROR that doesn't have a code. Is that something we really want? We can just warn here or infof since we don't have a code for this error.

confgenerator/self_logs.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func generateSelfLogsComponents(ctx context.Context, userAgent string) []fluentbit.Component {
+	out := make([]fluentbit.Component, 0)
+
+	out = append(out, generateFluentBitSelfLogsComponents(ctx)...)
+	out = append(out, generateHealthChecksLogsComponents(ctx)...)
+
+	healthLabels := structuredHealthLogsLabels()
+	out = append(out, stackdriverOutputComponent(healthLogsTag, userAgent, "", healthLabels))
```
rafaelwestphal, line 135: minor nit: we don't need the intermediate variable here

confgenerator/logging.go
```diff
@@ -45,7 +45,7 @@ func setLogNameComponents(ctx context.Context, tag, logName, receiverType string
 }
 
 // stackdriverOutputComponent generates a component that outputs logs matching the regex `match` using `userAgent`.
-func stackdriverOutputComponent(match, userAgent string, storageLimitSize string) fluentbit.Component {
+func stackdriverOutputComponent(match, userAgent string, storageLimitSize string, labels string) fluentbit.Component {
```
jefferbrecht, line 48: Nit: `labels` is logically a `map[string]string`, not a `string`, so I'd rather have the parameter be encoding-agnostic and let the function flatten everything to a string internally instead.

While we're at it, can you update the function header to briefly document `storageLimitSize` and `labels`?

integration_test/ops_agent_test.go
```diff
@@ -3001,8 +3001,14 @@ func TestLoggingSelfLogs(t *testing.T) {
...
 			t.Error(err)
 		}
+		
+		pkgVersion, err := agents.FetchPackageVersions(ctx, logger.ToMainLog(), vm, []agents.AgentPackage{agents.OpsPackage});
+		if err != nil {
+			t.Error(err)
+		}
 
-		if err := gce.WaitForLog(ctx, logger.ToMainLog(), vm, "ops-agent-health-checks", time.Hour, `severity="INFO"`); err != nil {
+		query := fmt.Sprintf(`severity="INFO" AND labels."agent.googleapis.com/health/agentKind"="ops-agent" AND labels."agent.googleapis.com/health/agentVersion"="%s" AND labels."agent.googleapis.com/health/schemaVersion"="v1"`, pkgVersion[agents.OpsPackage])
```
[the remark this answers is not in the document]
franciscovalentecastro: I modified the `TestPortsAndAPIHealthChecks` to also query for the `MonApiScopeErr` code.

internal/logs/logs.go
```diff
@@ -142,6 +140,18 @@ func (f ZapStructuredLogger) Errorf(format string, v ...any) {
 	f.logger.Errorf(format, v...)
 }
 
+func (f ZapStructuredLogger) Infow(format string, v ...any) {
```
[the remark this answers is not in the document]
franciscovalentecastro: Added comment.

confgenerator/self_logs.go
```diff
@@ -0,0 +1,158 @@
...
+			{"debug", "DEBUG"},
+			{"error", "ERROR"},
+			{"info", "INFO"},
+			{"warn", "WARNING"},
+		})...,
+	)
+	return out
+}
+
+func generateHealthLogsComponent(ctx context.Context) []fluentbit.Component {
```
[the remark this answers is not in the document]
franciscovalentecastro: Added comment.

confgenerator/logging.go
franciscovalentecastro: After debugging bunch of `Fluent-bit + Windows` test failures I realized I couldn't have 3 strackdriver (1 for customer logs, 1 for fluent-bit logs, 1 for health logs) outputs in the config (this seems to be a fluent-bit bug in windows related to threading, synced with @braydonk ).

I opted instead to use a `Modify` filter to add the labels. Some details on the options on adding labels to `ops-agent-health` logs:
- `Output Stackdriver Plugin` : 
    - The only way to add labels in the output plugin without breaking Windows would be to add them for both `ops-agent-health` and `ops-agent-fluent-bit` logs.
    - Adding labels in the output plugin is probably faster that using a `modify` filter.
- `Modify Filter` : 
    - To avoid changing in any way the `ops-agent-fluent-bit` logs, I opted to add all labels from a modify filter.
    - The filter was already being used for setting `severity` and `sourceLocation`.

internal/logs/logs.go
```diff
@@ -142,6 +140,18 @@ func (f ZapStructuredLogger) Errorf(format string, v ...any) {
 	f.logger.Errorf(format, v...)
 }
 
+func (f ZapStructuredLogger) Infow(msg string, keysAndValues ...any) {
```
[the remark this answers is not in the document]
franciscovalentecastro: Added comments in the interface and next to every implementation of `Infow` about it's usage.

confgenerator/self_logs.go
franciscovalentecastro: Remove the addition of labels from the output plugin.

confgenerator/logging.go
franciscovalentecastro: Update : Rewrote some details of last comment as per changes of last commits.

franciscovalentecastro: Synced offline with @jefferbrecht , @rafaelwestphal , using `Ops Agent - Modify Fields` for the labels.

rafaelwestphal (approved):

PR #1306, opened 2023-06-27
Use T2A machine type for ARM tests

jefferbrecht:
## Description
ARM distros need to run on the T2A machine series. We should have enough quota to use the same (CPU, RAM) configuration as E2.

## Related issue
b/274913389

## How has this been tested?
Will let the presubmits run.

## Checklist:
- Unit tests
  - [x] Unit tests do not apply.
  - [ ] Unit tests have been added/modified and passed for this PR.
- Integration tests
  - [ ] Integration tests do not apply.
  - [x] Integration tests have been added/modified and passed for this PR.
- Documentation
  - [x] This PR introduces no user visible changes.
  - [ ] This PR introduces user visible changes and the corresponding documentation change has been made.
- Minor version bump
  - [x] This PR introduces no new features.
  - [ ] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.
  - [ ] This PR bumps the version.

PR #1309, opened 2023-06-28
Fix ARM integration tests

jefferbrecht:
## Description
This fixes the failing tests from #1306:

* Failing Ops Agent tests:
  * ~TestPrometheusMetricsWithJSONExporter/debian-11-arm64~ (fixed)
* Failing 3P tests:
  * ~TestThirdPartyApps/debian-11-arm64/mysql~ (skipped in favour of mariadb)
  * ~TestThirdPartyApps/debian-11-arm64/mysql5.7~ (skipped)
  * ~TestThirdPartyApps/debian-11-arm64/mongodb3.6~ (skipped)
  * ~TestThirdPartyApps/debian-11-arm64/cassandra~ (fixed)
  * ~TestThirdPartyApps/debian-11-arm64/mongodb~ (fixed)
  * ~TestThirdPartyApps/debian-11-arm64/couchbase~ (skipped)
  * ~TestThirdPartyApps/debian-11-arm64/oracledb~ (skipped)
* WIP:
  * TestThirdPartyApps/debian-11-arm64/mariadb

## Related issue
b/274913389

## How has this been tested?
Will let presubmits run.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


quentinmit:
Debian 11 ships native MySQL on ARM, why aren't we testing that?

https://packages.debian.org/bullseye/mariadb-server

jefferbrecht:
> Debian 11 ships native MySQL on ARM, why aren't we testing that?
> 
> https://packages.debian.org/bullseye/mariadb-server

I definitely forgot about mariadb and I can work on adding this.

integration_test/third_party_apps_data/applications/cassandra/debian_ubuntu/install
```diff
@@ -3,12 +3,17 @@ set -e
...
+if [[ "${VERSION_ID}" == 23* || "$(uname -m)" == aarch64 ]]; then
     # There is a python dependency that fails for any older versions of cassandra on Ubuntu 23.04, so explicitly install Cassandra 4.1x.
+    # Additionally, cassandra doesn't have arm64 binaries for 2.2x, but does for 4.1x.
     echo "deb https://debian.cassandra.apache.org 41x main" | sudo tee -a /etc/apt/sources.list.d/cassandra.sources.list
 
     sudo apt update
-    sudo apt install -y openjdk-11-jdk cassandra
+
+    # cassandra doesn't run on Java 17, but on some platforms (debian-11-arm64) it has a transitive dependency on Java 17 (???).
+    # Installing openjdk-11-jdk first ensures that it takes precedence over Java 17 on PATH.
```
quentinmit, line 14: This isn't supposed to be how Debian handles conflicting binary names... each package that provides a binary (like `java`) has a priority, and every time you install or remove one of those packages it chooses the highest-priority version out of the currently installed ones.

Are you sure it actually has a transitive dependency on Java 17, and not just a dependency on a JRE that apt is resolving to Java 17 by default? If it's the latter, you can just specify both package names on a single `apt install`.

integration_test/third_party_apps_data/applications/couchbase/metadata.yaml
```diff
@@ -199,8 +199,10 @@ minimum_supported_agent_version:
 supported_operating_systems: linux
 platforms_to_skip:
   # couchbase is not supported on rhel9, Debian 11, Debian 12, or Ubuntu 23.04.
```
quentinmit, line 201: This comment is no longer correct; Debian 11 is now supported.

integration_test/third_party_apps_data/applications/mongodb/debian_ubuntu/install
```diff
@@ -2,17 +2,27 @@ set -e
 
 sudo apt-get update
 sudo apt-get install -y gnupg wget
-wget --no-verbose --output-document=- https://www.mongodb.org/static/pgp/server-5.0.asc | sudo apt-key add -
+wget --no-verbose --output-document=- https://www.mongodb.org/static/pgp/server-6.0.asc | sudo apt-key add -
```
quentinmit, line 5: FWIW `apt-key add` is deprecated and will be removed in Debian 12. You're supposed to just place the file in `/etc/apt/trusted.gpg.d/` directly instead (i.e. pass `--output-document=/etc/apt/trusted.gpg.d/mongodb.asc` to `wget`).

integration_test/third_party_apps_data/applications/cassandra/debian_ubuntu/install
jefferbrecht: I'm not a fan of this code either, but putting them in a single `apt install` simply doesn't work. Cassandra just picks whatever JRE is on PATH, which ends up being Java 17 when installed this way, which it doesn't support:

```
$ sudo apt install -y openjdk-11-jdk cassandra
$ /usr/sbin/cassandra
OpenJDK 64-Bit Server VM warning: Option UseBiasedLocking was deprecated in version 15.0 and will likely be removed in a future release.
Unrecognized VM option 'UseConcMarkSweepGC'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
$ java --version
openjdk 17.0.7 2023-04-18
OpenJDK Runtime Environment (build 17.0.7+7-Debian-1deb11u1)
OpenJDK 64-Bit Server VM (build 17.0.7+7-Debian-1deb11u1, mixed mode, sharing)
```

If I do `apt depends cassandra` then I get:

```
$ sudo apt depends cassandra                                                                                                                       
cassandra                                                                                                                                                                   
 |Depends: <openjdk-8-jre-headless>                                                                                                                                         
 |Depends: <java8-runtime>                                                                                                                                                  
    default-jre                                                                                                                                                             
    openjdk-11-jre                                                                                                                                                          
    openjdk-17-jre                                                                                                                                                          
 |Depends: openjdk-11-jre-headless
  Depends: <java11-runtime>
    default-jre
    openjdk-11-jre
    openjdk-17-jre
...
```

Now, if I do `apt install openjdk-11-jre-headless cassandra` instead, _then_ it works. So there's something incorrect about using `openjdk-11-jdk` I guess?

integration_test/third_party_apps_data/applications/couchbase/metadata.yaml
jefferbrecht: Removed `debian-11` from skipped platforms.

integration_test/third_party_apps_data/applications/mongodb/debian_ubuntu/install
jefferbrecht: Removed all occurrences of `apt-get add`.

apps/mysql.go
```diff
@@ -284,10 +305,32 @@ func (p LoggingProcessorMysqlGeneral) Components(ctx context.Context, tag string
...
+							"tid": "integer",
+						},
+					},
+				},
+				{
+					// MariaDB uses the same timestamp format here as old versions do for the error log:
+					// https://mariadb.com/kb/en/error-log/#format
+					// Sample line: 230707  1:41:38     40 Query    select table_catalog, table_schema, table_name from information_schema.tables
+					// Sample line:                      5 Connect  root@localhost on  using Socket
+					// (When a timestamp is present, it is followed by a single tab character.
```
jefferbrecht, line 325:
```suggestion
					// When a timestamp is present, it is followed by a single tab character.
```

apps/mysql.go
```diff
@@ -299,12 +342,20 @@ func (p LoggingProcessorMysqlGeneral) Components(ctx context.Context, tag string
 			{
 				StateName: "start_state",
 				NextState: "cont",
-				Regex:     `\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d+Z`,
+				Regex: fmt.Sprintf(
+					`^(%s|%s|\t\t)`,
```
jefferbrecht, line 346: Does `timeRegexMariaDBNew` need to be included in here too? Or are those logs not multiline?

apps/mysql.go
```diff
@@ -358,74 +414,218 @@ func (p LoggingProcessorMysqlSlow) Components(ctx context.Context, tag string, u
...
 			{
 				StateName: "start_state",
+				NextState: "comment",
+				Regex: fmt.Sprintf(
+					`^# (User@Host: |Time: (%s|%s))`,
+					timeRegexMySQLNew,
+					timeRegexOld,
+				),
+			},
+			// Explicitly consume the next line, which might be User@Host.
```
jefferbrecht, line 607: I assume it's not possible for `User@Host` to be the only line in an entire entry then? We'll always have a "next line" to consume?

integration_test/third_party_apps_data/applications/mariadb/debian_ubuntu/install
```diff
@@ -0,0 +1,44 @@
...
+# mysql_error receiver doesn't work with the default MariaDB configuration on
+# Debian 11.
+sudo tee /etc/mysql/mariadb.conf.d/99-replicas.cnf >/dev/null <<EOF
+[mysqld]
+general-log
+slow-query-log
+log-slow-verbosity = query_plan,innodb
+long-query-time = 0
+log_error
+
```
jefferbrecht, line 25: Nit:

```suggestion
```

jefferbrecht (commented):
LGTM, just minor comments.

We'll need to remove the do-not-submit from https://github.com/GoogleCloudPlatform/ops-agent/pull/1309/commits/dfb8c3f793ec30b785bbd3a0d6abd36a9f25a22f before merging, but let's wait for the last round of tests to finish first since I pushed some changes.

igorpeshansky (commented):
Some preliminary comments…

integration_test/third_party_apps_data/applications/mongodb/centos_rhel/install
```diff
@@ -1,25 +1,26 @@
 set -e
 
-cat << EOF > mongodb-org-5.0.repo
```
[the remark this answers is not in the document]
jefferbrecht: IIRC MongoDB 5 isn't available on Debian 11 ARM, so for simplicity I just upgraded all the tests to 6.0. This also unlocks testing on RHEL 9 which had no coverage before.

I don't want to keep piling more scope onto this PR so I'd rather not start a `mongodb5.0` test set here. If MongoDB 5 testing on RHEL is still valuable then perhaps we can special-case it so that 6.0 gets installed on RHEL 9 but 5.0 gets installed on other distros, as an interim measure while we work on a separate PR to add a `mongodb5.0` test set? WDYT?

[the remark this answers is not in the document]
igorpeshansky: I'll leave it to you and ObservIQ (@dehaansa?) to determine how important it actually is to have 5.0 in our testing matrix. FWIW, I wasn't suggesting special-casing RHEL 9. It should be fairly straightforward to copy the old `mongodb` directory into `mongodb5.0` and exclude both `debian-11-arm64` and `rocky-linux-9` via `platforms_to_skip`, but we could do that in a separate PR if we decide it's worth it.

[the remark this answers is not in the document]
dehaansa: Are there significant breaking changes between 5.0 and 6.0? If not, I don't see a requirement to continue to test 5.0, we aren't attempting to cover the full scope of supported versions in integration tests for other integrations, just versions where we encountered significant issues or know of breaking changes between versions. That said, if we want to add it as another supported integration to maintain support we can add 5.0 in the described way.

integration_test/testdata/prometheus/setup_json_exporter.sh
```diff
@@ -1,8 +1,17 @@
...
 # Download JSON exporter and extract to /opt/
+unset ARCH
+case "$(uname -m)" in
+  "x86_64")
+    ARCH="amd64"
+    ;;
+  "aarch64")
+    ARCH="arm64"
+    ;;
+esac
```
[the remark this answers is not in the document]
jefferbrecht: `dpkg` isn't portable though, so that simplicity would be offset by the complexity of finding equivalents on other distros...

The map version is nice because it's still portable yet simpler than what I had already, so I'll go with that for now unless there's an even simpler alternative that works on all distros.

integration_test/third_party_apps_data/applications/couchbase/debian_ubuntu/install
```diff
@@ -9,7 +9,7 @@ curl -O https://packages.couchbase.com/releases/couchbase-release/couchbase-rele
 sudo dpkg -i couchbase-release-1.0-amd64.deb
```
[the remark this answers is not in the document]
jefferbrecht: Yup, I'll change this to `${ARCH}`. And since this is the `debian_ubuntu` installer then I can use `$(dpkg --print-architecture)` too.

integration_test/testdata/prometheus/setup_json_exporter.sh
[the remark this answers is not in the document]
igorpeshansky: You're right, I'm so used to Debian using its own architecture constants that I didn't notice that this script is cross-distro.

integration_test/third_party_apps_data/applications/mongodb/centos_rhel/install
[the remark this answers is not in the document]
igorpeshansky: Good point. I don't see anything in the [documentation](https://www.mongodb.com/docs/manual/release-notes/6.0-compatibility/) that could be potentially related to the way we collect metrics or logs, so maybe switching to 6.0 is not as big of a deal as I initially thought. Let's hold off on adding 5.0-specific testing until we discover breakage that wasn't caught by these.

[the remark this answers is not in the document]
jefferbrecht: SGTM 👍

confgenerator/logging_modify_fields.go
```diff
@@ -193,6 +193,12 @@ function process(tag, timestamp, record)
 			conv = "math.floor(tonumber(v))"
 		case "float":
 			conv = "tonumber(v)"
+		case "YesNoBoolean":
```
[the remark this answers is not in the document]
jefferbrecht: No, the validation rules on `ModifyField.Type` prevent users from configuring this. It wouldn't be visible in any user-facing configurations.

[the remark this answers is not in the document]
igorpeshansky: Want to add a comment to that effect, e.g.: `// internal-only type; external use blocked by field validation` or something?

integration_test/third_party_apps_data/applications/mariadb/debian_ubuntu/install
```diff
@@ -0,0 +1,44 @@
...
+# information on how to configure multiple mysql instances on the same
+# machine.
+
+# set up replication source to validate replica metrics
+# set main target (replica) to use binary logging for replication to work
+
+# TODO: MariaDB is configured to use journald by default on at least Debian 11;
+# log_error forces the log to be written to a file instead, but this means the
+# mysql_error receiver doesn't work with the default MariaDB configuration on
+# Debian 11.
```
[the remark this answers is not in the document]
jefferbrecht: Added a note to `metadata.yaml`.

apps/mysql.go
jefferbrecht: Nevermind, looks like `timeRegexMariaDBNew` only applies to the error log.

integration_test/third_party_apps_test.go
```diff
@@ -463,7 +463,7 @@ func runLoggingTestCases(ctx context.Context, logger *logging.DirectoryLogger, v
 	return err
 }
 
-func runMetricsTestCases(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM, metrics []*metadata.ExpectedMetric, fc *feature_tracking_metadata.FeatureTrackingContainer) error {
+func runMetricsTestCases(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM, app string, metrics []*metadata.ExpectedMetric, fc *feature_tracking_metadata.FeatureTrackingContainer) error {
```
igorpeshansky, line 466: I don't see `app` being used anywhere. Am I missing something?

jefferbrecht: Forgot to remove this in 478d040, will fix.

igorpeshansky (approved):
LGTM :shipit:

apps/mysql.go
quentinmit: Correct, there's always at least one (and usually several) comment lines, and then at least one non-comment line before the next log line.

PR #1310, opened 2023-06-28
Combine unit tests into a single directory

quentinmit:
## Description
We previously had a separate set of unit tests for each platform; now that we're adding platforms (3 already, a 4th about to be added for ARM) it's starting to become unwieldy. Instead, this PR combines all the unit tests into a single directory, (attempting to) run every unit test on every platform. The Windows-only tests will run on Linux too, generating the corresponding error message (thus obviating the need for the previous `invalid` tests to make sure they don't work on Linux), etc.

To preserve the existing behavior of making sure that `invalid` tests always fail, if the test name begins with `invalid-` the code will expect that the test should fail on every platform.

## Related issue
b/274913389 and b/283086085

## How has this been tested?

Unit tests continue to pass. I carefully moved the existing golden files into place and verified the diffs by hand before using `go test . -update` to generate the missing golden files.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/confgenerator_test.go
```diff
@@ -109,71 +107,29 @@ var (
...
+				pc := pc
+				t.Run(pc.name, func(t *testing.T) {
+					testDir := filepath.Join(goldensDir, testName)
+					got, err := generateConfigs(pc, testDir)
+					if strings.HasPrefix(testName, "invalid-") {
+						assert.Assert(t, err != nil, "expected test config to be invalid, but was successful")
+					}
+					if testName != "builtin" {
+						delete(got, builtinConfigFileName)
+					}
```
jefferbrecht, line 128: Is this still needed?

quentinmit: Yes, this is needed to avoid having a duplicate golden built-in config file in every single test directory.

jefferbrecht (approved):

PR #1320, opened 2023-07-06
confgenerator: redact passwords from 3P integrations

ridwanmsharif:
## Description
This change replaces all the Passwords to use the `secret.String` type so when the config is logged, it doesn't leak sensitive data.

## Related issue
[b/290250205](http://b/290250205) | OracleDB Metrics integration logs password in plain text

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/apps/activemq.go
+++ b/apps/activemq.go
@@ -17,14 +17,15 @@ package apps
 import (
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
+	"github.com/GoogleCloudPlatform/ops-agent/internal/secret"
 )
 
 type MetricsReceiverActivemq struct {
 	confgenerator.ConfigComponent `yaml:",inline"`
 
-	Endpoint                               string `yaml:"endpoint" validate:"omitempty,hostname_port|startswith=service:jmx:"`
-	Username                               string `yaml:"username" validate:"required_with=Password"`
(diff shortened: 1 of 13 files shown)


rafaelwestphal (approved):

apps/activemq.go
```diff
@@ -17,14 +17,15 @@ package apps
...
 
 type MetricsReceiverActivemq struct {
 	confgenerator.ConfigComponent `yaml:",inline"`
 
-	Endpoint                               string `yaml:"endpoint" validate:"omitempty,hostname_port|startswith=service:jmx:"`
-	Username                               string `yaml:"username" validate:"required_with=Password"`
-	Password                               string `yaml:"password" validate:"required_with=Username"`
+	Endpoint                               string        `yaml:"endpoint" validate:"omitempty,hostname_port|startswith=service:jmx:"`
+	Username                               string        `yaml:"username" validate:"required_with=Password"`
+	Password                               secret.String `yaml:"password" validate:"required_with=Username"`
```
igorpeshansky, line 28: Are these fields ignored?

ridwanmsharif: I think so, I put up https://github.com/GoogleCloudPlatform/ops-agent/pull/1322 to get rid of them. Thought it was best to split that out to its own thing since its unrelated to the redaction. I think this is a bug - even when the username and password are currently specified, these ignored fields are set and the jmx struct ones are not. And as a result, the generated otel config doesn't actually use them.

PR #1325, opened 2023-07-10
Add Support for Configuring Simulacra using YAML + Change stdout logger to Directory Logger

shafinsiddique:
## Description

Currently, Simulacra can only be configured using command line arguments. This is not extensible in the long run because 
a) our configurations are only saved as Unix commands and b) as our requirements for our VMs grow, our commands become extremely long and easy to mess up. 

This PR adds support for Simulacra reading configurations from YAML files. If the user wants to pass in Simulacra configs using YAML, they can now just run simulacra binary with the --config argument set to the path of the YAML file. 

This is optional. You can still run Simulacra without creating any YAML files. However, if the --config argument is specified, then only configurations from the YAML file are used and all other command line arguments are ignored. 

This PR also changes the logger used in Simulacra from a Stdout Logger to a Directory Logger. More explained below. 

#
Here's an example of how to use Simulacra with YAMl. 

To run Simulacra with YAML config, create a local YAML file with the following content: 

`config_file_path: >-
  /usr/local/google/home/shafinsiddique/ops-agent/integration_test/simulacra/config.yaml
name: test-instance-simulacra
`

Then we can run the following command to run simulacra 

`go run -tags=integration_test "./cmd/simulacra/simulacra.go" --config "./config.yaml"`

We can also run Simulacra without YAML config: 

`go run -tags=integration_test "./cmd/simulacra/simulacra.go" --ops_agent_config "/usr/local/google/home/shafinsiddique/ops-agent/integration_test/simulacra/config.yaml"`

## Changes to Logger

This PR changes the logger SImulacra is using from a Stdout logger to a Directory Logger. This decision was made because we believe it'll be cleaner if there are separate files for logs. For example, one file for the logs from VM initialization and another for the logs from the installation commands. We still log the errors to Stdout so that the user immediately knows and does not have to tail any files. 

## How has this been tested?

- Ran the commands described above and inspected the logs as it installed all the required applications. This is an example VM that was created using SImulacra; https://pantheon.corp.google.com/compute/instancesDetail/zones/us-central1-a/instances/simulacra-vm-instance-50aafdf1-8ce3-4966-bacd-1cc52c0855d8?project=experimental-braydonk. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/cmd/simulacra/simulacra.go
+++ b/cmd/simulacra/simulacra.go
@@ -30,9 +30,34 @@ import (
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
 	"github.com/GoogleCloudPlatform/ops-agent/integration_test/agents"
 	"github.com/GoogleCloudPlatform/ops-agent/integration_test/gce"
+	"github.com/GoogleCloudPlatform/ops-agent/integration_test/logging"
 	"github.com/binxio/gcloudconfig"
 	"github.com/google/uuid"
+	"gopkg.in/yaml.v2"
 )
+
+const (
+	defaultPlatform           = "debian-11"
+	defaultThirdPartyAppsPath = "./integration_test/third_party_apps_data"
+	vmInitLogFile             = "vm_initialization.txt"
+)
+
+// Config represents the configuration for Simulacra. Most of the fields specify requirements about the VM that
+// Simulacra will instantiate.
+type Config struct {
+	// The OS for the VM.
+	Platform string `yaml:"platform"`
+	// Path to the Ops Agent Config File.
+	ConfigFilePath string `yaml:"config_file_path"`
+	// The Project Simulacra will be using to instantiate the VM.
+	Project string `yaml:"project"`
+	// Zone for the VM.
+	Zone string `yaml:"zone"`
+	// Name for the VM.
+	Name string `yaml:"name"`
+	// Path to Third Party Apps folder
+	ThirdPartyAppsPath string `yaml:"third_party_apps_path"`
+}
 
 func distroFolder(platform string) (string, error) {
 	if gce.IsWindows(platform) {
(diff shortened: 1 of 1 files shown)


cmd/simulacra/simulacra.go
```diff
@@ -178,43 +154,100 @@ func configureFromGCloud(project *string, zone *string) error {
...
+	vm, err := gce.CreateInstance(ctx, logger.ToFile(vmInitLogFile), options)
 	if err != nil {
-		logger.Fatalf("Failed to create GCE instance %v", err)
+		log.Default().Fatalf("Failed to create GCE instance %v", err)
 	}
 	// Install Ops Agent on VM.
-	if err := setupOpsAgent(ctx, vm, logger, *configFile); err != nil {
-		logger.Fatalf("Failed to install Ops Agent %v", err)
+	log.Default().Print("Installing Ops Agent, check main_log.txt for details")
+	if err := setupOpsAgent(ctx, vm, logger.ToMainLog(), config.ConfigFilePath); err != nil {
```
shafinsiddique, line 239: We can put logs for Ops Agent installation in a separate file but because there's not much that usually arises i figured it's best to just keep it to main_log.txt

cmd/simulacra/simulacra.go
```diff
@@ -149,11 +124,12 @@ func installApps(ctx context.Context, vm *gce.VM, logger *log.Logger, configFile
 
 	for _, app := range receivers {
 		if scriptContent, err := os.ReadFile(path.Join(installPath, "applications", app, folder, "install")); err == nil {
-			logger.Printf("Installing %s to VM", app)
-			if _, err := runScriptRemotely(ctx, logger, vm, string(scriptContent), make(map[string]string)); err != nil {
+			logger.ToMainLog().Printf("Installing %s to VM", app)
+			log.Default().Printf("Installing %s to VM", app)
+			if _, err := gce.RunScriptRemotely(ctx, logger, vm, string(scriptContent), nil, make(map[string]string)); err != nil {
```
shafinsiddique, line 129: gce.RunScriptRemotely() logs everything to main_log.txt by default so all our installation logs have to go there.

shafinsiddique:
Left some comments where I could use some insight on whether it is the best course of action

PR #1326, opened 2023-07-10
Add support for configuring Simulacra using YAML + Change Stdout logger to Directory Logger

shafinsiddique:
## Description

Currently, Simulacra can only be configured using command line arguments. This is not extensible in the long run because 
a) our configurations are only saved as Unix commands and b) as our requirements for our VMs grow, our commands become extremely long and easy to mess up. 

This PR adds support for Simulacra reading configurations from YAML files. If the user wants to pass in Simulacra configs using YAML, they can now just run simulacra binary with the --config argument set to the path of the YAML file. 

This is optional. You can still run Simulacra without creating any YAML files. However, if the --config argument is specified, then only configurations from the YAML file are used and all other command line arguments are ignored. 

This PR also changes the logger used in Simulacra from a Stdout Logger to a Directory Logger. More explained below. 

#
Here's an example of how to use Simulacra with YAMl. 

To run Simulacra with YAML config, create a local YAML file with the following content: 

`config_file_path: >-
  /usr/local/google/home/shafinsiddique/ops-agent/integration_test/simulacra/config.yaml
name: test-instance-simulacra
`

Then we can run the following command to run simulacra 

`go run -tags=integration_test "./cmd/simulacra/simulacra.go" --config "./config.yaml"`

We can also run Simulacra without YAML config: 

`go run -tags=integration_test "./cmd/simulacra/simulacra.go" --ops_agent_config "/usr/local/google/home/shafinsiddique/ops-agent/integration_test/simulacra/config.yaml"`

## Changes to Logger

This PR changes the logger SImulacra is using from a Stdout logger to a Directory Logger. This decision was made because we believe it'll be cleaner if there are separate files for logs. For example, one file for the logs from VM initialization and another for the logs from the installation commands. We still log the errors to Stdout so that the user immediately knows and does not have to tail any files. 

## How has this been tested?

- Ran the commands described above and inspected the logs as it installed all the required applications. This is an example VM that was created using SImulacra; https://pantheon.corp.google.com/compute/instancesDetail/zones/us-central1-a/instances/simulacra-vm-instance-50aafdf1-8ce3-4966-bacd-1cc52c0855d8?project=experimental-braydonk. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/cmd/simulacra/simulacra.go
+++ b/cmd/simulacra/simulacra.go
@@ -135,7 +135,7 @@ func installApps(ctx context.Context, vm *gce.VM, logger *logging.DirectoryLogge
 			logger.ToMainLog().Printf("Installing %s to VM", app)
 			log.Default().Printf("Installing %s to VM", app)
 			if _, err := gce.RunScriptRemotely(ctx, logger, vm, string(scriptContent), nil, make(map[string]string)); err != nil {
-				return err
+				return fmt.Errorf("Failed to install app %s %v", app, err)
 			}
 			logger.ToMainLog().Printf("Done Installing %s", app)
 
@@ -225,7 +225,7 @@ func runCustomScripts(ctx context.Context, vm *gce.VM, logger *logging.Directory
 			logger.ToMainLog().Printf("Running script from %s", scriptPath)
 			log.Default().Printf("Running script from %s", scriptPath)
 			if _, err := gce.RunScriptRemotely(ctx, logger, vm, string(scriptContent), nil, make(map[string]string)); err != nil {
-				return err
+				return fmt.Errorf("Script with path %s failed to run %v", scriptPath, err)
 			}
 			logger.ToMainLog().Printf("Done Running Script from  %s", scriptPath)
 
--- a/cmd/simulacra/simulacra.go
+++ b/cmd/simulacra/simulacra.go
@@ -57,6 +57,8 @@ type Config struct {
 	Name string `yaml:"name"`
 	// Path to Third Party Apps folder
 	ThirdPartyAppsPath string `yaml:"third_party_apps_path"`
+	// Path to script files that will be run on the VM.
+	Scripts []string `yaml:"scripts"`
 }
 
 func distroFolder(platform string) (string, error) {
@@ -123,7 +125,9 @@ func installApps(ctx context.Context, vm *gce.VM, logger *logging.DirectoryLogge
 	receivers := getAllReceivers(config)
 
 	for _, app := range receivers {
-		if scriptContent, err := os.ReadFile(path.Join(installPath, "applications", app, folder, "install")); err == nil {
+		if scriptContent, err := os.ReadFile(path.Join(installPath, "applications", app, folder, "install")); err != nil {
+			return err
+		} else {
 			logger.ToMainLog().Printf("Installing %s to VM", app)
 			log.Default().Printf("Installing %s to VM", app)
 			if _, err := gce.RunScriptRemotely(ctx, logger, vm, string(scriptContent), nil, make(map[string]string)); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/simulacra/simulacra.go
```diff
@@ -178,43 +154,100 @@ func configureFromGCloud(project *string, zone *string) error {
...
+	vm, err := gce.CreateInstance(ctx, logger.ToFile(vmInitLogFile), options)
 	if err != nil {
-		logger.Fatalf("Failed to create GCE instance %v", err)
+		log.Default().Fatalf("Failed to create GCE instance %v", err)
 	}
 	// Install Ops Agent on VM.
-	if err := setupOpsAgent(ctx, vm, logger, *configFile); err != nil {
-		logger.Fatalf("Failed to install Ops Agent %v", err)
+	log.Default().Print("Installing Ops Agent, check main_log.txt for details")
+	if err := setupOpsAgent(ctx, vm, logger.ToMainLog(), config.ConfigFilePath); err != nil {
```
shafinsiddique, line 239: the logs for ops agent installation are only a few lines so keeping them to main_log.txt as it's probably unergonomic for the user to read several files.

cmd/simulacra/simulacra.go
```diff
@@ -149,11 +124,12 @@ func installApps(ctx context.Context, vm *gce.VM, logger *log.Logger, configFile
 
 	for _, app := range receivers {
 		if scriptContent, err := os.ReadFile(path.Join(installPath, "applications", app, folder, "install")); err == nil {
-			logger.Printf("Installing %s to VM", app)
-			if _, err := runScriptRemotely(ctx, logger, vm, string(scriptContent), make(map[string]string)); err != nil {
+			logger.ToMainLog().Printf("Installing %s to VM", app)
+			log.Default().Printf("Installing %s to VM", app)
+			if _, err := gce.RunScriptRemotely(ctx, logger, vm, string(scriptContent), nil, make(map[string]string)); err != nil {
```
shafinsiddique, line 129: gce.RunScriptRemotely can only write to main_log.txt. So all our installation logs have to go there.

cmd/simulacra/simulacra.go
```diff
@@ -178,43 +158,127 @@ func configureFromGCloud(project *string, zone *string) error {
...
-		logger.Fatalf("Failed to install Ops Agent %v", err)
+	log.Default().Print("Installing Ops Agent, check main_log.txt for details")
+	if err := setupOpsAgent(ctx, vm, logger.ToMainLog(), config.ConfigFilePath); err != nil {
+		log.Default().Fatalf("Failed to install Ops Agent %v", err)
 	}
 
 	// Install Third Party Appliations based on Ops Agent Config.
-	if err := installApps(ctx, vm, logger, *configFile, *thirdPartyAppsPath); err != nil {
-		logger.Fatalf("Failed to install apps %v", err)
+	if len(config.ConfigFilePath) > 0 {
```
ridwanmsharif, line 265: Why did we need this condition? If we have the default config be used (equivalent to no config specified), we should still run the scripts (installing the apps is a noop since none are detected in the config)

cmd/simulacra/simulacra.go
```diff
@@ -178,43 +158,127 @@ func configureFromGCloud(project *string, zone *string) error {
...
+	log.Default().Printf("Creating VM Instance, check %s for details", vmInitLogFile)
+	vm, err := gce.CreateInstance(ctx, logger.ToFile(vmInitLogFile), options)
 	if err != nil {
-		logger.Fatalf("Failed to create GCE instance %v", err)
+		log.Default().Fatalf("Failed to create GCE instance %v", err)
 	}
 	// Install Ops Agent on VM.
-	if err := setupOpsAgent(ctx, vm, logger, *configFile); err != nil {
-		logger.Fatalf("Failed to install Ops Agent %v", err)
+	log.Default().Print("Installing Ops Agent, check main_log.txt for details")
```
ridwanmsharif, line 259: Add the full log file path here, if I use simulacra several times, the log files for each are going to separate directories. Just add `loggingDir` to the log here

cmd/simulacra/simulacra.go
```diff
@@ -178,43 +158,127 @@ func configureFromGCloud(project *string, zone *string) error {
...
+		if err := installApps(ctx, vm, logger, config.ConfigFilePath, config.ThirdPartyAppsPath); err != nil {
+			log.Default().Printf("Failed to install apps %v", err)
+		}
+	}
+
+	// Run custom Scripts on the VM
+	if len(config.Scripts) > 0 {
+		log.Default().Print("Running Custom Scripts on the VM, check main_log.txt for details")
+		if err := runCustomScripts(ctx, vm, logger, config.Scripts); err != nil {
+			log.Default().Fatalf("Error executing custom script on the VM %v", err)
```
ridwanmsharif, line 276: Can we give more info on what script failed? Most easily done by having `runCustomScripts` just wrap the error with more info.

ridwanmsharif (commented):
Just a few minor comments. Overall LGTM

cmd/simulacra/simulacra.go
shafinsiddique: Basically what my thinking process was that, if we take out the condition then the function 'installApps' returns with an error "no file with path "" found" (smth like that). Now its not exactly an 'error' since not specifying the config file was intentional. This is why I wanted to only execute the function if configFilePath is actually explicitly specified.

shafinsiddique: oh i see your comment about just short circuiting the function itself, that makes sense :)

braydonk (approved):
This PR has some others stacked on it, so I think you should:
1. Merge this PR
2. Merge the stacked PRs
3. Make a followup addressing the comments you would like to address.

PR #1327, opened 2023-07-10
Add support for BMS machines

alex-basinov:
## Description
This PR adds support for [BMS](https://cloud.google.com/bare-metal/) machines.

The PR also adds a new resource detector. The detector verifies that BMS_PROJECT_ID, BMS_LOCATION and BMS_INSTANCE_ID env variable are set and based on that creates BMSResource struct that implements the Resource interface. Note that this is temporary. Resource detection logic will move to the Otel binary at a later point. That's the main reason why the code goes into a non-master branch. 
If the ops_agent_engine detects that it runs on BMS, it then generates a BMS-specific otel config. The only difference from the standard config is that a) BMS-specific one is missing resourcedetection processor and b) it has a resource attribute processor in order to pass BMS_* env variables to the otel binary. The generated config looks like this:

```
processors:
...
  resource/hostmetrics_3:
    attributes:
    - action: upsert
      key: cloud.platform
      value: gcp_bare_metal_solution
    - action: upsert
      key: cloud.project
      value: <value from $BMS_PROJECT_ID env variable>
    - action: upsert
      key: cloud.region
      value: <value from $BMS_LOCATION env variable>
    - action: upsert
      key: host.id
      value: <value from $BMS_INSTANCE_ID env variable>
```

## Related issue
[b/266788401](http://b/266788401), [b/267347168](http://b/267347168)

## How has this been tested?
Tested manually on a test BMS machine using the following procedure:

1. Built `google_cloud_ops_agent_engine` binary compiled from [this PR](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/641) and then scp'ed it to a remote BMS machine:
   ```
   CGO_ENABLED=0 go build -o /tmp/google_cloud_ops_agent_engine cmd/google_cloud_ops_agent_engine/main.go && scp /tmp/google_cloud_ops_agent_engine  <test-bms-instance>:~/
   ```
2. Ran the binary to generate Otel config file:
   ```
   env BMS_PROJECT_ID=<test-bms-project> BMS_LOCATION=<bms-location> BMS_INSTANCE_ID=<test-bms-instance> ./google_cloud_ops_agent_engine --service=otel --in=config.yaml --out=/tmp/generated_otel_conf/
   ```
3. Then manually ran `otelopscol` and pointed it to the generated config
   ```
   env GOOGLE_APPLICATION_CREDENTIALS=/etc/google/auth/application_default_credentials.json ./otelopscol --config /tmp/generated_otel_conf/otel.yaml
   ```
4. Navigated to the Metrics Explorer in Pantheon and confirmed that the host metrics are available for the `test-bms-instance`

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/testdata/bms/builtin/linux/builtin/golden/b4a0dead382dce7b4fe011d3f59fdb6d.lua
```diff
@@ -0,0 +1,17 @@
...
+  local nestedRecord = {}
+  local parseKey = "message"
+  for k, v in pairs(record) do
+      if k ~= parseKey then
+          nestedRecord[k] = v
+      end
+  end
+
+  local result = {}
+  result[parseKey] = record[parseKey]
```
shakeebshams, line 12: This assumes "message" is in record. Might be good to wrap it around in an error handling clause

confgenerator/testdata/bms/builtin/linux/builtin/golden/f120d4527bd717cab023dbbe5fbdc332.lua
```diff
@@ -0,0 +1,42 @@
...
+local v = __field_0;
+(function(value)
+if record["logging.googleapis.com/labels"] == nil
+then
+record["logging.googleapis.com/labels"] = {}
+end
+record["logging.googleapis.com/labels"]["agent.googleapis.com/log_file_path"] = value
+end)(v)
+local v = __field_1;
+if v == nil then v = "" end;
```
shakeebshams, line 28: there are a few examples of this in this file. could be simplified to using lua's `or` operator

confgenerator/testdata/bms/builtin/linux/builtin/golden/b4a0dead382dce7b4fe011d3f59fdb6d.lua
alex-basinov: I didn't write these lua files. They were auto-generated by "go test -mod=mod github.com/GoogleCloudPlatform/ops-agent/confgenerator -update -v" command. They appear to be identical to the ones generated for GCE instances. Considering this, I think it would be more appropriate to address your concerns within the component responsible for their generation (I don't know what component it is though).

confgenerator/testdata/bms/builtin/linux/builtin/golden/f120d4527bd717cab023dbbe5fbdc332.lua
alex-basinov: The same applies here. I didn't write these lua files.

submodules/fluent-bit
igorpeshansky, line 1: I think you need to run `git submodule update --init` to pick up the latest changes here…

igorpeshansky:
Forgot to mention — I've updated the base branch (`bms-support`) to match the latest `master`. @alex-basinov Let's avoid rebasing the PR branch from this point — we can always squash the unnecessary commits when we merge.

confgenerator/testdata/bms/builtin/linux/builtin/golden/f120d4527bd717cab023dbbe5fbdc332.lua
igorpeshansky: This is all auto-generated code. I think the code in question comes from [here](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/filter/internal/ast/ast.go#L127).

confgenerator/testdata/bms/builtin/linux/builtin/golden/b4a0dead382dce7b4fe011d3f59fdb6d.lua
igorpeshansky: I believe it's [here](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/fluentbit/parser_deduplication.go#L33). Luckily, this Lua bit is verbatim in the code, so it should be easy to send a PR.

quentinmit (changes requested):
@igorpeshansky asked me to take over the code review. Apologies if I'm missing any context here; please feel free to point me at any design docs or comment threads that fill it in.

There are still some components that need special resource detection:
- Prometheus receiver
- OTLP receiver (see comment in `modular.go`)
- fluent-bit/logs

confgenerator/otel/modular.go
```diff
@@ -173,16 +177,28 @@ func (c ModularConfig) Generate() (string, error) {
...
 			name := processor.name(fmt.Sprintf("%s_%d", prefix, i))
 			processorNames = append(processorNames, name)
 			processors[name] = processor.Config
 		}
-
-		rdm := receiverPipeline.ResourceDetectionModes[pipeline.Type]
-		if name, ok := resourceDetectionProcessorNames[rdm]; ok {
-			processorNames = append(processorNames, name)
-			processors[name] = resourceDetectionProcessors[rdm].Config
+		if resource.GetType() == "bms" {
```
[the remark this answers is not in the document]
quentinmit: Yeah, I didn't comment on it since this line of code would go away with my suggested changes.

confgenerator/otel/modular.go
```diff
@@ -178,11 +169,20 @@ func (c ModularConfig) Generate() (string, error) {
...
-			processors[name] = resourceDetectionProcessors[rdm].Config
+		if !skipResourceDetection {
+			resourceDetectionProcessors := map[ResourceDetectionMode]Component{
+				Override: GCPResourceDetector(true),
+				Upsert:   GCPResourceDetector(false),
+			}
+			resourceDetectionProcessorNames := map[ResourceDetectionMode]string{
+				Override: resourceDetectionProcessors[Override].name("_global_0"),
+				Upsert:   resourceDetectionProcessors[Upsert].name("_global_1"),
+			}
```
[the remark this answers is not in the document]
alex-basinov: Fixed. I've accidentally moved them inside the for loop.

confgenerator/resourcedetector/detector.go
```diff
@@ -32,6 +32,8 @@ func GetResource() (Resource, error) {
 	switch {
 	case gcp_metadata.OnGCE():
 		return GetGCEResource()
+	case testOnBMS():
```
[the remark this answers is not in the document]
alex-basinov: Ack

confgenerator/resourcedetector/gce_detector.go
```diff
@@ -31,6 +31,7 @@ const (
 	label
 	interfaceIPv4
 	defaultScopes
+	location
```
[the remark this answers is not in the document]
alex-basinov: Done

confgenerator/testdata/goldensBMS/builtin/windows/builtin/input.yaml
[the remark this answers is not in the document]
alex-basinov: I removed the tests for windows.

systemd/google-cloud-ops-agent-opentelemetry-collector.service
```diff
@@ -23,6 +23,7 @@ RuntimeDirectory=google-cloud-ops-agent-opentelemetry-collector
 StateDirectory=google-cloud-ops-agent/opentelemetry-collector
 LogsDirectory=google-cloud-ops-agent
 Type=simple
+EnvironmentFile=-/etc/default/google-cloud-ops-agent
```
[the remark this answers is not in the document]
alex-basinov: Ok that makes sense. Thank you for the feedback. I removed the EnvironmentFile directive from the systemd files.

confgenerator/otel/modular.go
```diff
@@ -143,13 +147,13 @@ func (c ModularConfig) Generate() (string, error) {
...
 
-	resourceDetectionProcessors := map[ResourceDetectionMode]Component{
-		Override: GCPResourceDetector(true),
-		Upsert:   GCPResourceDetector(false),
-	}
-	resourceDetectionProcessorNames := map[ResourceDetectionMode]string{
-		Override: resourceDetectionProcessors[Override].name("_global_0"),
-		Upsert:   resourceDetectionProcessors[Upsert].name("_global_1"),
+	if r, ok := resource.(resourcedetector.BMSResource); ok {
+		processors["resource/bms"] = AddResourceAttributes(map[string]string{
```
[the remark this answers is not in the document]
alex-basinov: Do you have any suggestions for an easy way to detect when to not override resource attributes? Or, should I just create a static list of receivers for which we don't want to modify attributes?

confgenerator/otel/modular.go
```diff
@@ -116,7 +120,7 @@ type ModularConfig struct {
 //	receivers: [hostmetrics/mypipe]
 //	processors: [filter/mypipe_1, metrics_filter/mypipe_2, resourcedetection/_global_0]
 //	exporters: [googlecloud]
-func (c ModularConfig) Generate() (string, error) {
+func (c ModularConfig) Generate(resource resourcedetector.Resource) (string, error) {
```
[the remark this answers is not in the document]
alex-basinov: Done

internal/healthchecks/network_check.go
```diff
@@ -87,9 +88,15 @@ func (c NetworkCheck) Name() string {
...
 }
 
-func (c NetworkCheck) RunCheck(logger logs.StructuredLogger) error {
+func (c NetworkCheck) RunCheck(logger logs.StructuredLogger, resource resourcedetector.Resource) error {
 	var networkErrors []error
+
 	for _, r := range requests {
+		// Skip request to the GCE Metadata server if running on a BMS machine.
+		// BMS environment doesn't have Metadata servers.
+		if r.name == "GCE Metadata Server" && resource.GetType() == resourcedetector.BMS {
```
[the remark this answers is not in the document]
alex-basinov: Done.

confgenerator/confgenerator.go
```diff
@@ -115,6 +119,22 @@ func (uc *UnifiedConfig) GenerateOtelConfig(ctx context.Context) (string, error)
 	if uc.Metrics.Service.LogLevel == "" {
 		uc.Metrics.Service.LogLevel = "info"
 	}
+	if res.GetType() == "bms" {
```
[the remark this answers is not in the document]
alex-basinov: Done.

confgenerator/resourcedetector/bms_detector.go
```diff
@@ -0,0 +1,70 @@
...
+// The data provider interface for BMS environment
+type bmsDataProvider interface {
+	getProject() string
+	getLocation() string
+	getInstanceID() string
+}
+
+type BMSResourceBuilderInterface interface {
+	GetResource() (Resource, error)
+}
```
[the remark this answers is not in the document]
alex-basinov: Removed

confgenerator/resourcedetector/bms_detector.go
```diff
@@ -0,0 +1,70 @@
...
+
+type BMSResourceBuilder struct {
+	provider bmsDataProvider
```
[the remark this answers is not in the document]
alex-basinov: I removed the struct altogether

confgenerator/confgenerator.go
```diff
@@ -28,9 +28,12 @@ import (
 
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/resourcedetector"
 	"github.com/GoogleCloudPlatform/ops-agent/internal/platform"
 )
 
+const BMSCloudPlatformAttribute = "gcp_bare_metal_solution"
```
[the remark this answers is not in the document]
alex-basinov: Yes, it does. I added a comment to clarify that.

confgenerator/confgenerator_test.go
```diff
@@ -288,33 +315,36 @@ func testGeneratedFiles(t *testing.T, generatedFiles map[string]string, testDir
...
-		MachineType:   "test-machine-type",
-		Metadata:      map[string]string{"test-key": "test-value", "test-escape": "${foo:bar}"},
-		Label:         map[string]string{"test-label-key": "test-label-value"},
-		InterfaceIPv4: map[string]string{"test-interface": "test-interface-ipv4"},
+	for _, t := range testResources {
+		fmt.Printf("============= Running all tests for %v =============\n", t.name)
+		// Set up the test environment with mocked data.
+		confgenerator.MetadataResource = t.testResource
+		// Run test suite.
+		if thisr := m.Run(); r == 0 {
```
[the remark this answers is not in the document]
alex-basinov: I reverted the changes in this file and rearranged the code a bit so this is no longer relevant.

confgenerator/files.go
```diff
@@ -47,7 +47,7 @@ func ReadUnifiedConfigFromFile(ctx context.Context, path string) (*UnifiedConfig
 	return uc, nil
 }
 
-func (uc *UnifiedConfig) GenerateFilesFromConfig(ctx context.Context, service, logsDir, stateDir, outDir string) error {
+func (uc *UnifiedConfig) GenerateFilesFromConfig(ctx context.Context, service, logsDir, stateDir, outDir string, resource resourcedetector.Resource) error {
```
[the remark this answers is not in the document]
alex-basinov: Done.

submodules/fluent-bit
alex-basinov: Done

confgenerator/otel/modular.go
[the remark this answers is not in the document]
quentinmit: That part is already done in this code (the two cases of `Override` and `Upsert`) - just reuse the same structure.

confgenerator/files.go
```diff
@@ -52,6 +52,10 @@ func (uc *UnifiedConfig) GenerateFilesFromConfig(ctx context.Context, service, l
 	case "": // Validate-only.
 		return nil
 	case "fluentbit":
+		// short circuit in case of BMS
+		if platform.FromContext(ctx).ResourceOverride != nil {
```
quentinmit, line 56: This will catch other resource overrides, too. (and I assume this is a TODO, right? you're going to eventually support logging?)

Please add an explicit `SkipLogging` bool to the Platform struct, that you can set on BMS.

confgenerator/otel/modular.go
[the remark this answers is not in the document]
quentinmit: More specifically, move this up to line 157, and replace the `resourceDetectionProcessors` map with two new processors. Then you also don't need the added processor at line 198.

Also, this doesn't need to be BMS-specific. I think I would add a `OTelResourceAttributes() map[string]string` method to the `Resource` interface, which you can just call from here.

confgenerator/files.go
```diff
@@ -64,11 +68,6 @@ func (uc *UnifiedConfig) GenerateFilesFromConfig(ctx context.Context, service, l
 	case "otel":
 		// Fetch resource information from the metadata server.
 		var err error
-		MetadataResource, err = resourcedetector.GetResource()
```
quentinmit, line 67: This needs to stay, for now, for the benefit of the prometheus receiver (until that receiver is fixed).

confgenerator/otel/processors.go
```diff
@@ -386,3 +386,23 @@ func GCPResourceDetector(override bool) Component {
 		Config: config,
 	}
 }
+
+// ResourceAttribute returns a Component that applies changes on resource attributes.
```
quentinmit, line 390: It doesn't actually return a `Component` (though it should)

confgenerator/resourcedetector/gce_detector.go
```diff
@@ -14,25 +14,6 @@
 
 package resourcedetector
 
-type gceAttribute int
```
quentinmit, line 17: Why did you need to make this common? I don't see any new uses of this type/these constants.

confgenerator/otel/modular.go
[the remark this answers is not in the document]
alex-basinov: I apologize for such simple questions, I'm still new to the opentelemetry world, but can you provide an example when we would not want to override existing resource attributes ? Is this primarily needed for cases where libraries are sending telemetry data to a collector running on a different host ?

I'm unsure how to reuse the same structure because, unlike the resource detection processor, the resource processor doesn't support the "override" parameter.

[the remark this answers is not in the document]
alex-basinov: Please ignore the last part of my previous comment. I think I got it now. I didn't see your previous comment when I posted mine.

confgenerator/resourcedetector/gce_detector.go
alex-basinov: Igor proposed this because the constants are used by both the GCEResource and BMSResource structures.

confgenerator/otel/modular.go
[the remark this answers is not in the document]
alex-basinov: Done. PTAL ?

confgenerator/files.go
alex-basinov: Done.

confgenerator/otel/processors.go
alex-basinov: Done.

confgenerator/files.go
alex-basinov: Yes, I'll add a support for logging later.

confgenerator/files.go
```diff
@@ -52,6 +53,10 @@ func (uc *UnifiedConfig) GenerateFilesFromConfig(ctx context.Context, service, l
 	case "": // Validate-only.
 		return nil
 	case "fluentbit":
+		// TODO(alex-basinov) add support for logging on BMS
+		if platform.FromContext(ctx).SkipLogging != true {
```
quentinmit, line 57: I think this condition is backwards. (You're skipping when `SkipLogging` is false?)

Also, this is just going to skip config generation. Won't fluent-bit then fail to start because its config is missing?

confgenerator/otel/modular.go
```diff
@@ -164,6 +174,9 @@ func (c ModularConfig) Generate() (string, error) {
 			continue
 		}
 		for i, processor := range p {
+			if pl.ResourceOverride != nil && processor.Type == "resourcedetection" {
```
quentinmit, line 177: Why is this needed? The resource detection processors shouldn't be in `p` to begin with.

confgenerator/resourcedetector/gce_detector.go
quentinmit: I don't see where any of these are being used in the `BMSResource` structure, can you point it out to me?

internal/platform/platform.go
```diff
@@ -72,6 +75,13 @@ func detect() Platform {
 	p := Platform{
 		HostInfo: info,
 	}
+	if resourcedetector.OnBMS() {
+		p.ResourceOverride, err = resourcedetector.GetResource()
```
quentinmit, line 79: Why aren't you calling `resourcedetector.GetBMSResource()`?

(Also, beware typed vs untyped `nil`)

confgenerator/files.go
alex-basinov: it's a leftover from testing which I forgot to update. 
Yes, it would fail to start. I haven't thought about it.

Ok, I removed the SkipLogging field and updated GenerateFluentBitConfigs() to generate a BMS-specific config if ResourceOverride is set to the BMSResource. 
PTAL ?

confgenerator/resourcedetector/gce_detector.go
alex-basinov: oops, project and location constants were used by BMSResourceBuilderInterface at some point but then I made some changes and removed the interface and forgot to move the constants back to gce_detector.go.

confgenerator/otel/modular.go
alex-basinov: Because OTLP receiver injects a resource detection processor (named resourcedetection/otlp_0) in the config

https://github.com/GoogleCloudPlatform/ops-agent/blob/022c5dcaeba98de99b210aa2f7b81647071b6dfb/apps/otlp.go#L53

There might be other receivers doing the same.

internal/platform/platform.go
alex-basinov: I accidentally mis-typed the method name. Fixed.

confgenerator/otel/modular.go
alex-basinov: I was experimenting with various receivers and at some point discovered a resourcedetection receiver in the config. I didn't know where it came from and decided to simply remove  at the config generation phase. Turns out it comes from the otlp receiver when the metrics_mode is set to googlemanagedprometheus. I updated apps/otlp.go and replaced resourcedetection processor with otel.ResourceTransform() when on BMS.  PTAL ?

alex-basinov:
Quentin, can you please help understand why some of the tests failing ? I don't quite understand what's wrong with, for example, "Build (Rocky Linux 9 ARM64)" and "Build (Ubuntu 20.04 Focal ARM64)". Are these tests just flaky ? Are the SLES tests also flaky, or there's something I need to fix ?

Any idea why the TestOTLPMetricsGMP and TestOTLPMetricsGCM  failing as well? I looked through the logs but I don't see anything that would point to the source of the problem. That said, I might be looking at the wrong log files.

jefferbrecht:
The `TestOTLP` failures also look like the same corrupt rebase. Your branch is missing the change that was merged in #1398.

jefferbrecht:
Ah, just realized this PR is merging into `bms-support`, not `master`. `bms-support` is quite behind and ought to be updated. After that you can rebase your branch and those test failures should go away.

alex-basinov:
Thank you. Could someone please help update the base branch (bms-support) to match the latest master ?

apps/otlp.go
```diff
@@ -42,45 +59,54 @@ func (r ReceiverOTLP) Type() string {
...
-						// cluster = "__gce__"
-						stmt(`attributes["cluster"]`, `"__gce__"`),
-						// instance_name = host.name
-						stmt(`attributes["instance_name"]`, `resource.attributes["host.name"]`),
-						// machine_type = host.type
-						stmt(`attributes["machine_type"]`, `resource.attributes["host.type"]`),
-						// job and instance should be automatically set from service labels
-					},
-				}},
+	bmsResource := platform.FromContext(ctx).ResourceOverride
```
quentinmit, line 64: This is not the right way to handle this. Whether Ops Agent is running on a BMS machine is unrelated to whether the OTLP data it's processing is coming from a BMS machine.

If you want to add support for BMS->GMP resource conversion, you'll need to add the new statements *in addition* to the existing ones, not instead of them. And probably also do so unconditionally, regardless of whether Ops Agent is running on a BMS machine.

apps/otlp.go
```diff
@@ -42,45 +59,54 @@ func (r ReceiverOTLP) Type() string {
...
+		return components(bmsAttributes, otel.ResourceTransform(bmsResource.OTelResourceAttributes()))
 	}
+	return components(gceAttributes, otel.GCPResourceDetector(false))
 }
 
-func (r ReceiverOTLP) metricsProcessors() (otel.ExporterType, otel.ResourceDetectionMode, []otel.Component) {
+func (r ReceiverOTLP) metricsProcessors(ctx context.Context) (otel.ExporterType, otel.ResourceDetectionMode, []otel.Component) {
 	if r.MetricsMode != "googlecloudmonitoring" {
-		return otel.GMP, otel.None, r.gmpResourceProcessors()
+		p := r.gmpResourceProcessors(ctx)
```
quentinmit, line 108: Why did this change from one line to two?

confgenerator/resourcedetector/detector.go
```diff
@@ -46,6 +49,12 @@ func (UnrecognizedPlatformResource) GetType() string {
 	return "unrecognized platform"
 }
 
+func (UnrecognizedPlatformResource) OTelResourceAttributes() map[string]string {
+	return map[string]string{
```
quentinmit, line 53: Just return nil; we shouldn't be setting `cloud.platform` at all in this case.

confgenerator/otel/modular.go
```diff
@@ -143,13 +146,20 @@ func (c ModularConfig) Generate() (string, error) {
...
 		Upsert:   GCPResourceDetector(false),
 	}
-	resourceDetectionProcessorNames := map[ResourceDetectionMode]string{
-		Override: resourceDetectionProcessors[Override].name("_global_0"),
-		Upsert:   resourceDetectionProcessors[Upsert].name("_global_1"),
+
+	if pl.ResourceOverride != nil {
+		resourceProcessors = map[ResourceDetectionMode]Component{
+			Override: ResourceTransform(pl.ResourceOverride.OTelResourceAttributes()),
+			Upsert:   ResourceTransform(pl.ResourceOverride.OTelResourceAttributes()),
```
quentinmit, line 157: It looks like you're still not handling the two cases differently. That said, it also looks like I've named these enum cases poorly. The "override" case should be replacing the existing resource attributes (which for the `resource` processor means `"action": "upsert"`), and the other case should be only setting the resource if it's not already set (which for the `resource` processor means `"action": "insert"`). See https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/attributesprocessor/README.md

So change `ResourceTransform` to also take a bool parameter like `GCPResourceDetector` that changes what it sets `action` to, and if you want rename `Upsert` to `SetIfMissing` or something like that to make it clearer.

confgenerator/otel/modular.go
```diff
@@ -143,13 +146,20 @@ func (c ModularConfig) Generate() (string, error) {
 		"service":    service,
 	}
 
-	resourceDetectionProcessors := map[ResourceDetectionMode]Component{
+	resourceProcessors := map[ResourceDetectionMode]Component{
```
quentinmit, line 149: I would prefer keeping this named `resourceDetectionProcessors` (especially since you're going to switch back to a resource detection processor when it's supported upstream, right?)

confgenerator/logging.go
```diff
@@ -70,6 +72,13 @@ func stackdriverOutputComponent(match, userAgent string, storageLimitSize string
 		// Mute these errors until https://github.com/fluent/fluent-bit/issues/4473 is fixed.
 		"net.connect_timeout_log_error": "False",
 	}
+	r := platform.FromContext(ctx).ResourceOverride
+	if r != nil {
+		if r, ok := r.(resourcedetector.BMSResource); ok {
```
quentinmit, line 77: The resource type and labels should come from the `ResourceOverride`, instead of being hardcoded here. Maybe that means there should be a second method in the interface (`GCMResource() (string, map[string]string)` or something like that?)

apps/otlp.go
alex-basinov: This is a leftover from testing.

confgenerator/otel/modular.go
alex-basinov: Done.

confgenerator/logging.go
alex-basinov: Okay, I've created a new method, albeit with a slightly different name. It returns a comma-separated string containing resource labels. This approach simplifies the process by avoiding the need to return a map and then convert it to a comma-separated string while ensuring element order consistency.

I've also update GetType() to return the resource type which is used here in logging.go and in api_check.py.

apps/otlp.go
alex-basinov: Okay, but If the ops-agent is running on a BMS machine and its config contains OTLP receiver with metric_mode set to googlemanagedprometheus then the generated otel config will have a GCP resourcedetection processor which won't work on a BMS machine. Shouldn't I make sure that the resourcedetection processor doesn't get included into the otel config ?

confgenerator/logging.go
```diff
@@ -70,6 +71,11 @@ func stackdriverOutputComponent(match, userAgent string, storageLimitSize string
 		// Mute these errors until https://github.com/fluent/fluent-bit/issues/4473 is fixed.
 		"net.connect_timeout_log_error": "False",
 	}
+	r := platform.FromContext(ctx).ResourceOverride
+	if r != nil {
```
quentinmit, line 75: You can construct the `resource_labels` string here from the monitored resources label map in the resource; there's no reason to duplicate that code in each of the resource implementations.

e.g.
```go
if r := platform.FromContext(ctx).ResourceOverride; r != nil {
  mr = r.MonitoredResource()
  var labels []string
  for k, v := range mr.Labels {
    labels = append(labels, fmt.Sprintf("%s=%s", k, v))
  }
  sort.Strings(labels)
  config["resource"] = mr.Type
  config["resource_labels"] = strings.Join(labels, ",")
}
```

apps/otlp.go
```diff
@@ -44,10 +44,10 @@ func (r ReceiverOTLP) Type() string {
...
-	stmt := func(target, source string) string {
+	stmt := func(target, source, platform string) string {
 		// if cloud.provider == "gcp" && cloud.platform == "gcp_compute_engine"
-		return fmt.Sprintf(`set(%s, %s) where %s != nil and resource.attributes["cloud.platform"] == "gcp_compute_engine"`,
-			target, source, source)
+		return fmt.Sprintf(`set(%s, %s) where %s != nil and resource.attributes["cloud.platform"] == "%s"`,
+			target, source, source, platform)
 	}
 	return []otel.Component{
 		otel.GCPResourceDetector(false),
```
quentinmit, line 53: I think you asked about this line of code. You need to replace `otel.GCPResourceDetector(false)` with `otel.ResourceTransform(pl.ResourceOverride.OTelResourceAttributes(), false)` when there's a resource override specified. (Effectively, the `ResourceOverride` is taking the place of the `GCPResourceDetector`)

confgenerator/otel/processors.go
```diff
@@ -386,3 +386,32 @@ func GCPResourceDetector(override bool) Component {
 		Config: config,
 	}
 }
+
+// ResourceTransform returns a Component that applies changes on resource attributes.
+func ResourceTransform(attributes map[string]string, GCPResourceDetector bool) Component {
```
quentinmit, line 391: I think the bool parameter can just be called something like `override bool`. I don't understand why you've called it `GCPResourceDetector`?

internal/healthchecks/api_check.go
```diff
@@ -66,19 +56,16 @@ func createMonitoringPingRequest(gceMetadata resourcedetector.GCEResource) *moni
...
 	req := &monitoringpb.CreateTimeSeriesRequest{
-		Name: "projects/" + gceMetadata.Project,
+		Name: "projects/" + resource.GetProject(),
 		TimeSeries: []*monitoringpb.TimeSeries{{
 			MetricKind: metricpb.MetricDescriptor_GAUGE,
 			ValueType:  metricpb.MetricDescriptor_INT64,
 			Metric: &metricpb.Metric{
 				Type: metricType,
 			},
 			Resource: &monitoredres.MonitoredResource{
```
quentinmit, line 66: If you make the interface change I suggested, this can just become `Resource: resource.MonitoredResource()`

apps/otlp.go
quentinmit: You're right that the resourcedetection processor itself needs to change. I accidentally created a new thread elaborating on that, so I'll mark this one resolved.

confgenerator/resourcedetector/detector.go
```diff
@@ -22,6 +22,10 @@ import (
 // available attributes about the current monitoring resource.
 type Resource interface {
 	GetType() string
+	OTelResourceAttributes() map[string]string
+	FluentBitLabels() string
```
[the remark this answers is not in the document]
alex-basinov: Sure, but I had to modify the method's name slightly to "ProjectName()" since both GCEResource and BMSResource structs have a field named "Project", and it's not possible to have both a struct field and a method with the same name.

quentinmit (approved):

internal/healthchecks/network_check.go
```diff
@@ -103,8 +107,16 @@ func (c NetworkCheck) Name() string {
 
 func (c NetworkCheck) RunCheck(logger logs.StructuredLogger) error {
 	var networkErrors []error
-	for _, r := range requests {
+	ctx := context.TODO()
+	p := platform.FromContext(ctx)
+	for _, r := range commonRequests {
 		networkErrors = append(networkErrors, r.SendRequest(logger))
 	}
+	if p.ResourceOverride == nil {
```
igorpeshansky, line 115: @alex-basinov Note that this will need to be fixed before merging back into `master`, since GCE checks will also need to run if the resource is explicitly specified as `"gce_instance"`.

igorpeshansky: Actually, it's done as part of #1539, so we can pick this up on the next merge.

PR #1342, opened 2023-07-20
Add Support For Custom Service Accounts in Simulacra

shafinsiddique:
## Description

Currently, the only way we can set custom service accounts for VMs is by setting an environment variable. Simulacra reads all config from either the command line or from a YAML file. Given this, an environment variable is inconsistent and thus unergonomic for the user. This PR adds support for passing in Service Accounts to Simulacra through either command line argument or through a field in the YAML config. The service account value is passed to the command that initializes the VM.  

## How has this been tested?

I ran Simulacra with the --service_account command line arg set to "gce-instance-write-logs-and-me@experimental-braydonk.iam.gserviceaccount.com". It created this VM : https://pantheon.corp.google.com/compute/instancesDetail/zones/us-central1-a/instances/simulacra-vm-instance-d1b011de-2d4d-4ea5-8e3d-ef3815051aa3?project=experimental-braydonk&tab=details&pageState=(%22duration%22:(%22groupValue%22:%22PT1H%22,%22customValue%22:null))

If we inspect the Service Account in Pantheon, we see that it is the same value as we passed in through the command line. 
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump
  - [] This PR bumps the version.

### DIFF
--- a/cmd/simulacra/simulacra.go
+++ b/cmd/simulacra/simulacra.go
@@ -59,6 +59,8 @@ type Config struct {
 	ThirdPartyAppsPath string `yaml:"third_party_apps_path"`
 	// Path to script files that will be run on the VM.
 	Scripts []string `yaml:"scripts"`
+	// A Service Account for the VM.
+	ServiceAccount string `yaml:"service_account"`
 }
 
 func distroFolder(platform string) (string, error) {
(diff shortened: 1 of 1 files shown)


cmd/simulacra/simulacra.go
```diff
@@ -252,16 +274,9 @@ func main() {
...
-		Platform:    config.Platform,
-		MachineType: agents.RecommendedMachineType(config.Platform),
-		Name:        config.Name,
-		Project:     config.Project,
-		Zone:        config.Zone,
-	}
 	// Create VM Instance.
 	log.Default().Printf("Creating VM Instance, check %s for details", vmInitLogFile)
-	vm, err := gce.CreateInstance(ctx, logger.ToFile(vmInitLogFileName), options)
+	vm, err := createInstance(ctx, &config, logger.ToFile(vmInitLogFileName))
```
braydonk, line 279: nit: maybe `config` should already be a `*Config` rather than passing a reference specifically here?

braydonk (approved):

PR #1350, opened 2023-07-26
Add tzdata-java to install scripts

dehaansa:
## Description

Attempting to fix test issues like the following that cropped up overnight for rockylinux & centos builds.

```
        + /opt/kafka/bin/kafka-topics.sh --create --topic quickstart-events --bootstrap-server localhost:9092 --partitions 3 --replication-factor 1
        log4j:WARN Failed to set property [conversionPattern] to value "[%d] %p %m (%c)%n". 
        java.lang.reflect.InvocationTargetException
...
        Caused by: java.lang.Error: java.io.FileNotFoundException: /usr/lib/jvm/java-11-openjdk-11.0.20.0.8-2.el9.x86_64/lib/tzdb.dat (No such file or directory)
...
        Caused by: java.io.FileNotFoundException: /usr/lib/jvm/java-11-openjdk-11.0.20.0.8-2.el9.x86_64/lib/tzdb.dat (No such file or directory)
...
        [AdminClient clientId=adminclient-1] Connection to node -1 (localhost/127.0.0.1:9092) could not be established. Broker may not be available.
```


igorpeshansky (approved):
LGTM :shipit:

PR #1354, opened 2023-07-27
attempt to fix mariadb debian 12 hostname change

dehaansa:


dehaansa:
fluentbit doesn't support calling functions instead of just using variables in the config files.

PR #1357, opened 2023-08-01
Windows handles metric

avilevy18:
## Description

Add windows handle metric to host metric receiver 
## Related issue

[b/279901505](http://b/279901505)
## How has this been tested?

Integration and unit tests
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


apps/hostmetrics.go
```diff
@@ -301,6 +311,11 @@ func (r MetricsReceiverHostmetrics) Pipelines(_ context.Context) []otel.Receiver
 					otel.ToggleScalarDataType,
 					otel.AddLabel("process", "all"),
 				),
+				otel.RenameMetric(
```
igorpeshansky, line 314: Should this also happen only on Windows?

avilevy18: There should be no need, since it won't be enabled.

braydonk: Yeah it should be harmless to leave it here on both platforms, but might be confusing to future readers. The way this is written it's not elegant to exclude by just platform. What you'd need to do is change this to an array, add to the array conditionally based on platform, then spread it (with the `...` operator) into the arguments.

apps/hostmetrics.go
```diff
@@ -31,7 +32,18 @@ func (r MetricsReceiverHostmetrics) Type() string {
 	return "hostmetrics"
 }
 
-func (r MetricsReceiverHostmetrics) Pipelines(_ context.Context) []otel.ReceiverPipeline {
+func (r MetricsReceiverHostmetrics) Pipelines(ctx context.Context) []otel.ReceiverPipeline {
+	p := platform.FromContext(ctx)
+	s := map[string]interface{}{
```
braydonk, line 37: `s` feels like a non-obvious name here. I'd recommend changing it.

braydonk (approved):

apps/hostmetrics.go
igorpeshansky: It's confusing and pollutes the Linux configs, hence my suggestion.
As @braydonk said, you could do the following:
```
	transforms := []map[string]interface{}{
		// host transforms
		otel.RenameMetric(
			"system.cpu.time",
			"cpu/usage_time",
			…
		),
		…
		otel.RenameMetric(
			"system.paging.operations",
			"swap/io",
			…
		),
		// process transforms
		otel.RenameMetric(
			"process.cpu.time",
			"processes/cpu_time",
			…
		),
		…
		otel.RenameMetric(
			"process.memory.virtual",
			"processes/vm_usage",
			…
		),
	}
	if p.Type == platform.Windows {
		transforms = append(
			transforms, 
			otel.RenameMetric(
				"process.handles",
				"processes/windows/handles",
				otel.AddLabel("process", "all"),
			),
		)
	}
	transforms = append(
		transforms, 
		otel.AddPrefix("agent.googleapis.com"),
	)
	return []otel.ReceiverPipeline{{
		…
		Processors: map[string][]otel.Component{"metrics": {
			…
			otel.MetricsTransform(transforms...),
		}},
	}}
```

apps/hostmetrics.go
```diff
@@ -76,6 +313,7 @@ func (r MetricsReceiverHostmetrics) Pipelines(ctx context.Context) []otel.Receiv
 					},
 				},
 			},
+			//transforms...,
```
igorpeshansky, line 316: Is this left over from earlier debugging?

avilevy18: Whoops, thanks for the catch.

igorpeshansky (approved):
LGTM :shipit:

PR #1363, opened 2023-08-03
Integrating Simulacra with Diagnostic Tool Output

shafinsiddique:
## Description

We don’t want the Ops Agent on-call team to manually have to create this configuration file. Currently, we have a diagnostic tool that our customers run whenever their Ops Agent is not behaving accordingly. The diagnostic tool collects various information about the customer's environment and outputs it into different files. 

We can integrate the diagnostic tool with Simulacra so that the Simulacra configuration file can be automatically created from the information that the diagnostic tool generates. This would automate the process of manually creating Simulacra config files. 

This PR adds support in Simulacra for reading diagnostic tool output and configuring itself from that output. 

## How has this been tested?
Used a diagnostic tool output and passed it to Simulacra. It created the VM correctly. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


cmd/simulacra/simulacra.go
```diff
@@ -204,8 +212,43 @@ func getConfigFromYaml(configPath string) (*Config, error) {
...
+	}
+
+	var metadata Metadata
+	if err := json.Unmarshal([]byte(metadataFile), &metadata); err != nil {
+		return nil, err
+	}
+
+	config := &Config{
+		Platform:           metadata.Image,
+		Project:            metadata.ProjectName,
```
braydonk, line 234: You definitely won't want this. The `project_name` is going to be the customer's project. This should still be the project where we Simulacra will actually deploy stuff

cmd/simulacra/simulacra.go
```diff
@@ -204,8 +212,43 @@ func getConfigFromYaml(configPath string) (*Config, error) {
...
 }
 
+func getConfigFromDiagnosticOutput(outputDir string) (*Config, error) {
+	type Metadata struct {
+		Image       string `json:"image"`
+		ProjectName string `json:"project_name"`
+		Zone        string `json:"zone"`
+	}
+
+	metadataFile, err := ioutil.ReadFile(path.Join(outputDir, "vm_config.json"))
```
braydonk, line 222: `ioutil` is deprecated, this should be `os.ReadFile` instead.

braydonk, line 222: `ioutil` is deprecated, this should be `os.ReadFile` instead.

cmd/simulacra/simulacra.go
```diff
@@ -204,8 +212,43 @@ func getConfigFromYaml(configPath string) (*Config, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var metadata Metadata
+	if err := json.Unmarshal([]byte(metadataFile), &metadata); err != nil {
```
braydonk, line 228: `metadataFile` will already be a `[]byte` right?

integration_test/gce/gce_testing.go
```diff
@@ -1053,13 +1053,25 @@ func addFrameworkLabels(inputLabels map[string]string) (map[string]string, error
...
 }
 
+func isFullImageName(name string) bool {
+	return strings.HasPrefix(name, "projects/")
+}
+
+func getImageFamilyName(image string) string {
+	if isFullImageName(image) {
+		components := strings.Split(image, "/")
+		return components[len(components)-1]
```
braydonk, line 1063: So the last component of the array is the image family name? I thought that was the image name itself

shafinsiddique: okay this was definitely poor naming on my part, 

this function essentially does 2 things : it either gets the image name if prefix "/project" is in the string OR it gets family name if that prefix is not there.

martijnvans: could I get an example of what `image` looks like in a case where it starts with `projects/`?

I have a feeling that you will need to specify options.ImageProject in cases like this...

integration_test/gce/gce_testing.go
```diff
@@ -1053,13 +1053,25 @@ func addFrameworkLabels(inputLabels map[string]string) (map[string]string, error
...
+}
+
 // attemptCreateInstance creates a VM instance and waits for it to be ready.
 // Returns a VM object or an error (never both). The caller is responsible for
 // deleting the VM if (and only if) the returned error is nil.
 func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOptions) (vmToReturn *VM, errToReturn error) {
 	vm := &VM{
 		Project:  options.Project,
-		Platform: options.Platform,
+		Platform: getImageFamilyName(options.Platform),
```
martijnvans, line 1074: I would prefer keeping this line as is (Platform: options.Platform), and doing something like this down below (line 1130ish):

```
	if isFullImageName(options.Platform) {
		components := strings.Split(options.Platform, "/")
		imageOrImageFamilyFlag = "--image=" + components[len(components)-1]
		// maybe parse out the image project too? if so, this chunk and lines 1124-1127 should move up above line 1107. Unless I'm misunderstanding and the image project isn't part of this string in the first place.
	}
```

integration_test/gce/gce_testing.go
shafinsiddique: Yep! So this is what the image looks like most of the time coming from the metadata file : " projects/debian-cloud/global/images/debian-11-bullseye-v20230711"

cmd/simulacra/simulacra.go
shafinsiddique: oh this is so true, this is supposed to infer from Gcloud

integration_test/gce/gce_testing.go
shafinsiddique: I think if we did this, it would become inconsistent throughout the rest of the function. The vm.Platform value would be like this : " projects/debian-cloud/global/images/debian-11-bullseye-v20230711". Then in line 1213/1227/1246, we would get a runtime error because they are expecting the vm.platform value to be just the image family name but instead it's that weird format from the metadata output. This is why i think we need to set vm.Platform to the image family name when we create the vm struct from the options input.

integration_test/gce/gce_testing.go
martijnvans: OK so yeah the "debian-cloud" part should get passed in as options.ImageProject. The "global" part should probably get passed in to --image-family-scope, though there's no VMOptions for that yet.

I think we should parse this string `projects/debian-cloud/global/images/debian-11-bullseye-v20230711` in simulacra.go, adding some things to VMOptions in the process:

- a TODO to rename VMOptions.Platform -> ImageFamily (overdue)
- VMOptions.Image: an alternative to VMOptions.Platform
- VMOptons.ImageFamilyScope: doesn't seem very important, but it's probably technically correct to include it. It's kind of described here: https://screenshot.googleplex.com/BEPtcVek32VbZnr

Then we'd need to change some things at the start of attemptCreateInstance() to:
- check that one of VMOptions.Platform and VMOptions.Image are specified, not 0 or 2 of them
- put the nonempty one into VM.Platform (similar to today, vm.Platform would continue to hold either an image family or a specific image),
- change the code that computes imageOrImageFamilyFlag to look at the VMOptions fields instead of VM.Platform
- add some code to turn VMOptions.ImageFamilyScope into --image-family-scope.

In addition, I think you'd need some light modifications to simulacra.go:
- new fields in Config called Image and ImageProject, I guess ImageFamilyScope too
- some code around here to parse metadata.image into Image/ImageProject/ImageFamilyScope: https://screenshot.googleplex.com/7MVkQHDG6AUV6oK

sorry to ask for relatively big changes. I think it's important to make gce_testing.go as "dumb" as we can and avoid smuggling features into it if we can.

Happy to discuss and/or help out.

integration_test/gce/gce_testing.go
martijnvans: true true. I thought about this a lot, see my other comment.

martijnvans (changes requested):

integration_test/gce/gce_testing.go
shafinsiddique: This is a great idea! I'm working on implementing these rn, thank you for describing it in such details

shafinsiddique: Hey @martijnvans , I have made all the suggested changes with the exception of renaming Platform in VMOption to 'ImageFamily.'

Since other files use that field name, it would be a big change for this PR. 

I have created a task https://buganizer.corp.google.com/issues/295174028 and assigned it to myself to make that change in a seperate PR.

martijnvans: Thanks! I left a comment about a suggested 3-step approach.

martijnvans (approved):

cmd/simulacra/simulacra.go
```diff
@@ -215,21 +217,17 @@ func getConfigFromYaml(configPath string) (*Config, error) {
...
-func parseMetadataImage(name string) (string, string) {
+func parseMetadataImage(name string) (string, string, string, error) {
 	components := strings.Split(name, "/")
+	if len(components) < 5 {
+		return "", "", "", fmt.Errorf("image name from metadata must be of format 'projects/debian-cloud/global/images/debian-11-bullseye-v20230711' ")
+	}
 	scope := components[2]
 	image := components[4]
-	return scope, image
+	imgProject := components[1]
```
martijnvans, line 229: nit: move this up a couple of lines, above components[2] and components[4].

integration_test/gce/gce_testing.go
```diff
@@ -1109,18 +1130,30 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
 		return nil, fmt.Errorf("attemptCreateInstance() could not construct valid labels: %v", err)
 	}
 
-	imageOrImageFamilyFlag := "--image-family=" + vm.Platform
+	imageOrImageFamilyFlag := "--image=" + vm.Platform
```
martijnvans, line 1133: Would slightly prefer for this to pull from options.Image instead of vm.Platform because that's IMO less confusing.

Same thing with lines 1136 and 1139 (options.Platform this time instead of vm.Platform).

cmd/simulacra/simulacra.go
```diff
@@ -181,16 +194,16 @@ func getInstanceName() string {
...
 	if err != nil {
 		return nil, err
 	}
 	if err := yaml.Unmarshal(file, &config); err != nil {
 		return nil, err
 	}
 
-	if config.Platform == "" {
-		config.Platform = defaultPlatform
+	if config.ImageFamily == "" && config.Image == "" {
```
braydonk, line 205: So `ImageFamily` is okay to be empty if `Image` is set?

cmd/simulacra/simulacra.go
```diff
@@ -204,9 +217,61 @@ func getConfigFromYaml(configPath string) (*Config, error) {
 	return &config, nil
 }
 
+// Parse the metadata image name with format 'projects/debian-cloud/global/images/debian-11-bullseye-v20230711'
+// and return the scope and image name.
+func parseMetadataImage(name string) (string, string, string, error) {
+	components := strings.Split(name, "/")
+	if len(components) < 5 {
+		return "", "", "", fmt.Errorf("image name from metadata must be of format 'projects/debian-cloud/global/images/debian-11-bullseye-v20230711' ")
```
braydonk, line 225: nit: if you aren't formatting anything, you can use `errors.New` instead of `fmt.Errorf`.

cmd/simulacra/simulacra.go
```diff
@@ -204,9 +217,61 @@ func getConfigFromYaml(configPath string) (*Config, error) {
 	return &config, nil
 }
 
+// Parse the metadata image name with format 'projects/debian-cloud/global/images/debian-11-bullseye-v20230711'
```
braydonk, line 220: How certain are we that the image will always be in this format? Have we looked at any custom images to make sure they follow this format too?

cmd/simulacra/simulacra.go
```diff
@@ -204,9 +217,61 @@ func getConfigFromYaml(configPath string) (*Config, error) {
...
+	image := components[4]
+	return imgProject, scope, image, nil
+}
+
+func getConfigFromDiagnosticOutput(outputDir string) (*Config, error) {
+	type Metadata struct {
+		Image string `json:"image"`
+	}
+
+	metadataFile, err := os.ReadFile(path.Join(outputDir, "vm_config.json"))
```
braydonk, line 238: Use `filepath.Join` instead for cross-platform compatibility. See [explanation in package docs](https://pkg.go.dev/path).

integration_test/gce/gce_testing.go
```diff
@@ -1053,13 +1053,34 @@ func addFrameworkLabels(inputLabels map[string]string) (map[string]string, error
...
+
+	if image != "" {
+		return image, nil
+	}
+
+	if platform != "" {
+		return platform, nil
+	}
+
+	return "", fmt.Errorf("Atleast one of image or platform must be specified.")
```
braydonk, line 1069:
```suggestion
	return "", errors.New("at least one of image or platform must be specified")
```

braydonk: * Using `errors.New` since we're not formatting anything,
* `Atleast` -> `at least`
* Error messages should have no punctuation and not be capitalized

braydonk: (I'm aware we may not be following this pattern properly in a lot of places, IDE tools are supposed to be set up to warn us about this but it doesn't always catch it)

cmd/simulacra/simulacra.go
braydonk:
```suggestion
		return "", "", "", errors.New("image name from metadata must be of format 'projects/debian-cloud/global/images/debian-11-bullseye-v20230711'")
```

cmd/simulacra/simulacra.go
shafinsiddique: yep, it's an either or. If both are set, there's actually an error we send from the integration testing side.

cmd/simulacra/simulacra.go
shafinsiddique: Discussed offline. What we ended up doing was if the image name in the metadata file does not have the prefix "/projects", then we ask the user (using command line input) for the image family.

PR #1376, opened 2023-08-15
Fix: Network Manager race condition

avilevy18:
## Related issue

b/294246198

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -4144,6 +4144,31 @@ func TestNoNvmlOtelReceiverWithoutGpu(t *testing.T) {
 	})
 }
 
+func TestRestartVM(t *testing.T) {
+	t.Parallel()
+	gce.RunForEachPlatform(t, func(t *testing.T, platform string) {
+		t.Parallel()
+
+		ctx, logger, vm := agents.CommonSetup(t, platform)
+		if err := agents.SetupOpsAgent(ctx, logger.ToMainLog(), vm, ""); err != nil {
+			t.Fatal(err)
+		}
+
+		if err := gce.RestartInstance(ctx, logger, vm); err != nil {
+			t.Fatal(err)
+		}
+
+		cmdOut, err := gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", getRecentServiceOutputForPlatform(vm.Platform))
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		checkExpectedHealthCheckResult(t, cmdOut.Stdout, "Network", "PASS", "")
+		checkExpectedHealthCheckResult(t, cmdOut.Stdout, "Ports", "PASS", "")
+		checkExpectedHealthCheckResult(t, cmdOut.Stdout, "API", "PASS", "")
+	})
+}
+
 func TestMain(m *testing.M) {
 	code := m.Run()
 	gce.CleanupKeysOrDie()


integration_test/ops_agent_test.go
```diff
@@ -4195,6 +4196,42 @@ func TestNoNvmlOtelReceiverWithoutGpu(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		// time.Sleep(60 * time.Second)
+
+		cmdOut, err = gce.RunRemotely(ctx, logger.ToMainLog(), vm, "", getRecentServiceOutputForPlatform(vm.Platform))
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		checkExpectedHealthCheckResult(t, cmdOut.Stdout, "Network", "PASS", "")
```
rafaelwestphal, line 4229: I know this is a draft, but can you also check that the error doesn't occur?
I'm not sure if this would pick up the "PASS" from the previous health-checks or not.

internal/healthchecks/network_check.go
```diff
@@ -29,7 +30,15 @@ type networkRequest struct {
...
 func (r networkRequest) SendRequest(logger logs.StructuredLogger) error {
-	response, err := http.Get(r.url)
+	var response *http.Response
+	var err error
+	for i := 0; i < 5; i++ {
+		response, err = http.Get(r.url)
+		if err == nil && response.StatusCode == http.StatusOK {
+			break
+		}
+		time.Sleep(500 * time.Millisecond)
```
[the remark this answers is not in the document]
avilevy18: Adjusting to 1 second per cycle

internal/healthchecks/network_check.go
```diff
@@ -29,7 +30,15 @@ type networkRequest struct {
 }
 
 func (r networkRequest) SendRequest(logger logs.StructuredLogger) error {
-	response, err := http.Get(r.url)
+	var response *http.Response
+	var err error
+	for i := 0; i < 5; i++ {
```
ridwanmsharif, line 35: Why did we need to do this 5 times if we're using systemd to make `network-online` a prerequisite target?

systemd/google-cloud-ops-agent.service
```diff
@@ -14,7 +14,8 @@
 
 [Unit]
 Description=Google Cloud Ops Agent
-Wants=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opentelemetry-collector.service google-cloud-ops-agent-diagnostics.service
+Wants=google-cloud-ops-agent-fluent-bit.service google-cloud-ops-agent-opentelemetry-collector.service google-cloud-ops-agent-diagnostics.service network-online.target
+After=network-online.target
```
ridwanmsharif, line 18: The subagents also set `Requires=network-online.target`. Any reason we don't need that here? 

Also, can we remove the dependency from the subagents since that dep is added here and the subagents depends on this service?

ridwanmsharif (commented):
Is there something we need to do in Windows to also avoid this race condition?

systemd/google-cloud-ops-agent.service
jefferbrecht: I'd prefer to leave the subagent dependency there since they depend on it for their own reasons. Also helps reduce the number of things we need to change if we ever move the startup checks into a different service.

ridwanmsharif: Fair enough, I agree

internal/healthchecks/network_check.go
jefferbrecht: I assume it's for error 5xx flakes but in that case a comment in the code is probably warranted

rafaelwestphal:
This should refer to b/295772265 instead.

internal/healthchecks/network_check.go
avilevy18: If a machine has multiple network interfaces, `network-online` indicates that only one of them has come up so far. 
Adding `network-online.target` is a best effort not a full solution.

internal/healthchecks/network_check.go
```diff
@@ -29,7 +30,15 @@ type networkRequest struct {
...
+		response, err = http.Get(r.url)
+		if err == nil && response.StatusCode == http.StatusOK {
+			break
+		}
+		time.Sleep(1 * time.Second)
```
quentinmit, line 40: Exponential backoff?

quentinmit (approved):

internal/healthchecks/network_check.go
avilevy18: Added

PR #1415, opened 2023-09-13
healthchecks : Retry `monitoringPing` when `InvalidArgument` error.

franciscovalentecastro:
## Description
The *API Check* from the `healthchecks` package returns an `InvalidArgument` error when the Ops Agent is restarted very quickly :

```
[API Check] Result: ERROR, Detail: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: One or more points were written more frequently than the maximum sampling period configured for the metric.
```

This is due to the [maximum rate of "one point each 5 seconds" at which data can be written to a single time series](https://cloud.google.com/monitoring/quotas#custom_metrics_quotas). This PR retries `monitoringPing` 2 times with a 6 second backoff to avoid going above the rate.

## Related issue
b/291631906

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/internal/healthchecks/api_check.go
+++ b/internal/healthchecks/api_check.go
@@ -36,6 +36,7 @@ const (
 	ServiceDisabled              = "SERVICE_DISABLED"
 	AccessTokenScopeInsufficient = "ACCESS_TOKEN_SCOPE_INSUFFICIENT"
 	IamPermissionDenied          = "IAM_PERMISSION_DENIED"
+	MaxMonitoringPingRetries     = 3
 )
 
 func getGCEMetadata() (resourcedetector.GCEResource, error) {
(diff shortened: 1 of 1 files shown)


avilevy18 (approved):

internal/healthchecks/api_check.go
```diff
@@ -49,11 +50,12 @@ func getGCEMetadata() (resourcedetector.GCEResource, error) {
...
 }
 
-// monitoringPing reports whether the client's connection to the monitoring service and the
-// authentication configuration are valid. To accomplish this, monitoringPing writes a
-// time series point with empty values to an Ops Agent specific metric.
-// This method mirrors the "(c *Client) Ping" method in "cloud.google.com/go/logging".
-func monitoringPing(ctx context.Context, client monitoring.MetricClient, gceMetadata resourcedetector.GCEResource) error {
+func isInvalidArgumentErr(err error) bool {
+	apiErr, ok := err.(*apierror.APIError)
+	return ok && apiErr.GRPCStatus().Code() == codes.InvalidArgument
```
avilevy18, line 55: Do we want to consider matching the description as well as the code?

franciscovalentecastro: `InvalidArgument` can appear in a malformed metric write request. But in this case, the `monitoringPing` is generated to be correct. So it may have value checking for the specific error `One or more points were written more frequently ...` messaged.

The error `APIError` has the method `Reason()` that could match the text. But I'm unsure if that would be maintainable if the message changes in the future. We are also retrying once, so this won't extend long in an edge case.

PR #1426, opened 2023-09-18
Template assumes `aarch64` for `.Arch`

jefferbrecht:
## Description
`.Arch` comes directly from the user-provided env variable into the helper script, which I'd prefer to be `aarch64` for the sake of being the more common ARM counterpart to `x86_64`.

That being said, all of the ARM configs here are already suffixed with `_arm64`, so it makes sense for the generated files to still contain `arm64`, hence the conversion inside the template.

## Related issue
b/292100890

## How has this been tested?
Will test on my internal CL after this is merged.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


martijnvans (changes requested):
This feels weird because it doesn't generalize well at all to having more than two architectures. I'll leave a suggestion on your script CL.

PR #1433, opened 2023-09-20
healthchecks: Improve `monitoringPing` retry logic.

franciscovalentecastro:
## Description
The previously merged PR https://github.com/GoogleCloudPlatform/ops-agent/pull/1415 fixed the following error when restarting the Ops Agent very quickly :
```
[API Check] Result: ERROR, Detail: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: One or more points were written more frequently than the maximum sampling period configured for the metric.
```

This PR improves the retry logic to avoid waiting 6 seconds after the second metric write request.

## Related issue
b/291631906

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/internal/healthchecks/api_check.go
+++ b/internal/healthchecks/api_check.go
@@ -100,19 +100,21 @@ func createMonitoringPingRequest(gceMetadata resourcedetector.GCEResource) *moni
 func monitoringPing(ctx context.Context, client monitoring.MetricClient, gceMetadata resourcedetector.GCEResource) error {
 	// The retry logic fixes b/291631906 when the `CreateTimeSeries` request is retried very quickly
 	// resulting in an `InvalidArgument` error due a maximum write rate of one point every 5 seconds.
-	// https://cloud.google.com/monitoring/quotas 
+	// https://cloud.google.com/monitoring/quotas
 	pingBackoff := backoff.WithMaxRetries(backoff.NewConstantBackOff(6 * time.Second), MaxMonitoringPingRetries)
-	ticker := backoff.NewTicker(pingBackoff)
-
-	var err error
-	for range ticker.C {
+	pingOperation := func() error {
 		err := client.CreateTimeSeries(ctx, createMonitoringPingRequest(gceMetadata))
-		if err == nil || !isInvalidArgumentErr(err) {
-			ticker.Stop()
-			break
+		if !isInvalidArgumentErr(err) {
+			// This is a permanent API misconfiguration error.
+			return backoff.Permanent(err)
 		}
+		return err
 	}
 
+	err := backoff.Retry(pingOperation, pingBackoff)
+	if errors.Is(err, *backoff.PermanentError) {
+		return err.Unwrap()
+	}
 	return err
 }
 
--- a/internal/healthchecks/api_check.go
+++ b/internal/healthchecks/api_check.go
@@ -51,11 +51,6 @@ func getGCEMetadata() (resourcedetector.GCEResource, error) {
 	return resourcedetector.GCEResource{}, fmt.Errorf("not in GCE")
 }
 
-func isInvalidArgumentErr(err error) bool {
-	apiErr, ok := err.(*apierror.APIError)
-	return ok && apiErr.GRPCStatus().Code() == codes.InvalidArgument
-}
-
 func createMonitoringPingRequest(gceMetadata resourcedetector.GCEResource) *monitoringpb.CreateTimeSeriesRequest {
 	metricType := "agent.googleapis.com/agent/ops_agent/enabled_receivers"
 	now := &timestamppb.Timestamp{
@@ -102,15 +97,8 @@ func monitoringPing(ctx context.Context, client monitoring.MetricClient, gceMeta
 	// be called multiple times in quick succession, the first attempted request to `CreateTimeSeries`
 	// may fail. We can retry the request >5 seconds later in such cases.
 	// https://cloud.google.com/monitoring/quotas
-	pingBackoff := backoff.WithMaxRetries(backoff.NewConstantBackOff(6 * time.Second), MaxMonitoringPingRetries)
-	pingOperation := func() error {
-		err := client.CreateTimeSeries(ctx, createMonitoringPingRequest(gceMetadata))
-		if !isInvalidArgumentErr(err) {
-			// This is a permanent API misconfiguration error.
-			return backoff.Permanent(err)
-		}
-		return err
-	}
+	pingBackoff := backoff.WithMaxRetries(backoff.NewConstantBackOff(6*time.Second), MaxMonitoringPingRetries)
+	pingOperation := func() error { return client.CreateTimeSeries(ctx, createMonitoringPingRequest(gceMetadata)) }
 	return backoff.Retry(pingOperation, pingBackoff)
 }
 
--- a/internal/healthchecks/api_check.go
+++ b/internal/healthchecks/api_check.go
@@ -110,7 +110,6 @@ func monitoringPing(ctx context.Context, client monitoring.MetricClient, gceMeta
 		}
 		return err
 	}
-
 	return backoff.Retry(pingOperation, pingBackoff)
 }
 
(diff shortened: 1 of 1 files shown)


internal/healthchecks/api_check.go
```diff
@@ -97,18 +97,15 @@ func createMonitoringPingRequest(gceMetadata resourcedetector.GCEResource) *moni
 // time series point with empty values to an Ops Agent specific metric.
 // This method mirrors the "(c *Client) Ping" method in "cloud.google.com/go/logging".
 func monitoringPing(ctx context.Context, client monitoring.MetricClient, gceMetadata resourcedetector.GCEResource) error {
-	var err error
-	for i := 0; i < MaxMonitoringPingRetries; i++ {
```
jefferbrecht, line 101: Looks like we can delete `MaxMonitoringPingRetries` now?

internal/healthchecks/api_check.go
```diff
@@ -97,18 +97,15 @@ func createMonitoringPingRequest(gceMetadata resourcedetector.GCEResource) *moni
...
+	err := client.CreateTimeSeries(ctx, createMonitoringPingRequest(gceMetadata))
+	if err == nil || !isInvalidArgumentErr(err) { 
+		return err
 	}
-	return err
+	// This fixes b/291631906 when the monitoringPing is retried very quickly resulting
+	// in an `InvalidArgument` error due a maximum write rate of one point every 5 seconds.
+	// https://cloud.google.com/monitoring/quotas 
+	time.Sleep(6 * time.Second)
+	return client.CreateTimeSeries(ctx, createMonitoringPingRequest(gceMetadata))
```
jefferbrecht, line 108: Generally we should use `backoff` for implementing retry logic.

jefferbrecht (changes requested):

internal/healthchecks/api_check.go
```diff
@@ -97,17 +98,21 @@ func createMonitoringPingRequest(gceMetadata resourcedetector.GCEResource) *moni
 // time series point with empty values to an Ops Agent specific metric.
 // This method mirrors the "(c *Client) Ping" method in "cloud.google.com/go/logging".
 func monitoringPing(ctx context.Context, client monitoring.MetricClient, gceMetadata resourcedetector.GCEResource) error {
+	// The retry logic fixes b/291631906 when the `CreateTimeSeries` request is retried very quickly
+	// resulting in an `InvalidArgument` error due a maximum write rate of one point every 5 seconds.
+	// https://cloud.google.com/monitoring/quotas 
+	pingBackoff := backoff.WithMaxRetries(backoff.NewConstantBackOff(6 * time.Second), MaxMonitoringPingRetries)
+	ticker := backoff.NewTicker(pingBackoff)
```
jefferbrecht, line 105: `backoff.Retry` is probably simpler than using a ticker + loop. Example: https://github.com/GoogleCloudPlatform/ops-agent/blob/ef7383bce689b86bd28596a9e37400877b6fcb86/integration_test/gce/gce_testing.go#L957-L961

franciscovalentecastro: This is my thought process for not using `backoff.Retry` and going for the `ticker + loop` instead.

1. Not every `error` returned from the `client.CreateTimeSeries(ctx, createMonitoringPingRequest(gceMetadata))` operation implies the need for a retry, only `InvalidArgument` needs a retry. All other errors are helpful to determine the API config state.
2. We would need to use `PermanentError` to determine `good` and `bad` (retriable) errors.
3. There would be a need for `Unwrap` logic for the `good` errors.
4. `ticker + loop` is simpler to read and understand.

internal/healthchecks/api_check.go
franciscovalentecastro: Thought about this a little decided to use `backoff` to implement the retry logic, instead of what's presented here. Using `backoff` solves the issue from the previous implementation, waiting 6 seconds after last retry and it is more flexible if in the future the retry logic has to be modified.

internal/healthchecks/api_check.go
franciscovalentecastro: Well, took a crack at it and it looks nicer with `backoff.Retry` since there is no need to `Unwrap` the errors because `runMonitoringCheck` uses `As` to verify the error chain.

internal/healthchecks/api_check.go
franciscovalentecastro: Setting `MaxMonitoringPingRetries = 1` now with the `backoff.Retry` implementation.

internal/healthchecks/api_check.go
```diff
@@ -97,18 +98,19 @@ func createMonitoringPingRequest(gceMetadata resourcedetector.GCEResource) *moni
...
 // This method mirrors the "(c *Client) Ping" method in "cloud.google.com/go/logging".
 func monitoringPing(ctx context.Context, client monitoring.MetricClient, gceMetadata resourcedetector.GCEResource) error {
-	var err error
-	for i := 0; i < MaxMonitoringPingRetries; i++ {
-		err = client.CreateTimeSeries(ctx, createMonitoringPingRequest(gceMetadata))
-		if err == nil || !isInvalidArgumentErr(err) { 
-			break
+	// The retry logic fixes b/291631906 when the `CreateTimeSeries` request is retried very quickly
+	// resulting in an `InvalidArgument` error due a maximum write rate of one point every 5 seconds.
+	// https://cloud.google.com/monitoring/quotas
```
jefferbrecht, line 103: Nit: the public documentation describing quotas is sufficient to explain the code here, we don't need to reference the internal ticket.

```suggestion
	// Points written to a time series must be at least 5 seconds apart. Because `monitoringPing` might
	// be called multiple times in quick succession, the first attempted request to `CreateTimeSeries`
	// may fail. We can retry the request >5 seconds later in such cases.
	// https://cloud.google.com/monitoring/quotas
```

internal/healthchecks/api_check.go
```diff
@@ -97,18 +98,19 @@ func createMonitoringPingRequest(gceMetadata resourcedetector.GCEResource) *moni
...
+	pingBackoff := backoff.WithMaxRetries(backoff.NewConstantBackOff(6 * time.Second), MaxMonitoringPingRetries)
+	pingOperation := func() error {
+		err := client.CreateTimeSeries(ctx, createMonitoringPingRequest(gceMetadata))
+		if !isInvalidArgumentErr(err) {
+			// This is a permanent API misconfiguration error.
+			return backoff.Permanent(err)
```
jefferbrecht, line 109: Optional: the overhead of retrying the request in the case of a permanent error is not significant. We could simply remove both this `if{}` and the `isInvalidArgumentErr` function and I don't think it would ever affect anybody. Up to you.

jefferbrecht (approved):
LGTM with some additional minor/optional comments

internal/healthchecks/api_check.go
franciscovalentecastro: I removed the `if{}` and `isInvalidArgumentErr` so all errors are retried. This helps to make sure all other errors are permanent by retrying once.

After the change `TestNetworkHealthCheck` started to fail. I set `Unavailable` grpc response as a connection issue to fix the failure. With the retries its clear that `Unvailable` is the correct way to detect no connection (when firewall blocks connection), since with the current default 12 sec timeout it returned `DeadlineExceeded`. Keeping both as `MonApiConnErr` seems to be ok. The `Network Check` has `Exponential Backoff`, so no need to do the same in the `API Check`.

```
case codes.Unavailable:
	return MonApiConnErr
```

jefferbrecht (approved):
LGTM

internal/healthchecks/api_check.go
```diff
@@ -97,18 +93,13 @@ func createMonitoringPingRequest(gceMetadata resourcedetector.GCEResource) *moni
...
-		// https://cloud.google.com/monitoring/quotas 
-		time.Sleep(6 * time.Second)
-	}
-	return err
+	// Points written to a time series must be at least 5 seconds apart. Because `monitoringPing` might
+	// be called multiple times in quick succession, the first attempted request to `CreateTimeSeries`
+	// may fail. We can retry the request >5 seconds later in such cases.
+	// https://cloud.google.com/monitoring/quotas
+	pingBackoff := backoff.WithMaxRetries(backoff.NewConstantBackOff(6*time.Second), MaxMonitoringPingRetries)
+	pingOperation := func() error { return client.CreateTimeSeries(ctx, createMonitoringPingRequest(gceMetadata)) }
```
igorpeshansky, line 101: @franciscovalentecastro Am I missing something, or will we now also retry `InvalidArgument` errors?

jefferbrecht: See https://github.com/GoogleCloudPlatform/ops-agent/pull/1433#discussion_r1333368553

igorpeshansky: Thanks, missed that.

PR #1438, opened 2023-09-22
Add RPM version string substitutions to only allow valid characters.

igorpeshansky:
## Description
Add version string substitutions when building RPM packages to only allow valid characters.

According to the [Fedora Packaging Guildelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_upstream_uses_invalid_characters_in_the_version), only "ASCII letters (upper and lower case), digits and periods" are allowed in RPM version strings.

## Related issue
[b/267347168](http://b/267347168)

## How has this been tested?

Temporarily pushed a commit with an invalid RPM version string in `VERSION`, and observed that the automated builds [succeeded](https://source.cloud.google.com/results/invocations/b227af4e-2480-4284-9bc4-924473ac2928). Without these changes, the build failed with:
```
+ rpmbuild --define '_source_filedigest_algorithm md5' --define 'package_version 2.41.0-extrastuff' --define '_sourcedir /work' --define '_rpmdir /work' -ba pkg/rpm/google-cloud-ops-agent.spec
error: line 16: Illegal char '-' in: Version: 2.41.0-extrastuff
```

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


sophieyfang (approved):

pkg/rpm/google-cloud-ops-agent.spec
```diff
@@ -38,7 +38,7 @@ The Google Cloud Ops Agent collects metrics and logs from the system.
 %install
 cd %{_sourcedir}
 build_distro=%{dist}
-CODE_VERSION=%{version} BUILD_DISTRO=${build_distro#.} DESTDIR="%{buildroot}" ./build.sh
```
quentinmit, line 41: Why are you eliminating this logic? This is to make sure that the RPM version gets passed into build.sh; without this it's possible that the RPM version and the version that Ops Agent reports at runtime with will differ.

igorpeshansky: See the commit message of 3328aa6d3adf8e08c33e5c9f1b8675f39bb4a717. Previously, `CODE_VERSION` was always the same as `PKG_VERSION` in our automated builds, so it was a no-op from the automated build standpoint. With this PR, the RPM version is actually a transformed `PKG_VERSION`, which makes the `CODE_VERSION` inconsistent between RPM and DEB builds.

If you want to override the version in manual builds, you could just edit `PKG_VERSION`, right? Can you please give an example of a use case where you'd want the RPM package version (and thus the version reported in the agent metrics) to differ from `PKG_VERSION`?

quentinmit: The package version will already be different from the Ops Agent version; distros sometimes add various prefixes when building packages (such as `~focal`). I guess the question is whether we want metrics to report the final package version or the underlying source version. IMO we should report the package version. Also it just feels wrong to have two different layers of the stack independently determine the version number - it's supposed to be the package manager's job to track the version number.

igorpeshansky: The logic I ripped out actually stripped off those extra suffixes. So, regardless of my opinion below, this PR is preserving the original behavior for numeric versions. The only question is whether we should allow overriding the `PKG_VERSION` variable at build time.

I believe that the version compiled into the agent itself really ought to be the agent version, not the package version. Other tools (like [OS inventory manager](https://cloud.google.com/compute/docs/instances/os-inventory-management)) should be available to report the package versions. Looking at it another way, the metrics ought to report the version of the _code_, not the _package_.

quentinmit: I guess I'm thinking about cases where the package patches the code. I don't think it's necessarily a valid assumption that the code is the same across all package builds. But I think this is a nit, and we can always change this in the future, so *shrug*.

quentinmit (approved):

PR #1489, opened 2023-11-02
Fix sles tests

LujieDuan:
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


avilevy18 (approved):
Spoke offline about 1 nit, feel free to disregard if you don't think its useful

LujieDuan:
On SLES 15 ARM where we install Python311 from the vendored repo, it will try to get other dependencies `libpython3_11-1_0 python311 python311-base shared-python-startup` from the system repo and can timeout. Adding those to the test-repo and will retry the tests.

integration_test/gce/gce_testing.go
```diff
@@ -1500,57 +1500,66 @@ func InstallGsutilIfNeeded(ctx context.Context, logger *log.Logger, vm *VM) erro
...
-		// Use a vendored repo to reduce flakiness of the external repos.
-		// See http://go/sdi/releases/build-test-release/vendored for details.
-		repoArch := "x86-64"
-		if IsARM(vm.Platform) {
-			repoArch = "aarch64"
+	gcloudArch := "x86_64"
+	if IsARM(vm.Platform) {
+		gcloudArch = "arm"
+	}
+	gcloudPkg := "google-cloud-cli-453.0.0-linux-" + gcloudArch + ".tar.gz"
```
igorpeshansky, line 1507: Do they maintain a direct link to the latest tarball? If not, is there a way to retrieve the latest version programmatically?
We might be adding yet another chore to periodically upgrade gcloud on SLES/OpenSUSE or let that setup diverge from other distros.

integration_test/gce/gce_testing.go
```diff
@@ -1500,57 +1500,66 @@ func InstallGsutilIfNeeded(ctx context.Context, logger *log.Logger, vm *VM) erro
...
-	}
-
-	installCmd := `set -ex
+		}
 
+		installCmd = `set -ex
 ` + repoSetupCmd + `
-sudo zypper --non-interactive install --force-resolution --capability 'python>=3.6'
-sudo zypper --non-interactive install python3-certifi
+sudo zypper --non-interactive install python311 python3-certifi
```
igorpeshansky, line 1546: Why did we call the vendored package `python311` rather than just `python3`?

integration_test/gce/gce_testing.go
LujieDuan: Good point! Though I couldn't find a direct link for the tarball installation method. Modified the script to run `gcloud components update` after the installation to keep the cli to the latest version. I think alternatively we could also use the installer method to download the latest version as before, and use the VERSION there to download the latest tarball.

integration_test/gce/gce_testing.go
LujieDuan: for SLES 15 and OpenSUSE Leap, in the system repo, `Python3` only has version 3.6 available and `Python311` is an alternative installation. `yum upload` would keep the name when uploading to our repo. Is there anyway we can change the name of the downloaded RPM?

PR #1492, opened 2023-11-02
prometheus: add CustomUnmarshaler

braydonk:
## Description
This PR upgrades `go-yaml` to get new functionality that allows you to register custom unmarshalers for specific data types. This allows us to add a custom marshaler when we're marshalling Prometheus structs to ensure we render the actual value of `config.Secret` fields.

## Related issue
[b/306723825](http://b/306723825)

## How has this been tested?
Unit tests

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


braydonk:
I was unable to update to the newest version of go-yaml because of a breaking change that I didn't have time to track down

confgenerator/otel/modular.go
```diff
@@ -97,7 +98,12 @@ func configToYaml(config interface{}) ([]byte, error) {
 	if err := mapstructure.Decode(config, &outMap); err != nil {
 		return nil, err
 	}
-	return yaml.Marshal(outMap)
+	return yaml.MarshalWithOptions(
+		outMap,
+		yaml.CustomMarshaler[commonconfig.Secret](func(s commonconfig.Secret) ([]byte, error) {
```
igorpeshansky, line 103: I just realized that [`modular.go:97`](https://github.com/GoogleCloudPlatform/ops-agent/blob/2b7466370c9aad268489f3ce4e1b43bdc218aef1/confgenerator/otel/modular.go#L97) may be addressing the same problem a different way. Should we be using a uniform approach? Want to at least add a TODO?

confgenerator/testdata/goldens/metrics-receiver_prometheus_basic_auth/golden/windows-2012/4d6012ff003886818fb9b9285b4af962.lua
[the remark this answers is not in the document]
braydonk: This is automatic; all platforms are always added now. I don't remember if there's a way not to.

confgenerator/otel/modular.go
braydonk: I think you're right; the output is identical if I remove the mapstructure decode first. So I think I'll remove that here and update the comment.

braydonk: Actually, maybe I should ask @ridwanmsharif what that was here for and if there's a reason for it I'm not seeing.

igorpeshansky: Is the output also identical if you remove the custom marshaller here (and only use `mapstructure.Decode`)? I suspect the main issue was that `deepCopy` marshalled the unmodified config struct…

braydonk: In that scenario (no custom unmarshaler, just mapstructure decode), the output still has `secret`.

igorpeshansky: Ah, ok. We can clean this up another time.

igorpeshansky (approved):
LGTM :shipit:

confgenerator/otel/modular.go
ridwanmsharif: I'm not aware of any reason why that was there. I suspect it has something to do with us using OTel structs directly in the receiver configs (like https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b39f83d56853ca03fc8edd196eedf65e5046b9fb/receiver/apachereceiver/config.go#L16-L20) and if they use `mapstructure` annotations, they would need to be decoded appropriately using the annotations before we marshal them into yaml.

With that said, there is nothing prometheus specific about this, and wasn't there because of the prometheus structs. I suspect if none of our golden files change when we remove this `mapstructure.Decode` that might just be because the structs in the pipeline don't have any complicated decoding rules defined. But that is by convention and keeping things as is will keep this less error prone.

braydonk: Thanks Ridwan, sounds like we should leave it in there for now.

braydonk:
The Windows unit test is Coveralls complaining for (as far as I can tell) no reason.

SLES are all failing because of the Cloud SDK breakage.

Debian 11 failure looks like a flake but I am not 100% sure.

braydonk:
I found out why upgrading to the latest yaml library was breaking. They change how they handle cases where there is stray whitespace at the beginning or end of the string by instead adding quotes. Not sure why they opted to break their behaviour in that way tbh. But I've instead just removed the stray whitespace from our one metric definition. This allows us to get up to date with the latest version of `go-yaml`.

igorpeshansky (approved):
LGTM :shipit:

igorpeshansky (approved):

PR #1505, opened 2023-11-08
Fixes or skips Debian 12 Aarch64 Tests

StefanKurek:
## Description
Fixes or skips Debian 12 Aarch64 tests. Here are all tests passing with the Bullseye overwrite hack in place
image

Hack has since been removed.

## Related issue
NA

## How has this been tested?
NA

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/third_party_apps_data/applications/mariadb/metadata.yaml
+++ b/integration_test/third_party_apps_data/applications/mariadb/metadata.yaml
@@ -149,6 +149,7 @@ expected_metrics:
     unavailable_on:
       - ubuntu-2310-amd64
       - ubuntu-2310-arm64
+      - debian-12-arm64
   - type: workload.googleapis.com/mysql.sorts
     value_type: INT64
     kind: CUMULATIVE
--- a/integration_test/third_party_apps_data/applications/aerospike/metadata.yaml
+++ b/integration_test/third_party_apps_data/applications/aerospike/metadata.yaml
@@ -132,6 +132,7 @@ supported_operating_systems: linux
 platforms_to_skip:
   # Unable to install Aerospike on various distros.
   - debian-12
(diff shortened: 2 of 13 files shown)


integration_test/third_party_apps_data/applications/mariadb/metadata.yaml
```diff
@@ -149,6 +149,7 @@ expected_metrics:
     unavailable_on:
       - ubuntu-2310-amd64
       - ubuntu-2310-arm64
+      - debian-12-arm64
```
quentinmit, line 152: Why is `mariadb` being skipped on `debian-12-arm64`? It runs on `debian-12` (non-`arm64`), and since `mysql` is also skipped this means we have no MySQL/MariaDB coverage.

StefanKurek: @quentinmit just this one metric seems to be an issue for debian-12-arm64 (and I should only be skipping the one metric and not the entire test). 

From what I understood, there was a known discussion about this metric being problematic at one point so I assumed this wouldn't be a problem to skip. Let me know if this rings any bells or if you want to further discuss.

quentinmit: Ah, the diff hunk was misleading and I thought it was talking about all the metrics, not just one. Line 44 is actually skipping MariaDB entirely on `debian-12`, with a different reason. If the tests otherwise pass aside from this one metric now, LGTM.

quentinmit (approved):

StefanKurek:
@quentinmit there seems to be a persistent windows test that both failing and unrelated to these changes. I think it should be good to merge when you are ready.

PR #1506, opened 2023-11-09
Testing: Update the _zones_for_arch in common.gcl to use name x86_64

LujieDuan:
## Description
~~Remove the `arch` from the Mantic x86_64 gcls and use the default value of `amd64`. `_zones_for_arch` uses `amd64` and `aarch64` as keys to pick the zones.~~

Update the `_zones_for_arch` in common.gcl to use name `x86_64`. 

## Related issue
[b/305930083](http://b/305930083)

## How has this been tested?
Mantic x86_64's integration tests and 3P tests are passing. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/third_party_apps_data/applications/apache/metadata.yaml
+++ b/integration_test/third_party_apps_data/applications/apache/metadata.yaml
@@ -22,7 +22,6 @@ description: |-
   number of open connections or incoming requests. The integration also collects
   access and error logs. Access logs are parsed into a JSON payload focused on
   request details, and error logs are parsed for their error code and message.
-  test
 configure_integration: |-
   To collect telemetry from your Apache Web Server, you must configure the
   server's `httpd.conf` file to enable the
--- a/integration_test/third_party_apps_data/applications/apache/metadata.yaml
+++ b/integration_test/third_party_apps_data/applications/apache/metadata.yaml
@@ -22,6 +22,7 @@ description: |-
   number of open connections or incoming requests. The integration also collects
   access and error logs. Access logs are parsed into a JSON payload focused on
   request details, and error logs are parsed for their error code and message.
+  test
 configure_integration: |-
   To collect telemetry from your Apache Web Server, you must configure the
   server's `httpd.conf` file to enable the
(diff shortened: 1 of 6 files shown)


kokoro/config/test/ops_agent/mantic_x86_64.gcl
```diff
@@ -4,6 +4,5 @@ import '../image_lists.gcl' as image_lists
 config build = common.ops_agent_test {
   params {
     platforms = image_lists.mantic_x86_64.presubmit
```
sophieyfang, line 6: I see all the arm64 gcl files still keep the arch field and only this one removes it. Why is it?

jefferbrecht:
If these configs were generated by the script then we should also fix the underlying cause -- e.g. we might need to update the template files (https://github.com/GoogleCloudPlatform/ops-agent/blob/master/kokoro/config/test/ops_agent/template.gcl.tmpl and a few others) to only output the arch if it's `!= 'x86_64`'.

kokoro/config/test/ops_agent/mantic_x86_64.gcl
LujieDuan: For the x86 gcls, if we omit the `arch` parameter, the final configuration will use the `arch = 'amd64'` value from `common.gcl` and then pick the zones. For arm64 gcls we need to keep the `arch` parameter in each gcl, so that it will override the default value in `common.gcl`.

LujieDuan:
Discussed with @martijnvans it is okey for us to change the name in `_zones_for_arch` from `amd64` to `x86_64` and we don't need to update the kokoro tool or the templates. Reverting the previous change the updating the name.

integration_test/third_party_apps_data/applications/apache/metadata.yaml
```diff
@@ -22,6 +22,7 @@ description: |-
   number of open connections or incoming requests. The integration also collects
   access and error logs. Access logs are parsed into a JSON payload focused on
   request details, and error logs are parsed for their error code and message.
+  test
```
martijnvans, line 25: please revert

martijnvans (approved):
1 thing to revert before merging

PR #1519, opened 2023-11-17
Using exclude_metrics to turn off receivers

LujieDuan:
## Description
Add the feature that built-in receivers can be turned off by specifying all of the metrics in a `exclude_metrics` processor. 

## Related issue
[b/305224523](http://b/305224523)

## How has this been tested?
Unit tests added/updated:
1. `metrics-default_overrides_disable_all` now does not generate `nvml` receiver anymore; 
2. `metrics-processor_three_exclude_metrics_processors_not_disable_nvml` tests with three `exclude_metrics` processors in the default pipeline; the second processor excludes all metrics from the `nvml` receiver, but this config won't turn off the `nvml` receiver since the processor merging will stop after the first processor; 
3. `metrics-processor_two_exclude_metrics_processors` tests with two processors that can exclude all `nvml` metrics in combination, but this configuration won't turn off the receiver since we merge the processors one at a time; 
4. `metrics-receiver_exclude_nvml_from_hostmetrics` tests the escape hatch to turn off the receiver by excluding all GPU metrics; 
5. `metrics-receiver_exclude_subset_of_nvml_metrics` tests the receiver is not turned off when only a subset of its metrics is excluded. 
6. `metrics-processor_exclude_all_nvml_metrics_individually` tests the `nvml` receiver is not generated when we exclude the metrics individually without the glob pattern. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


apps/hostmetrics.go
```diff
@@ -367,6 +386,19 @@ func (r MetricsReceiverHostmetrics) Pipelines(ctx context.Context) []otel.Receiv
 	return pipelines
 }
 
+// TODO:
+func canTurnOffReceiver(expectedMetrics []string, metricsPattern []string) bool {
+	matched := set.Set[string]{}
+	for _, m := range expectedMetrics {
+		for _, rp := range metricsPattern {
+			if r, _ := regexp.MatchString(strings.ReplaceAll(rp, "*", ".*"), m); r {
```
quentinmit, line 394: Why not make this a method on the `MetricsProcessorExcludeMetrics`?

Don't use regexp for this, because that will cause all sorts of weird behavior with metacharacters like `.`. Just use `filepath.Match(rp, m)`, and it can be as simple as:

```
func (p MetricsProcessorExcludeMetrics) AllMetricsExcluded(metrics ...string) bool {
	OUTER:
	for _, metric := range metrics {
		for _, excludePattern := range p.MetricsPattern {
			if filepath.Match(excludePattern, metric) {
				continue OUTER
			}
		}
		return false
	}
	return true
}
```

confgenerator/confgenerator.go
```diff
@@ -147,7 +147,17 @@ func (uc *UnifiedConfig) generateOtelPipelines(ctx context.Context) (map[string]
 	outR := make(map[string]otel.ReceiverPipeline)
 	outP := make(map[string]otel.Pipeline)
 	addReceiver := func(pipelineType, pID, rID string, receiver OTelReceiver, processorIDs []string) error {
-		for i, receiverPipeline := range receiver.Pipelines(ctx) {
+		pipelines := receiver.Pipelines(ctx)
```
quentinmit, line 150: This code needs a bit of refactoring to support all possible paths through, but for now (in a way that will only work if receivers are in a single pipeline) it can be as simple as:

```go
if pipelineType == "metrics" {
	for len(processorIDs) > 0 {
		if mr, ok := receiver.(MetricsProcessorMerger); ok {
			mergedReceiver, ok := mr.MergeMetricsProcessor(m.Processors[processorIDs[0]])
			if ok {
				receiver = mergedReceiver
				processorIDs = processorIDs[1:]
			}
		}
		break
	}
}
```

with no other changes to the function.

confgenerator/config.go
```diff
@@ -574,6 +578,10 @@ type OTelReceiver interface {
 	Pipelines(ctx context.Context) []otel.ReceiverPipeline
 }
 
+type ModifiableOtelReceiver interface {
+	MergeWithProcessor(ctx context.Context, processor MetricsProcessor) []otel.ReceiverPipeline
```
quentinmit, line 582: If an interface has one method it should be named after the method (maybe `MetricsProcessorMerger` in this case)

confgenerator/confgenerator.go
```diff
@@ -251,26 +261,13 @@ func contains(s []string, str string) bool {
 	return false
 }
 
-func processUserDefinedMultilineParser(i int, pID string, receiver LoggingReceiver, processor LoggingProcessor, receiverComponents []fluentbit.Component, processorComponents []fluentbit.Component) error {
-	var multilineParserNames []string
+func validateMultilineProcessor(i int, pID string, receiver LoggingReceiver, processor LoggingProcessor, receiverComponents []fluentbit.Component, processorComponents []fluentbit.Component) error {
```
quentinmit, line 264: Let's punt logging support to a separate PR and revert the logging changes here

confgenerator/confgenerator.go
LujieDuan: There is a case where `exclude_metrics` can have multiple patterns targeting the `nvml` receiver and the `hostmetrics` receiver:
```
metrics:
  processors:
    metrics_filter:
      type: exclude_metrics
      metrics_pattern:
      - agent.googleapis.com/gpu/*
      - agent.googleapis.com/cpu/*
```
Which means we will still need to add the `exclude_metrics` processor to the generated config. To me, we can choose to:
1. Handle this in the `MergeMetricsProcessor` function, where we first disable the nvml receiver, and then add this processor to the `hostmetrics` receiver pipeline; 
2. Let `MergeMetricsProcessor` disable the nvml receiver only; in `generateOtelPipelines()` we do not skip the processor and we add the processor as before, even if  `MergeMetricsProcessor()` returns a new receiver. 

I tried 2 here since I feels it is slightly simpler to implement; but as we are adding the processor as before, `MergeMetricsProcessor()` might be confusion since the processor is not completely merged into the receiver. I changed the interface to 
```
type ModifiableReceiver interface {
	ModifyReceiver(p MetricsProcessor) MetricsReceiver
}
```

confgenerator/config.go
LujieDuan: Done!

apps/hostmetrics.go
```diff
@@ -33,6 +36,10 @@ func (r MetricsReceiverHostmetrics) Type() string {
 }
 
 func (r MetricsReceiverHostmetrics) Pipelines(ctx context.Context) []otel.ReceiverPipeline {
+	return r.MergeWithProcessor(ctx, nil)
+}
+
+func (r MetricsReceiverHostmetrics) MergeWithProcessor(ctx context.Context, processor confgenerator.MetricsProcessor) []otel.ReceiverPipeline {
```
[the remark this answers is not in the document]
LujieDuan: This looks good! Followed this to update the PR. 
1. Changed `MergeMetricsProcessor` to `ModifyReceiver`, as discussed in the other comment; 
2. Changed `disableGPU` to be an exported property; our feature tracking uses recursive reflections on the fields - having a nonexported field here would cause the feature extraction to fail.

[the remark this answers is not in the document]
quentinmit: 2 is only relevant if we care about extracting this as a feature, though. Do we?

apps/hostmetrics.go
quentinmit: Hm, yeah, good catch on the different behavior of `*`. Refactoring to use the same logic makes sense.

confgenerator/confgenerator.go
quentinmit: The interface and method name need to be Metrics-specific because it's taking a `MetricsProcessor`; when we add logging support that's going to need to be a different method with a different name (taking a `LoggingProcessor`). So it can't just be `ModifyReceiver`.

Your new signature wouldn't allow for cases where we do need to suppress the processor (such as in logging). I actually think the unwritten option 3 is the best here:

```go
// MergeMetricsProcessor attempts to merge p into the current receiver.
// It returns the new receiver and a processor or nil.
MergeMetricsProcessor(p MetricsProcessor) (MetricsReceiver, MetricsProcessor)
```

Then we can have this use-case return `r, p` if we can't be sure the processor is a noop, and other use cases (like merging multiline) can return `r, nil`.

apps/metric_processors.go
```diff
@@ -51,6 +45,37 @@ func (p MetricsProcessorExcludeMetrics) Processors() []otel.Component {
 	)}
 }
 
+// globToRegex converts metrics glob patterns to regex patterns
+// For generating otel config, `$` needs to be escaped:
+// https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricstransformprocessor#rename-multiple-metrics-using-substitution
+func globToRegex(glob string, escapeForOtel bool) string {
+	dollarSign := "$"
```
quentinmit, line 52: Why not do the `strings.Replace` in `Processors()` above, and eliminate the `escapeForOtel` parameter here?

apps/metric_processors.go
```diff
@@ -51,6 +45,37 @@ func (p MetricsProcessorExcludeMetrics) Processors() []otel.Component {
...
+	var literals []string
+	for _, g := range strings.Split(glob, "*") {
+		literals = append(literals, strings.Replace(regexp.QuoteMeta(g), "$", dollarSign, -1))
+	}
+	return fmt.Sprintf(`^%s%s`, strings.Join(literals, `.*`), dollarSign)
+
+}
+
+// AllMetricsExcluded checks if its MetricsPattern list can match all of the
+// input metrics which would indicate all of the metics will be excluded
```
quentinmit, line 65: Typo: "metics" instead of "metrics"

confgenerator/confgenerator.go
```diff
@@ -147,6 +147,13 @@ func (uc *UnifiedConfig) generateOtelPipelines(ctx context.Context) (map[string]
 	outR := make(map[string]otel.ReceiverPipeline)
 	outP := make(map[string]otel.Pipeline)
 	addReceiver := func(pipelineType, pID, rID string, receiver OTelReceiver, processorIDs []string) error {
+		if pipelineType == "metrics" {
+			for _, processorId := range processorIDs {
+				if mr, ok := receiver.(ModifiableReceiver); ok {
+					receiver = mr.ModifyReceiver(m.Processors[processorId])
```
quentinmit, line 153: You need to abort this loop as soon as you fail to successfully merge a processor. (Otherwise a hypothetical rename a metric + exclude a metric pipeline might incorrectly exclude the wrong metric, for instance.)

apps/hostmetrics.go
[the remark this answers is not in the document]
LujieDuan: Yes, this field does not need to be tracked; changed this back to `disableGPUMetrics` and modified feature tracking to skip unexported fields.

confgenerator/confgenerator.go
LujieDuan: Option 3 makes sense! Updated to implement this with minor modification: 
```
MergeMetricsProcessor(p MetricsProcessor) (MetricsReceiver, MetricsProcessor, bool)
```
where the third return values can be used to break the loop when merging a sequence of processors.

apps/metric_processors.go
LujieDuan: Done!

apps/metric_processors.go
LujieDuan: Done!

confgenerator/confgenerator.go
LujieDuan: Done! Updated the loop so that it will only continue to merge the next processor when: the receiver is still a `ModifiableReceiver` and the merging is successful.

confgenerator/confgenerator.go
quentinmit: I think you can use `processor != nil` as the exit condition, can't you? What is the circumstance where your `bool` would have a different value:?

LujieDuan: With three returns values this can be a bit confusing for the callers; in this interface the second and third return values are for different purposes:
1. The bool indicates whether this `MetricsProcessorMerger` knows how to merge this processor; the hostmetrics knows how to merge with exclude_metrics but can't (shouldn't?) merge with metrics renaming processor when we have one; so we can use this to exit the loop;
2. The returned processor is used to indicate whether the processor can be skipped, in the case when the merging was done successfully; `hostmetrics` will always return the exclude_metrics processor unchanged, since the processor can have exclusion patterns targeting the otel hostmetrics receiver pipeline; `processor != nil` means the processor still need to be added to the final config, but we can continue the loop to merge the next processor; one case would be `metrics-processor_three_exclude_metrics_processors_disable_nvml`;

We could use the first return value to check for the exit condition and drop the boolean:  
```
	// MergeMetricsProcessor attempts to merge p into the current receiver.
	// It returns:
        // a new receiver; if nil, then can't merge the processor
        // a processor: if nil, then p has been merged into the new receiver and 
        // doesn't need to be added again
	MergeMetricsProcessor(p MetricsProcessor) (MetricsReceiver, MetricsProcessor, bool)
```

quentinmit: > 2\. The returned processor is used to indicate whether the processor can be skipped, in the case when the merging was done successfully; `hostmetrics` will always return the exclude_metrics processor unchanged, since the processor can have exclusion patterns targeting the otel hostmetrics receiver pipeline; `processor != nil` means the processor still need to be added to the final config, but we can continue the loop to merge the next processor; one case would be `metrics-processor_three_exclude_metrics_processors_disable_nvml`;

I think you have this backwards. The only time when it's possible to continue merging processors is when the returned processor is `nil`. In any other case, there is still a processor doing some amount of work, and that work might change the set of metrics (or logs, or traces) that the next processor receives. Only if there is no longer a processor do you know for sure what input will be going to the next processor in the list, and you can safely merge the next processor as well.

LujieDuan: Yes I got it backwards and this makes sense now. 
Reverted my change back to what was suggested here earlier - with the interface defined as 
```
MergeMetricsProcessor(p MetricsProcessor) (MetricsReceiver, bool)
```
the merging logic:
- can be used as a escape hatch to turn off the `nvml` receiver, as long as all the `nvml` metrics are excluded in the first processor (which is the case for the built-in pipeline);
- can handle the case where the processor exclude metrics from both `nvml` and `hostmetrics`.

quentinmit: I think you still want to return a new processor instead of just a bool, because you might have partially merged the processor but still have more work to do. No?

quentinmit: Thinking about it, the receiver can always save a pointer to the processor in its own struct, so I believe that covers every case where you might want to return a modified processor.

confgenerator/testdata/feature/golden.csv
```diff
@@ -186,6 +186,7 @@ App,Field,Override,
 *apps.MetricsReceiverHbase,confgenerator.ConfigComponent.Type,
 *apps.MetricsReceiverHbase,confgenerator.MetricsReceiverSharedCollectJVM.CollectJVMMetrics,
 *apps.MetricsReceiverHostmetrics,confgenerator.ConfigComponent.Type,
+*apps.MetricsReceiverHostmetrics,disableGPUMetrics,
```
quentinmit, line 189: Is this supposed to be here, since the field is skipped?

quentinmit (approved):

PR #1521, opened 2023-11-22
OTel logging support

quentinmit:
## Description
This adds very basic OTel logging support, including:
- `files` receiver
- `parse_json` processor
- `modify_fields` processor
- Transformation tests run simultaneously on OTel and fluent-bit
- `otel_logging` experiment flag that enables `logging.service.experimental_otel_logging: true` in config

TODO before I'm comfortable merging:
- [x] Merge https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/193
- [x] Merge https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/29929
- [x] Run otel transformation test as part of package build
- [x] Add `error` returns for config errors like the `operations` field
- [x] Validate that regexes are RE2-compatible
- [x] Add `exclude_logs` processor
- [x] Fix trace and span ID handling using `trace_id.string` and `span_id.string` (incl. stripping `projects/.*/traces/.*/` prefix)
- [x] Fix handling of log_file_name (set the right label, make sure it uses the correct choice of file name or file path)
- [x] Ignore errors in transformprocessor (such as time/number parsing) while still checking for them in transformation tests
- [ ] Configure file checkpoints/buffering
- [x] Wait for otel logs to shut down the collector instead of just sleeping for 5 seconds
- [ ] Check that otel config can be generated as part of validation

## Related issue
b/307783213

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/config.go
```diff
@@ -812,27 +825,38 @@ func (uc *UnifiedConfig) Validate(ctx context.Context) error {
...
 		log.Print(`The "logging.exporters" field is no longer needed and will be ignored. This does not change any functionality. Please remove it from your configuration.`)
 	}
 	if l.Service == nil {
 		return nil
 	}
+	validReceivers := map[string]bool{}
+	for k := range l.Receivers {
+		validReceivers[k] = true
+	}
+	if uc.Combined != nil {
```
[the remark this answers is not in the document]
quentinmit: Why do we have a helper for `HasCombined` that just does this? It's even more characters than the `!=nil` check, and it's used in exactly one place right now.

(BTW, I prefer seeing an explicit `!= nil` here because that's actually testing what is important - we need the `uc.Combined.Receivers` on the next line to not hit a nil-pointer dereference. `HasCombined` might have more complicated logic that doesn't necessarily correlate with `nil`.)

confgenerator/logging.go
```diff
@@ -47,6 +48,27 @@ func setLogNameComponents(ctx context.Context, tag, logName, receiverType string
 	}.Components(ctx, tag, "setlogname")
 }
 
+func otelSetLogNameComponents(_ context.Context, logName, hostName string) []otel.Component {
+	statements := []string{
+		// TODO: Prepend `receiver_id.` if it already exists, like the `fluent_forward` receiver?
+		fmt.Sprintf(`set(attributes["gcp.log_name"], %q) where attributes["gcp.log_name"] == nil`, logName),
+		fmt.Sprintf(`set(attributes["compute.googleapis.com/resource_name"], %q) where attributes["compute.googleapis.com/resource_name"] == nil`, hostName),
```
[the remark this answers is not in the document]
quentinmit: This was some of the first code I wrote for this PR, before any of the helpers existed. I've replaced this with a call to `LoggingProcessorModifyFields` just like the fluent-bit version above.

confgenerator/config.go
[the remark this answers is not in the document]
avilevy18: `HasCombined` was created to be consistent with `HasLogging` & `HasMetrics`.  

We might want to consider removing all 3 methods or, make sure we are using it consistently.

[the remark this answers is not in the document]
quentinmit: It looks like the `Has` versions of all three are barely being used; `cmd/ops_agent_windows/run_windows.go` is incorrectly using them to mean "has a non-empty logging/metrics section", while the implementions just check for the section existing at all (even if there are zero receivers or pipelines). The only other place they're used is `confgenerator/feature_tracking.go` which is using them in lieu of `nil` checks.

I'd rather not put that in this PR since it's already so large, but I think all three could be cleaned up and that would be a net improvement.

confgenerator/logging.go
```diff
@@ -47,6 +48,20 @@ func setLogNameComponents(ctx context.Context, tag, logName, receiverType string
...
+	return LoggingProcessorModifyFields{
+		Fields: map[string]*ModifyField{
+			// TODO: Prepend `receiver_id.` if it already exists, like the `fluent_forward` receiver?
+			"logName": {
+				DefaultValue: &logName,
+			},
+			`labels."compute.googleapis.com/resource_name"`: {
+				DefaultValue: &hostName,
+			},
+		},
```
jefferbrecht, line 61: Don't we need to move `agent.googleapis.com/log_file_path` from `jsonPayload` to a label here? https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/logging.go#L40-L42

quentinmit: No, that's only in the legacy config because fluent-bit populates the field in `jsonPayload`:

https://github.com/GoogleCloudPlatform/ops-agent/blob/7e24f4bca67a63cf43b6cf30bd326280263b0edf/confgenerator/logging_receivers.go#L117-L119

This is handled as part of the receiver pipeline in the otel version (though looking at the code it seems like it might be buggy - I should add some unit tests, since this isn't covered by transformation tests)

jefferbrecht: It needs to end up being a label at any rate, so SGTM if you want to handle it in the receiver pipeline instead. (I assume you mean changing [this line](https://github.com/GoogleCloudPlatform/ops-agent/pull/1521/files#diff-92305d0589ed0ffcedd1368edad5c31bb62f7bd7098aa1150c02625391db628dR198) to write to an attribute instead of body.)

confgenerator/logging_processors.go
```diff
@@ -126,6 +212,45 @@ func (p LoggingProcessorParseJson) Components(ctx context.Context, tag, uid stri
...
+		return nil, err
+	}
+
+	fromAccessor, err := m.OTTLAccessor()
+	if err != nil {
+		return nil, err
+	}
+
+	statements := ottl.NewStatements(
+		ottl.LValue{"body"}.SetIf(ottl.ParseJSON(fromAccessor), fromAccessor.IsPresent()),
```
jefferbrecht, line 231: This produces a runtime error and drops the log if the incoming log has a string-valued `Body`:

```
Jan 18 11:39:00 ops-agent otelopscol[11652]: 2024-01-18T11:39:00.197-0500#011error#011logs/processor.go:54#011failed processing logs#011{"error": "failed to execute statement: set(body, ParseJSON(body[\"message\"])) where (body != nil and body[\"message\"] != n
il), log bodies of type Str cannot be indexed"}
```

The log shouldn't be dropped if it can't be parsed, but beyond that, there should also be a supported way to parse JSON out of a string-valued body. The idea that immediately comes to mind is that we should allow `parse_json` to use the `textPayload` field since that's where a string-valued body would end up in the LogEntry.

We should fix the dropped logs in this PR but we can defer supporting `textPayload` to a later PR.

confgenerator/filter/internal/ast/ast.go
```diff
@@ -43,6 +46,22 @@ type Attrib interface{}
 // Each element of the slice is not yet unescaped (if needed).
 type Target []string
 
+var logEntryRootValueMapToOTel = map[string][]string{
+	"severity": {"severity_text"},
```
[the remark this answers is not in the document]
jefferbrecht: Strawman to resolve: if we parse that field and write to `severity_text`, we should also zero out `severity_number` and let the exporter fill in the missing number using its own default mapping: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/logs.go#L439-L441

confgenerator/confgenerator.go
```diff
@@ -191,12 +196,26 @@ func (uc *UnifiedConfig) generateOtelPipelines(ctx context.Context) (map[string]
...
+				} else if pipelineType == "logs" {
+					var p LoggingProcessor
+					p, ok = l.Processors[pID]
+					if ok {
+						processor, ok = p.(OTelProcessor)
+					}
+				}
+				// TODO: Add trace processors?
 				if !ok {
 					return fmt.Errorf("processor %q not found", pID)
```
jefferbrecht, line 212: `!ok` here can mean either "processor not found" or "processor is not an OTel processor"; can we distinguish the error message better?

confgenerator/confmerger.go
```diff
@@ -47,8 +48,8 @@ func MergeConfFiles(ctx context.Context, userConfPath string, builtInConfStructs
 
 	// Ensure the merged config struct fields are valid.
 	v := newValidator()
-	if err := v.Struct(result); err != nil {
-		panic(err)
+	if err := v.StructCtx(ctx, result); err != nil {
+		log.Fatalf("merged config failed to validate: %v", err)
```
jefferbrecht, line 52: Why change this from a panic to a log? It should be impossible for the merged config to be invalid if the user config is valid (and we've already validated the user config by this point).

quentinmit: `log.Fatalf` still panics, it just prints the panic message in a nicer format.

confgenerator/logging_processors.go
quentinmit: I think the existing code would actually work if you configured `field: jsonPayload` on the processor. We have to default to `jsonPayload.message` for backwards compatibility.

As far as runtime errors go, I already have a TODO in this PR to make sure that we ignore parse failures. And I can add `textPayload` as an alias for `body`, but I think that's about the best I can do. We can't make `parse_json` automagically work because we already promised it would look for `jsonPayload.message`.

jefferbrecht: `field: jsonPayload` feels weird since users will see the log in `textPayload` if left unparsed. Agreed that we shouldn't change the default behaviour when `field` is left unspecified, I actually meant allowing users to specify `field: textPayload` (which I think is consistent with your idea of aliasing `textPayload` to `body`).

confgenerator/filter/internal/ast/ast.go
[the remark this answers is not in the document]
quentinmit: Oh, oops. I misread the exporter code. I thought it overwrote `severityNumber` if `SeverityText` is present, but now I see there's a `severityNumber == 0` guard.

Arguably the `severity_text` setter should itself clear or transform `severity_number`, but they don't appear to do that today: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/contexts/ottllog/log.go#L310

confgenerator/logging.go
quentinmit: I fixed the receiver, and added an integration test for this.

confgenerator/logging_processors.go
quentinmit: I've added `textPayload` as a valid `field`.

confgenerator/confgenerator.go
quentinmit: Yep, I added a better error message.

confgenerator/filter/internal/ast/ast.go
```diff
@@ -393,13 +486,57 @@ func (r Restriction) FluentConfig(tag, key string) ([]fluentbit.Component, strin
...
+		expr = ottl.IsMatch(lhs, r.RHS)
+		// TODO: Support Ruby regex syntax
+
+		if r.Operator == "!~" {
+			expr = ottl.Not(expr)
+		}
+	case "=", "!=":
+		// equality, case insensitive
+		expr = ottl.IsMatch(lhs, fmt.Sprintf(`(?i)^%s$`, regexp.QuoteMeta(r.RHS)))
+		if r.Operator == "!~" {
```
jefferbrecht, line 518:
```suggestion
		if r.Operator == "!=" {
```

confgenerator/otel/modular.go
```diff
@@ -114,6 +114,12 @@ type ModularConfig struct {
 	Pipelines         map[string]Pipeline
 
 	Exporters map[ExporterType]Component
+
+	// Test-only options:
+	// Don't generate any self-metrics (
```
jefferbrecht, line 119:
```suggestion
	// Don't generate any self-metrics
```

confgenerator/otel/ottl/ottl.go
```diff
@@ -0,0 +1,253 @@
...
+	)
+	return out
+}
+
+// Delete removes a (potentially nested) key from its parent maps, if that key exists.
+func (a LValue) Delete() Statements {
+	parent := a[:len(a)-1]
+	child := a[len(a)-1]
+	return Statements{
+		statementf(`delete_key(%s, %q) where %s`, parent, child, a.IsPresent()),
```
jefferbrecht, line 216: Isn't `child` already a `Value` here (so, `%s` instead of `%q`)?

jefferbrecht: Or I think I got it backwards -- `parent` is a string, so the `%s` should be a `%q` I think?

jefferbrecht: Nevermind, one's a path and one's a string for some reason.

jefferbrecht (approved):

confgenerator/logging_receivers.go
```diff
@@ -491,6 +493,109 @@ func (r LoggingReceiverWindowsEventLog) Components(ctx context.Context, tag stri
...
+							ottl.Equals(keyword0, ottl.StringLiteral("Audit Failure")),
+						)),
+					)
+				},
+			},
+			// TODO: Fix OTel receiver to provide raw non-parsed messages.
+			"jsonPayload.Message":      {CopyFrom: "jsonPayload.message"},
+			"jsonPayload.Qualifiers":   {CopyFrom: "jsonPayload.event_id.qualifiers"},
+			"jsonPayload.RecordNumber": {CopyFrom: "jsonPayload.record_id"},
+			"jsonPayload.Sid":          {CopyFrom: "jsonPayload.security.user_id"},
```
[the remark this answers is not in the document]
quentinmit: It was just as much work to just set the default value to an empty string, so I just did that.

jefferbrecht (approved):

PR #1525, opened 2023-11-27
Testing: Replace RunScriptRemotely with RunRemotely in `ops_agent_test.go`

martijnvans:
## Description
On Windows, `RunRemotely` has better error handling semantics unless the script explicitly sets `$ErrorActionPreference` to `Stop`. For an example of bad error handling, see the unnoticed errros in https://storage.googleapis.com/ops-agents-public-buckets-test-logs/prod/stackdriver_agents/testing/consumer/ops_agent_release/windows_x86_64/117/20231116-055824/logs/TestOTLPMetricsGMP_windows-2022/main_log.txt

On Linux, `RunRemotely` has nearly identical behavior. The only difference I've found is that it will print the command it's running (i.e. the full script) into the logs, instead of something like `Running bash -x c7259b2f-3ad3-482d-940f-bea51b4ac67a.sh`.

I'm looking into ways to improve `RunScriptRemotely`, and failing that, limiting its use to the relatively few places that need it. This CL is part of that.

## Related issue
none, just bothers me and confuses people sometimes.

## How has this been tested?
automated tests should be enough.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/ops_agent_test.go
```diff
@@ -99,7 +99,7 @@ func workDirForPlatform(platform string) string {
 	if gce.IsWindows(platform) {
 		return `C:\work`
 	}
-	return "/root/work"
```
martijnvans, line 102: RunRemotely runs without sudo (unlike RunScriptRemotely), so it was hitting errors about lacking sudo. Instead of sprinkling sudo everywhere, I just moved this directory to somewhere accessible without sudo.

integration_test/ops_agent_test.go
```diff
@@ -3261,9 +3259,9 @@ func testPrometheusMetrics(t *testing.T, opsAgentConfig string, testChecks []moc
 
 		// 4. Start the go http server
 		setupScript := `sudo systemctl daemon-reload && sudo systemctl enable http-server-for-prometheus-test && sudo systemctl restart http-server-for-prometheus-test`
-		setupOut, err := gce.RunRemotely(ctx, logger, vm, "", setupScript)
+		_, err = gce.RunRemotely(ctx, logger, vm, "", setupScript)
 		if err != nil {
-			t.Fatalf("failed to start the http server in VM via systemctl with err: %v, stderr: %s", err, setupOut.Stderr)
+			t.Fatalf("failed to start the http server in VM via systemctl with err: %v", err)
```
igorpeshansky, line 3264: Just curious, why are we now ignoring the stderr? Doesn't seem related to this PR…

martijnvans: You're right, it's just something I noticed and cleaned up along the way. `err` already has stderr (technically, interleaved stdout and stderr) in it, so there's no need to print both `err` and `setupOut.Stderr`.

This is consistent with other callsites to RunRemotely, which all pretty much just print `err`.

igorpeshansky (approved):
LGTM :shipit:

PR #1531, opened 2023-11-28
Change transformation tests to use LogEntry format

avilevy18:
## Description

Updated transformation logs to use LogEntry format instead.
## Related issue

b/308821542 
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


transformation_test/transformation_test.go
```diff
@@ -90,50 +88,42 @@ func TestTransformationTests(t *testing.T) {
...
+					entry := e.(map[string]interface{})
+					date := entry["timestamp"].(string)
+					timestamp, err := time.Parse(time.RFC3339Nano, date)
+					if err != nil {
+						t.Fatal(err)
 					}
+					if timestamp.After(testStartTime) {
+						entry["timestamp"] = "now"
+					}
+					data = append(data, entry)
```
quentinmit, line 121: I think we might have already discussed this, but is it worth tracking how those entries are divided into requests? (i.e. should we put `d` in `entries` instead of each individual entry?)

transformation_test/transformation_test.go
```diff
@@ -90,50 +88,42 @@ func TestTransformationTests(t *testing.T) {
...
-					date := time.UnixMicro(int64(date * 1e6)).UTC()
-					if date.After(testStartTime) {
-						data[i]["date"] = "now"
-					} else {
-						data[i]["date"] = date.Format(time.RFC3339Nano)
+
+			var entries []any
+			// Only search for entries if stdout is not null
+			if val, ok := d["entries"].([]any); ok {
+				entries = val
```
quentinmit, line 110: I think this will fail if fluent-bit decides to use multiple WriteLogsRequests, no? (Probably you will fail at the `yaml.Unmarshal`, but if you don't, you'll just keep the entries from the last request here.)

quentinmit (commented):
It looks like you took a bunch of these tests from `ops_agent_test.go`. Can you please rename them to match the test name from `ops_agent_test` so that we can easily tell which tests have been ported? I used names like `ops_agent_test-TestLogEntrySpecialFields` in my PR, if you're okay with matching that naming convention.

Then it'll be easy to come through later and remove the redundant tests in `ops_agent_test.go`.

transformation_test/transformation_test.go
```diff
@@ -204,9 +202,19 @@ func generateFluentBitConfigs(ctx context.Context, transformationTest transforma
...
 		Config: map[string]string{
-			"Name":   "stdout",
-			"Match":  "*",
-			"Format": "json",
+			"Match":                         "*",
+			"Name":                          "stackdriver",
+			"Retry_Limit":                   "3",
+			"http_request_key":              "logging.googleapis.com/httpRequest",
+			"net.connect_timeout_log_error": "False",
+			"resource":                      "gce_instance",
```
quentinmit, line 210: Is fluent-bit going to try to do resource detection (and thus make this test fail on a cloudtop because it adds resource labels)?

transformation_test/transformation_test.go
```diff
@@ -155,7 +152,8 @@ func checkOutput(t *testing.T, name string, got []map[string]any) {
 	if err := yaml.Unmarshal(wantBytes, &want); err != nil {
 		t.Fatal(err)
 	}
-	if diff := cmp.Diff(got, want); diff != "" {
+
+	if diff := cmp.Diff(gotBytes, wantBytes); diff != "" {
```
quentinmit, line 156: Why are you diffing gotBytes and wantBytes instead of got and want? I believe that will make the diff much less readable.

transformation_test/transformation_test.go
```diff
@@ -90,50 +89,48 @@ func TestTransformationTests(t *testing.T) {
...
-			out, _ := io.ReadAll(stdout)
-			_ = eg.Wait()
+			var d []map[string]any
+			var data []map[string]interface{}
 
-			err := yaml.Unmarshal(out, &data)
-			if err != nil {
-				t.Log(string(out))
-				t.Fatal(err)
+			sc := bufio.NewScanner(bytes.NewReader(stdout.Bytes()))
```
quentinmit, line 103: You should use a `json.Decoder` to read the sequential JSON maps, instead of using a `bufio.Scanner` + `yaml.Unmarshal`

transformation_test/transformation_test.go
```diff
@@ -90,50 +89,48 @@ func TestTransformationTests(t *testing.T) {
...
+						date := entry["timestamp"].(string)
+						timestamp, err := time.Parse(time.RFC3339Nano, date)
+						if err != nil {
+							t.Fatal(err)
+						}
+						if timestamp.After(testStartTime) {
+							entry["timestamp"] = "now"
+						}
 					}
+					data = append(data, req)
```
quentinmit, line 128: Maybe put this outside the if? IDK when we'd get a request with no entries, but it seems like we should capture that in the golden.

transformation_test/transformation_test.go
avilevy18: Tested it out with no issues.

quentinmit (approved):

PR #1537, opened 2023-12-04
Clean up fluent-bit component construction

igorpeshansky:
## Description

Remove explicit construction of fluent-bit components from all third-party application receivers and processors. Factor out all fluent-bit components into helper functions and/or logging processors.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/logging_processors.go
```diff
@@ -275,6 +275,55 @@ func (p LoggingProcessorNestWildcard) Components(ctx context.Context, tag, uid s
 	}
 }
 
+type LoggingProcessorLift struct {
```
quentinmit, line 278: Please don't name these logging processors; we're unlikely to ever expose them as-is, and someone in the future might think that this was already designed and try to use it publicly.

Grep and lift belong as functions (or possibly still structs, for convenience) in the `fluentbit` package.

confgenerator/logging_processors.go
```diff
@@ -275,6 +275,55 @@ func (p LoggingProcessorNestWildcard) Components(ctx context.Context, tag, uid s
...
+		},
+	}
+
+	return []fluentbit.Component{
+		filter,
+	}
+}
+
+type LoggingProcessorGrep struct {
+	Field   string `validate:"required"`
```
quentinmit, line 301: These validate tags will never be used, so it's probably more confusing than helpful to have them here.

confgenerator/self_logs.go
```diff
@@ -55,6 +56,40 @@ func healthChecksLogsPath() string {
 	return path.Join("${logs_dir}", "health-checks.log")
 }
 
+type LoggingProcessorSampleLogs struct {
```
quentinmit, line 59: This is bespoke enough that it's fine staying as raw fluent-bit config. I don't think it's ever going to be reused in another processor.

confgenerator/logging_processors.go
igorpeshansky: I was actually hoping to eventually expose the actual grep, nest, and lift processors to third-party receivers/processors, and use them to construct auxiliary pipelines. This change gets us part of the way there. I am happy to change them to functions again, and lift them into processors as part of the pipelines change (I'd have to move the other proto-processors that were already present, like NestWildcard).

confgenerator/logging_processors.go
igorpeshansky: I suppose you're right — I was using them as pseudo-documentation to indicate that Regex and Exclude are mutually-exclusive, but I can just say that in a comment…

confgenerator/self_logs.go
igorpeshansky: Again, I was hoping to actually turn this into an internal logging processor that was used in an actual pipeline…

confgenerator/logging_processors.go
quentinmit: I think this suggests that there needs to a new category of processors: internal-only processors that can exist as objects in pipelines but can't be used in user config. Can you just move these into the `fluentbit` package and change the names slightly (maybe just call it `fluentbit.Lift`?)? I think they should still be usable as processors since they'll still implement the interface, and it will be clear they're not for user config.

(I think these are fluentbit-specific because nest/lift/etc. are all just done as workarounds specific to fluent-bit, and for an otel version of these third-party processors we wouldn't use them.)

igorpeshansky: I suspect we can achieve internal-only processors by defining the `Type()` method, but not calling `RegisterType`. Done.

While these translate directly into fluent-bit filters, the _actions_ performed by those are more general (extract values that are nested under a field into a top-level field). That should be translatable to an otel logging filter as well.

quentinmit: I do not think that avoiding `RegisterType` is sufficient, because this is a loaded foot-gun. Also, we already have a clear pattern for how these should be implemented - as functions in `confgenerator/fluentbit/processors.go` that return `[]fluentbit.Component` (and which AFAICT you are not touching in this PR).

While it's theoretically possible to translate these actions into otel filters, we would never do that because they are specific workarounds for fluent-bit bugs. There are better ways to do all of these things in otel without combining these specific action primitives.

PR #1541, opened 2023-12-06
Ingest Dataproc GCE instance metadata attributes as logging labels.

igorpeshansky:
## Description

Adds default ingestion of Dataproc-specific GCE instance metadata attributes as logging labels, namely:
- `dataproc-cluster-name`
- `dataproc-cluster-uuid`
- `dataproc-region`

## Related issue

[b/315343759](http://b/315343759)

## How has this been tested?

Installed the agent built from this branch on a Dataproc node, and received a log entry annotated as follows:
```
"labels": {
  "compute.googleapis.com/attributes/dataproc-cluster-name": "fluent-bit-dataproc-test",
  "compute.googleapis.com/attributes/dataproc-region": "us-central1",
  "compute.googleapis.com/attributes/dataproc-cluster-uuid": "c10ce82e-15c6-4a6a-bcad-3db8a1d9557e",
  "compute.googleapis.com/resource_name": "fluent-bit-dataproc-test-m"
}
```

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/testdata/goldens/builtin/golden/linux-dataproc/otel.yaml
```diff
@@ -0,0 +1,499 @@
+exporters:
```
quentinmit, line 1: I think you forgot to commit the `input.yaml` for this test

confgenerator/testdata/goldens/logging-processor_gce_metadata_attributes_precedence/input.yaml
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+
+logging:
+  receivers:
+    sample_logs:
+      type: files
+      include_paths: [/tmp/*.log]
+  processors:
+    metadata:
+      type: gce_metadata_attributes
```
quentinmit, line 22: How do you envision this processor actually being used?

quentinmit (commented):
This feels like a very specific processor to me, and I find it weird that it's implemented as a processor when all other resource detection is automagical. If it needs to be exposed at all, my inclination is that this would be better served as config options under `logging.service`. But if we can just detect a dataproc context and add the correct labels automatically, I think I would prefer that even more.

igorpeshansky:
Repurposed this PR to only ingest a fixed set of Dataproc labels. PTAL.

confgenerator/logging_processors.go
```diff
@@ -364,3 +368,66 @@ end`, lua))...)
...
+	return false
+}
+
+const (
+	attributeLabelPrefix string = "compute.googleapis.com/attributes/"
+)
+
+func (p LoggingProcessorGceMetadataAttributes) Components(ctx context.Context, tag, uid string) []fluentbit.Component {
+	processorName := fmt.Sprintf("%s.%s.gce_metadata", tag, uid)
+	resource := MetadataResource
```
quentinmit, line 401: This global variable exists only for testing the Prometheus receiver; you shouldn't reuse it here for a different purpose. You should just call `resourcedetector.GetResource` (and arguably that should take a context argument and check for a resource override itself)

confgenerator/logging_processors.go
```diff
@@ -364,3 +368,66 @@ end`, lua))...)
...
+	if r := platform.FromContext(ctx).ResourceOverride; r != nil {
+		resource = r
+	}
+	gceMetadata, ok := resource.(resourcedetector.GCEResource)
+	if !ok {
+		// Not on GCE; no attributes to detect.
+		log.Printf("ignoring the gce_metadata_attributes processor outside of GCE: %T", resource)
+		return nil
+	}
+	modifications := make(map[string]*ModifyField)
```
quentinmit, line 411: Nit: prefer
```go
modifications := map[string]*ModifyField{}
```
(i.e., use an empty literal map instead of `make`)

confgenerator/logging_processors.go
```diff
@@ -364,3 +368,66 @@ end`, lua))...)
...
+// A LoggingProcessorGceMetadataAttributes annotates logs with labels
+// corresponding to instance attributes from the GCE metadata server.
+type LoggingProcessorGceMetadataAttributes struct {
+	ConfigComponent `yaml:",inline"`
+	Include         []string `yaml:"include" validate:"dive,gt=0,glob"`
+	Exclude         []string `yaml:"exclude" validate:"dive,gt=0,glob"`
+	Static          *bool `yaml:"static" validate:"required,eq=true"`
+}
+
+func (p LoggingProcessorGceMetadataAttributes) Type() string {
```
quentinmit, line 381: This doesn't need a `Type()` since it isn't being used in pipelines (yet?)

confgenerator/logging_processors.go
```diff
@@ -364,3 +368,66 @@ end`, lua))...)
 func init() {
 	LoggingProcessorTypes.RegisterType(func() LoggingProcessor { return &LoggingProcessorExcludeLogs{} })
 }
+
+// A LoggingProcessorGceMetadataAttributes annotates logs with labels
+// corresponding to instance attributes from the GCE metadata server.
+type LoggingProcessorGceMetadataAttributes struct {
+	ConfigComponent `yaml:",inline"`
+	Include         []string `yaml:"include" validate:"dive,gt=0,glob"`
```
quentinmit, line 376: These validations will never be used, AFAICT

confgenerator/confgenerator_test.go
```diff
@@ -149,6 +149,43 @@ func TestGoldens(t *testing.T) {
...
+	dataprocMetadata := map[string]string{
+		"dataproc-cluster-name": "test-cluster",
+		"dataproc-cluster-uuid": "test-uuid",
+		"dataproc-region": "test-region",
+	}
+
+	t.Run(testName, func(t *testing.T) {
+		t.Parallel()
+		pc := testPlatforms[0]
+		assert.Equal(t, pc.name, "linux")
```
quentinmit, line 166: This is going to bite someone in the future who is changing the `testPlatforms` array. Just put the first platform into its own global variable and use it in both the array and here.

confgenerator/logging_processors.go
```diff
@@ -364,3 +368,66 @@ end`, lua))...)
...
+	Exclude         []string `yaml:"exclude" validate:"dive,gt=0,glob"`
+	Static          *bool `yaml:"static" validate:"required,eq=true"`
```
quentinmit, line 378: You can just use `oneof` with a single enum value. No need to add a new validation type.

confgenerator/logging_processors.go
igorpeshansky: Now that `ResourceOverride` has been added, do we even need `MetadataResource` anymore?

confgenerator/logging_processors.go
igorpeshansky: I'd like to ensure that all `LoggingProcessor*` structs conform to the `LoggingProcessor` interface, which does require a `Type()`.

confgenerator/logging_processors.go
igorpeshansky: Updated to remove `MetadataResource`, and use `platform.GetResource()` instead. I couldn't put that code in `resourcedetector`, as that would result in an import cycle.

confgenerator/logging_processors.go
quentinmit: I specifically want them to *not* conform to the interface to prevent them from accidentally leaking into external configs.

confgenerator/logging_processors.go
quentinmit: There's no way to enable that without also enabling use in user configs.

igorpeshansky: Removed the validations.

confgenerator/logging_processors.go
igorpeshansky: That's what the "internal:" prefix check is for… Ideally, we would be able to represent most of the configs as Ops Agent pipelines (whether or not visible to the user), so making them conform to the interface makes sense to me.

quentinmit: I don't think adding that prefix really changes anything. If you really care about having a valid processor, you already have one in the form of `LoggingProcessorModifyFields`, and you can just make this a factory function that returns a `LoggingProcessorModifyFields`.

```go
type addGceMetadataAttributes struct {
  Include, Exclude []string
}
 func (p addGceMetadataAttributes) Processor(ctx context.Context) LoggingProcessor {
  // body identical
  return LoggingProcessorModifyFields{
		Fields: modifications,
  }
}
```
Also, put it in a separate file (maybe `internal_processors.go`) so it's not commingled with user-facing processors.

igorpeshansky: Removed the processor code.

confgenerator/logging.go
```diff
@@ -48,12 +48,17 @@ func setLogNameComponents(ctx context.Context, tag, logName, receiverType string
 }
 
 // stackdriverOutputComponent generates a component that outputs logs matching the regex `match` using `userAgent`.
-func stackdriverOutputComponent(ctx context.Context, match, userAgent string, storageLimitSize string) fluentbit.Component {
+func stackdriverOutputComponent(ctx context.Context, match, userAgent string, storageLimitSize string) (fluentbit.Component, error) {
+	r, err := platform.FromContext(ctx).GetResource()
```
quentinmit, line 52: The old code explicitly didn't do resource detection here because it didn't want to block on it; why is this change necessary?

confgenerator/resourcedetector/detector.go
```diff
@@ -22,17 +22,37 @@ import (
 // An implementation of the Resource interface will have fields represent
 // available attributes about the current monitoring resource.
 type Resource interface {
+	ProjectName() string
+	IsAutoDetected() bool
```
quentinmit, line 26: this shouldn't be part of the Resource interface; maybe it needs to be tracked somewhere (in `Platform`, perhaps?) but each resource shouldn't have to know about it.

confgenerator/logging.go
quentinmit: Line 83 in the old code did the resource override

qingling128 (approved):
LGTM

Sync'd offline. We might make some additional refactoring in following commits or a separate PR. But the content here is okay as is as well based on our understanding of the user visible impact.

PR #1547, opened 2023-12-14
Testing: Fix GPU driver installation on Rocky Linux 9

LujieDuan:
## Description
GPU driver installation is failing on Rocky Linux 9 due to missing matching kernel-devel package. 

This PR tries to fix the Rocky Linux 9 failure and also improve the centos/RHEL `install` script for DCGM/NVML:
- Add "try_install" logic: the script will have multiple installation methods and try the methods one at a time;;
- Currently the script includes two installation methods - runfile and network-rpm which tend to have higher success rate; will add more in the future when available; 
   - For Rocky Linux 9 - the installation order is network-rpm first and then runfile; the runfile method is not working with the current image (this is what is failing before this PR); the idea is if an underlying image change causes the first candidate to fail, there could be chances that the runfile method starts to work on the new image; 
- Remove:
  -  We won't test K80 model on RHEL/CentOS, so remove the logic to handle K80; 
  -  Remove the temporary fix for Rocky Linux 9.2/8.8 images, since the GCE images have been updated to the newest versions (9.3/8.9);

## Related issue
[b/316343810](http://b/316343810)
[b/312949832](http://b/312949832)

## How has this been tested?
Check the NVML/DCGM tests pass for RHEL/CentOS; make sure the tests use the first installation method specified for that distro at the moment. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,81 @@
...
+        set -e
+        if [ $exit_code -eq 0 ]; then
+            echo "NVIDIA driver has been installed successfully with $install_method."
+            return 
+        fi
+    done
+    return 1
+}
+
+handle_rehl7() {
```
igorpeshansky, line 48:
```suggestion
handle_rhel7() {
```
Also below and in `nvml/centos_rhel/install`…

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,81 @@
...
-        ;;
-esac
+VERSION_ID=${VERSION_ID%%.*}
 
-echo "Installing NVIDIA Data Center driver $DRIVER_VERSION"
-curl -fSsl -O https://us.download.nvidia.com/tesla/$DRIVER_VERSION/NVIDIA-Linux-x86_64-$DRIVER_VERSION.run
-sudo bash ./NVIDIA-Linux-x86_64-$DRIVER_VERSION.run --silent
+install_driver_from_runfile() {
+    # Ref: https://docs.nvidia.com/datacenter/tesla/tesla-installation-notes/index.html#runfile
+    # This method requires the match kernel-devel package to be installed, and
```
igorpeshansky, line 7:
```suggestion
    # This method requires the matching kernel-devel package to be installed, and 
```

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,81 @@
...
+    sudo yum install -y datacenter-gpu-manager
+    sudo systemctl --now enable nvidia-dcgm 
+
+    # check DCGM service running and load profiling module
+    dcgmi discovery --list
+}
+
+try_install() {
+    for install_method in "$@"
+    do
```
igorpeshansky, line 34:
```suggestion
    for install_method in "$@"; do
```

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,81 @@
...
+        echo "Installing NVIDIA driver with $install_method..."
+        # Run the installation method in a subshell with -e so that the function
+        # exit on error
```
igorpeshansky, line 37:
```suggestion
        # exits on error
```

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,81 @@
...
+        set +e ; ( set -e; $install_method ) ; exit_code=$?
+        set -e
+        if [ $exit_code -eq 0 ]; then
+            echo "NVIDIA driver has been installed successfully with $install_method."
+            return 
+        fi
+    done
+    return 1
```
igorpeshansky, line 45: This return code is not checked anywhere. Do we need to return it?

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,81 @@
...
-sudo yum install -y datacenter-gpu-manager
-sudo systemctl --now enable nvidia-dcgm
-
-# check DCGM service running and load profiling module
-dcgmi discovery --list
+install_dcgm() {
+    # Ref: https://docs.nvidia.com/datacenter/dcgm/latest/user-guide/getting-started.html#rhel-centos-rocky-linux
+    sudo yum install -y yum-utils
+    sudo yum-config-manager \
+    --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel$VERSION_ID/x86_64/cuda-rhel$VERSION_ID.repo
```
igorpeshansky, line 23: Let's indent the continuation line…
```suggestion
        --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel$VERSION_ID/x86_64/cuda-rhel$VERSION_ID.repo
```

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,81 @@
...
+    do
+        echo "Installing NVIDIA driver with $install_method..."
+        # Run the installation method in a subshell with -e so that the function
+        # exit on error
+        set +e ; ( set -e; $install_method ) ; exit_code=$?
+        set -e
+        if [ $exit_code -eq 0 ]; then
+            echo "NVIDIA driver has been installed successfully with $install_method."
+            return 
+        fi
```
igorpeshansky, line 43: Since your main goal here is to ignore the exit code, this could be made much simpler:
```suggestion
        $install_method && {
            echo "NVIDIA driver has been installed successfully with $install_method."
            return
        }
```

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,81 @@
...
+    install_driver_from_package_manager() {
+        # Ref: https://developer.nvidia.com/cuda-12-2-2-download-archive?target_os=Linux&target_arch=x86_64&Distribution=RHEL&target_version=8&target_type=rpm_network
+        sudo dnf config-manager \
+        --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel$VERSION_ID/x86_64/cuda-rhel$VERSION_ID.repo
+        sudo dnf clean all
+        sudo dnf -y module install nvidia-driver:latest
+
+        # check NVIDIA driver installation succeeded
+        nvidia-smi
+    }
```
igorpeshansky, line 73: This is _almost_ identical to `install_driver_from_package_manager` defined in `handle_rehl7`[sic]. The only differences are `s/yum/dnf` and `module install nvidia-driver:latest` vs `install nvidia-driver-latest-dkms`. Is there anything we can do to share more of the code (e.g., given that `yum` and `dnf` can pretty much be used interchangeably on newer systems)?

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
LujieDuan: Here I was also looking to get the behaviour that the function `$install_method` can exit on errors (do not run the remains commands in the function) using `set +e ; ( set -e; func) ;  set -e`. Added comments to better document this; is there a way to simplify the command while keep this?
```
 # Temporarily disable the "-e" option so that failures in the function
 # won't cause this script to exit; also run the installation method in a
 # subshell with -e so that the function itself will exit immediately if 
 # any command encounters error
 ```

File: integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install (hunk 5 above)
Comment (LujieDuan) [reply integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install:54]: My thought here is to have this return code works together with `set -e` - so that if none of the installation methods works, the script stops after `try_install` and do not run following commands.

File: integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install (hunk 6 above)
Comment (LujieDuan) [reply integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install]: Done!

File: integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install (hunk 4 above)
Comment (LujieDuan) [reply integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install]: Done!

File: integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install (hunk 3 above)
Comment (LujieDuan) [reply integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install]: Done!

File: integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install (hunk 1 above)
Comment (LujieDuan) [reply integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install]: Done!

File: integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install (hunk 7 above)
Comment (igorpeshansky) [reply integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install]: That's what I was saying — using `&&` will have the effect of temporarily disabling `errexit` for that one command.

File: integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install (hunk 5 above)
Comment (igorpeshansky) [reply integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install:54]: I see. Makes sense. I would then change line 50 to `return 0` to match.

File: integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install [hunk 9]
```diff
@@ -1,54 +1,84 @@
...
+handle_common() {
+    install_driver_from_package_manager() {
+        # Ref: https://developer.nvidia.com/cuda-12-2-2-download-archive?target_os=Linux&target_arch=x86_64&Distribution=RHEL&target_version=8&target_type=rpm_network
+        setup_repo
+        sudo yum -y module install nvidia-driver:latest
+        verify_driver
+    }
+
+    try_install install_driver_from_package_manager install_driver_from_runfile
+}
```
igorpeshansky, line 78: You can make it even shorter, e.g.:
```suggestion
install_driver_from_package_manager() {
    setup_repo
    install_driver_package
    verify_driver
}

handle_rhel7() {
    install_driver_package() {
        # Ref: https://developer.nvidia.com/cuda-12-2-2-download-archive?target_os=Linux&target_arch=x86_64&Distribution=RHEL&target_version=7&target_type=rpm_network
        sudo yum -y install nvidia-driver-latest-dkms
    }

    try_install install_driver_from_runfile install_driver_from_package_manager
}

handle_common() {
    install_driver_package() {
        # Ref: https://developer.nvidia.com/cuda-12-2-2-download-archive?target_os=Linux&target_arch=x86_64&Distribution=RHEL&target_version=8&target_type=rpm_network
        sudo yum -y module install nvidia-driver:latest
    }

    try_install install_driver_from_package_manager install_driver_from_runfile
}
```
Also, do we ever expect the driver _verification_ to fail after it's been installed (as opposed to the installation itself)? If not, you can pull out `verify_driver` to after the `try_install` step.

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,84 @@
...
+
+handle_rhel7() {
+    install_driver_from_package_manager() {
+        # Ref: https://developer.nvidia.com/cuda-12-2-2-download-archive?target_os=Linux&target_arch=x86_64&Distribution=RHEL&target_version=7&target_type=rpm_network
+        setup_repo
+        sudo yum -y install nvidia-driver-latest-dkms
+        verify_driver
+    }
+
+    try_install install_driver_from_runfile install_driver_from_package_manager
```
igorpeshansky, line 66: If there's some documentation that indicates that installing from a runfile is preferable to installing the system packages on rhel7, this would be a good place to cite it (e.g.:
```
    # Prefer installing from a runfile on RHEL 7: http://…
```
).

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
LujieDuan: Sorry here was what I wanted to ask: `&&` would disable `errexit` on calling the function, and evey commands of the function. So with my function
```
install_driver_from_runfile () {
   install kernel_devel
   install driver
}
``` 
`install driver` will still run if `install kernel_devel` fails, when using `install_driver_from_runfile && ...`, and would polluate the logs.
The behaviour I want to have is: disable `errexit` on calling the function, but enable `errexit` on evey commands of the function. 

Another way I can find is to do proper error checking in the function on the commands, so that we don't need to rely on errexit.

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
LujieDuan: Looked into this a bit more and it seems that unlike Rocky 8/9, Centos 7's repo would have all versions kernel source packages (https://pkgs.org/download/kernel-devel); installing GPU driver from yum would function perfectly on Centos 7 if we install the kernel source first. 

Updated the script so that we install from package managers for all three distros, and leave runfile as the backup method.

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
LujieDuan: Done! Thanks!

`verify_driver` can fail even when the installation returns no error. When this happens, we want `try_install` to try the next method (runfile). Added comments to `verify_driver` function with potential error messages. 

The runfile method would error out if it find an existing package manager driver installation, so in the latest commit I have modified the runfile method to try to remove yum installation first.

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
LujieDuan: Done!

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
igorpeshansky: Ah, you're right. I forgot that bash has this weird behavior that the conditional/boolean context was propagated all the way down into functions and subshells. There's a great explanation at https://lists.gnu.org/archive/html/bug-bash/2012-12/msg00094.html if you wanted to quote it in the comment…

igorpeshansky: Actually, I did find a sort of shorter solution, which uses an explicit `bash` subcommand instead of a subshell. It's not as readable as what I was proposing, so I'll leave it to you to decide whether it's actually worth pursuing. I also added a proposed comment, which you can feel free to omit:
```diff
-        # Run the installation method in a subshell with -e so that the function
-        # exit on error
-        set +e ; ( set -e; $install_method ) ; exit_code=$?
-        set -e
-        if [ $exit_code -eq 0 ]; then
-            echo "NVIDIA driver has been installed successfully with $install_method."
-            return 
-        fi
+        export -f $install_method
+        # Can't use a subshell because of https://lists.gnu.org/archive/html/bug-bash/2012-12/msg00094.html
+        bash -$- -c $install_method && {
+            echo "NVIDIA driver has been installed successfully with $install_method."
+            return
+        }
```
FWIW, this solution is also robust with respect to putting the calling function/shell into the conditional/boolean context.

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,101 @@
...
+        # Temporarily disable the "-e" option so that failures in the function
+        # won't cause this script to exit; also run the installation method in a
+        # subshell with -e so that the function itself will exit immediately if 
+        # any command encounters error
+        set +e ; ( set -e; $install_method ) ; exit_code=$?
+        set -e
+        if [ $exit_code -eq 0 ]; then
+            echo "NVIDIA driver has been installed successfully with $install_method."
+            return 0
+        fi
```
igorpeshansky, line 63: [Optional] Revisiting this, see the other thread (which I, sadly, can't make part of this review):
```suggestion
        # Can't use a subshell because of https://lists.gnu.org/archive/html/bug-bash/2012-12/msg00094.html
        bash -$- -c $install_method && {
            echo "NVIDIA driver has been installed successfully with $install_method."
            return 0
        }
```
[Some supporting experimentation](https://tio.run/##vVFdasMwDH7PKYQppWH4Ag07TGrLscCVh@2wjrGzp3LXtSVLoYOyJ9vS9yN93vXZT5Mb2RSKDK6nQFw87snagJsWPhsAND6CAijpg3jowIV@yK8rraTn@pDxBkMDx4T2pDQmVM1Xg4e3mApoN5dvNhmljt2ZzRHW663q5rifviULHAvggYpq5@x3Kn6ZL9UzJo/GIFq06g@a5Komud9jyZUXlB09pi7BBDIk7ZQqotpUxIsgLtA7WVz92m@SfjCmnfx4Naon6JUGbZ6XVxwLJqHL@z92uw7qiCl7Ydyf6KmLLzhP0xE "Bash – Try It Online").

You would also need to export all necessary functions at the top level for the `bash` subprocess to inherit them:
```
export -f install_driver_from_package_manager install_driver_from_runfile setup_repo install_driver_package remove_driver_package verify_driver 
```

igorpeshansky (commented):
Mostly LGTM; a couple of minor comments (which apply across both files).

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
LujieDuan: Done! 

Updated to use bash subcommand. Thanks for the detailed suggestion and explanation!

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,101 @@
...
+
+handle_common() {
+    install_driver_package() {
+        # Ref: https://developer.nvidia.com/cuda-12-2-2-download-archive?target_os=Linux&target_arch=x86_64&Distribution=RHEL&target_version=8&target_type=rpm_network
+        sudo yum -y module install nvidia-driver:latest
+    }
+
+    remove_driver_package() {
+        # Ref: https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#removing-cuda-toolkit-and-driver
+        sudo dnf -y module remove --all nvidia-driver
```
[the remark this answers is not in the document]
LujieDuan: Done!

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
LujieDuan: Done!

Used the second command to export all functions; also exported the `VERSION_ID` env. Thanks!

igorpeshansky (commented):
One last tweak, otherwise LGTM.

integration_test/third_party_apps_data/applications/dcgm/centos_rhel/install
```diff
@@ -1,54 +1,99 @@
...
+    sudo systemctl --now enable nvidia-dcgm 
+
+    # check DCGM service running and load profiling module
+    dcgmi discovery --list
+}
+
+try_install() {
+    # Export all functions for the bash subprocess
+    eval "$(declare -F | sed 's/ -f / -fx /')"
+    export VERSION_ID=$VERSION_ID
```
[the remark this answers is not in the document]
LujieDuan: Done!

igorpeshansky (approved):
LGTM :shipit:

PR #1569, opened 2024-01-09
Internal: attempt to reduce Ubuntu test flakes

martijnvans:
## Description
This is an attempt to fix b/310644585 (`test flakes: "Could not get lock /var/lib/apt/lists/lock"`) or at least make it happen less often.

Alternatives considered:
- modifying the user-facing install script to force `apt` to wait for the lock instead of failing out. I'm inclined to not do this because it's a rather drastic change to the install script (`apt-get` -> `apt`) and it would cause the install script to spit out a lot of `WARNING: apt does not have a stable CLI interface. Use with caution in scripts.` garbage. Besides, it is my reading based on that warning and pages like [this one](https://manpages.debian.org/bullseye/apt/apt.8.en.html#SCRIPT_USAGE_AND_DIFFERENCES_FROM_OTHER_APT_TOOLS) that `apt-get` is supposed to be preferred from scripts anyway.
- modifying `apt`/`dpkg` settings in integration tests only to force `apt` to wait for the lock. I tried this a lot and couldn't get it to work. I got frustrated with how difficult it was to set global settings for `apt` and eventually gave up.

## Related issue
b/310644585

## How has this been tested?
Presubmits should be enough. Hopefully we will see reduced flake rates in our continuous builds over the next week too.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/gce/gce_testing.go
+++ b/integration_test/gce/gce_testing.go
@@ -1273,6 +1273,16 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
 		return nil, err
 	}
 
+	if isUbuntu(vm.Platform) {
+		// Taken from https://askubuntu.com/a/1039155 to fix b/310644585.
+		// This is unnecessary for newer versions of apt because apt was fixed
+		// (https://bugs.debian.org/754103) to wait for the lock instead of
+		// failing out immediately.
+		if _, err := RunRemotely(ctx, logger, vm, "", "sudo systemctl disable apt-daily.service && sudo systemctl disable apt-daily.timer && sudo systemctl disable apt-daily-upgrade.timer && sudo systemctl disable apt-daily-upgrade.service"); err != nil {
+			return nil, fmt.Errorf("attemptCreateInstance() failed to disable apt background updates: %v", err)
+		}
+	}
+
 	if IsSUSE(vm.Platform) {
 		// Set download.max_silent_tries to 5 (by default, it is commented out in
 		// the config file). This should help with issues like b/211003972.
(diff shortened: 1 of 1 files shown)


integration_test/gce/gce_testing.go
```diff
@@ -1273,6 +1273,16 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
 		return nil, err
 	}
 
+	if isUbuntu(vm.Platform) {
```
quentinmit, line 1276: This problem isn't exclusive to Ubuntu; it could happen on any apt-based distro.

integration_test/gce/gce_testing.go
```diff
@@ -1273,6 +1273,16 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
...
+		// Taken from https://askubuntu.com/a/1039155 to fix b/310644585.
+		// This is unnecessary for newer versions of apt because apt was fixed
+		// (https://bugs.debian.org/754103) to wait for the lock instead of
+		// failing out immediately.
+		if _, err := RunRemotely(ctx, logger, vm, "", "sudo systemctl disable apt-daily.service && sudo systemctl disable apt-daily.timer && sudo systemctl disable apt-daily-upgrade.timer && sudo systemctl disable apt-daily-upgrade.service"); err != nil {
```
quentinmit, line 1281: This is still racy because the service might be running when you try to disable it

quentinmit (commented):
I believe you can enable timeouts by default for all apt-get commands by running

```shell
echo "DPkg::Lock::Timeout 120;" > /etc/apt/apt.conf.d/99stackdriver
```

and you need to do it on all apt-based distros, not just Ubuntu.

PR #1568, opened 2024-01-09
A single kokoro build config for ops agent tests

Subbarker:
## Description

For now, this is Louhi-only

Once verified, we can replace our per-platform/arch boilerplate kokoro configs with a single one.

## Checklist:
- Unit tests
  - [x] Unit tests do not apply.
  - [ ] Unit tests have been added/modified and passed for this PR.
- Integration tests
  - [x] Integration tests do not apply.
  - [ ] Integration tests have been added/modified and passed for this PR.
- Documentation
  - [x] This PR introduces no user visible changes.
  - [ ] This PR introduces user visible changes and the corresponding documentation change has been made.
- Minor version bump
  - [x] This PR introduces no new features.
  - [ ] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.
  - [ ] This PR bumps the version.

PR #1570, opened 2024-01-09
A single kokoro build config for third party tests

Subbarker:
## Description

similar to https://github.com/GoogleCloudPlatform/ops-agent/pull/1568

For now, this is Louhi-only

Once verified, we can replace our per-platform/arch boilerplate kokoro configs with a single one.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


kokoro/config/test/third_party_apps/release.gcl
```diff
@@ -0,0 +1,30 @@
...
+    distro = parts[3]
+    arch = parts[4]
+  }
+}.ret
+
+config build = common.ops_agent_test {
+  params {
+    local _tag = _parse_louhi_tag(params.environment._LOUHI_TAG_NAME)
+    local _distro_arch = _tag.distro + '_' + _tag.arch
+    platforms = platforms = third_party_image_lists.get(_distro_arch, image_lists.get(_distro_arch, 'invalid_distro')).release
```
martijnvans, line 17: This line is messed up in a few ways:

*  `third_party_image_lists` -> `image_lists_3p_override`
*  `platforms = platforms =`

It could also use a comment that `image_lists_3p_override` is used first, with a fallback to `image_lists` if it's not there.

kokoro/config/test/image_lists_3p_override.gcl
```diff
@@ -0,0 +1,85 @@
+import 'image_lists.gcl' as image_lists
+
```
martijnvans, line 2: I think I would prefer if we could make 2 things happen here:

*  put presubmits in this file as well, just like image_lists.gcl
*  make this file hold all distros in one tuple/map instead of having it hold just the overrides and having to fall back to image_lists.gcl if it isn't found. This will make it a lot easier to refer to this file from the various presubmit files (such as https://github.com/GoogleCloudPlatform/ops-agent/blob/master/kokoro/config/test/third_party_apps/bullseye_x86_64.gcl)

The solution I have in mind will also dodge an issue that I suspect will happen when doing `image_lists.get(_distro_arch, 'invalid_distro')`: I suspect that `.get()` won't work because `image_lists` is an imported file and not a proper tuple. I mean maybe `.get` will work, I dunno.

But anyway, the solution I have in mind is to put everything in both files into tuples, called I dunno `distros`:

```
distros = {
  bookworm_aarch64 = distro {
    ...
  }
  mantic_x86_64 = distro {
    ...
  }
```

Then in image_lists_3p.gcl, we would write:

```
distros = image_lists.distros {
  mantic_x86_64 {
    presubmit = [
      'foo'
    ] ,
     release = [
      'ubuntu-2310-amd64',
    ]
  }
  mantic_aarch64 {
    ...
  }
}
```

Then `.get` is guaranteed to work because `distros` is a tuple, and it allows us to refer to `image_lists_3p.distros` without needing to fall back to `image_lists`.

Subbarker: I'm going to try doing something else entirely here, and use the `presubmit` distros for 3p tests.

martijnvans (approved):

PR #1583, opened 2024-01-19
Increase service startup delay for tests that fail the health checks

jefferbrecht:
## Description
The health checks can take up to 30s to time out when they fail. We use a service startup delay of 20s. which is adequate in passing scenarios. However, for tests that deliberately fail the health checks, we need to make sure that the full 30s can lapse before validating the health check results.

## Related issue
b/321001728

## How has this been tested?
Will let presubmits run.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test.go
+++ b/integration_test/ops_agent_test.go
@@ -4452,6 +4452,11 @@ func TestPortsAndAPIHealthChecks(t *testing.T) {
 			t.Fatal(err)
 		}
 
+		// Use a longer service startup delay to allow the health check backoffs to complete.
+		// Health check backoffs take up to 30s, and SetupOpsAgent already waits for 20s,
+		// so wait an additional 20s for some extra buffer.
+		time.Sleep(20 * time.Second)
+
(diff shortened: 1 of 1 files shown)


integration_test/ops_agent_test.go
```diff
@@ -4505,6 +4510,10 @@ func TestNetworkHealthCheck(t *testing.T) {
 			t.Fatal(err)
 		}
 
+		// Use a longer service startup delay to allow the health check backoffs to complete.
```
franciscovalentecastro, line 4513: The previous remote command `startCommandForPlatform` won't continue until the `Health Checks` finish executing. So, there is not need to account for the `network check` backoffs within this tests. 

AFAIU, the flakiness arises (as @jefferbrecht pointed offline) when querying for the service status within `getRecentServiceOutputForPlatform()`. We could still add some delays, to be sure the status reaches the `Windows Event Log` or `journald`.

ridwanmsharif:
Not happening anymore.

PR #1586, opened 2024-01-22
Disable PacApiConnErr check

jefferbrecht:
## Description
The endpoint seems to return 404 if you request it too early during VM startup. Disable it for now until we figure out the root cause to unblock the integration tests.

## Related issue
b/321220138

## How has this been tested?
Will let presubmits run.

## Checklist:
- Unit tests
- Integration tests
- Documentation
  - [ ] This PR introduces user visible changes and we're leaving the documentation as-is for now because 1) users will still be running older versions of the agent that produce PacApiConnErr, and 2) we intend for this to be a brief, temporary change
- Minor version bump


franciscovalentecastro (approved):
LGTM

PR #1610, opened 2024-02-06
Support postgres delay rename

dehaansa:
## Description
Update the postgres pipeline to maintain backwards compatibility as the replication delay metric is renamed & changed from integer to float.

## How has this been tested?
Untested, first pass is allowing this to run through kokoro without the featuregate enabled, then will enable the featuregate and see that the metrics still come through.


dehaansa:
image
All checks passing when modified to only collect the new delay metric & drop back to integer and rename to old metric.

jefferbrecht:
LGTM; I'm also fine with the more minimal alternative as discussed in chat.

jefferbrecht (approved):

dehaansa:
Debian 10 failure is known issue with rabbitMQ, fixed in https://github.com/GoogleCloudPlatform/ops-agent/pull/1629

confgenerator/otel/processors.go
```diff
@@ -191,6 +191,11 @@ func ConvertGaugeToSum(metricName string) TransformQuery {
 	return TransformQuery(fmt.Sprintf(`convert_gauge_to_sum("cumulative", true) where metric.name == "%s"`, metricName))
 }
 
+// ConvertFloatToInt returns an expression where a gauge metric can be converted into a sum
```
jefferbrecht, line 194: Nit:

```suggestion
// ConvertFloatToInt returns an expression where a float-valued metric can be converted to an int
```

PR #1623, opened 2024-02-12
Internal: Move ZONES logic into go_test

Subbarker:
## Description

When running tests from Louhi, it is too late to change the zones as-is. By moving the zone setting logic into go_test.sh. It should work for Louhi as well.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/kokoro/scripts/test/go_test.sh
+++ b/kokoro/scripts/test/go_test.sh
@@ -64,6 +64,15 @@ else:
  print(data)"
 }
 
+# A helper function for joining a bash array.
+# Ex. join_by , a b c -> a,b,c
+function join_by {
+  local d=${1-} f=${2-}
+  if shift 2; then
+    printf %s "$f" "${@/#/$d}"
+  fi
+}
+
 function set_platforms() {
   # if PLATFORMS is defined, do nothing
   if [[ -n "${PLATFORMS:-}" ]]; then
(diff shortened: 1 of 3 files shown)


kokoro/scripts/test/go_test.sh
```diff
@@ -64,6 +64,15 @@ else:
  print(data)"
 }
 
+# A helper function for joining a bash array.
+# Ex. join_by , a b c -> a,b,c
+function join_by {
+  local d=${1-} f=${2-}
+  if shift 2; then
+    printf %s "$f" "${@/#/$d}"
```
martijnvans, line 72: I found this very hard to understand. At the very least this needds comments and d/f need to be renamed.

I had to read this whole article to get what it's doing:
https://dev.to/meleu/how-to-join-array-elements-in-a-bash-script-303a

Plus it doesn't work in the admittedly strange case of having a `&` in the delimiter. Which I discovered only after reading all of the comments on https://stackoverflow.com/a/17841619. This gotcha reinforces my feeling that this is using too much bash magic.

How about:
```
function join_by() {
  delim="$1"
  for (( i = 2; i <= $#; i++)); do
    printf "${!i}"  # The ith positional argument
    if [[ $i -ne $# ]]; then
      printf "$delim"
    fi
  done
}
```

Subbarker: Agreed on it being hard to understand. Used your version.

kokoro/scripts/test/go_test.sh
```diff
@@ -136,17 +139,18 @@ function set_zones() {
...
       us-central1-f
     )
   else
-    zones=(
+    zone_list=(
       invalid_zone
     )
   fi
-  export ZONES="$(join_by , ${zones[@]})"
+  zones=$(join_by , "${zone_list[@]}")
```
martijnvans, line 152: I was thinking

```
ZONES="$(join_by , "${zone_list[@]}")"
export ZONES
```

but what you have is OK too.

Note that I added an extra pair of quotes (in case the joined zones list somehow had a space), which is a best practice and somehow works fine and doesn't confuse bash, yay.

martijnvans (approved):

kokoro/scripts/test/go_test.sh
Subbarker: Gosh but it does confuse me.

PR #1641, opened 2024-03-12
Internal: use gotestsum in go_test.sh

Subbarker:
## Description

`gotestsum` is a [wrapper](https://github.com/gotestyourself/gotestsum) for `go test`.

We can rerun flaky tests easily with the `--rerun-fails`, which should significantly speed up our release builds where we currently rerun the entire suite (saving quota too!).

This also means we no longer need `go-junit-report`, as `gotestsum` provides identical functionality. This may let us upgrade our version of go we use in our test runner.

## Related issue
https://github.com/GoogleCloudPlatform/ops-agent/pull/1644

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/kokoro/scripts/test/go_test.sh
+++ b/kokoro/scripts/test/go_test.sh
@@ -212,8 +212,10 @@ if [[ "${SHORT:-false}" == "true" ]]; then
   args+=( "-test.short" )
 fi
 
+go install gotest.tools/gotestsum@latest
+
 TEST_UNDECLARED_OUTPUTS_DIR="${LOGS_DIR}" \
-  go test -v "${TEST_SUITE_NAME}.go" \
+  gotestsum --rerun-fails -v "${TEST_SUITE_NAME}.go" \
   "${args[@]}" \
   2>&1 \
   | tee "${STDERR_STDOUT_FILE}"
--- a/kokoro/scripts/test/go_test.sh
+++ b/kokoro/scripts/test/go_test.sh
@@ -215,7 +215,7 @@ fi
 go install gotest.tools/gotestsum@latest
 
 TEST_UNDECLARED_OUTPUTS_DIR="${LOGS_DIR}" \
-  gotestsum --rerun-fails --format=standard-verbose "${TEST_SUITE_NAME}.go" \
+  gotestsum --rerun-fails -- -v "${TEST_SUITE_NAME}.go" \
   "${args[@]}" \
   2>&1 \
   | tee "${STDERR_STDOUT_FILE}"
--- a/kokoro/scripts/test/go_test.sh
+++ b/kokoro/scripts/test/go_test.sh
@@ -207,11 +207,9 @@ if [[ "${SHORT:-false}" == "true" ]]; then
   args+=( "-test.short" )
 fi
 
-# "${GOTESTSUM_ARGS:-}"
-
 TEST_UNDECLARED_OUTPUTS_DIR="${LOGS_DIR}" \
   gotestsum \
-  --rerun-fails \
+  "${GOTESTSUM_ARGS:-}" \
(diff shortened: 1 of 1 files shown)


kokoro/scripts/test/go_test.sh
```diff
@@ -193,11 +193,6 @@ if [[ "${TEST_SUITE_NAME}" == "os_config_test" ]]; then
 fi
 
 STDERR_STDOUT_FILE="${KOKORO_ARTIFACTS_DIR}/test_stderr_stdout.txt"
-function produce_xml() {
-  cat "${STDERR_STDOUT_FILE}" | "$(go env GOPATH)/bin/go-junit-report" > "${LOGS_DIR}/sponge_log.xml"
-}
-# Always run produce_xml on exit, whether the test passes or fails.
-trap produce_xml EXIT
```
igorpeshansky, line 200: This part might still be relevant, right? If the test crashes/panics, shouldn't we still get an XML file?

Subbarker: As long as the `gotestsum` doesn't crash, we still will. If that crashes, then we should be looking in the `go_test` log.

igorpeshansky: I would confirm with @martijnvans — I seem to recall that this trap was originally added because otherwise we ended up with radio silence from CI runs where the test runner crashed. But maybe it's no longer needed?

martijnvans: I believe that if the test panics, we won't get an xml file (before this CL). Afterwards? Anyone's guess. We can try adding a panic and see how `gotestsum` reacts. I bet it does something reasonable though.

kokoro/scripts/test/go_test.sh
```diff
@@ -213,7 +208,11 @@ if [[ "${SHORT:-false}" == "true" ]]; then
 fi
 
 TEST_UNDECLARED_OUTPUTS_DIR="${LOGS_DIR}" \
-  go test -v "${TEST_SUITE_NAME}.go" \
+  gotestsum \
+  "${GOTESTSUM_ARGS:-}" \
```
martijnvans, line 212: This is failing with: `flag provided but not defined: -packages`

I think your GOTESTSUM_ARGS being `''` is messing it up.

I recommend skipping gotestsum_args for this CL and we can discuss what you want GOTESTSUM_ARGS to do in a follow up CL.

kokoro/scripts/test/go_test.sh
```diff
@@ -213,7 +208,11 @@ if [[ "${SHORT:-false}" == "true" ]]; then
...
+  --packages=./"${TEST_SUITE_NAME}.go" \
+  --junitfile "${LOGS_DIR}/sponge_log.xml" \
+  -- -v  \
```
martijnvans, line 215: I think we shouldd use gotestsum flag `--format=standard-verbose` to replace this use of `-v` here. Looking at a passing log, I don't think this `-v` is doing anything for us anymore, and the output it used to be printing has been lost: https://btx-internal.corp.google.com/invocations/decb3aed-be92-4f20-9615-eef393aff7ad/log

kokoro/scripts/test/go_test.sh
Subbarker: Done

kokoro/scripts/test/go_test.sh
Subbarker: Done.

kokoro/scripts/test/go_test.sh
Subbarker: Since the purpose of this PR just swapping to `gotestsum`, and it sounds like this case is being handled the same as before. I'll close this.

martijnvans (approved):

PR #1643, opened 2024-03-14
Internal: Set image projects via configs

Subbarker:
## Description

Right now we use interpret image names passed in from configs to determine what project they come from, and where they are an "Image" or an "Image Family". This logic has grown in complexity over time and most importantly, requires changes to our testing logic whenever a new corner case shows up.

This PR implements storing project names alongside the image/family names with test configs in a way that is simple to manage for future tests.

E.g.:

Image Family / Project: `<project>:<family>`

Specific Image / Project: `<project>=<image>`

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


kokoro/config/test/image_lists.gcl
```diff
@@ -15,149 +15,149 @@ local template _distro {
 // DEB Linux distros. (Do not modify this comment.)
 bookworm_aarch64 = _distro {
   release = [
-      'debian-12-arm64',
```
igorpeshansky, line 18: Can we do something like this instead:
```
local template _distro {
  // Shared image project.
  image_project = external
  // List of release images.
  release_images = external
  // The representative images to be used in presubmit (defaults to none).
  // Presubmits usually smoke test only the oldest and newest OS families from
  // a certain OS to speed up presubmit tests and reduce flakiness. The full
  // set will be tested by the nightly jobs.
  presubmit_images = []
  // List of release image specs.
  release = map(lambda i_: cond(match(i_, '[:=]'), i_, "%image_project%:%i_%"), release_images)
  // The representative image specs to be used in presubmit.
  presubmit = map(lambda i_: cond(match(i_, '[:=]'), i_, "%image_project%:%i_%"), presubmit_images)
}
```
and then just use something like:
```
bookworm_aarch64 = _distro {
  image_project = 'debian-cloud'
  release_images = [
      'debian-12-arm64',
  ]
  presubmit_images = [
      'debian-12-arm64',
  ]
}
```
?

kokoro/config/test/image_lists.gcl
```diff
@@ -15,149 +15,149 @@ local template _distro {
...
 }
 sles15_x86_64 = _distro {
   release = [
-    'sles-15',
-    'sles-15-sp2-sap',
-    'sles-15-sp5-sap',
-    'opensuse-leap',
-    'opensuse-leap-15-4',
+    'suse-cloud:sles-15',
+    'suse-sap-cloud:rhel-sap-cloud:sles-15-sp2-sap',
```
igorpeshansky, line 148: ???

kokoro/config/test/image_lists.gcl
```diff
@@ -15,149 +15,149 @@ local template _distro {
...
+    'suse-sap-cloud:rhel-sap-cloud:sles-15-sp5-sap',
+    'opensuse-cloud:opensuse-leap',
+    'opensuse-cloud=opensuse-leap-15-4-v20231208-x86-64',
```
igorpeshansky, line 151: IMO, this exposes unnecessary implementation details. The key here is that we're testing on `opensuse-leap-15-4` — the specific image is somewhat irrelevant. I wish GCE exposed this as a separate image family, but it doesn't, so we invented one. We might still want to override a specific image (family), even for one of the existing ones (e.g., to temporarily pin the image if a broken one is rolled out) without changing the intent of the tests.

kokoro/config/test/image_lists.gcl
Subbarker: that one's a typo, Fixed.

kokoro/config/test/image_lists.gcl
Subbarker: This doesn't expose anything we aren't already exposing?

kokoro/config/test/image_lists.gcl
Subbarker: Several distros are tested on images living in multiple projects. So the structure would have to be somewhere different. That being said, What I'd really like to do is remove `image_lists.gcl` and make `project.yaml` the source of truth.

kokoro/config/test/image_lists.gcl
Subbarker: I agree that it's not a great situation to be in, but I think it's far better to specify the image in one place, rather than two.

igorpeshansky: Initially, this provided a layer of abstraction that allowed separating the concerns of "what we intend to test" vs "what that actually translates to". Conflating them could make the former harder to discover. That said, having the override in the same file _is_ nicer, so maybe the right solution here is to rely on a comment instead…

PR #1670, opened 2024-04-10
Fix all go-staticcheck warnings

braydonk:
## Description
This PR fixes all present `go-staticcheck` warnings in the codebase.

Everyone who I've requested a PR from was the last editor of code that was addressed according to a git blame.

## How has this been tested?
Confgenerator unit tests.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


quentinmit (commented):
Why are there changes in the golden files? If the output is changing, that's not just a lint change.

A number of these changes appear to be creating inconsistencies between similar functions (e.g. it appears to have removed unused function parameters, so now there are some functions that take `(a, b, c)` and other functions that just take `(a, c)` based on the current implementation of that function. I don't think that's actually an improvement.

braydonk:
> Why are there changes in the golden files? 

Most of the goldens that are changing are [removing punctuation from errors](https://staticcheck.dev/docs/checks/#ST1005). The only valid output that changed is oracledb because there was trailing whitespace in the file that my editor trims automatically.

> A number of these changes appear to be creating inconsistencies between similar functions

This is because of the unused parameters check in `gopls`. These functions weren't using the context that was passed in, and since they aren't bound by any interface I removed them from the argument list.

I'm fine with turning off the error formatting check to restore goldens to their original form. For the unused parameters, we could use `_` for those, but it seems arbitrary if they aren't bound by satisfying some interface. But using `_` for context arguments is probably a fine heuristic.

PR #1687, opened 2024-04-25
Add MS SQL Server Errorlog Support

Caleb-Hurshman:
## Description

This PR adds support for Microsoft SQL Server error logs. 

## Related issue

Code for this PR references this [PR](https://github.com/GoogleCloudPlatform/ops-agent/pull/490) from 03/2022

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


apps/mssql.go
```diff
@@ -18,10 +18,80 @@ import (
...
+type LoggingReceiverMssqlLog struct {
+	LoggingProcessorMssqlLog                `yaml:",inline"`
+	confgenerator.LoggingReceiverFilesMixin `yaml:",inline" validate:"structonly"`
+}
+
+func (r LoggingReceiverMssqlLog) Components(ctx context.Context, tag string) []fluentbit.Component {
+	if len(r.IncludePaths) == 0 {
+		r.IncludePaths = []string{
+			"/var/opt/mssql/log/errorlog",
+			"Program Files\\Microsoft SQL Server\\MSSQL.n\\MSSQL\\LOG\\ERRORLOG",
```
dehaansa, line 74: By default these should be in the `C:\\` drive. Also the `MSSQL.n` value isn't actually a literal `n`, but the id of the SQL Server instance, so we want to use a wildcard instead.

apps/mssql.go
```diff
@@ -18,10 +18,80 @@ import (
...
+			"Program Files\\Microsoft SQL Server\\MSSQL.n\\MSSQL\\LOG\\ERRORLOG.n",
```
dehaansa, line 75: We don't want to monitor the ERROLOG.n files, as those are rotated error logs that will have already been read before being rotated. In this case the `n` is a number representing how old the log file is (ERRORLOG.0, .1, .2, etc)

apps/mssql.go
```diff
@@ -18,10 +18,80 @@ import (
...
+type LoggingProcessorMssqlLog struct {
+	confgenerator.ConfigComponent `yaml:",inline"`
+}
+
+func (LoggingProcessorMssqlLog) Type() string {
+	return "mssql_errorlog"
+}
+
+func (p LoggingProcessorMssqlLog) Components(ctx context.Context, tag string, uid string) []fluentbit.Component {
+	c := confgenerator.LoggingProcessorParseRegex{
```
dehaansa, line 35: In addition to the Regex, we need the parser shared section that tells fluentbit how to parse the timestamp. the one below is copied from saphana so it's not the right field/format.
```
ParserShared: confgenerator.ParserShared{
			TimeKey:    "time",
			TimeFormat: "%Y-%m-%d %H:%M:%S.%L",
}

File: integration_test/third_party_apps_test/applications/mssql/enable [hunk 4]
```diff
@@ -11,6 +11,20 @@ Copy-Item -Path 'C:\Program Files\Google\Cloud Operations\Ops Agent\config\confi
 
 # Configure the Ops Agent.
 Add-Content 'C:\Program Files\Google\Cloud Operations\Ops Agent\config\config.yaml' "
+logging:
+  receivers:
+    mssql_v2:
```
dehaansa, line 16: the logging receiver `type` in this case would use the name defined in the Type() function - which would be `mssql_errorlog`. We also want to use the default config wherever possible in integration tests, so we wouldn't specify `include_paths` unless they are nonstandard. `receiver_version` is also not relevant to the logging receiver, it's specific to metrics.

The v2 language is because there have been two major iterations of MSSQL metrics support, v2 shouldn't be used in the language for logging.

apps/mssql.go
Caleb-Hurshman: The original PR calls this out:
```
// Not sending the log timestamp "logtime" from above in as the Time_Key as there will be time zone issues.
// MS SQL Server writes logs in server time, not a standard time zone such as UTC
```
I can add these comments in

dehaansa: We'll add the comment and see how the ops-agent team would like to handle it.

dehaansa (approved):

jefferbrecht:
Just giving an update from our side chats: this is blocked on not having UTF-16 support in Fluent Bit, and that support has not yet been prioritized. We'll give another update here once that changes.

PR #1703, opened 2024-05-13
Internal: fix install gsutil issues on suse distros

Subbarker:
## Related issue

b/337015263

## Checklist:
- Unit tests
  - [x] Unit tests do not apply.
  - [ ] Unit tests have been added/modified and passed for this PR.
- Integration tests
  - [ ] Integration tests do not apply.
  - [x] Integration tests have been added/modified and passed for this PR.
- Documentation
  - [x] This PR introduces no user visible changes.
  - [ ] This PR introduces user visible changes and the corresponding documentation change has been made.
- Minor version bump
  - [ ] This PR introduces no new features.
  - [ ] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.
  - [ ] This PR bumps the version.

PR #1720, opened 2024-05-22
Testing: Update `IsSUSE` and call to `prepareSLES`.

franciscovalentecastro:
## Description
This PR attempts to update the uses of `IsSUSE`, calls to `prepareSLES` and overall preparation requisites to run tests correctly on a SUSE VM.

Additionally fixes the following error by calling `prepareSLES` when the `ImageSpec` is  `stackdriver-test-143416:sles-15-sp4-sap-saphana` which is a special case, which wasn't failing before when using `Platform`  :
```
Problem retrieving files from 'SLE-Module-Basesystem15-SP4-Pool'.
Credentials are invalid. For details see "/var/log/cloudregister". Re-register the system with "registercloudguest --force-new"
Please see the above error message for a hint.
```

Related PRs : 
- https://github.com/GoogleCloudPlatform/ops-agent/pull/1685
- https://github.com/GoogleCloudPlatform/ops-agent/pull/1703
- https://github.com/GoogleCloudPlatform/ops-agent/pull/1716

## Related issue
b/339815782

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/gce/gce_testing.go
+++ b/integration_test/gce/gce_testing.go
@@ -1252,17 +1252,17 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
 		logger.Printf("Unable to retrieve information about the VM's boot disk: %v", err)
 	}
 
-	if err := waitForStart(ctx, logger, vm); err != nil {
-		return nil, err
-	}
-
 	if os, err := getOS(ctx, logger, vm); err != nil {
 		return nil, err
 	} else {
 		vm.OS = *os
 	}
 
-	if vm.OS.ID == "opensuse" || vm.OS.ID == "opensuse-leap" || vm.OS.ID == "sles" || vm.OS.ID == "sles_sap" {
+	if err := waitForStart(ctx, logger, vm); err != nil {
+		return nil, err
+	}
+
+	if IsSUSE(vm.OS.ID) {
 		// Set download.max_silent_tries to 5 (by default, it is commented out in
 		// the config file). This should help with issues like b/211003972.
 		if _, err := RunRemotely(ctx, logger, vm, "sudo sed -i -E 's/.*download.max_silent_tries.*/download.max_silent_tries = 5/g' /etc/zypp/zypp.conf"); err != nil {
@@ -1270,13 +1270,13 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
 		}
 	}
 
-	if strings.HasPrefix(vm.ImageSpec, "sles-") {
+	if IsSLES(vm.OS.ID) {
 		if err := prepareSLES(ctx, logger, vm); err != nil {
 			return nil, fmt.Errorf("%s: %v", prepareSLESMessage, err)
 		}
 	}
 
-	if IsSUSE(vm.ImageSpec) {
+	if IsSUSE(vm.OS.ID) {
 		// Set ZYPP_LOCK_TIMEOUT so tests that use zypper don't randomly fail
 		// because some background process happened to be using zypper at the same time.
 		if _, err := RunRemotely(ctx, logger, vm, `echo 'ZYPP_LOCK_TIMEOUT=300' | sudo tee -a /etc/environment`); err != nil {
@@ -1302,8 +1302,12 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
 	return vm, nil
 }
 
-func IsSUSE(imageSpec string) bool {
-	return strings.HasPrefix(imageSpec, "suse-") || strings.HasPrefix(imageSpec, "opensuse-") || strings.Contains(imageSpec, "sles-")
+func IsSLES(vm_OS_ID string) bool {
+	return vm_OS_ID == "sles" || vm_OS_ID == "sles_sap"
+}
+
+func IsSUSE(vm_OS_ID string) bool {
+	return vm_OS_ID == "opensuse" || vm_OS_ID == "opensuse-leap" || vm_OS_ID == "sles" || vm_OS_ID == "sles_sap"
 }
 
 func IsCentOS(imageSpec string) bool {
@@ -1351,7 +1355,7 @@ func CreateInstance(origCtx context.Context, logger *log.Logger, options VMOptio
 			// windows-*-core instances sometimes fail to be ssh-able: b/305721001
 			(IsWindowsCore(options.ImageSpec) && strings.Contains(err.Error(), windowsStartupFailedMessage)) ||
 			// SLES instances sometimes fail to be ssh-able: b/186426190
-			(IsSUSE(options.ImageSpec) && strings.Contains(err.Error(), startupFailedMessage)) ||
+			strings.Contains(err.Error(), startupFailedMessage) ||
 			strings.Contains(err.Error(), prepareSLESMessage)
 	}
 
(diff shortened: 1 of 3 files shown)


integration_test/gce/gce_testing.go
```diff
@@ -1262,15 +1264,15 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
...
+	if IsSUSE(vm.ImageSpec) {
 		// Set download.max_silent_tries to 5 (by default, it is commented out in
 		// the config file). This should help with issues like b/211003972.
 		if _, err := RunRemotely(ctx, logger, vm, "sudo sed -i -E 's/.*download.max_silent_tries.*/download.max_silent_tries = 5/g' /etc/zypp/zypp.conf"); err != nil {
 			return nil, fmt.Errorf("attemptCreateInstance() failed to configure retries in zypp.conf: %v", err)
 		}
 	}
 
-	if strings.HasPrefix(vm.ImageSpec, "sles-") {
+	if vm.OS.ID == "sles" || vm.OS.ID != "sles_sap" {
```
Subbarker, line 1275: This seems wrong, won't this almost always be true?

franciscovalentecastro: Yes, that's wrong i made a typo when setting this `vm.OS.ID != "sles_sap"`. Should be `vm.OS.ID == "sles_sap"`, since i'm trying to do an `IsSLES` function.

integration_test/gce/gce_testing.go
```diff
@@ -1302,8 +1302,12 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
...
 }
 
-func IsSUSE(imageSpec string) bool {
-	return strings.HasPrefix(imageSpec, "suse-") || strings.HasPrefix(imageSpec, "opensuse-") || strings.Contains(imageSpec, "sles-")
+func IsSLES(vm_OS_ID string) bool {
+	return vm_OS_ID == "sles" || vm_OS_ID == "sles_sap"
+}
+
+func IsSUSE(vm_OS_ID string) bool {
+	return vm_OS_ID == "opensuse" || vm_OS_ID == "opensuse-leap" || vm_OS_ID == "sles" || vm_OS_ID == "sles_sap"
```
Subbarker, line 1310: this can call IsSLES, to reduce redundant code, e.g.
return vm_OS_ID == "opensuse" || vm_OS_ID == "opensuse-leap" || IsSLES(VM_OS_ID)

integration_test/gce/gce_testing.go
```diff
@@ -1550,7 +1554,7 @@ func InstallGsutilIfNeeded(ctx context.Context, logger *log.Logger, vm *VM) erro
 
 	// SUSE seems to be the only distro without gsutil, so what follows is all
 	// very SUSE-specific.
-	if vm.OS.ID != "opensuse" && vm.OS.ID != "opensuse-leap" && vm.OS.ID != "sles" && vm.OS.ID != "sles_sap" {
+	if IsSUSE(vm.OS.ID) {
```
Subbarker, line 1557: Shouldn't this be `!IsSUSE(vm.OS.ID)` ?

integration_test/gce/gce_testing.go
```diff
@@ -1351,7 +1355,7 @@ func CreateInstance(origCtx context.Context, logger *log.Logger, options VMOptio
 			// windows-*-core instances sometimes fail to be ssh-able: b/305721001
 			(IsWindowsCore(options.ImageSpec) && strings.Contains(err.Error(), windowsStartupFailedMessage)) ||
 			// SLES instances sometimes fail to be ssh-able: b/186426190
-			(IsSUSE(options.ImageSpec) && strings.Contains(err.Error(), startupFailedMessage)) ||
```
Subbarker, line 1354: This error message is very generic, so we may overhit this code path

integration_test/gce/gce_testing.go
```diff
@@ -1302,8 +1302,12 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
 	return vm, nil
 }
 
-func IsSUSE(imageSpec string) bool {
-	return strings.HasPrefix(imageSpec, "suse-") || strings.HasPrefix(imageSpec, "opensuse-") || strings.Contains(imageSpec, "sles-")
+func IsSLES(vm_OS_ID string) bool {
```
igorpeshansky, line 1305: Should these functions operate on `vm` or `vm.OS`?

integration_test/gce/gce_testing.go
igorpeshansky: +1

integration_test/gce/gce_testing.go
```diff
@@ -1302,8 +1302,12 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
...
+	return vm_OS_ID == "sles" || vm_OS_ID == "sles_sap"
+}
+
+func IsSUSE(vm_OS_ID string) bool {
+	return vm_OS_ID == "opensuse" || vm_OS_ID == "opensuse-leap" || vm_OS_ID == "sles" || vm_OS_ID == "sles_sap"
 }
 
 func IsCentOS(imageSpec string) bool {
```
igorpeshansky, line 1313: Should we plan to change the rest of these functions in a subsequent PR? Or fold it into this one?

If we change them all to operate on the `vm` instead of `vm.ImageSpec`, we can clean them up one-by-one later…

integration_test/gce/gce_testing.go
Subbarker: I think we should pass `vm`, it seems the most future-proof.

franciscovalentecastro: I created `IsSUSEVM(vm *VM) bool` that operates directly on `vm`.

integration_test/gce/gce_testing.go
franciscovalentecastro: I created `IsSUSEImageSpec(imageSpec string) bool` that operates on the `ImageSpec` to keep the same logic.

integration_test/gce/gce_testing.go
franciscovalentecastro: Fixed.

integration_test/gce/gce_testing.go
franciscovalentecastro: Done.

integration_test/gce/gce_testing.go
franciscovalentecastro: **> Should we plan to change the rest of these functions in a subsequent PR?**
I think we should update the rest of the functions, but i do expect unforseen consequences (other failures), so i propose to do the other updates in a followup PR.

Subbarker (dismissed):

Subbarker:
It looks like the soak test launcher test has a failure related to your change

franciscovalentecastro:
> It looks like the soak test launcher test has a failure related to your change

Yes, they were from my change. It was because i tried to call `getOS` before `waitForStart`, which i realized it won't work since `getOS` ssh's and run a command. I reverted the order and made sure within `waitForStartLinux` we use `IsSUSEImageSpec`, because the vm.OS.ID is not set at that moment.

Most `SLES` tests are passing (`SLES 15 AArch64` seems like a flake) and most importantly `third party apps integration test (SAP HANA x86_64)` passed, which was the one i wanted to fix. 

There is failure in `Ops Agent integration test (Rocky Linux 8 x86_64)` and `third party apps integration test (CentOS 7 x86_64)`, but i think they are flakes unrelated to this PR.

Subbarker (approved):

franciscovalentecastro:
Most tests on https://github.com/GoogleCloudPlatform/ops-agent/pull/1720/commits/4f19034c50df887836afd0c5e10f663548ad14a7 passed, only `Ops Agent integration test (Ubuntu 23.10 Mantic)` which didn't fail in previous run, so it's most probably a flake.

PR #1724, opened 2024-05-24
Ensure all start_states for multiline logs look for start characters

dehaansa:
## Description
1. Ensure that all multiline regexes specify the start of line character for their start_state. This resolves an ongoing customer issue with postgres audit logs.
2. Adds parser for SLES12 default log format for postgresql
3. Fixes existing postgresql parser to appropriately parse database and user from log.

## Related issue
https://metabug.corp.google.com/entity/145251722?audience=43
[b/316922937](http://b/316922937)

## How has this been tested?
Tested in integration tests in kokoro.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/apps/postgresql.go
+++ b/apps/postgresql.go
@@ -111,15 +111,32 @@ func (LoggingProcessorPostgresql) Type() string {
 func (p LoggingProcessorPostgresql) Components(ctx context.Context, tag string, uid string) []fluentbit.Component {
 	c := confgenerator.LoggingProcessorParseMultilineRegex{
 		LoggingProcessorParseRegexComplex: confgenerator.LoggingProcessorParseRegexComplex{
+			// Limited logging documentation: https://www.postgresql.org/docs/10/runtime-config-logging.html
 			Parsers: []confgenerator.RegexParser{
 				{
-					// Limited logging documentation: https://www.postgresql.org/docs/10/runtime-config-logging.html
+					// This parser matches most distributions' defaults by our testing
+					// log_line_prefix = '%m [%p] '
+					// log_line_prefix = '%m [%p] %u@%d '
 					// Sample line: 2022-01-12 20:57:58.378 UTC [26241] LOG:  starting PostgreSQL 14.1 (Debian 14.1-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
 					// Sample line: 2022-01-12 20:59:25.169 UTC [27445] postgres@postgres FATAL:  Peer authentication failed for user "postgres"
 					// Sample line: 2022-01-12 21:49:13.989 UTC [27836] postgres@postgres LOG:  duration: 1.074 ms  statement: select *
 					//    from pg_database
 					//    where 1=1;
-					Regex: `^(?<time>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}.\d{3,} \w+)\s*\[(?<tid>\d+)\](?:\s+(?<role>\S*)@(?<user>\S*))?\s*(?<level>\w+):\s+(?<message>[\s\S]*)`,
+					Regex: `^(?<time>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}.\d{3,} \w+)\s*\[(?<tid>\d+)\](?:\s+(?<user>\S*)@(?<database>\S*))?\s*(?<level>\w+):\s+(?<message>[\s\S]*)`,
+					Parser: confgenerator.ParserShared{
+						TimeKey:    "time",
+						TimeFormat: "%Y-%m-%d %H:%M:%S.%L %z",
+						Types: map[string]string{
+							"tid": "integer",
+						},
+					},
+				},
+				{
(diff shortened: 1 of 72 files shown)


dehaansa:
image
Only one postgres failure, re-running to see if it was a fluke or if I need to understand something SLES12 specific.

jefferbrecht:
> image Only one postgres failure, re-running to see if it was a fluke or if I need to understand something SLES12 specific.

Looks like `level` and `tid` are missing from `jsonPayload`:

```json
{
  "jsonPayload": {
    "message": "2024-05-30 15:34:26.572 UTC postgres postgres [23958]STATEMENT:  INSERT INTO \n\ttest2 (id) VALUES('1970-01-01 00:00:00.123 UTC');"
  },
...
```

dehaansa:
Based on testing, SLES12 has a different default log_line_prefix than other distributions. Added a separate parser for this.

Also discovered that the other default was incorrectly labeling database name as role, and had the order of user & database reversed. See documentation https://www.postgresql.org/docs/10/runtime-config-logging.html

```
Escape | Effect | Session only
-- | -- | --
%a | Application name | yes
%u | User name | yes
%d | Database name | yes
%r | Remote host name or IP address, and remote port | yes
%h | Remote host name or IP address | yes
%p | Process ID | no
%t | Time stamp without milliseconds | no
%m | Time stamp with milliseconds | no
%n | Time stamp with milliseconds (as a Unix epoch) | no
%i | Command tag: type of session's current command | yes
%e | SQLSTATE error code | no
%c | Session ID: see below | no
%l | Number of the log line for each session or process, starting at 1 | no
%s | Process start time stamp | no
%v | Virtual transaction ID (backendID/localXID) | no
%x | Transaction ID (0 if none is assigned) | no
%q | Produces no output, but tells non-session processes to stop at this point in the string; ignored by session processes | no
%% | Literal % | no
```

dehaansa:
The only failure on this last run was a fluke on Cassandra on Debian 10 which has passed previously 

```
=== FAIL: integration_test/third_party_apps_test TestThirdPartyApps/debian-cloud:debian-10/cassandra (298.98s)
    main_test.go:971: Test logs: https://console.cloud.google.com/storage/browser/ops-agents-public-buckets-test-logs/prod/stackdriver_agents/testing/consumer/third_party_apps/test/buster_x86_64/2167/20240530-111732/logs/TestThirdPartyApps_debian-cloud:debian-10_cassandra 
    main_test.go:1002: Instance Log: https://console.cloud.google.com/logs/viewer?resource=gce_instance%2Finstance_id%2F8841878630816906742&project=stackdriver-test-143416
    main_test.go:1007: Attempt 1 of debian-cloud:debian-10 test of cassandra finished with err=WaitForMetric(metric="workload.googleapis.com/cassandra.storage.load.count", extraFilters=[]): rpc error: code = Canceled desc = CANCELLED, retryable=false
    main_test.go:1013: Non-retryable error: WaitForMetric(metric="workload.googleapis.com/cassandra.storage.load.count", extraFilters=[]): rpc error: code = Canceled desc = CANCELLED
    ```

**Comment by dehaansa** [comment]
One fluke unrelated test before merging in master.

image

**Review by jefferbrecht — APPROVED** [review]

File: apps/postgresql.go [hunk 1]
```diff
@@ -130,13 +131,27 @@ func (p LoggingProcessorPostgresql) Components(ctx context.Context, tag string,
 						},
 					},
 				},
+				{
+					// This parser matches the default log_line_prefix from SLES12 in our testing
+					// log_line_prefix = '%m %d %u [%p]'
```
igorpeshansky, line 136: Is there really no space at the end? Sigh!

apps/postgresql.go
```diff
@@ -130,13 +131,27 @@ func (p LoggingProcessorPostgresql) Components(ctx context.Context, tag string,
...
+					// Sample line: 2024-05-30 15:34:26.572 UTC postgres postgres [23958]STATEMENT:  INSERT INTO
+					//					test2 (id) VALUES('1970-01-01 00:00:00.123 UTC');
+					Regex: `^(?<time>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}.\d{3,} \w+)\s*(?:\s+(?<database>\S*)\s+(?<user>\S*))?\s*\[(?<tid>\d+)\]\s*(?<level>\w+):\s+(?<message>[\s\S]*)`,
```
igorpeshansky, line 139: [Optional] Both of these regexes treat the user/database pair as optional, so I wonder if it makes sense to combine them into one and avoid adding a new parser…

igorpeshansky (approved):
LGTM, with one optional comment.

apps/postgresql.go
dehaansa: Yup! Confirmed.

apps/postgresql.go
dehaansa: I was concerned that fluentbit might not support two different capture groups with the same name, but didn't find anything in the documentation. Also concerned if it's not documented about the potential that it could change in the future.

igorpeshansky: You're right, this may not work in fluent-bit.

igorpeshansky (approved):
LGTM :shipit:

PR #1739, opened 2024-06-18
Testing: Add TEST_SELECTOR to Kokoro scripts to allow running a subset of the tests.

igorpeshansky:
## Description
Add a `TEST_SELECTOR` environment variable to Kokoro scripts to allow running a subset of integration or third-party application tests.

Also change third-party application test to be less aggressive about filtering tests by modified files and update integration test docs to mention `IMAGE_SPECS` instead of `PLATFORMS`.

## How has this been tested?
Manual invocations: [`rockylinux9_x86_64`](https://btx.cloud.google.com/invocations/bde1a673-499c-4a1a-a9df-f4b20ead2cdf/targets).

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/README.md
```diff
@@ -130,6 +130,11 @@ The test is designed so that simply modifying files in the
 right thing. But we do expect that we will need to make big changes to both the
 data directory and the test runner before it is really meeting our needs.
 
+Note that, by default, the test will skip any applications that were not
+impacted by the currently modified set of files. To force the test to include
+all available applications, regardless of the current modifications, set the
+`ALL_APPS` environment variable to `true`.
```
jefferbrecht, line 136: If I'm specifying `TEST_SELECTOR`, it's really unlikely that I'll ever want anything other than `ALL_APPS=true`. In that case, it'd be simpler if we just removed `ALL_APPS` and instead added a check for the presence of `-test.run` when deciding whether to skip the app filters.

For the use case of "I just want to run everything", it seems that `TEST_SELECTOR=.*` is more or less equivalent to `ALL_APPS=true` anyway (if we were to implement the previously suggested logic).

If you insist on having a separate toggle for skipping app filtering anyway, then I think I'd still prefer for it to be opt-out rather than opt-in, i.e. the default behaviour should be `ALL_APPS=true` and we override it in the presubmits with `ALL_APPS=false` -- let the burden of verbose invocations be on the automation rather than the manual user.

igorpeshansky: I had the same thought, and looked for a way to check for the presence of `-test.run` from within the test, but couldn't find any. I suppose I could check directly for `TEST_SELECTOR` in the environment, but `TEST_SELECTOR` is a Kokoro thing, not a test driver thing, and this wouldn't work if tests were invoked locally.

We could definitely toggle the default behavior, but I thought that was added specifically to speed up third-party tests during development (i.e., for the manual use case). Maybe it's just too aggressive… How about we treat the case where none of the files related to third-party app testing are modified as `ALL_APPS=true`, and only apply the current behavior when there are modifications to the apps themselves? Even in that case, I might be able to justify a separate toggle (e.g., if you want to run the apps you haven't modified, for comparison), but we can start there. WDYT?

igorpeshansky: Done in https://github.com/GoogleCloudPlatform/ops-agent/pull/1739/commits/c14c96a9fa080b4b65df5342a018056fda27b93a.

jefferbrecht: `test.run` is an ordinary flag that can be queried with `flag.Lookup("test.run") != nil`.

igorpeshansky: Heh, so it is. I think there's room for both, though.

jefferbrecht (approved):
LGTM; we can still add the check for `test.run` to further increase the likelihood of getting exactly the tests intended, but that can be a separate PR if you'd like.

integration_test/README.md
igorpeshansky: A check for `test.run` was added in https://github.com/GoogleCloudPlatform/ops-agent/pull/1739/commits/291f5b0134034770c6ec87e67a751104d8e02184.

jefferbrecht (approved):

PR #1747, opened 2024-07-03
Change syslog dependency to Recommends

braydonk:
## Description
While a syslog daemon is required for full desired Ops Agent functionality, our previous reliance on the presence of a syslog daemon was provided by the dependency specified in the `google-compute-engine` package, which is a `Recommends` dependency.

This PR changes the syslog daemon dependency in the Ops Agent deb package to `Recommends` instead of `Depends`.

## Related issue
[b/339061751](http://b/339061751)

## How has this been tested?
Manually tested the workflow where we install Ops Agent without installing `rsyslog`:

`make build` to get a local `deb` package.  
Created a new Debian 12 GCE VM.  
Remove `rsyslog` from the VM:
```
sudo apt purge rsyslog
```
Copied the deb package to the VM. Install it without installing Recommended packages:
```
sudo apt install --no-install-recommends ./google-cloud-ops-agent_2.49.0~debian12_amd64.deb
```
`rsyslog` was not installed, verified with:
```
dpkg -l | grep rsyslog
```

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -7,9 +7,8 @@ Standards-Version: 3.9.4
 
 Package: google-cloud-ops-agent
 Architecture: any
-Depends: rsyslog | system-log-daemon,
-         ${shlibs:Depends},
-         ${misc:Depends}
+Depends: ${shlibs:Depends}, ${misc:Depends}
+Recommends: rsyslog | system-log-daemon
 Conflicts: stackdriver-agent, google-fluentd
 Description: Google Cloud Ops Agent
  The Google Cloud Ops Agent collects metrics and logs from the system.


debian/control
```diff
@@ -7,9 +7,8 @@ Standards-Version: 3.9.4
 
 Package: google-cloud-ops-agent
 Architecture: any
-Depends: rsyslog | system-log-daemon,
-         ${shlibs:Depends},
-         ${misc:Depends}
+Depends: ${shlibs:Depends}, ${misc:Depends}
+Recommends: rsyslog | system-log-daemon
```
braydonk, line 11: There was discussion about swapping the order of these to put the virtual package `system-log-daemon` first. I found [these docs from Debian](https://www.debian.org/doc/debian-policy/ch-relationships.html#virtual-packages-provides) that suggest to use this ordering (virtual package second).

igorpeshansky: Are you referring to the following paragraph:
> To specify which of a set of real packages should be the default to satisfy a particular dependency on a virtual package, list the real package as an alternative before the virtual one.

? Any reason to not just use the virtual package as the dependency?

braydonk: I am referring to that paragraph.

If we just specify a virtual package, how will it know which one to install? I was under the impression we had to give it at least one real package to specify with the virtual as an alternative.

igorpeshansky: Makes sense.

igorpeshansky (approved):
LGTM :shipit:

braydonk:
Debian 11 failure is a flake. The rest of the failures are expected.

@igorpeshansky do you think this should be merged now or should we wait til those CI failures are resolved? This is slated for 2.49.0 but so are the fixes causing the CI failures in the PR anyway.

igorpeshansky (approved):
LGTM :shipit:

PR #1757, opened 2024-07-16
Install terraform in test containers

hsmatulisgoogle:
## Description
This will install terraform in test containers so that we may run our CI on the terraform modules

### DIFF
--- a/kokoro/scripts/test/go_test.Dockerfile
+++ b/kokoro/scripts/test/go_test.Dockerfile
@@ -15,6 +15,12 @@ RUN apt-get update
 
 RUN apt-get install --yes python3-yaml
 
+RUN apt-get update && apt-get install -y gnupg software-properties-common
+
+RUN wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg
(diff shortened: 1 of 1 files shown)


kokoro/scripts/test/go_test.Dockerfile
```diff
@@ -15,6 +15,12 @@ RUN apt-get update
 
 RUN apt-get install --yes python3-yaml
 
+RUN apt-get update && apt-get install -y gnupg software-properties-common
+
+RUN wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg
+RUN echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | tee /etc/apt/sources.list.d/hashicorp.list
+RUN apt update && apt install terraform
```
martijnvans, line 22: it's a bit unclear why this command uses apt instead of apt-get like the other commands.

martijnvans (approved):

kokoro/scripts/test/go_test.Dockerfile
igorpeshansky: For non-interactive usage like container builds, `apt-get` is preferred. `apt` is just a wrapper that provides a nicer interactive experience.

hsmatulisgoogle: Fixed

PR #1750, opened 2024-07-11
Add the ability to specify gcloud configuration directory in tests.

igorpeshansky:
## Description
Add the ability to specify a unique `gcloud` configuration directory per test.
This should prevent errors like:
```
ERROR: (gcloud.beta.compute.instances.create) You do not currently have an active account selected.
```

Also fix a compilation bug in `agents.go` that manifests with newer Go versions.

## Related issue
[b/352054649](http://b/352054649)

## How has this been tested?
Integration tests.

## Checklist:
- Unit tests
  - [x] Unit tests do not apply.
  - [ ] Unit tests have been added/modified and passed for this PR.
- Integration tests
  - [ ] Integration tests do not apply.
  - [x] Integration tests have been added/modified and passed for this PR.
- Documentation
  - [x] This PR introduces no user visible changes.
  - [ ] This PR introduces user visible changes and the corresponding documentation change has been made.
- Minor version bump
  - [x] This PR introduces no new features.
  - [ ] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.
  - [ ] This PR bumps the version.

PR #1761, opened 2024-07-25
Copy the existing gcloud config directory if it exists.

igorpeshansky:
## Description
Copy the existing `gcloud` config directory if it exists. This will avoid errors in situations where testing relies on this configuration, e.g., when running integration tests locally with cached credentials. A follow-on to #1750.

## Related issue
[b/352054649](http://b/352054649)

## How has this been tested?
Integration tests.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/gce/gce_testing.go
```diff
@@ -739,6 +739,95 @@ func runCommand(ctx context.Context, logger *log.Logger, stdin io.Reader, args [
...
+		"active_config",
+		"access_tokens.db",
+		"credentials.db",
+		"default_configs.db",
+	}
+	currentConfigDir, err := getGcloudConfigDir(ctx)
+	if err != nil {
+		return err
+	}
+	sfi, err := os.Stat(currentConfigDir)
```
martijnvans, line 786: this is a lot of code! [Stackoverflow](https://stackoverflow.com/a/77849003) says that go 1.23 will likely get a function to do (approximately) this, but in the meantime, why not just runCommand() a "cp --recursive" command? Do we really need to copy only these relevantFiles or something?

integration_test/gce/gce_testing.go
```diff
@@ -739,6 +739,95 @@ func runCommand(ctx context.Context, logger *log.Logger, stdin io.Reader, args [
 	return output, err
 }
 
+// getGcloudConfigDir returns the current gcloud configuration directory.
+func getGcloudConfigDir(ctx context.Context) (string, error) {
+	null, err := os.Open(os.DevNull)
+	if err != nil {
+		panic(fmt.Sprintf("unable to open os.DevNull: %v", err))
```
martijnvans, line 746: why not just return the error normally instead of panicing? panicing causes stuff like test cleanup to not run, which I guess isn't so bad now that we have automatic VM cleanup, but anyway. Maybe it causes stuff to show up worse in ResultStore? I'd rather just have a normal error be returned.

integration_test/gce/gce_testing.go
```diff
@@ -739,6 +739,95 @@ func runCommand(ctx context.Context, logger *log.Logger, stdin io.Reader, args [
...
+
+// SetupGcloudConfigDir sets up a new gcloud configuration directory.
+// This copies the "configurations" subdirectory of the context-specified
+// configuration directory into the new directory.
+func SetupGcloudConfigDir(ctx context.Context, directory string) error {
+	relevantFiles := []string{
+		"active_config",
+		"access_tokens.db",
+		"credentials.db",
+		"default_configs.db",
```
martijnvans, line 780: I worry that this list will go stale, since we are reaching into gcloud internals here. Why not just copy everything?

integration_test/gce/gce_testing.go
```diff
@@ -739,6 +739,95 @@ func runCommand(ctx context.Context, logger *log.Logger, stdin io.Reader, args [
 	return output, err
 }
 
+// getGcloudConfigDir returns the current gcloud configuration directory.
+func getGcloudConfigDir(ctx context.Context) (string, error) {
+	null, err := os.Open(os.DevNull)
+	if err != nil {
+		panic(fmt.Sprintf("unable to open os.DevNull: %v", err))
+	}
+	out, err := RunGcloud(ctx, log.New(null, "", 0), "", []string{"info", "--format=value[terminator=''](config.paths.global_config_dir)"})
```
martijnvans, line 748: If we run this once in init() and save the result, we don't have to run it a bunch of times and don't have to have extra code for logging the output to /dev/null either.

Plus, we can do whatever we need to do to "warm up" the configuration directory on Kokoro, so that gcloud doesn't need to regenerate the same 4 files hundreds of times per test run. Unless the "relevantFiles" are already generated somehow by this point when running on Kokoro...

integration_test/gce/gce_testing.go
```diff
@@ -739,6 +739,95 @@ func runCommand(ctx context.Context, logger *log.Logger, stdin io.Reader, args [
...
+	if err != nil {
+		return "", fmt.Errorf("error running gcloud info: %w", err)
```
martijnvans, line 750: nit: RunGcloud's returned error will already include the full arguments list, no need to reiterate what it tried to do in the error message.

integration_test/gce/gce_testing.go
igorpeshansky: This was one of those "should never ever happen" scenarios. However, it's even better to use `io.Discard`, so done.

integration_test/gce/gce_testing.go
igorpeshansky: The whole point is to create a unique per-test configuration directory, to prevent data races between tests. I agree that _within a test_ we shouldn't have to re-create the directory. But this is only invoked in code that is used at the start of the tests…

integration_test/gce/gce_testing.go
igorpeshansky: Good point. Done.

integration_test/gce/gce_testing.go
martijnvans: fair enough

integration_test/gce/gce_testing.go
martijnvans: I think I was proposing:
- cp -r for now, with a TODO I guess
- replace that with os.CopyFS when it becomes available, because it's less hacky

This saves us about 50+ lines of code of hand-rolled directory copying code. Like for example, `defer in.Close()` drops errors from closing files, but we don't need to hash that out if you are just calling `cp -r`. So yeah I think `cp -r` is worth doing, even if only to save time going back and forth in reviews.

igorpeshansky: Done.

integration_test/gce/gce_testing.go
```diff
@@ -739,6 +739,31 @@ func runCommand(ctx context.Context, logger *log.Logger, stdin io.Reader, args [
...
+// This copies the "configurations" subdirectory of the context-specified
+// configuration directory into the new directory.
+// This only works on Linux.
+func SetupGcloudConfigDir(ctx context.Context, directory string) error {
+	currentConfigDir, err := getGcloudConfigDir(ctx)
+	if err != nil {
+		return err
+	}
+	// TODO: Replace with os.CopyFS() once available.
+	if out, err := runCommand(ctx, log.New(io.Discard, "", 0), nil, []string{"cp", "-r", filepath.Join(currentConfigDir, "."), filepath.Join(directory, ".")}, nil); err != nil {
```
martijnvans, line 761: why are you adding a `/.` on the end of `currentConfigDir` and `directory`?

Seems unnecessary

integration_test/gce/gce_testing.go
```diff
@@ -739,6 +739,31 @@ func runCommand(ctx context.Context, logger *log.Logger, stdin io.Reader, args [
...
+		return fmt.Errorf("error copying directory contents: %s (%w)", out.Stderr, err)
```
martijnvans, line 762: runCommand's returned `err` already contains the command and also out.Stderr (technically, interleaved stdout+stderr which is usually more useful). So again there is no need to repeat that stuff here, you can just return `err`.

integration_test/gce/gce_testing.go
```diff
@@ -739,6 +739,31 @@ func runCommand(ctx context.Context, logger *log.Logger, stdin io.Reader, args [
...
+func getGcloudConfigDir(ctx context.Context) (string, error) {
+	out, err := RunGcloud(ctx, log.New(io.Discard, "", 0), "", []string{"info", "--format=value[terminator=''](config.paths.global_config_dir)"})
+	if err != nil {
+		return "", err
+	}
+	return out.Stdout, nil
+}
+
+// SetupGcloudConfigDir sets up a new gcloud configuration directory.
+// This copies the "configurations" subdirectory of the context-specified
```
martijnvans, line 752: this comment is stale, it copies the whole thing.

martijnvans (approved):
some nits left, nothing much

integration_test/gce/gce_testing.go
igorpeshansky: Done.

integration_test/gce/gce_testing.go
igorpeshansky: The `/.` on the source directory is to ensure that only the contents are copied (https://askubuntu.com/a/1151667). The `/.` on the target directory is to ensure that it exists (and is a directory).

integration_test/gce/gce_testing.go
igorpeshansky: Done.

jefferbrecht (approved):

PR #1772, opened 2024-08-02
Clean up and update application metadata files.

igorpeshansky:
## Description
Bring the third-party application `metadata.yaml` files up-to-date with the public docs, in preparation for eventually using them as the source of truth.

Also add some more YAML validation and improve test output.

## Related issue
[b/240700717](http://b/240700717)

## How has this been tested?
Except for the new code, which was tested by unit tests, all of the `metadata.yaml` changes have been run through an automatic doc generator and compared with the existing documentation. Some of the changes also update the scripts that run in the third-party application tests — those have passed.

## Checklist:
- Unit tests
  - [ ] Unit tests do not apply.
  - [x] Unit tests have been added/modified and passed for this PR.
- Integration tests
  - [ ] Integration tests do not apply.
  - [x] Integration tests have been added/modified and passed for this PR.
- Documentation
  - [x] This PR introduces no user visible changes.
  - [ ] This PR introduces user visible changes and the corresponding documentation change has been made.
- Minor version bump
  - [x] This PR introduces no new features.
  - [ ] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.
  - [ ] This PR bumps the version.

PR #1773, opened 2024-08-04
Add description, units, and notes to metrics/labels in application metadata files.

igorpeshansky:
## Description
This builds on #1772 to add `description` and `notes` fields to metrics and metric labels, as well as a `unit` field to metrics.

As an example of how to use this, descriptions, units, and notes are added to the `dcgm/metadata.yaml` file. Other files can be filled in as a follow-up.

In addition, it makes the following changes to the MySQL/MariaDB metadata (currently not used as the source of truth for the docs):
- Add notes to logs and log fields to reflect the dependency on the MySQL/MariaDB logging options.
- Remove mention of stale `application_version` configuration option.
- Document correct defaults for `endpoint` and `username`.
- Remove extraneous test-only log collection for MariaDB.

## Related issue
[b/354042783](http://b/354042783)

## How has this been tested?
The `dcgm` metrics can now render as follows:
image

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


mtabasko (approved):
LGTM! Thanks.

integration_test/third_party_apps_test/main_test.go
```diff
@@ -624,26 +624,41 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
...
-		for _, m := range metadata.ExpectedMetrics {
+		for _, m := range integrationMetadata.ExpectedMetrics {
 			// The windows metrics that do not target workload.googleapis.com cannot set
 			// the instrumentation_* labels
 			if strings.HasPrefix(m.Type, "agent.googleapis.com") {
 				continue
 			}
 			if m.Labels == nil {
-				m.Labels = map[string]string{}
+				m.Labels = []*metadata.MetricLabel{}
```
jefferbrecht, line 634: Wouldn't this code be much shorter if we kept `m.Labels` as a map?

igorpeshansky: It would, but then we wouldn't be able to reuse labels via anchors/aliases.

jefferbrecht: Do you mean YAML anchors/aliases? I didn't see any such reuse in this PR, so it feels like added complexity without realizing its benefits.

igorpeshansky: You missed them in `dcgm/metadata.yaml` (e.g., `model_label`).

jefferbrecht: Whoops, you're right -- my mistake!

integration_test/metadata/integration_metadata.go
```diff
@@ -209,31 +233,35 @@ func AssertMetric(metric *ExpectedMetric, series *monitoringpb.TimeSeries) error
...
-	// Only expected labels must be present
 	var err error
+	// Only expected labels must be present
+	expectedLabels := make(map[string]bool)
+	for _, expectedLabel := range metric.Labels {
+		expectedLabels[expectedLabel.Name] = true
+	}
 	for actualLabel, actualValue := range series.Metric.Labels {
-		if _, ok := metric.Labels[actualLabel]; !ok {
+		if _, ok := expectedLabels[actualLabel]; !ok {
```
jefferbrecht, line 243: Since we're only populating `expectedLabels` with `true` values:

```suggestion
		if !expectedLabels[actualLabel]; {
```

jefferbrecht (approved):
LGTM, just a minor non-blocking comment.

integration_test/metadata/integration_metadata.go
jefferbrecht: Sigh, I typo'd the suggestion -- that trailing semicolon shouldn't be there.

igorpeshansky: Fixed.

integration_test/metadata/integration_metadata.go
```diff
@@ -87,11 +87,15 @@ type LogField struct {
 	// A list of platforms that this field is not available on.
 	// Examples: debian-11.
 	UnavailableOn []string `yaml:"unavailable_on,omitempty"`
+	// Annotations/footnotes about the field.
+	Notes []string `yaml:"notes,omitempty" validate:"omitempty,unique"`
```
jefferbrecht, line 91: Does this need to be `[]*string` too, for anchors to work?

integration_test/third_party_apps_test/applications/dcgm/metadata.yaml
```diff
@@ -115,8 +115,8 @@ expected_metrics:
       - *gpu_number_label
       - &direction_label
         name: direction
-        value_regex: transmit|receive
-        description: "Direction of the link traffic, one of [transmit, receive]."
+        value_regex: tx|rx
```
jefferbrecht, line 118: Why didn't our existing tests fail here?

integration_test/third_party_apps_test/applications/mariadb/metadata.yaml
```diff
@@ -364,17 +383,14 @@ configuration_options:
           default: null
           description: This value must be `mysql`.
         - name: endpoint
-          default: localhost:3306
-          description: The url exposed by MySQL.
+          default: /var/run/mysqld/mysqld.sock
+          description: The `hostname:port` or Unix socket path starting with `/` used to connect to the MySQL server.
```
jefferbrecht, line 387: If the goal is to match the public docs, then should this be reverted? The docs say `localhost:3306` today: https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/mariadb#MariaDB-metrics

If this change was intentional, then should we change the nomenclature here from `MySQL server` to `MariaDB server`?

integration_test/third_party_apps_test/applications/dcgm/metadata.yaml
jefferbrecht: Nevermind, I was using the "commits since last review" diff and I think there was an intermediate commit where this said `transmit|receive`.

integration_test/metadata/integration_metadata.go
igorpeshansky: Not really, it would work either way. The only thing using pointers would buy you is that YAML aliases would be aliased in Go, rather than copied. Since these are strings, I didn't bother. None of the notes use pointers.

integration_test/third_party_apps_test/applications/mariadb/metadata.yaml
igorpeshansky: You're right, sent a change to update the existing docs manually (since our current MySQL/MariaDB docs aren't generated). After that change is submitted and staged, this will match the docs.

> If this change was intentional, then should we change the nomenclature here from `MySQL server` to `MariaDB server`?

Good catch. Done.

mtabasko: Fix CL submitted! Good catch, Jeff. Thanks.

mtabasko (approved):
LGTM!

PR #1779, opened 2024-08-15
fluent-bit: update to fluent bit 3.1.6

ridwanmsharif:
## Description
This change picks up these changes from Fluent Bit:
fluent/fluent-bit#9193  - fixes issue when large log payloads are parsed to json 
https://github.com/fluent/fluent-bit/pull/9201 - fixes DST/ST timezone issue
https://github.com/fluent/fluent-bit/pull/9209 - fixes a memory leak during `in_forward` exception handling
https://github.com/fluent/fluent-bit/pull/9182 - adds a latency metric to self metrics - not affecting the ops agent since we don't ingest it yet

There are some additional bug fixes in the `in_forward` plugin that patches some memory leaks but we're likely unaffected by it since it leaks only when dealing with traces and metrics.

## Related issue
Related bug: [b/359600424](http://b/359600424)

### DIFF
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -61,7 +61,7 @@ RUN $files = (iwr -UseBasicParsing https://raw.githubusercontent.com/slproweb/op
     iwr -UseBasicParsing -Uri \"$($installer.url)\" -OutFile /local/openssl.msi; `
     $ActualHash = $(Get-FileHash /local/openssl.msi -Algorithm SHA256).Hash.ToLower(); `
     if ($ActualHash -ne \"$($installer.sha256)\") { `
-    throw \"OpenSSL hash mismatch. Expected: $($installer.sha256), Actual: $ActualHash\" `
+      throw \"OpenSSL hash mismatch. Expected: $($installer.sha256), Actual: $ActualHash\" `
     }; `
     Start-Process msiexec.exe -Wait -ArgumentList \"/i C:\local\openssl.msi /quiet\";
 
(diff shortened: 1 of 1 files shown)


igorpeshansky (approved):
This also picks up https://github.com/fluent/fluent-bit/commit/c7c0c09dc1670c3827b09a043e8686b76d1791cc and a couple of `in_forward` fixes. Worth mentioning?

Otherwise LGTM :shipit:

Dockerfile.windows
```diff
@@ -61,7 +61,7 @@ RUN $files = (iwr -UseBasicParsing https://raw.githubusercontent.com/slproweb/op
     iwr -UseBasicParsing -Uri \"$($installer.url)\" -OutFile /local/openssl.msi; `
     $ActualHash = $(Get-FileHash /local/openssl.msi -Algorithm SHA256).Hash.ToLower(); `
     if ($ActualHash -ne \"$($installer.sha256)\") { `
-      throw \"OpenSSL hash mismatch. Expected: $($installer.sha256), Actual: $ActualHash\" `
+    throw \"OpenSSL hash mismatch. Expected: $($installer.sha256), Actual: $ActualHash\" `
```
igorpeshansky, line 64: Any reason this was unindented?

Dockerfile.windows
```diff
@@ -100,7 +100,79 @@ COPY submodules/fluent-bit /work/submodules/fluent-bit
 
 WORKDIR /work/submodules/fluent-bit/build
 
-RUN cmake -G "'Visual Studio 16 2019'" -DCMAKE_BUILD_TYPE=RELWITHDEBINFO -DFLB_OUT_KINESIS_STREAMS=OFF ../;
```
igorpeshansky, line 103: [Optional] Should we have kept `-DFLB_OUT_KINESIS_STREAMS=Off`?

igorpeshansky (approved):
A couple of minor questions, otherwise LGTM :shipit:

Dockerfile.windows
ridwanmsharif: My IDE auto formats dockerfiles, but I don't know why it insists that this shouldn't be indented. I added it back cause it looks right

Dockerfile.windows
ridwanmsharif: I chose to keep things as close to the options before the upgrade to 3.16 and before the upgrade the flag we were passing in was ignored and out_kinesis was always on.

I'll separately create a bug to clean up what plugins are built with our build - there is potentially more room for cleanup.

PR #1805, opened 2024-09-25
Add dpkg cache lock timeout when setting up VM

jefferbrecht:
## Description
Trying to address failures to capture the dpkg cache lock when installing packages during an integration test.

## Related issue
[b/369656678](http://b/369656678)

## How has this been tested?
Will let presubmits run.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/agents/agents.go
+++ b/integration_test/agents/agents.go
@@ -738,10 +738,10 @@ func InstallOpsAgent(ctx context.Context, logger *log.Logger, vm *gce.VM, locati
 	}
 
 	preservedEnvironment := map[string]string{
-		"REPO_SUFFIX":              location.repoSuffix,
-		"REPO_CODENAME":            location.repoCodename,
-		"ARTIFACT_REGISTRY_PROJECT":             location.artifactRegistryProject,
-		"ARTIFACT_REGISTRY_REGION": location.artifactRegistryRegion,
+		"REPO_SUFFIX":               location.repoSuffix,
+		"REPO_CODENAME":             location.repoCodename,
+		"ARTIFACT_REGISTRY_PROJECT": location.artifactRegistryProject,
+		"ARTIFACT_REGISTRY_REGION":  location.artifactRegistryRegion,
 	}
 
 	if gce.IsWindows(vm.ImageSpec) {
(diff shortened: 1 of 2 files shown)


integration_test/agents/agents.go
```diff
@@ -910,7 +910,9 @@ func InstallPackageFromGCS(ctx context.Context, logger *log.Logger, vm *gce.VM,
 	// 1. install stable package from Rapture
 	// 2. install just-built package from GCS
 	// Nor do I know why apt considers that sequence to be a downgrade.
-	if _, err := gce.RunRemotely(ctx, logger, vm, "sudo apt install --allow-downgrades --yes --verbose-versions /tmp/agentUpload/*"); err != nil {
+	// TODO(b/369656678): We see frequent errors like "Waiting for cache lock:
+	//   Could not get lock /var/lib/dpkg/lock-frontend", so add a generous timeout.
+	if _, err := gce.RunRemotely(ctx, logger, vm, "sudo apt install --allow-downgrades --yes --verbose-versions --option DPkg::Lock::Timeout=300 /tmp/agentUpload/*"); err != nil {
```
igorpeshansky, line 915: Should we take the opportunity to turn this into `apt-get`?

jefferbrecht:
@igorpeshansky: An alternative that might be more robust: what if we instead do something like `echo 'DPkg::Lock::Timeout=300' > /etc/dpkg/dpkg/cfg.d/cache-lock-timeout.cfg` during VM setup? That'd also help avoid e.g. running into this during 3P tests that need to install stuff at startup.

igorpeshansky:
> @igorpeshansky: An alternative that might be more robust: what if we instead do something like `echo 'DPkg::Lock::Timeout=300' > /etc/dpkg/dpkg/cfg.d/cache-lock-timeout.cfg` during VM setup? That'd also help avoid e.g. running into this during 3P tests that need to install stuff at startup.

Yes, this does sound like a more general approach.

integration_test/gce/gce_testing.go
```diff
@@ -1394,6 +1402,10 @@ func IsARM(imageSpec string) bool {
 	return strings.Contains(imageSpec, "arm64")
 }
 
+func IsUsingDPKG(imageSpec string) bool {
```
igorpeshansky, line 1405: Might be worth renaming this to `IsDebianBased` and updating line 316…

igorpeshansky (approved):
LGTM :shipit:

PR #1818, opened 2024-10-24
Test Ubuntu 24.10 Oracular support

XuechunHou:
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1024,7 +1024,7 @@ ARG OPENJDK_MAJOR_VERSION
 RUN set -x; apt-get update && \
 		DEBIAN_FRONTEND=noninteractive apt-get -y install git systemd \
 		autoconf libtool libcurl4-openssl-dev libltdl-dev libssl-dev libyajl-dev \
-		build-essential cmake bison flex file libsystemd-dev \
+		build-essential cmake bison flex file systemd-dev debhelper libsystemd-dev \
 		devscripts cdbs pkg-config openjdk-${OPENJDK_MAJOR_VERSION}-jdk zip
 
 SHELL ["/bin/bash", "-c"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1014,6 +1014,107 @@ FROM scratch AS noble
 COPY --from=noble-build /tmp/google-cloud-ops-agent.tgz /google-cloud-ops-agent-ubuntu-noble.tgz
 COPY --from=noble-build /google-cloud-ops-agent*.deb /
 
+# ======================================
+# Build Ops Agent for ubuntu-oracular
+# ======================================
+
+FROM ubuntu:oracular AS oracular-build-base
+ARG OPENJDK_MAJOR_VERSION
+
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -1014,6 +1014,107 @@ FROM scratch AS noble
...
+# Build Ops Agent for ubuntu-oracular
+# ======================================
+
+FROM ubuntu:oracular AS oracular-build-base
+ARG OPENJDK_MAJOR_VERSION
+
+RUN set -x; apt-get update && \
+		DEBIAN_FRONTEND=noninteractive apt-get -y install git systemd \
+		autoconf libtool libcurl4-openssl-dev libltdl-dev libssl-dev libyajl-dev \
+		build-essential cmake bison flex file libsystemd-dev \
```
franciscovalentecastro, line 1027: Adding `systemd-dev` to this list should fix the `RUN ./systemd.sh /work/cache/` build error.

franciscovalentecastro: Also it requires to add `debhelper` here and update to fluent-bit v3.1.9 so the build works.

dockerfiles/compile.go
```diff
@@ -201,6 +201,18 @@ RUN ln -fs /usr/lib/systemd /lib/systemd` + installJava + installCMake,
 		tar_distro_name:   "ubuntu-noble",
 		package_extension: "deb",
 	},
+	{
+		from_image:  "ubuntu:oracular",
+		target_name: "oracular",
+		install_packages: `RUN set -x; apt-get update && \
+		DEBIAN_FRONTEND=noninteractive apt-get -y install git systemd \
+		autoconf libtool libcurl4-openssl-dev libltdl-dev libssl-dev libyajl-dev \
+		build-essential cmake bison flex file libsystemd-dev \
```
franciscovalentecastro, line 210: Need to add `systemd-dev` and `debhelper` here. It is also need a newer version of fluent-bit. I tried 3.1.9 and it works. Created https://github.com/GoogleCloudPlatform/ops-agent/pull/1819 for the fluent-bit update.

Dockerfile
franciscovalentecastro: Moved this comment to the correct place in `compile.go`.

dockerfiles/compile.go
franciscovalentecastro: No need to update fluent-bit, the current version in the repo 3.1.6 builds correctly in Ubuntu Oracular.

kokoro/config/test/ops_agent/presubmit/bullseye_aarch64.gcl
jefferbrecht, line 1: Note that bullseye-aarch64 went EOL and these tests didn't actually run (this might be why we're not seeing the postgres failures on this PR). We should apply these temporary changes to e.g. bookworm instead of bullseye (and we should also get rid of all of the bullseye-aarch64 stuff in a separate PR later).

PR #1858, opened 2025-01-15
A basic Ops Agent Plugin

XuechunHou:
## Description
1. Created the directory structure for Ops Agent plugin.
2.  Added a basic Ops Agent plugin that defines placeholder methods for RPC endpoints required by `plugin_comm.proto`.  Actual implementation will come later. 
3.  Have to maintain a copy of the [plugin_comm.proto](https://github.com/GoogleCloudPlatform/google-guest-agent/blob/main/internal/plugin/proto/plugin_comm.proto) in the Ops Agent repo for now, because UAP [hasn't exposed it for external dependency yet](https://github.com/GoogleCloudPlatform/google-guest-agent/blob/main/internal/plugin/proto/plugin_comm.proto). 

## Related issue
[b/380277488](https://b.corp.google.com/issues/380277488)

## How has this been tested?
This PR only includes Initial setup of the Ops Agent plugin with basic structure in place.  Tests will be developed alongside implementation in later PRs. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


cmd/ops_agent_uap_plugin/plugin.go
```diff
@@ -0,0 +1,70 @@
...
+	pb.RegisterGuestAgentPluginServer(server, ps)
+
+	// ctx := context.Background()
+	// ps.GetStatus(ctx, &pb.GetStatusRequest{})
+	// ps.Start(ctx, &pb.StartRequest{})
+	// ps.Start(ctx, &pb.StartRequest{})
+	// ps.GetStatus(ctx, &pb.GetStatusRequest{})
+	// ps.Stop(ctx, &pb.StopRequest{})
+	// ps.GetStatus(ctx, &pb.GetStatusRequest{})
+	// ps.Stop(ctx, &pb.StopRequest{})
```
jinghan-ma, line 64: These seem to be test code

cmd/ops_agent_uap_plugin/service.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// Start starts the plugin and initiates the plugin functionality.
+// Until plugin receives Start request plugin is expected to be not functioning
+// and just listening on the address handed off waiting for the request.
+func (ps *OpsAgentPluginServer) Start(ctx context.Context, msg *pb.StartRequest) (*pb.StartResponse, error) {
+	if ps.cancel != nil {
+		log.Printf("The Ops Agent plugin is started already, skipping the current request")
+		return nil, status.Errorf(1, "Ops Agent Plugin is started already")
+	}
```
jinghan-ma, line 36: Does UAP define what the plugin should do if a plugin that's already started receives another `Start` signal? In particular, should it return an error or should it only return an grpc error if there's a failure in the startup process?

(The same question applies to `Stop`)

XuechunHou: Thanks for pointing out.  UAP team says that ideally we don't need to worry about handling duplicate requests, but I am keeping the check just in case. 

I have updated the code to only log the duplicate requests, instead of throwing errors as it'd trigger retries.

cmd/ops_agent_uap_plugin/plugin.go
XuechunHou: deleted

jinghan-ma (approved):
LGTM. Getting this on master should help us with testing the release pipeline for the plugin.

PR #1861, opened 2025-01-17
Create `otel_logging_supported_config` feature tracking metric.

franciscovalentecastro:
## Description
Create `otel_logging_support` feature tracking metric.

```
{
    Module: "logging",
    Kind:   "service",
    Type:   "otel_logging",
    Key:    []string{"otel_logging_supported_config"},
    Value:  "true / false",
}
```

## Related issue
b/390671495

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -72,6 +72,17 @@ func googleManagedPrometheusExporter(userAgent string) otel.Component {
 	}
 }
 
+func (uc *UnifiedConfig) getOTelLogLevel() string {
+	logLevel := "info"
+	if uc.Metrics != nil && uc.Metrics.Service != nil && uc.Metrics.Service.LogLevel != "" {
+		logLevel = uc.Metrics.Service.LogLevel
+	}
+	if uc.Logging != nil && uc.Logging.Service != nil && uc.Logging.Service.LogLevel != "" && uc.Logging.Service.OTelLogging {
+		logLevel = uc.Logging.Service.LogLevel
+	}
+	return logLevel
+}
+
 func (uc *UnifiedConfig) GenerateOtelConfig(ctx context.Context) (string, error) {
 	p := platform.FromContext(ctx)
 	userAgent, _ := p.UserAgent("Google-Cloud-Ops-Agent-Metrics")
@@ -101,11 +112,8 @@ func (uc *UnifiedConfig) GenerateOtelConfig(ctx context.Context) (string, error)
 		ReceiverPipelineName: "fluentbit",
 	}
 
-	if uc.Metrics.Service.LogLevel == "" {
-		uc.Metrics.Service.LogLevel = "info"
-	}
 	otelConfig, err := otel.ModularConfig{
-		LogLevel:          uc.Metrics.Service.LogLevel,
+		LogLevel:          uc.getOTelLogLevel(),
 		ReceiverPipelines: receiverPipelines,
 		Pipelines:         pipelines,
 		Exporters: map[otel.ExporterType]otel.Component{
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -23,7 +23,6 @@ import (
 	"log"
 	"maps"
 	"path"
-	"reflect"
 	"regexp"
 	"sort"
 	"strconv"
@@ -197,7 +196,6 @@ func (p pipelineInstance) fluentBitComponents(ctx context.Context) (fbSource, er
 func (p pipelineInstance) otelComponents(ctx context.Context) (map[string]otel.ReceiverPipeline, map[string]otel.Pipeline, error) {
 	outR := make(map[string]otel.ReceiverPipeline)
 	outP := make(map[string]otel.Pipeline)
-	fmt.Println("p.receiver = ", p.receiver, "typeof = ", reflect.TypeOf(p.receiver))
 	receiver, ok := p.receiver.(OTelReceiver)
 	if !ok {
 		return nil, nil, fmt.Errorf("%q is not an otel receiver", p.rID)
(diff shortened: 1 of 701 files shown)


confgenerator/config.go
```diff
@@ -1139,6 +1151,46 @@ func (uc *UnifiedConfig) OTelLoggingReceivers(ctx context.Context) (map[string]O
...
 }
 
+func (uc *UnifiedConfig) OTelLoggingProcessors(ctx context.Context) (map[string]OTelProcessor, error) {
+	processors, err := uc.LoggingProcessors(ctx)
+	if err != nil {
+		return nil, err
+	}
+	validProcessors := map[string]OTelProcessor{}
+	for k, v := range processors {
+		if v, ok := v.(OTelProcessor); ok {
```
quentinmit, line 1161: This doesn't actually check if the processor is supported, only if it *might* be supported.

Instead of doing all this work to separately validate, I would just clone the config, set the experimental flag, and see if generating causes an error:

```go
func (uc *UnifiedConfig) OTelLoggingSupported() bool {
  uc2 := uc.DeepCopy(ctx)
  if uc2.Logging == nil {
    return false
  }
  if uc2.Logging.Service == nil {
    uc2.Logging.Service = &Service{}
  }
  uc2.Logging.Service.OTelLogging = true
  _, err := uc2.GenerateOtelConfig(ctx)
  return err == nil
}

```

confgenerator/feature_tracking.go
```diff
@@ -437,6 +439,23 @@ func getMetadata(field reflect.StructField) metadata {
...
+func getOTelLoggingSupportedConfig(uc *UnifiedConfig) Feature {
+	feature := Feature{
+		Module: "logging",
+		Kind:   "service",
+		Type:   "otel_logging",
+		Key:    []string{"otel_logging_supported_config"},
+		Value:  "false",
+	}
+
+	ctx := context.Background()
```
quentinmit, line 451: This needs to be the `Context` from the unit test, not `context.Background()`. It looks like that needs to be passed into `ExtractFeatures`.

confgenerator/confgenerator.go
```diff
@@ -101,11 +112,8 @@ func (uc *UnifiedConfig) GenerateOtelConfig(ctx context.Context) (string, error)
 		ReceiverPipelineName: "fluentbit",
 	}
 
-	if uc.Metrics.Service.LogLevel == "" {
```
quentinmit, line 104: Your current diff will cause the logging log level to override the metrics log level. I think we should stick with the existing behavior of only using the metrics log level, which you could do with

```go
logLevel := "info"
if uc.Metrics != nil && uc.Metrics.Service != nil && uc.Metrics.Service.LogLevel != "" {
  logLevel = uc.Metrics.Service.LogLevel
}
```

confgenerator/config.go
```diff
@@ -907,15 +907,17 @@ func (uc *UnifiedConfig) ValidateCombined() error {
 
 func (uc *UnifiedConfig) MetricsReceivers() (map[string]MetricsReceiver, error) {
 	validReceivers := map[string]MetricsReceiver{}
-	if uc.Metrics != nil {
+	if uc.Metrics != nil && uc.Metrics.Receivers != nil {
```
quentinmit, line 910: This is unnecessary; it's valid to iterate over a `nil` map.

confgenerator/config.go
```diff
@@ -907,15 +907,17 @@ func (uc *UnifiedConfig) ValidateCombined() error {
...
 		for k, v := range uc.Metrics.Receivers {
 			validReceivers[k] = v
 		}
 	}
 	if uc.Combined != nil {
 		for k, v := range uc.Combined.Receivers {
-			if _, ok := uc.Metrics.Receivers[k]; ok {
-				return nil, fmt.Errorf("metrics receiver %q has the same name as combined receiver %q", k, k)
+			if uc.Metrics != nil && uc.Metrics.Receivers != nil {
```
quentinmit, line 917: This only needs to be `if uc.Metrics != nil`; subscripting a `nil` map is valid.

confgenerator/config.go
franciscovalentecastro: Removed the `uc.Metrics.Receivers != nil`. Done.

confgenerator/config.go
franciscovalentecastro: Removed the uc.Metrics.Receivers != nil. Done.

confgenerator/confgenerator.go
franciscovalentecastro: Done.

confgenerator/config.go
franciscovalentecastro: Chose to do check the whole config and not consider individual pipelines for the current PR.

IMO, feature tracking about individual pipelines would not be very helpful for our goal (getting statistics of OTel logging supported configs), since it would require several additional metric labels to describe the pipeline in detail ( which receivers, processors or features are unsupported).

I think it would be simpler to `track specific unsupported receivers or processors` (not tied to a specific pipeline) if this data is needed in the future. I consider this additional labels about which specific feature is not supported is out of scope of this initial task / PR and could added later if need.

confgenerator/feature_tracking_test.go
```diff
@@ -593,8 +614,8 @@ func TestBed(t *testing.T) {
 	for _, test := range tests {
 		test := test
 		t.Run(test.Name, func(t *testing.T) {
-			t.Parallel()
-			actual, err := confgenerator.ExtractFeatures(test.Config)
+			//t.Parallel()
```
avilevy18, line 617: Un-comment?

internal/self_metrics/self_metrics.go
```diff
@@ -99,7 +99,7 @@ func InstrumentEnabledReceiversMetric(ctx context.Context, uc *confgenerator.Uni
 }
 
 func InstrumentFeatureTrackingMetric(uc *confgenerator.UnifiedConfig, meter metricapi.Meter) error {
-	features, err := confgenerator.ExtractFeatures(uc)
+	features, err := confgenerator.ExtractFeatures(context.Background(), uc)
```
avilevy18, line 102: I think you should pass the context from `CollectOpsAgentSelfMetrics` as an argument.

confgenerator/feature_tracking_test.go
franciscovalentecastro: Done.

quentinmit (approved):

avilevy18 (approved):

PR #1864, opened 2025-01-22
Ops Agent Plugin - Subagents Startup Logic

XuechunHou:
## Description
This PR enhanced the `Start()` method by adding functionality to start up subagents and the diagnostics service.

## Related issue
[b/380277488](https://b.corp.google.com/issues/380277488)

## How has this been tested?
Unit tests are added.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/builds/ops_agent_plugin.sh
+++ b/builds/ops_agent_plugin.sh
@@ -16,5 +16,5 @@
 set -x -e
 DESTDIR=$1
 mkdir -p "$DESTDIR/opt/google-cloud-ops-agent"
-go build -buildvcs=false -ldflags "-s -w" -o "$DESTDIR/opt/google-cloud-ops-agent/plugin" \
+go build -buildvcs=false -ldflags "-s -w" -o "$DESTDIR/opt/google-cloud-ops-agent/ops-agent-uap-plugin" \
   github.com/GoogleCloudPlatform/ops-agent/cmd/ops_agent_uap_plugin
--- a/builds/ops_agent_plugin.sh
+++ b/builds/ops_agent_plugin.sh
@@ -16,5 +16,5 @@
 set -x -e
 DESTDIR=$1
 mkdir -p "$DESTDIR/opt/google-cloud-ops-agent"
-go build -buildvcs=false -ldflags "-s -w" -o "$DESTDIR/opt/google-cloud-ops-agent/ops-agent-uap-plugin" \
+go build -buildvcs=false -ldflags "-s -w" -o "$DESTDIR/opt/google-cloud-ops-agent/plugin" \
   github.com/GoogleCloudPlatform/ops-agent/cmd/ops_agent_uap_plugin
--- a/builds/ops_agent_plugin.sh
+++ b/builds/ops_agent_plugin.sh
@@ -15,6 +15,6 @@
 
 set -x -e
 DESTDIR=$1
-mkdir -p "$DESTDIR/opt/google-cloud-ops-agent/libexec"
-go build -buildvcs=false -ldflags "-s -w" -o "$DESTDIR/opt/google-cloud-ops-agent/libexec/google_cloud_ops_agent_uap_plugin" \
+mkdir -p "$DESTDIR/opt/google-cloud-ops-agent"
+go build -buildvcs=false -ldflags "-s -w" -o "$DESTDIR/opt/google-cloud-ops-agent/plugin" \
   github.com/GoogleCloudPlatform/ops-agent/cmd/ops_agent_uap_plugin
--- a/cmd/ops_agent_uap_plugin/service_linux.go
+++ b/cmd/ops_agent_uap_plugin/service_linux.go
@@ -28,7 +28,6 @@ import (
 	"strings"
 	"sync"
 	"syscall"
-	"time"
 
 	"google.golang.org/grpc/status"
 
@@ -48,7 +47,7 @@ const (
 	FluentBitRuntimeDirectory   = "run/google-cloud-ops-agent-fluent-bit"
 	OtelRuntimeDirectory        = "run/google-cloud-ops-agent-opentelemetry-collector"
 	DefaultPluginStateDirectory = "/var/lib/google-guest-agent/plugins/ops-agent-plugin"
-	RestartLimit                = 3 // if a command crashes more than `restartLimit` times consecutively, the command will no longer be restarted.
+	RestartLimit                = 10 // if a command crashes more than restartLimit times consecutively, the command will no longer be restarted.
 )
 
 var (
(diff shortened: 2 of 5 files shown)


builds/ops_agent_plugin.sh
```diff
@@ -15,6 +15,6 @@
 
 set -x -e
 DESTDIR=$1
-mkdir -p "$DESTDIR/opt/google-cloud-ops-agent/libexec"
-go build -buildvcs=false -ldflags "-s -w" -o "$DESTDIR/opt/google-cloud-ops-agent/libexec/google_cloud_ops_agent_uap_plugin" \
+mkdir -p "$DESTDIR/opt/google-cloud-ops-agent"
+go build -buildvcs=false -ldflags "-s -w" -o "$DESTDIR/opt/google-cloud-ops-agent/plugin" \
```
jinghan-ma, line 19: Can we call it `uap_plugin`, just cause plugin can be used in too many different contexts.

cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -67,12 +81,24 @@ func (ps *OpsAgentPluginServer) Start(ctx context.Context, msg *pb.StartRequest)
 	pContext, cancel := context.WithCancel(context.Background())
 	ps.cancel = cancel
 
+	pluginInstallaPath, err := os.Executable()
+	if err != nil {
+		log.Printf("Start() failed, because it cannot determine the plugin install location: %s", err)
+		return nil, status.Error(1, err.Error())
+	}
+	pluginInstallaPath, err = filepath.EvalSymlinks(pluginInstallaPath)
```
jinghan-ma, line 89: rename to `pluginInstallPath` without `a` before `Path`

cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -82,21 +108,22 @@ func (ps *OpsAgentPluginServer) Start(ctx context.Context, msg *pb.StartRequest)
 	}
 
 	// Ops Agent config validation
-	if err := validateOpsAgentConfig(pContext, ps.runCommand, pluginStateDir); err != nil {
+	if err := validateOpsAgentConfig(pContext, pluginInstallDir, ps.runCommand); err != nil {
```
jinghan-ma, line 111: Is there a particular reason why we changed to use the Install directory instead of the state directory for the config now?

cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -130,6 +157,103 @@ func (ps *OpsAgentPluginServer) GetStatus(ctx context.Context, msg *pb.GetStatus
...
+		"--parser", path.Join(pluginStateDirectory, FluentBitRuntimeDirectory, "fluent_bit_parser.conf"),
+		"--storage_path", path.Join(pluginStateDirectory, FluentBitStateDiectory, "buffers"),
+	)
+	wg.Add(1)
+	go restartCommand(ctx, cancel, runFluentBitCmd, runCommand, &wg)
+
+	wg.Wait()
+}
+
+func restartCommand(ctx context.Context, cancel context.CancelFunc, cmd *exec.Cmd, runCommand RunCommandFunc, wg *sync.WaitGroup) {
```
jinghan-ma, line 210: Should this still be called `restartCommand` now that we don't actually restart the subagents after a crash?

cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -130,6 +157,103 @@ func (ps *OpsAgentPluginServer) GetStatus(ctx context.Context, msg *pb.GetStatus
...
+	if err != nil {
+		log.Printf("command: %s exited with errors, not restarting.\nCommand output: %s\n Command error:%s", cmd.Args, string(output), err)
+		cancel() // cancels the parent context which also stops other Ops Agent sub-binaries from running.
+		return
+	} else {
+		log.Printf("command: %s %s exited successfully.\nCommand output: %s", cmd.Path, cmd.Args, string(output))
+	}
+
+	wg.Add(1)
+	go restartCommand(ctx, cancel, cmd, runCommand, wg)
```
jinghan-ma, line 236: Why do we make a recursive call here? What's the scenario where we need this?

This line will only be executed if `runCommand` exited without errors. The only scenario I can think of for that is when the subagents exit due to an exit signal.

cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -130,6 +145,113 @@ func (ps *OpsAgentPluginServer) GetStatus(ctx context.Context, msg *pb.GetStatus
...
+	wg.Add(1)
+	go restartCommand(ctx, cancel, runFluentBitCmd, runCommand, RestartLimit, RestartLimit, &wg)
+
+	wg.Wait()
+}
+
+func restartCommand(ctx context.Context, cancel context.CancelFunc, cmd *exec.Cmd, runCommand RunCommandFunc, remainingRetry int, totalRetry int, wg *sync.WaitGroup) {
+	defer wg.Done()
+	if cmd == nil {
+		return
```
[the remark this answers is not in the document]
XuechunHou: It should technically never be nil, but this is a sanity check in case it somehow is.

cmd/ops_agent_uap_plugin/service_linux.go
XuechunHou: So the plugin install directory is different from the plugin State directory. We will have to run the conf generator binary to verify the provided Ops Agent config. The conf generator locates under the `pluginInstallDir` not the `pluginStateDir`.

cmd/ops_agent_uap_plugin/service_linux.go
XuechunHou: The command will be restarted when the previous execution exited without errors. 

I didn't think about when a command would exit without errors. This is just to catch that when it happens, and also trying to mimic the behaviour of `Restart=always` in systemd services.

builds/ops_agent_plugin.sh
XuechunHou: Done

cmd/ops_agent_uap_plugin/service_linux.go
XuechunHou: Done

builds/ops_agent_plugin.sh
XuechunHou: Actually https://source.corp.google.com/piper///depot/google3/cloud/cluster/guestplatform/demo_plugin/client.go;bpv=1;bpt=1;rcl=720287686;l=129 no, UAP requires that the plugin entrypoint be named 'plugin'.

cmd/ops_agent_uap_plugin/service_linux.go
jinghan-ma: I think it's better if we don't restart if the subagent exited without errors, because the fact that the subagent exited (aside from during shutdown) is itself an error, so not restarting (resulting in giving an error to UAP in GetStatus) can help us catch it.

Also, since we are delegating the entire logic of crash recovery and restarting to UAP, I think we no longer need to mimic systemd's behaviour.

cmd/ops_agent_uap_plugin/service_linux.go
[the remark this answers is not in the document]
jinghan-ma: Since we never expect it to be nil, should we raise an error if cmd is nil?

jinghan-ma (approved):

PR #1870, opened 2025-01-29
Integration Testing Ops Agent UAP Plugin

XuechunHou:
## Description
1. This PR updated the existing Ops agent integration tests and third party tests, making them compatible with the Ops Agent plugin. 
2. This PR also added a few integration tests for the Ops Agent Plugin, exclusively testing its `GetStatus` method. 

## Related issue
b/381246351

## How has this been tested?
Ops Agent integration tests should run twice per OS distro and architecture, once for the regular ops agent, once for the plugin. Both runs should pass. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/agents/agents.go
```diff
@@ -186,27 +189,53 @@ func RunOpsAgentDiagnostics(ctx context.Context, logger *logging.DirectoryLogger
...
+		} {
+			_, basename := path.Split(log)
+			gce.RunRemotely(ctx, logger.ToFile(basename+txtSuffix), vm, "sudo cat "+log)
+		}
 	}
 }
 
 func runOpsAgentDiagnosticsWindows(ctx context.Context, logger *logging.DirectoryLogger, vm *gce.VM) {
+	isUAPPlugin := IsOpsAgentUAPPlugin()
+	if isUAPPlugin {
```
rafaelwestphal, line 236: We can inline the IsOpsAgentUAPPlugin() here:
if IsOpsAgentUAPPlugin() {
return
}

integration_test/agents/agents.go
```diff
@@ -664,9 +693,16 @@ func InstallStandaloneWindowsMonitoringAgent(ctx context.Context, logger *log.Lo
 }
 
 func getRestartOpsAgentCmd(imageSpec string) string {
-	if gce.IsWindows(imageSpec) {
+	isUAPPlugin := IsOpsAgentUAPPlugin()
```
rafaelwestphal, line 696: I think the logic it will be easier to understand if we separated uap from the rest
if isUAPplugin {
    if windows { return "" }
    return grpcul ...
}

integration_test/agents/agents.go
```diff
@@ -786,23 +825,43 @@ func SetupOpsAgent(ctx context.Context, logger *log.Logger, vm *gce.VM, config s
...
+	}
+	return fmt.Sprintf("sudo nohup ~/plugin --address=localhost:%s --errorlogfile=errorlog.txt --protocol=tcp 1>/dev/null 2>/dev/null &", port)
+}
+
+// StartOpsAgentPlugin starts the Ops Agent Plugin gRPC server on the testing VM in the background.
+func StartOpsAgentPlugin(ctx context.Context, logger *log.Logger, vm *gce.VM, port string) error {
+	if _, err := gce.RunRemotely(ctx, logger, vm, getStartOpsAgentPluginCmd(vm.ImageSpec, port)); err != nil {
+		return fmt.Errorf("StartOpsAgentPlugin() failed to start the ops agent plugin: %v", err)
+	}
+	// Give agents time to shut down. Fluent-Bit's default shutdown grace period
```
rafaelwestphal, line 851: I don't understand this comment. Is this function starting the  agent? why do we wait for Fluent-bit shutdown

integration_test/agents/agents.go
```diff
@@ -874,6 +935,43 @@ func CommonSetupWithExtraCreateArgumentsAndMetadata(t *testing.T, imageSpec stri
...
+		return err
+	}
+	if err := gce.InstallGrpcurlIfNeeded(ctx, logger, vm); err != nil {
+		return err
+	}
+
+	if _, err := gce.RunRemotely(ctx, logger, vm, "sudo gsutil cp -r "+gcsPath+"/* /tmp/agentUpload"); err != nil {
+		return fmt.Errorf("error copying down agent package from GCS: %v", err)
+	}
+	if _, err := gce.RunRemotely(ctx, logger, vm, "mv /tmp/agentUpload/*.tar.gz /tmp/agentPlugin || echo nothing to move"); err != nil {
```
rafaelwestphal, line 958: We need to error here if there is no UAP plugin .tar.gz at this point. The || echo is preventing the command for erroring out.

integration_test/agents/agents.go
```diff
@@ -874,6 +935,43 @@ func CommonSetupWithExtraCreateArgumentsAndMetadata(t *testing.T, imageSpec stri
...
+		return err
+	}
+	// Print the contents of /tmp/agentPlugin into the logs.
+	if _, err := gce.RunRemotely(ctx, logger, vm, "ls -la /tmp/agentPlugin"); err != nil {
+		return err
+	}
+	if _, err := gce.RunRemotely(ctx, logger, vm, "sudo find /tmp/agentPlugin -maxdepth 1 -name \"google-cloud-ops-agent-plugin*.tar.gz\" -print0 | xargs -0 -I {} sudo tar -xzf {} --no-overwrite-dir -C ~/ && ls -la"); err != nil {
```
rafaelwestphal, line 965: I think doing tar like this will override the folder structure for the home directory and mess up the permissions

integration_test/agents/agents.go
```diff
@@ -813,9 +872,11 @@ func SetupOpsAgentFrom(ctx context.Context, logger *log.Logger, vm *gce.VM, conf
 		if err := gce.UploadContent(ctx, logger, vm, strings.NewReader(config), util.GetConfigPath(vm.ImageSpec)); err != nil {
 			return fmt.Errorf("SetupOpsAgentFrom() failed to upload config file: %v", err)
 		}
-		if err := restartOpsAgent(ctx, logger, vm); err != nil {
+		if err := RestartOpsAgent(ctx, logger, vm); err != nil {
 			return err
 		}
+	} else if isUAPPluin {
```
rafaelwestphal, line 878: I think this is better at the top of the if clause to prevent cases where ethe config > 0 and we are using UAP

integration_test/agents/agents.go
```diff
@@ -932,3 +1030,27 @@ func installWindowsPackageFromGCS(ctx context.Context, logger *log.Logger, vm *g
 	}
 	return nil
 }
+
+func GetOtelConfigPath(imageSpec string) string {
+	isUAPPlugin := IsOpsAgentUAPPlugin()
```
rafaelwestphal, line 1035: I think we can improve the if structure here by nesting UAP cases together.
if UAP {
...
}
<original code>

integration_test/gce/gce_testing.go
```diff
@@ -1637,6 +1637,45 @@ func RestartInstance(ctx context.Context, logger *log.Logger, vm *VM) error {
...
+
+	arch := "x86_64"
+	if IsARM(vm.ImageSpec) {
+		arch = "arm64"
+	}
+
+	switch arch {
+	case "x86_64":
+		installCmd = `
+sudo curl -sSL "https://github.com/fullstorydev/grpcurl/releases/download/v1.8.6/grpcurl_1.8.6_linux_x86_64.tar.gz" | sudo tar -xz -C /usr/local/bin
```
rafaelwestphal, line 1663: Let's refactor the command to a common string  and just replace the arch.

integration_test/ops_agent_test/main_test.go
```diff
@@ -96,17 +95,39 @@ func workDirForImage(imageSpec string) string {
...
+		return "grpcurl -plaintext -d '{}' localhost:1234 plugin_comm.GuestAgentPlugin/Start"
+	}
+
 	// Return a command that works for both < 2.0.0 and >= 2.0.0 agents.
 	return "sudo service google-cloud-ops-agent start || sudo systemctl start google-cloud-ops-agent"
 }
 
 func stopCommandForImage(imageSpec string) string {
-	if gce.IsWindows(imageSpec) {
+	isUAPPlugin := agents.IsOpsAgentUAPPlugin()
```
rafaelwestphal, line 117: I think we can improve the if structure here by nesting UAP cases together.
if UAP {
...
}

integration_test/ops_agent_test/main_test.go
```diff
@@ -96,17 +95,39 @@ func workDirForImage(imageSpec string) string {
 }
 
 func startCommandForImage(imageSpec string) string {
-	if gce.IsWindows(imageSpec) {
+	isUAPPlugin := agents.IsOpsAgentUAPPlugin()
+
```
rafaelwestphal, line 99: I think we can improve the if structure here by nesting UAP cases together.
if UAP {
...
}

integration_test/agents/agents.go
```diff
@@ -186,27 +189,53 @@ func RunOpsAgentDiagnostics(ctx context.Context, logger *logging.DirectoryLogger
...
+	if isUAPPlugin {
+		gce.RunRemotely(ctx, logger.ToFile("systemctl_status_for_ops_agent.txt"), vm, fmt.Sprintf("grpcurl -plaintext -d '{}' localhost:%s plugin_comm.GuestAgentPlugin/GetStatus", OpsAgentPluginServerPort))
+	} else {
+		gce.RunRemotely(ctx, logger.ToFile("systemctl_status_for_ops_agent.txt"), vm, "sudo systemctl status google-cloud-ops-agent*")
+	}
 
 	gce.RunRemotely(ctx, logger.ToFile("journalctl_output.txt"), vm, "sudo journalctl -xe")
+	if isUAPPlugin {
+		for _, log := range []string{
+			gce.SyslogLocation(vm.ImageSpec),
```
rafaelwestphal, line 202: Can we refactor this code avoid the duplication of the files list? It will make it easier in the feature when we need to remove/add/replace anything.

integration_test/third_party_apps_test/applications/aerospike/enable
```diff
@@ -2,6 +2,14 @@
 
 set -e
 
+# Create a back up of the existing file so existing configurations are not lost.
```
rafaelwestphal, line 5: Remove all the places where this comment was copied since we don't do that here.

integration_test/third_party_apps_test/applications/apache/enable
```diff
@@ -30,5 +38,5 @@ logging:
           - apache_error
 EOF
 
-sudo service google-cloud-ops-agent restart
-sleep 60
+
```
rafaelwestphal, line 41: nit: Can we remove the extra lines and on the other files that they got left.

integration_test/agents/agents.go
rafaelwestphal: Nvm me! -no-overwrite-dir prevents that

integration_test/third_party_apps_test/applications/apache/enable
XuechunHou: Done

integration_test/ops_agent_test/main_test.go
XuechunHou: Done

integration_test/agents/agents.go
XuechunHou: you are right, this function only starts the agent, removed the 10 seconds wait.

integration_test/agents/agents.go
XuechunHou: Thank you for the comments. You're correct, they are the same set of files. However, they have different paths, and I haven't yet found a good way to generalize them. Even within the Ops Agent plugin itself, not all of these files are located in the same directories.

Genmini suggested something like the following: 

```
filePaths := map[string]string{
                "health-checks.log":             "/var/log/google-cloud-ops-agent/health-checks.log",
                "config.yaml":                    "/etc/google-cloud-ops-agent/config.yaml",
                "subagents/logging-module.log":   "/var/log/google-cloud-ops-agent/subagents/logging-module.log",
                "subagents/metrics-module.log":   "/var/log/google-cloud-ops-agent/subagents/metrics-module.log",
                "nvidia-installer.log":           "/var/log/nvidia-installer.log",
                "fluent_bit_main.conf":          "/run/google-cloud-ops-agent-fluent-bit/fluent_bit_main.conf",
                "fluent_bit_parser.conf":        "/run/google-cloud-ops-agent-fluent-bit/fluent_bit_parser.conf",
                "otel.yaml":                      "/run/google-cloud-ops-agent-opentelemetry-collector/otel.yaml",
        }

        if isUAPPlugin {
                for file, filePath := range filePaths {
                        filePaths[file] = path.Join("/var/lib/google-guest-agent/agent_state/plugins/ops-agent-plugin", filePath)
                        // Adjust for run and log directories.
                        if file == "fluent_bit_main.conf" || file == "fluent_bit_parser.conf" || file == "otel.yaml" {
                                filePaths[file] = path.Join("/var/lib/google-guest-agent/agent_state/plugins/ops-agent-plugin/run", path.Base(filePath))
                        } else if file == "health-checks.log" || file == "subagents/logging-module.log" || file == "subagents/metrics-module.log" || file == "nvidia-installer.log" {
                                filePaths[file] = path.Join("/var/lib/google-guest-agent/agent_state/plugins/ops-agent-plugin/log", path.Base(filePath))
                        } else if file == "config.yaml" {
                                filePaths[file] = path.Join("/var/lib/google-guest-agent/agent_state/plugins/ops-agent-plugin", path.Base(filePath))
                        }
                }
        }
```

However, that adds too many if-else statements, and it also reduces the readability of the file paths for the plugin.

integration_test/agents/agents.go
XuechunHou: Emm I think my logic here is right. 

regular ops agent is started once installed. However, ops agent plugin needs to be explicitly started after install.


when `len(config) >0`, both the regular and the plugin Ops Agent need a (re)start.  A `restart` is needed for the regular agent to apply the config. A `start` is needed for the plugin agent since it has never been started yet. 


When `len(config) == 0` only the plugin Ops Agent need to be started.

integration_test/agents/agents.go
```diff
@@ -813,9 +868,11 @@ func SetupOpsAgentFrom(ctx context.Context, logger *log.Logger, vm *gce.VM, conf
 		if err := gce.UploadContent(ctx, logger, vm, strings.NewReader(config), util.GetConfigPath(vm.ImageSpec)); err != nil {
 			return fmt.Errorf("SetupOpsAgentFrom() failed to upload config file: %v", err)
 		}
-		if err := restartOpsAgent(ctx, logger, vm); err != nil {
+		if err := RestartOpsAgent(ctx, logger, vm); err != nil {
 			return err
 		}
+	} else if IsOpsAgentUAPPlugin() {
+		return RestartOpsAgent(ctx, logger, vm)
```
jinghan-ma, line 875: Why do we need to restart the ops agent plugin here specifically?

integration_test/gce/gce_testing.go
```diff
@@ -1637,6 +1637,34 @@ func RestartInstance(ctx context.Context, logger *log.Logger, vm *VM) error {
...
+		return nil
+	}
+
+	logger.Printf("grpcurl not found, installing it...")
+
+	arch := "x86_64"
+	if IsARM(vm.ImageSpec) {
+		arch = "arm64"
+	}
+	if arch != "x86_64" && arch != "arm64" {
```
jinghan-ma, line 1657: Given how we define `arch` above, will we ever go into this if statement?

integration_test/ops_agent_test/main_test.go
```diff
@@ -4427,6 +4553,13 @@ func checkExpectedHealthCheckResult(t *testing.T, output string, name string, ex
 }
 
 func getRecentServiceOutputForImage(imageSpec string) string {
+	if agents.IsOpsAgentUAPPlugin() {
+		if gce.IsWindows(imageSpec) {
+			return ""
+		}
+		return "grpcurl -plaintext -d '{}' localhost:1234 plugin_comm.GuestAgentPlugin/GetStatus"
+	}
```
jinghan-ma, line 4561: I wonder if it makes sense for us to decouple the UAP status call from `ServiceOutput`. They currently are somewhat similar, but we expect them to evolve in very different ways -- so the way we test for it will also start to deviate.

I feel like it makes it clearer if we remove the abstraction and explicitly call something like `GetPluginStatus` in many places, so that it is clear to the reader that __only UAP has this behaviour__. Examples of this includes the current check for an Unhealthy status when plugin is stopped, or future tests such as checking for error codes (which is exclusive to UAP GetStatus).

integration_test/ops_agent_test/main_test.go
```diff
@@ -4491,9 +4632,19 @@ func TestPortsAndAPIHealthChecks(t *testing.T) {
...
+			cmdOut, err = gce.RunRemotely(ctx, logger, vm, getHealthCheckLogsForUAPPluginByImage(vm.ImageSpec))
+			if err != nil {
+				t.Fatal(err)
+			}
+
+		} else {
+			cmdOut, err = gce.RunRemotely(ctx, logger, vm, getRecentServiceOutputForImage(vm.ImageSpec))
+			if err != nil {
+				t.Fatal(err)
+			}
```
jinghan-ma, line 4647: This block seem to be repeated for each of the health check tests. Let's abstract this by introducing a new function `getHealthCheckResultsForImage` that calls one of `getHealthCheckLogsForUAPPluginByImage` or `getRecentServiceOutputForImage`.

integration_test/ops_agent_test/main_test.go
jinghan-ma: I think the root of the problem is that there's a layer of abstraction missing here. 
Currently, `getRecentServiceOutputForImage` is actually overloaded to perform two jobs:
1. Get service status (i.e. check if ops agent service is running), and
2. Get health check results.

These two happen to come from the same place when the Ops Agent is running as a service. So we can just call `getRecentServiceOutputForImage` and check both.

But the plugin version of the Ops Agent separates these two (more specifically, health checks is directly logged to a file). So it doesn't make sense having such a function for the UAP version.

I think we can make a quick change to fix this: Introduce 2 new functions. `getServiceStatusForImage` and `getHealthCheckResultsForImage` (this one is also mentioned in the comment below).

For regular Ops Agent, these functions will still call `getRecentServiceOutputForImage`. For UAP, it'll call the appropriate UAP equivalent.

Then, for each of the existing calls to `getRecentServiceOutputForImage`, replace it with one of the new functions above based on what exactly we are checking.

integration_test/third_party_apps_test/main_test.go
```diff
@@ -595,6 +595,11 @@ func runSingleTest(ctx context.Context, logger *logging.DirectoryLogger, vm *gce
 		return nonRetryable, fmt.Errorf("error enabling %s: %v", app, err)
 	}
 
+	if err := agents.RestartOpsAgent(ctx, logger.ToMainLog(), vm); err != nil {
+		return nonRetryable, fmt.Errorf("error restarting agent: %v", err)
+	}
+	time.Sleep(60 * time.Second)
+
```
jinghan-ma, line 602: Why do we need to restart the ops agent here? Also, do we need to sleep for a whole minute?

jinghan-ma (changes requested):

integration_test/third_party_apps_test/main_test.go
XuechunHou: If you look at the `if` block right above this one: 

```if _, err = runScriptFromScriptsDir(ctx, logger.ToMainLog(), vm, path.Join("applications", app, "enable"), nil); err != nil {
		return nonRetryable, fmt.Errorf("error enabling %s: %v", app, err)
	}```


that block executes what's inside the `enable` script for each third party app. The enable script populates `config.yaml` with some custom config. That's why Ops agent needs to be restarted in order to apply the config.


https://github.com/GoogleCloudPlatform/ops-agent/pull/1870/files#diff-db0ec4d7bd3cd0a566b078eb57aa07cee85ca6c3ee560bd68206d4d471db9211 if you look at the changes I made to the `enable` scripts, you will see that the `sleep` statements are removed from the script and added here in `main_test.go`. The sleep seconds range from 0 to 60 seconds, so I decided to sleep for 60 seconds in `main_test.go` universally.

File: integration_test/agents/agents.go [hunk 19]
```diff
@@ -664,6 +692,13 @@ func InstallStandaloneWindowsMonitoringAgent(ctx context.Context, logger *log.Lo
 }
 
 func getRestartOpsAgentCmd(imageSpec string) string {
+	if IsOpsAgentUAPPlugin() {
+		if gce.IsWindows(imageSpec) {
+			return ""
+		}
+		return fmt.Sprintf("grpcurl -plaintext -d '{}' localhost:%s plugin_comm.GuestAgentPlugin/Stop && grpcurl -plaintext -d '{}' localhost:%s plugin_comm.GuestAgentPlugin/Start", OpsAgentPluginServerPort, OpsAgentPluginServerPort)
```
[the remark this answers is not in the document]
XuechunHou: The test is not interacting with UAP at all, but we've asked this question to the UAP team and the answer is no. 

"Or will it start the plugin on an entirely different process?" process or port? I think both are possible.

integration_test/agents/agents.go
XuechunHou: Not like the regular ops agent, the agent plugin is not `started` automatically after installation. It needs to be explicitly started. 

The point of calling`RestartOpsAgent` function is to trigger a `start` of the agent. 

When len(config)> 0, the regular agent needs to be re-started, and the plugin agent needs to be started to apply the custom config. 

When len(config) ==0, only the plugin agent needs to be started.

integration_test/gce/gce_testing.go
XuechunHou: You are right

integration_test/ops_agent_test/main_test.go
XuechunHou: Done

integration_test/agents/agents.go
```diff
@@ -186,27 +189,52 @@ func RunOpsAgentDiagnostics(ctx context.Context, logger *logging.DirectoryLogger
 	gce.RunRemotely(metricsCtx, logger.ToFile("fluent_bit_metrics.txt"), vm, "sudo curl -s localhost:20202/metrics")
 	gce.RunRemotely(metricsCtx, logger.ToFile("otel_metrics.txt"), vm, "sudo curl -s localhost:20201/metrics")
 
-	gce.RunRemotely(ctx, logger.ToFile("systemctl_status_for_ops_agent.txt"), vm, "sudo systemctl status google-cloud-ops-agent*")
+	isUAPPlugin := IsOpsAgentUAPPlugin()
+	if isUAPPlugin {
+		gce.RunRemotely(ctx, logger.ToFile("systemctl_status_for_ops_agent.txt"), vm, fmt.Sprintf("grpcurl -plaintext -d '{}' localhost:%s plugin_comm.GuestAgentPlugin/GetStatus", OpsAgentPluginServerPort))
```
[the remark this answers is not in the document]
XuechunHou: updated.

integration_test/agents/agents.go
XuechunHou: Done

integration_test/agents/agents.go
```diff
@@ -874,6 +931,43 @@ func CommonSetupWithExtraCreateArgumentsAndMetadata(t *testing.T, imageSpec stri
...
+	if err := gce.InstallGrpcurlIfNeeded(ctx, logger, vm); err != nil {
+		return err
+	}
+
+	if _, err := gce.RunRemotely(ctx, logger, vm, "sudo gsutil cp -r "+gcsPath+"/* /tmp/agentUpload"); err != nil {
+		return fmt.Errorf("error copying down agent package from GCS: %v", err)
+	}
+	if _, err := gce.RunRemotely(ctx, logger, vm, "mv /tmp/agentUpload/*.tar.gz /tmp/agentPlugin"); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
XuechunHou: Done

integration_test/ops_agent_test/main_test.go
XuechunHou: https://github.com/GoogleCloudPlatform/ops-agent/pull/1870/commits/6078a3e97213d88be317fdef526984e3f8c883e2 Update PTAL.

XuechunHou: `getRecentServiceOutputForImage` does different things for each kind of the Ops Agent:

For the regular agent: it retrieves health check results from `systemctl status`
For the agent plugin: it retrieves the response of the `GetStatus` call. 


In my commit https://github.com/GoogleCloudPlatform/ops-agent/commit/6078a3e97213d88be317fdef526984e3f8c883e2, a new function `getUAPPluginStatusForImage ` is added to run exclusively for the agent plugin to retrieve the gRPC response of the `GetStatus` call. 

`getRecentServiceOutputForImage`  is simplified to work only for the regular ops agent. 

WDYT? Thanks

integration_test/ops_agent_test/main_test.go
```diff
@@ -756,6 +769,87 @@ func TestCustomLogFile(t *testing.T) {
...
+		if !strings.Contains(cmdOut.Stdout, "The Ops Agent Plugin is running ok.") {
+			t.Error("expected the plugin to report that the Ops Agent is running")
+		}
+
+		_, processName, err := fetchPIDAndProcessName(ctx, logger, vm, []string{"fluent-bit"})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		logger.Printf("TestPluginGetStatusReturnsUnhealthyStatusOnSubAgentTermination: Found %s", processName)
```
rafaelwestphal, line 826: We can remove this print.

integration_test/ops_agent_test/main_test.go
```diff
@@ -4822,29 +4974,59 @@ func TestRestartVM(t *testing.T) {
...
+			checkExpectedHealthCheckResult(t, cmdOut.Stdout, "Network", "PASS", "")
+			checkExpectedHealthCheckResult(t, cmdOut.Stdout, "Ports", "PASS", "")
+			checkExpectedHealthCheckResult(t, cmdOut.Stdout, "API", "PASS", "")
+		}
 
 		logger.Printf(`Restarting instance. For details, see "VM_restart.txt".`)
 		if err := gce.RestartInstance(ctx, dirLog.ToFile("VM_restart.txt"), vm); err != nil {
 			t.Fatal(err)
 		}
+		if isUAPPlugin {
```
rafaelwestphal, line 5003: nit: Can we bundle this 2 ifs?

rafaelwestphal (approved):

integration_test/agents/agents.go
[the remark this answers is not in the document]
jinghan-ma: I think it might be better to mimic UAP's operations more closely in our own tests by killing the wrapper and starting a new one?

Not a big issue either way.

integration_test/agents/agents.go
jinghan-ma: In that case, is it possible to be more explicit here and call `StartOpsAgentPlugin` instead?

integration_test/ops_agent_test/main_test.go
```diff
@@ -4822,29 +4972,57 @@ func TestRestartVM(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		cmdOut, err := gce.RunRemotely(ctx, logger, vm, getRecentServiceOutputForImage(vm.ImageSpec))
-		if err != nil {
-			t.Fatal(err)
-		}
+		isUAPPlugin := agents.IsOpsAgentUAPPlugin()
+		if isUAPPlugin {
```
jinghan-ma, line 4976: Would it make more sense to skip this test if we're in UAP world? Different from the systemd version, we don't actually expect the UAP plugin to be started automatically after a restart, so the behaviour this test checks for doesn't really apply. Maybe it's less confusing if we explicitly skip this test.

jinghan-ma (approved):

integration_test/ops_agent_test/main_test.go
XuechunHou: Emm somehow I cant reply to your comment above. Let me reply here.
"In that case, is it possible to be more explicit here and call StartOpsAgentPlugin instead?"
- `StartOpsAgentPlugin` and `RestartOpsAgent` do different things for the agent plugin; `StartOpsAgentPlugin` starts the plugin grpc server itself. However, `RestartOpsAgent` sends a gRPC request to the plugin's `Start()` endpoint, to start the Ops Agent. 
In this case the plugin gRPC server is live already, I only need to start the Ops Agent. 





I am unsure what the RestartVM test is actually testing. If it's verifying that the agent installation always works even after a VM restart, then it makes sense to test for the plugin as well.

integration_test/agents/agents.go
XuechunHou: `StartOpsAgentPlugin` and `RestartOpsAgent` do different things for the agent plugin; `StartOpsAgentPlugin` starts the plugin grpc server itself. However, `RestartOpsAgent` sends a gRPC request to the plugin's `Start()` endpoint, to start the Ops Agent. 
In this case the plugin gRPC server is live already, I only need to start the Ops Agent.

integration_test/agents/agents.go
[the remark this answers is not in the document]
XuechunHou: re-starting the plugin and re-starting the Ops Agent through the UAP plugin is different. Here I am explicitly re-starting the Ops Agent itself. In the UAP world, this is done by sending a `Stop` and then a `Start` gRPC request to the plugin gRPC server.

PR #1872, opened 2025-02-03
confgenerator : Add otel logging support for `systemd_journald` receiver.

franciscovalentecastro:
## Description
Add otel logging support for `systemd_journald` receiver.

## Related issue
b/394593541

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/logging_receivers.go
```diff
@@ -682,6 +682,61 @@ func (r LoggingReceiverSystemd) Components(ctx context.Context, tag string) []fl
 	return input
 }
 
+func (r LoggingReceiverSystemd) Pipelines(ctx context.Context) ([]otel.ReceiverPipeline, error) {
+	operators := []map[string]any{}
+	operators = append(operators, map[string]any{
```
quentinmit, line 687: I think we should use `ModifyFields` here instead of using the deprecated `operators` option to do these manipulations.

integration_test/ops_agent_test/main_test.go
```diff
@@ -2261,6 +2261,16 @@ func TestWindowsEventLogWithNonDefaultTimeZone(t *testing.T) {
...
+	t.Parallel()
+	t.Run("fluent-bit", func(t *testing.T) {
+		testSystemdLog(t, false)
+	})
+	t.Run("otel", func(t *testing.T) {
+		testSystemdLog(t, true)
+	})
+}
+
+func testSystemdLog(t *testing.T, otel bool) {
```
quentinmit, line 2273: I'd like to see this actually test the schema of the log entries, instead of just testing that one exists.

Probably that means we should send a particular canary message to journald and then check how it is represented in Cloud Logging.

confgenerator/logging_receivers.go
franciscovalentecastro: Yes, using `ModifyFields` is better. I was having a hard time trying to use the `operators` and with `ModifyFields` it was a lot simpler. Done.

confgenerator/logging_receivers.go
```diff
@@ -682,6 +682,63 @@ func (r LoggingReceiverSystemd) Components(ctx context.Context, tag string) []fl
...
+	receiver_config := map[string]any{
+		"start_at": "beginning",
+		"priority": "debug",
+	}
+
+	modify_fields_processors, err := LoggingProcessorModifyFields{
+		Fields: map[string]*ModifyField{
+			`severity`: {
+				CopyFrom: "jsonPayload.PRIORITY",
+				// This mapping sets SeverityText in otel logs data model which naming differs from Cloud Logging model.
```
quentinmit, line 695: We should be accepting the Cloud Logging values in severityText. Users can be writing this field directly, either with their own modify_fields or directly with JSON, and they can filter on those values. I think we should try to get this fixed in the exporter.

Note that the OTel spec allows (and encourages) arbitrary strings as severityText: https://opentelemetry.io/docs/specs/otel/logs/data-model/#severity-fields

For now I'd put this back to LogSeverity, change the test to only use severities where the name matches, and then file a bug to add support for the rest of the severities in the exporter.

quentinmit (approved):

confgenerator/logging_receivers.go
franciscovalentecastro: Done. I created this b/400435104 and i'll start working on updating the `googlecloudexporter`.

PR #1874, opened 2025-02-03
Fixed a Race Condition in the UAP Plugin Implementation

XuechunHou:
## Description
This PR identifies and fixes a race condition in the Ops Agent UAP plugin implementation. The setting (read) and resetting (write) of the `cancel` function in the `Start()` and `Stop()` methods should be atomic and protected by a mutex, which has been added in this PR.

## Related issue
b/380277488
## How has this been tested?
`go test -mod=mod -coverpkg="./..." -coverprofile=covprofile ./...` command triggered in the presubmit flakes due to this race condition.

## Checklist:
- Unit tests
- Integration tests
  - [] Integration tests have been added/modified and passed for this PR.
- Documentation
- Minor version bump

### DIFF
--- a/cmd/ops_agent_uap_plugin/service_linux.go
+++ b/cmd/ops_agent_uap_plugin/service_linux.go
@@ -72,14 +72,17 @@ func (ps *OpsAgentPluginServer) Apply(ctx context.Context, msg *pb.ApplyRequest)
 // Until plugin receives Start request plugin is expected to be not functioning
 // and just listening on the address handed off waiting for the request.
 func (ps *OpsAgentPluginServer) Start(ctx context.Context, msg *pb.StartRequest) (*pb.StartResponse, error) {
+	ps.mu.Lock()
 	if ps.cancel != nil {
 		log.Printf("The Ops Agent plugin is started already, skipping the current request")
+		ps.mu.Unlock()
 		return &pb.StartResponse{}, nil
 	}
 	log.Printf("Received a Start request: %s. Starting the Ops Agent", msg)
 
 	pContext, cancel := context.WithCancel(context.Background())
 	ps.cancel = cancel
+	ps.mu.Unlock()
 
 	pluginInstallPath, err := os.Executable()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


rafaelwestphal (approved):
Thanks for looking into this!

cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -101,24 +104,21 @@ func (ps *OpsAgentPluginServer) Start(ctx context.Context, msg *pb.StartRequest)
 	// Find existing ops agent installation, and conflicting legacy agent installation.
 	foundConflictingInstallations, err := findPreExistentAgents(pContext, ps.runCommand, AgentSystemdServiceNames)
 	if foundConflictingInstallations || err != nil {
-		ps.cancel()
-		ps.cancel = nil
+		ps.Stop(ctx, &pb.StopRequest{Cleanup: false})
```
franciscovalentecastro, line 107: Why is `ps.Stop()` substituting `ps.cancel()` and `ps.cancel = nil` ?

cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -72,14 +72,17 @@ func (ps *OpsAgentPluginServer) Apply(ctx context.Context, msg *pb.ApplyRequest)
 // Until plugin receives Start request plugin is expected to be not functioning
 // and just listening on the address handed off waiting for the request.
 func (ps *OpsAgentPluginServer) Start(ctx context.Context, msg *pb.StartRequest) (*pb.StartResponse, error) {
+	ps.mu.Lock()
 	if ps.cancel != nil {
 		log.Printf("The Ops Agent plugin is started already, skipping the current request")
+		ps.mu.Unlock()
```
franciscovalentecastro, line 78: Why to use `defer ps.mu.Unlock()` the same as in `Stop` ?

XuechunHou: Because the mutex unlock on L78 should not wait until the `Start()` method is about to return. `Start()` is a long function, unlock `Stop()` and `GetStatus()`

The only thing `Stop()` and `GetStatus()` do is to access the field `cancel`, that's not the case with `Start()`. `Start` does additional stuff like: run config generator, start up subagents etc.

cmd/ops_agent_uap_plugin/service_linux.go
XuechunHou: Because `ps.Stop()` literally does the same thing. Since a lock must be obtained before accessing the cancel function, extra lock and unlock statements would need to be added everywhere that calls `ps.cancel() and ps.cancel = nil`. Because `ps.Stop()` achieves the same result, I decided to replace the calls to `ps.cancel() and ps.cancel = nil` with a call to `ps.Stop()`.

franciscovalentecastro (approved):
LGTM! Thank you for all the explanations!

PR #1879, opened 2025-02-10
[Draft] confgenerator : Improve detection of unsupported OTel Logging Receivers and Processors.

franciscovalentecastro:
## Description
A proposed solution to improve detection of unsupported OTel Logging Receivers and Processors. 

### Explanation
Currently the detection of supported `Receivers` and `Processors` relies on [determining if an instantiated struct implements](https://github.com/GoogleCloudPlatform/ops-agent/blob/5770e42ca0ae90a2d9f8084dcd0e668084ecc478/confgenerator/confgenerator.go#L204) the [OTelReceiver](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/config.go#L582) or [OTelProcessor](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/config.go#L775) interfaces. This is based on the assumption that OTel Logging unsupported `Receivers` and `Processors` don't implement the `Pipelines` and `Processors` methods. 

This assumption fails, for example, when a Receiver (or Processor) has a (nested) `embedded struct` which does implement a method (e.g. Pipeline), but it is not directly implemented in the "parent" struct. See [LoggingReceiverFlink](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/apps/flink.go#L144C1-L147C2), in this example the `flink` receiver relies on a `parse_regex` processor which is unsupported, but our current detection logic can't determine this because the embedded `files` receiver does implement the interface.

One solution could be to use the `reflect` golang package and "determine" if the implemented `Pipeline` method is `promoted` from the embedded struct or directly implemented in the (parent) struct (e.g. comparing method pointers). This could be very fragile and does seem unintended comparing with the Go language definitions ( https://github.com/golang/go/issues/21162 ).

### Proposed Solution
Add a the missing `Pipelines` and `Processors` methods to all Logging Receiver and Processors with a clear `error` message when a Receiver or Processor is not supported. This indeed breaks the previous more elegant solution/approach of detection of support by checking missing methods (does or doesn't implement interface) into a fully descriptive approach (every method determines if its supported).

### Details
- Add `Pipelines` and `Processors` methods to all Logging Receiver and Processors. Return error when not supported by OTel Logging.
- Detect otel logging support from Feature Tracking using the Merged config instead of the User config. This was causing issues in the detection when detecting unsupported processors in the `default_pipeline`.

## Related issue
b/390671495

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/apps/active_directory_ds.go
+++ b/apps/active_directory_ds.go
@@ -62,6 +62,10 @@ func (r LoggingReceiverActiveDirectoryDS) Type() string {
 	return "active_directory_ds"
 }
 
+func (r LoggingReceiverActiveDirectoryDS) Pipelines(context.Context) ([]otel.ReceiverPipeline, error) {
+	return nil, confgenerator.GetUnsupportedOtelLogReceiverError()
+}
+
 func (r LoggingReceiverActiveDirectoryDS) Components(ctx context.Context, tag string) []fluentbit.Component {
 	l := confgenerator.LoggingReceiverWindowsEventLog{
 		Channels: []string{"Directory Service", "Active Directory Web Services"},
--- a/apps/apache.go
+++ b/apps/apache.go
@@ -78,6 +78,10 @@ func (LoggingProcessorApacheError) Type() string {
 	return "apache_error"
 }
 
+func (p LoggingProcessorApacheError) Processors(ctx context.Context) ([]otel.Component, error) {
+	return nil, confgenerator.GetUnsupportedOtelLogProcessorError()
+}
+
 func (p LoggingProcessorApacheError) Components(ctx context.Context, tag string, uid string) []fluentbit.Component {
 	c := confgenerator.LoggingProcessorParseRegex{
 		// Documentation: https://httpd.apache.org/docs/current/logs.html#errorlog
@@ -135,6 +139,10 @@ type LoggingProcessorApacheAccess struct {
 	confgenerator.ConfigComponent `yaml:",inline"`
 }
 
+func (p LoggingProcessorApacheAccess) Processors(ctx context.Context) ([]otel.Component, error) {
+	return nil, confgenerator.GetUnsupportedOtelLogProcessorError()
+}
+
 func (p LoggingProcessorApacheAccess) Components(ctx context.Context, tag string, uid string) []fluentbit.Component {
 	return genericAccessLogParser(ctx, p.Type(), tag, uid)
 }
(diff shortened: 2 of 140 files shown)


franciscovalentecastro:
> Shouldn't the new top-level Pipelines and Processors functions defer to the underlying embedded structs instead of directly returning an error?

Yes, that is how eventually every `Receiver` and `Processor` is going to be implemented. Returning an `error` is a proposed solution on how to correctly determine if a receiver / processor is supported or not by otel logging.

Currently the determination between `supported` or `unsupported` relies on this methods (Pipelines and Processors) to be `implemented` or `missing`, respectively. This indeed is the more elegant solution for the "supported check", but the assumptions for this fail when some of the embedded structs are supported and others aren't. For example, (as presented in the PR explanation), the current `flink` receiver is determined  as `supported`, but it relies on the `regex` parser which is unsupported.

jefferbrecht:
> Currently the determination between `supported` or `unsupported` relies on this methods (Pipelines and Processors) to be `implemented` or `missing`, respectively. This indeed is the more elegant solution for the "supported check", but the assumptions for this fail when some of the embedded structs are supported and others aren't. For example, (as presented in the PR explanation), the current `flink` receiver is determined as `supported`, but it relies on the `regex` parser which is unsupported.

I meant to have the deferral take place inside the top-level function itself. confgenerator won't have general knowledge of how compound components (like most 3P receivers) are structured, but the components themselves do. Something like this:

```go
func (r LoggingReceiverFlink) Pipelines(ctx context.Context) ([]otel.ReceiverPipeline, error) {
	mixinPipelines, err := r.LoggingReceiverFilesMixin.Pipelines(ctx)
	if err != nil {
		return nil, err
	}
	// Now try compiling LoggingProcessorFlink to []otel.Component, which should error out if otel isn't supported
	// Finally, merge the compiled processors with mixinPipelines
}
```

The more fundamental issue here seems to be that we have no intermediate representation for compound logging components -- we don't actually have a way to try compiling `LoggingProcessorFlink` to OTel components, because the underlying constituent components are being erased on lines 117 & 138.

franciscovalentecastro:
This makes sense! I agree! Having an intermediate structure for the 3P app logging receivers and processors which would be able to determine if the underlying OTel components are implemented (or not implemented) is going to help in the long run.

@quentinmit implemented a prototype with this idea in mind in branch [quentin-generic-receivers](https://github.com/GoogleCloudPlatform/ops-agent/tree/quentin-generic-receivers).

I decided to try a simpler solution by just change the struct elements from `embedded structs` to `fields`, so the methods are not promoted (avoiding the misdetection as OtelReceiver), and consider a bigger change of the structure of `Composited LoggingReceiver` (e.g. using [generics](https://github.com/GoogleCloudPlatform/ops-agent/tree/quentin-generic-receivers)) for a followup PR.
```
type LoggingReceiverFlink struct {
	LoggingProcessorFlink                                    loggingProcessorFlink          `yaml:",inline"`
	confgenerator.LoggingReceiverFilesMixin    loggingReceiverFilesMixin    `yaml:",inline" validate:"structonly"`
}
```

PR #1902, opened 2025-03-05
UAP Windows Plugin - Full Implementation

XuechunHou:
## Description
This PR implements the following:
1. Full Windows Plugin: Start(), Stop(), GetStatus()
2. Removed config validation step from the Linux plugin implementation, as the config generation step already does validation.
3. The `RunCommand` method was removed from being a struct method because its method signatures are incompatible between Windows and Linux.
4.  `GetUserAndMergedConfigs` function from the diagnostics service impl is exported and used in the UAP plugin. 

## Related issue
b/399170437

## How has this been tested?
Unit tests are added.
Integration tests will come later. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/cmd/ops_agent_uap_plugin/service_windows.go
+++ b/cmd/ops_agent_uap_plugin/service_windows.go
@@ -21,16 +21,39 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 	"os/exec"
+	"path/filepath"
 
 	"golang.org/x/sys/windows/svc/mgr"
 	"google.golang.org/grpc/status"
 
+	"github.com/GoogleCloudPlatform/ops-agent/apps"
+	"github.com/GoogleCloudPlatform/ops-agent/confgenerator"
+	"github.com/kardianos/osext"
+	"golang.org/x/sys/windows/svc/debug"
+	"golang.org/x/sys/windows/svc/eventlog"
+
 	pb "github.com/GoogleCloudPlatform/ops-agent/cmd/ops_agent_uap_plugin/google_guest_agent/plugin"
 )
 
+const (
+	GeneratedConfigsOutDir           = "generated_configs"
+	LogsDirectory                    = "log"
+	RuntimeDirectory                 = "run"
+	OpsAgentUAPPluginEventID  uint32 = 8
+	DiagnosticsEventID        uint32 = 2
+	WindowsEventLogIdentifier        = "google-cloud-ops-agent-uap-plugin"
+	AgentWrapperBinary               = "google-cloud-ops-agent-wrapper.exe"
+	DiagnosticsBinary                = "google-cloud-ops-agent-diagnostics.exe"
+	FluentbitBinary                  = "fluent-bit.exe"
+	OtelBinary                       = "google-cloud-metrics-agent_windows_amd64.exe"
+)
+
 var (
-	AgentWindowsServiceName = []string{"StackdriverLogging", "StackdriverMonitoring", "google-cloud-ops-agent"}
+	AgentWindowsServiceName       = []string{"StackdriverLogging", "StackdriverMonitoring", "google-cloud-ops-agent"}
+	DefaultPluginStateDirectory   = filepath.Join(os.Getenv("PROGRAMDATA"), "Google/Compute Engine/google-guest-agent/agent_state/plugins/ops-agent-plugin")
+	OpsAgentConfigLocationWindows = filepath.Join(os.Getenv("PROGRAMDATA"), "Google/Cloud Operations/Ops Agent/config/config.yaml")
 )
 
 // Apply applies the config sent or performs the work defined in the message.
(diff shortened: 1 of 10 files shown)


rafaelwestphal (approved):
I can' t attest much for the windows specific details, but the rest of the implementation follows the same pattern for the Linux impl, so LGTM.

cmd/ops_agent_uap_plugin/service_windows.go
```diff
@@ -53,9 +195,280 @@ func (ps *OpsAgentPluginServer) Stop(ctx context.Context, msg *pb.StopRequest) (
...
+func (c *windowsServiceManagerConn) ListServices() ([]string, error) {
+	return c.mgr.ListServices()
+}
+
+func (c *windowsServiceManagerConn) Disconnect() error {
+	return c.mgr.Disconnect()
+}
+
+// findPreExistentAgents checks if any of the Ops Agent and legacy agents are already installed as Window Services.
+func findPreExistentAgents(mgr serviceManager, agentWindowsServiceNames []string) (bool, error) {
```
jinghan-ma, line 243: Would it be more generic if we changed `findPreExistentAgents` to return a `[]string` instead of bool? i.e. It would:

1. Return an empty array if no conflicting agent is installed.
2. Return an array of all conflicting agents
3. Return an error if any issues arise when trying to list services

Then, in Start, we can check if the returned array is empty or not instead of the bool.

This way, if we need to re-use it in the future (e.g. part of diagnostics/troubleshooting), we wouldn't have to parse an error object to distinguish between failing to list services vs having a conflicting agent installed.

cmd/ops_agent_uap_plugin/service_windows.go
```diff
@@ -53,9 +195,280 @@ func (ps *OpsAgentPluginServer) Stop(ctx context.Context, msg *pb.StopRequest) (
...
+		"--storage_path", path.Join(pluginStateDirectory, "run/buffers"),
+	)
+	wg.Add(1)
+	go runSubAgentCommand(ctx, cancel, runFluentBitCmd, runCommand, &wg, jobHandle)
+
+	wg.Wait()
+	windows.CloseHandle(jobHandle)
+}
+
+func runDiagnosticsService(ctx context.Context, cancel context.CancelFunc, otelErrorHandler otel.ErrorHandler, wg *sync.WaitGroup) {
```
jinghan-ma, line 393: Why do we not assign the diagnostics service to the JobHandle. Do we plan on manually ensuring it's killed?

cmd/ops_agent_uap_plugin/service_windows.go
```diff
@@ -32,19 +84,109 @@ func (ps *OpsAgentPluginServer) Apply(ctx context.Context, msg *pb.ApplyRequest)
...
+		log.Printf("Start() failed, because it failed to create Windows event logger: %s", err)
+		return nil, status.Error(1, err.Error())
+	}
+
+	// Subagents config validation and generation.
+	if err := generateSubAgentConfigs(ctx, OpsAgentConfigLocationWindows, pluginStateDir, windowsEventLogger); err != nil {
+		ps.Stop(ctx, &pb.StopRequest{Cleanup: false})
+		windowsEventLogger.Close()
+		log.Printf("Start() failed at the subagent config validation and generation step: %s", err)
+		return nil, status.Error(1, err.Error())
```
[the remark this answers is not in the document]
XuechunHou: Done

cmd/ops_agent_uap_plugin/service_windows.go
```diff
@@ -53,9 +195,280 @@ func (ps *OpsAgentPluginServer) Stop(ctx context.Context, msg *pb.StopRequest) (
...
+	return &pb.Status{Code: 0, Results: []string{"The Ops Agent Plugin is running ok."}}, nil
+}
+
+// serviceManager is an interface to abstract the Windows service manager. This is used to facilitate testing.
+type serviceManager interface {
+	Connect() (serviceManagerConnection, error)
+}
+
+// serviceManagerConnection is an interface to abstract the connection to the Windows service manager. This is used to facilitate testing.
+type serviceManagerConnection interface {
```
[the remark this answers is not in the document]
XuechunHou: The methods `Connect`, `ListServices`, and `Disconnect` cannot reside in the same interface. ListServices and Disconnect are instance methods, requiring a valid instance created by Connect to be invoked.


In the golang mgr package:
`func Connect() (*Mgr, error)`

The `Connect` function is not a struct method, it's a standalone public method that does not implement any interfaces. 

```
The `Mgr` struct has to struct methods: `ListServices()` and `Disconnect()`:


type Mgr struct {
	Handle windows.Handle
}

func (m *Mgr) ListServices() ([]string, error)
func (m *Mgr) Disconnect() error 
```

`Connect()` needs to be invoked first to obtain a `Mgr` instance, and then `ListServices` and `Disconnect` can be invoked through the `Mgr` instance.

cmd/ops_agent_uap_plugin/service_windows.go
XuechunHou: Although the runDiagnosticsService function runs in a goroutine, it does not start any os processes, unlike the otel and fluentbit goroutines.

child goroutines are terminated when the parent program terminates. This is handled by go.

cmd/ops_agent_uap_plugin/service_windows.go
```diff
@@ -53,9 +195,280 @@ func (ps *OpsAgentPluginServer) Stop(ctx context.Context, msg *pb.StopRequest) (
...
+	healthCheckResults := gceHealthChecks.RunAllHealthChecks(healthCheckFileLogger)
+
+	// Log health check results to windows event log too.
+	healthCheckWindowsEventLogger := logs.WindowsServiceLogger{EventID: OpsAgentUAPPluginEventID, Logger: windowsEventLogger}
+	healthchecks.LogHealthCheckResults(healthCheckResults, healthCheckWindowsEventLogger)
+	windowsEventLogger.Info(OpsAgentUAPPluginEventID, "Health checks completed")
+}
+
+func createWindowsJobHandle() (windows.Handle, error) {
+	jobHandle, err := windows.CreateJobObject(nil, nil)
```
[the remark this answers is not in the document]
XuechunHou: I don't think the Job object is retrievable after the program exits, because all job handles will be closed, and the Job object might already be destroyed.

cmd/ops_agent_uap_plugin/service_windows.go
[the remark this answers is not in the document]
jinghan-ma: I see. I just saw how it's being used in the tests. Makes sense.

cmd/ops_agent_uap_plugin/service_windows.go
```diff
@@ -53,9 +189,257 @@ func (ps *OpsAgentPluginServer) Stop(ctx context.Context, msg *pb.StopRequest) (
...
+	defer ps.mu.Unlock()
+	if ps.cancel == nil {
+		log.Println("The Ops Agent plugin is not running")
+		return &pb.Status{Code: 1, Results: []string{"The Ops Agent Plugin is not running."}}, nil
+	}
+	log.Println("The Ops Agent plugin is running")
+	return &pb.Status{Code: 0, Results: []string{"The Ops Agent Plugin is running ok."}}, nil
+}
+
+// serviceManager is an interface to abstract the Windows service manager. This is used to facilitate testing.
```
jinghan-ma, line 203: Would it make sense to move this to a separate file in `utils`? I feel like it distracts from the core logic of this file.

cmd/ops_agent_uap_plugin/service_windows.go
```diff
@@ -53,9 +189,257 @@ func (ps *OpsAgentPluginServer) Stop(ctx context.Context, msg *pb.StopRequest) (
...
+		}
+	}
+
+	if len(alreadyInstalledAgentServiceNames) != 0 {
+		return alreadyInstalledAgentServiceNames, fmt.Errorf("conflicting installations identified: %v", alreadyInstalledAgentServiceNames)
+	}
+	return alreadyInstalledAgentServiceNames, nil
+}
+
+func createWindowsEventLogger() (debug.Log, error) {
```
jinghan-ma, line 267: One possibility is to move `createWindowsEventLogger()`, `otelErrorHandler`, and `runHealthChecks` into a separate file (e.g. `logging_windows.go`).

Could help keep the `service_windows.go` file focused on the Start/Stop/GetStatus functionalities, without needing to worry about small details like how logging implementation.

jinghan-ma (approved):
Left a few suggestions, but they are mostly opinions.

LGTM

PR #1904, opened 2025-03-07
integration_test : Add `TestAppHubLogLabels` to test AppHub integration from Ops Agent collected logs.

franciscovalentecastro:
## Description
This test does the following : 
- Create a test Instance Template and a Managed Instance Group (MIG).
- Creates a VM in the MIG that the test can SSH into.
- Register MIG to a test AppHub application `ops-agent-app-hub-integration-test-app` as a workload.
- Verify a custom `syslog` log contains the correct `apphub.*` labels in Cloud Logging.
- Cleanup resources created for test and stale resources (older than 24h) from previous test executions.

## Related issue
b/354756009

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/gce/gce_testing.go
+++ b/integration_test/gce/gce_testing.go
@@ -1442,7 +1442,7 @@ func attemptCreateManagedInstanceGroupInstance(ctx context.Context, logger *log.
 	}
 
 	createTemplateArgs := []string{
-		"beta", "compute", "instance-templates", "create", vm.Name + "-temp",
+		"beta", "compute", "instance-templates", "create", vm.Name + "-t",
 		"--project=" + vm.Project,
 		"--machine-type=" + vm.MachineType,
 		"--network=" + vm.Network,
@@ -1488,7 +1488,7 @@ func attemptCreateManagedInstanceGroupInstance(ctx context.Context, logger *log.
 		"--project=" + vm.Project,
 		"--zone=" + vm.Zone,
 		"--size=0",
-		"--template=" + vm.Name + "-temp",
+		"--template=" + vm.Name + "-t",
 		"--stateful-internal-ip=enabled",
 		"--format=json",
 	}
@@ -1961,7 +1961,7 @@ func DeleteManagedInstanceGroupInstance(logger *log.Logger, vm *VM) error {
 		attempt++
 		_, err = RunGcloud(ctx, logger, "",
 			[]string{
-				"compute", "instance-templates", "delete", vm.Name + "-temp",
+				"compute", "instance-templates", "delete", vm.Name + "-t",
 				"--project=" + vm.Project,
 			})
 		if err == nil {
--- a/integration_test/gce/gce_testing.go
+++ b/integration_test/gce/gce_testing.go
@@ -320,7 +320,7 @@ func (migVM ManagedInstanceGroupVM) ManagedInstanceGroupName() string {
 }
 
 func (migVM ManagedInstanceGroupVM) InstanceTemplateName() string {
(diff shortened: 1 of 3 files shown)


integration_test/gce/gce_testing.go
```diff
@@ -1229,32 +1243,80 @@ func attemptCreateInstance(ctx context.Context, logger *log.Logger, options VMOp
...
+	if IsDebianBased(vm.ImageSpec) {
+		if _, err := RunRemotely(ctx, logger, vm, "echo 'DPkg::Lock::Timeout=300' | sudo tee /etc/dpkg/dpkg.cfg.d/cache-lock-timeout.cfg"); err != nil {
+			return fmt.Errorf("setting increased dpkg cache lock timeout failed: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func addditionalCreateInstaceArgs(options VMOptions, vm *VM) ([]string, error) {
```
quentinmit, line 1309: Nit: I think you meant `additionalCreateInstanceArgs` (two typos)

integration_test/ops_agent_test/main_test.go
```diff
@@ -5102,6 +5114,159 @@ func TestLogCompression(t *testing.T) {
...
+
+func TestAppHubLogLabels(t *testing.T) {
+	t.Parallel()
+	gce.RunForEachImage(t, func(t *testing.T, imageSpec string) {
+		t.Parallel()
+
+		ctx, logger, migVM := setupMainLogAndManagedInstaceGroupVM(t, imageSpec)
+
+		// Cleanup stale resources.
+		t.Cleanup(func() { cleanupStaleResourcesForTestAppHubLogLabels(ctx, logger) })
```
quentinmit, line 5199: You probably don't want to do this for each image, because they'll all race with each other and try to delete the same resources multiple times. (I think it might be sufficient to just hoist this up to outside the `gce.RunForEachImage` call)

integration_test/ops_agent_test/main_test.go
```diff
@@ -5102,6 +5114,159 @@ func TestLogCompression(t *testing.T) {
...
+
+		// Setup Apphub #1 : Discover Managed Instance Group resource from AppHub API
```
quentinmit, line 5201: Do you need to discover the workload if you already know the identity of the MIG?

quentinmit (approved):
I'm not sure why we are shelling out to `gcloud` for all of this work instead of calling the API directly, but you seem to be copying the existing style. I think this code would be a lot simpler if it used the API directly.

integration_test/ops_agent_test/main_test.go
franciscovalentecastro: Yes, to create an AppHub workload it needs a specific `Discovered Workload ID` which can only be obtained from the calling the AppHub API [1].

[1] https://cloud.google.com/app-hub/docs/set-up-app-hub#register-services-and-workloads

integration_test/ops_agent_test/main_test.go
franciscovalentecastro: You are right. I updated the code to only run `cleanupStaleResourcesForTestAppHubLogLabels` once per execution of the `TestAppHubLogLabels` test, while keeping all logs in the same folder. This by only running the cleanup in the first imageSpec given to the test.

integration_test/gce/gce_testing.go
franciscovalentecastro: Fixed. Thank you!

PR #1919, opened 2025-03-28
Config Management via UAP

XuechunHou:
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/cmd/ops_agent_uap_plugin/plugin.go
+++ b/cmd/ops_agent_uap_plugin/plugin.go
@@ -25,6 +25,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/reflection"
+	"github.com/bufbuild/protoyaml-go"
 
 	pb "github.com/GoogleCloudPlatform/ops-agent/cmd/ops_agent_uap_plugin/google_guest_agent/plugin"
 )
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -176,6 +176,7 @@ RUN ./pkg/rpm/build.sh
 COPY cmd/ops_agent_uap_plugin cmd/ops_agent_uap_plugin
 COPY ./builds/ops_agent_plugin.sh .
 RUN ./ops_agent_plugin.sh /work/cache/
+RUN source VERSION && echo $PKG_VERSION  > /work/cache/OPS_AGENT_VERSION
```
jinghan-ma, line 179: Should we put this in `./ops_agent_plugin.sh`? This way we don't have to repeat it for each image, and also there's fewer Docker build stages, so fewer layers and faster build time.

cmd/ops_agent_uap_plugin/plugin.go
```diff
@@ -96,3 +98,38 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func writeCustomConfigToFile(req *pb.StartRequest, configPath string) error {
+	customConfig := []byte{}
+	switch req.GetServiceConfig().(type) {
+	case *pb.StartRequest_StringConfig:
+		customConfig = []byte(req.GetStringConfig())
+	case *pb.StartRequest_StructConfig:
```
jinghan-ma, line 107: If a user inputs a custom yaml and it's able to parse into a `StructConfig`, would the `StringConfig` also be sent, or only the parsed version?

If both are sent, wouldn't this case never trigger?

jinghan-ma: Also, do we even want to use the parsed config?

Since we simply write the config to a file and let the Ops Agent handle it, do we need to check for proper yaml at this stage?

jinghan-ma (approved):

cmd/ops_agent_uap_plugin/plugin.go
XuechunHou: https://source.corp.google.com/piper///depot/google3/third_party/guest_agent/dev/internal/plugin/proto/plugin_comm.proto;l=53 `service_config` is a oneof field, so either string_config or struct_config is sent per `StartRequest`, never both.

XuechunHou: > Also, do we even want to use the parsed config?
> 
> Since we simply write the config to a file and let the Ops Agent handle it, do we need to check for proper yaml at this stage?

"Also, do we even want to use the parsed config?" If the FE allows both, we will have to handle both.

struct_config is populated when users upload a config.yaml file at the FE, it will be parsed, validated and converted to a struct proto at the FE before us receiving it. 

string_config is populated when users paste a chunk of string at the FE, the string will be forwarded to us as is without any validation. 

I am not doing any validation at this stage really, I am just converting the struct proto back to a yaml and writes it to config.yaml file. Whether it's a valid Ops Agent config yaml is left to the conf generator...

Dockerfile
XuechunHou: Done

PR #1953, opened 2025-05-14
Add SecretManagerProvider to Ops Agent

XuechunHou:
## Description

1. This PR pulls the latest otel upgrade into Ops Agent. 
2. This PR modifies the configs used 3p integration tests, replacing the plaintext secrets with the google secret manager provider identifier. 

## Related issue
b/417214181

## How has this been tested?
integration tests 

## Checklist:
- Unit tests
- Integration tests
- Documentation
  - [X] This PR introduces user visible changes and the corresponding documentation change will be made.
- Minor version bump
  - [X] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.


integration_test/third_party_apps_test/applications/mysql/enable
```diff
@@ -19,7 +19,7 @@ metrics:
     mysql:
       type: mysql
       username: root
-      password: Ss123%321
+      password: {googlesecretmanager:projects/228437877067/secrets/ops-agent-3p-test-mysql-password/versions/1}
```
quentinmit, line 22: Won't this cause the test to fail when run on any other project?

Also, why is there not a `$` at the start of the value?

XuechunHou: When does the test run on other projects? 

"Also, why is there not a $ at the start of the value?" -> fixed, ty.

quentinmit: > When does the test run on other projects?

When it's run directly from a developer's workstation with `go test`

apps/oracledb.go
```diff
@@ -78,6 +79,9 @@ func (r MetricsReceiverOracleDB) Pipelines(_ context.Context) ([]otel.ReceiverPi
 	auth := url.QueryEscape(r.Username)
 	secretPassword := r.Password.SecretValue()
 	if len(secretPassword) > 0 {
+		if strings.Contains(secretPassword, "${googlesecretmanager") {
```
jinghan-ma, line 82: Nit. Wouldn't we want `secretPassword`  to start with `${googlesecretmanager`? (i.e. `HasPrefix`)

jinghan-ma (approved):

apps/oracledb.go
XuechunHou: Currently, we are trying to prevent users from using the secret provider in OracleDB-related configuration entries. 

This is because the password will be URI-escaped before being passed to the OTel collector, which results in the provider identifier being altered and consequently unable to map back to a secret entry in Google Secret Manager. 

We will remove this check later when we implement the fix in the sqlquery receiver.

braydonk: Here's the upstream issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40117

integration_test/third_party_apps_test/applications/mysql/enable
XuechunHou: The secrets are accessible as long as the SA attached to the test VM has `roles/secretmanager.secretAccessor` granted to the project hosting the secrets.

If the test is not running from within a VM, then the deverlopers `@google.com` email needs to be granted the iam role.

integration_test/third_party_apps_test/applications/couchbase/enable
```diff
@@ -13,13 +13,15 @@ fi
 sudo cp /etc/google-cloud-ops-agent/config.yaml /etc/google-cloud-ops-agent/config.yaml.bak
 
 # Configure the Ops Agent.
-sudo tee /etc/google-cloud-ops-agent/config.yaml > /dev/null << EOF
+sudo tee /etc/google-cloud-ops-agent/config.yaml > /dev/null << \EOF
```
braydonk, line 16: Why was the `\EOF` added in every enable script here?

integration_test/third_party_apps_test/applications/couchbase/enable
```diff
@@ -13,13 +13,15 @@ fi
...
 metrics:
   receivers:
     couchbase:
       type: couchbase
       username: admin
-      password: password
+      # Use plaintext password if you are not using the googlesecretmanager provider.
+      # password: password 
+      password: ${googlesecretmanager:projects/228437877067/secrets/ops-agent-3p-test-couchbase-password/versions/1}
```
braydonk, line 24: I am not a fan of the testing strategy in this PR because:

1. It overrides the plaintext password testcases reducing our coverage
2. It buries `googlesecretmanager` tests in random database app tests. It should really have its own dedicated test cases in the main test framework.

I think this should be redone in a way that addresses both of those points.

braydonk (commented):
We can leave it for now, but in the future a PR like this that does two distinctly different things (upgrades `otelopscol`, adds new test cases for `googlesecretmanager`) should be done in two separate PRs.

integration_test/third_party_apps_test/applications/couchbase/enable
XuechunHou: Adding \ so that the provider identifier is not interpreted as a shell variable. 

This content is eventually stored as a shell script in the test VM: 
[image](https://screenshot.googleplex.com/B5PWySZ5ciMRxP3)

integration_test/third_party_apps_test/applications/couchbase/enable
quentinmit: I agree with this, and I think it should be replaced with an integration test that also creates the secrets in the current project, so the test is not hardcoded to a single project.

integration_test/third_party_apps_test/applications/mysql/enable
quentinmit: And if the developer does not have an `@google.com` account? Ops Agent is open source software.

apps/oracledb.go
jinghan-ma: Based on my previous comment, I don't think deciding whether to implement escaping in a new provider or sql receiver should be blocking for the Secret Manager integration.

quentinmit: Also copied from the upstream issue:

> This isn't a question of the "ideal" solution - the proposed solution here can not work. Once the string is already concatenated into a single URL, it's no longer possible to know which part of that string needs encoding applied to it.

Also, the exact same problem applies to environment variable substitution, not just `googlesecretmanager`, so if you want to report a configuration error until this is supported upstream, you should be checking for `${` and not just `${googlesecretmanager`

jinghan-ma: That is a good point. This would also not work with env variable substitution. Maybe we do a regex match for  something like`${.*}`

Let's also change the error message to say that this integration does not support confmap providers.

XuechunHou: Done

integration_test/third_party_apps_test/applications/couchbase/enable
braydonk: > Adding \ so that the provider identifier is not interpreted as a shell variable.

This does not explain why `EOF` has a `\`, I'm still confused about this.

integration_test/third_party_apps_test/applications/couchbase/enable
XuechunHou: Done

braydonk (approved):

braydonk (approved):

PR #1957, opened 2025-05-27
[Draft] Testing updates to get Otel Logging `AgentSelfMetrics`.

franciscovalentecastro:
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/confgenerator/agentmetrics.go
+++ b/confgenerator/agentmetrics.go
@@ -63,6 +63,21 @@ func (r AgentSelfMetrics) OtelPipeline() otel.ReceiverPipeline {
 				"otelcol_process_memory_rss",
 				"grpc.client.attempt.duration",
 				"googlecloudmonitoring/point_count",
+				"otelcol_exporter_sent_log_records",
+				"otelcol_receiver_refused_log_records",
+				"otelcol_receiver_accepted_log_records",
+			),
+			otel.MetricsTransform(
+				otel.RenameMetric("grpc.client.attempt.duration", "grpc.client.attempt.duration.logging",
+					// TODO: below is proposed new configuration for the metrics transform processor
+					// ignore any non "google.monitoring" RPCs (note there won't be any other RPCs for now)
+					// - action: select_label_values
+					//   label: grpc_client_method
+					//   value_regexp: ^google\.monitoring
+					otel.RenameLabel("grpc.status", "state"),
+					// delete grpc_client_method dimension & service.version label, retaining only state
+					otel.AggregateLabels("sum", "state"),
+				),
 			),
 			otel.Transform("metric", "metric",
 				// create new count metric from histogram metric
--- a/confgenerator/agentmetrics.go
+++ b/confgenerator/agentmetrics.go
@@ -15,7 +15,6 @@
 package confgenerator
 
 import (
-	"context"
 	"fmt"
 	"path/filepath"
 	"time"
@@ -28,11 +27,15 @@ import (
 // It is never referenced in the config file, and instead is forcibly added in confgenerator.go.
 // Therefore, it does not need to implement any interfaces.
 type AgentSelfMetrics struct {
-	Version string
-	Port    int
+	MetricsVersionLabel string
+	LoggingVersionLabel string
+	FluentBitPort       int
+	OtelPort            int
+	OtelLoggingEnabled  bool
+	OtelRuntimeDir      string
 }
 
-func (r AgentSelfMetrics) MetricsSubmodulePipeline() otel.ReceiverPipeline {
+func (r AgentSelfMetrics) OtelPipeline() otel.ReceiverPipeline {
 	return otel.ReceiverPipeline{
 		Receiver: otel.Component{
 			Type: "prometheus",
@@ -43,7 +46,7 @@ func (r AgentSelfMetrics) MetricsSubmodulePipeline() otel.ReceiverPipeline {
 						"scrape_interval": "1m",
 						"static_configs": []map[string]interface{}{{
 							// TODO(b/196990135): Customization for the port number
-							"targets": []string{fmt.Sprintf("0.0.0.0:%d", r.Port)},
+							"targets": []string{fmt.Sprintf("0.0.0.0:%d", r.OtelPort)},
 						}},
 					}},
 				},
@@ -77,7 +80,7 @@ func (r AgentSelfMetrics) MetricsSubmodulePipeline() otel.ReceiverPipeline {
 				otel.RenameMetric("otelcol_process_uptime", "agent/uptime",
 					// change data type from double -> int64
 					otel.ToggleScalarDataType,
-					otel.AddLabel("version", r.Version),
+					otel.AddLabel("version", r.MetricsVersionLabel),
 					// remove service.version label
 					otel.AggregateLabels("sum", "version"),
 				),
@@ -107,7 +110,7 @@ func (r AgentSelfMetrics) MetricsSubmodulePipeline() otel.ReceiverPipeline {
 	}
 }
 
-func (r AgentSelfMetrics) LoggingSubmodulePipeline() otel.ReceiverPipeline {
+func (r AgentSelfMetrics) FluentBitPipeline() otel.ReceiverPipeline {
 	return otel.ReceiverPipeline{
 		Receiver: otel.Component{
 			Type: "prometheus",
@@ -119,7 +122,7 @@ func (r AgentSelfMetrics) LoggingSubmodulePipeline() otel.ReceiverPipeline {
 						"metrics_path":    "/metrics",
 						"static_configs": []map[string]interface{}{{
 							// TODO(b/196990135): Customization for the port number
-							"targets": []string{fmt.Sprintf("0.0.0.0:%d", r.Port)},
+							"targets": []string{fmt.Sprintf("0.0.0.0:%d", r.FluentBitPort)},
 						}},
 					}},
 				},
@@ -141,7 +144,7 @@ func (r AgentSelfMetrics) LoggingSubmodulePipeline() otel.ReceiverPipeline {
 				otel.RenameMetric("fluentbit_uptime", "agent/uptime",
 					// change data type from double -> int64
 					otel.ToggleScalarDataType,
-					otel.AddLabel("version", r.Version),
+					otel.AddLabel("version", r.LoggingVersionLabel),
 					// remove service.version label
 					otel.AggregateLabels("sum", "version"),
 				),
@@ -169,11 +172,11 @@ func (r AgentSelfMetrics) LoggingSubmodulePipeline() otel.ReceiverPipeline {
 	}
 }
 
-func OpsAgentSelfMetricsPipeline(ctx context.Context, outDir string) otel.ReceiverPipeline {
+func (r AgentSelfMetrics) OpsAgentPipeline() otel.ReceiverPipeline {
 	receiver_config := map[string]any{
 		"include": []string{
-			filepath.Join(outDir, "enabled_receivers_otlp.json"),
-			filepath.Join(outDir, "feature_tracking_otlp.json")},
+			filepath.Join(r.OtelRuntimeDir, "enabled_receivers_otlp.json"),
+			filepath.Join(r.OtelRuntimeDir, "feature_tracking_otlp.json")},
 		"replay_file":   true,
 		"poll_interval": time.Duration(60 * time.Second).String(),
 	}
--- a/confgenerator/agentmetrics.go
+++ b/confgenerator/agentmetrics.go
@@ -67,28 +67,22 @@ func (r AgentSelfMetrics) OtelPipeline() otel.ReceiverPipeline {
 				"otelcol_receiver_refused_log_records",
 				"otelcol_receiver_accepted_log_records",
 			),
-			otel.MetricsTransform(
-				otel.RenameMetric("grpc.client.attempt.duration", "grpc.client.attempt.duration.logging",
-					// TODO: below is proposed new configuration for the metrics transform processor
-					// ignore any non "google.monitoring" RPCs (note there won't be any other RPCs for now)
-					// - action: select_label_values
-					//   label: grpc_client_method
-					//   value_regexp: ^google\.monitoring
-					otel.RenameLabel("grpc.status", "state"),
-					// delete grpc_client_method dimension & service.version label, retaining only state
-					otel.AggregateLabels("sum", "state"),
-				),
-			),
 			otel.Transform("metric", "metric",
-				// create new count metric from histogram metric
-				ottl.ExtractCountMetric(true, "grpc.client.attempt.duration"),
+				[]ottl.Statement{
+					// Separate metric into logging and monitoring
+					ottl.CopyMetric("grpc.client.attempt.duration.logging", `grpc.client.attempt.duration.grpc_client_method ~= ^google\.logging`)[0],
+					ottl.CopyMetric("grpc.client.attempt.duration.monitoring", `grpc.client.attempt.duration.grpc_client_method ~= ^google\.monitoring`)[0],
+					// Create new count metrics from histogram metric
+					ottl.ExtractCountMetric(true, "grpc.client.attempt.duration.logging")[0],
+					ottl.ExtractCountMetric(true, "grpc.client.attempt.duration.monitoring")[0]},
 			),
 			otel.MetricsFilter(
 				"include",
 				"strict",
 				"otelcol_process_uptime",
 				"otelcol_process_memory_rss",
-				"grpc.client.attempt.duration_count",
+				"grpc.client.attempt.duration.logging_count",
+				"grpc.client.attempt.duration.monitoring_count",
 				"googlecloudmonitoring/point_count",
 			),
 			otel.MetricsTransform(
@@ -103,7 +97,17 @@ func (r AgentSelfMetrics) OtelPipeline() otel.ReceiverPipeline {
 					// remove service.version label
 					otel.AggregateLabels("sum"),
 				),
-				otel.RenameMetric("grpc.client.attempt.duration_count", "agent/api_request_count",
+				otel.RenameMetric("grpc.client.attempt.duration.monitoring_count", "agent/api_request_count",
+					// TODO: below is proposed new configuration for the metrics transform processor
+					// ignore any non "google.monitoring" RPCs (note there won't be any other RPCs for now)
+					// - action: select_label_values
+					//   label: grpc_client_method
+					//   value_regexp: ^google\.monitoring
+					otel.RenameLabel("grpc.status", "state"),
+					// delete grpc_client_method dimension & service.version label, retaining only state
+					otel.AggregateLabels("sum", "state"),
+				),
+				otel.RenameMetric("grpc.client.attempt.duration.logging_count", "agent/request_count",
 					// TODO: below is proposed new configuration for the metrics transform processor
 					// ignore any non "google.monitoring" RPCs (note there won't be any other RPCs for now)
 					// - action: select_label_values
--- a/confgenerator/agentmetrics.go
+++ b/confgenerator/agentmetrics.go
@@ -98,21 +98,11 @@ func (r AgentSelfMetrics) OtelPipeline() otel.ReceiverPipeline {
 					otel.AggregateLabels("sum"),
 				),
 				otel.RenameMetric("grpc.client.attempt.duration.monitoring_count", "agent/api_request_count",
-					// TODO: below is proposed new configuration for the metrics transform processor
-					// ignore any non "google.monitoring" RPCs (note there won't be any other RPCs for now)
-					// - action: select_label_values
-					//   label: grpc_client_method
-					//   value_regexp: ^google\.monitoring
 					otel.RenameLabel("grpc.status", "state"),
 					// delete grpc_client_method dimension & service.version label, retaining only state
 					otel.AggregateLabels("sum", "state"),
 				),
 				otel.RenameMetric("grpc.client.attempt.duration.logging_count", "agent/request_count",
-					// TODO: below is proposed new configuration for the metrics transform processor
-					// ignore any non "google.monitoring" RPCs (note there won't be any other RPCs for now)
-					// - action: select_label_values
-					//   label: grpc_client_method
-					//   value_regexp: ^google\.monitoring
 					otel.RenameLabel("grpc.status", "state"),
 					// delete grpc_client_method dimension & service.version label, retaining only state
 					otel.AggregateLabels("sum", "state"),
--- a/confgenerator/otel/ottl/ottl.go
+++ b/confgenerator/otel/ottl/ottl.go
@@ -202,6 +202,13 @@ func IsNotNil(a Value) Value {
 	return valuef(`%s != nil`, a)
 }
 
+// CopyMetric creates a copy of a given metric with a new name
+func CopyMetric(metricName string, condition string) Statements {
+	return Statements{
+		statementf(`copy_metric(%q) where %s`, metricName, condition),
+	}
+}
+
 // ExtractCountMetric creates a new metric based on the count value of a Histogram metric
 func ExtractCountMetric(monotonic bool, metricName string) Statements {
 	monotonicStr := "false"
--- a/confgenerator/otel/processors.go
+++ b/confgenerator/otel/processors.go
@@ -345,6 +345,21 @@ func RenameMetric(old, new string, operations ...map[string]interface{}) map[str
 	return out
 }
 
+// RenameMetricRegexp returns a config snippet that renames metrics matching the input regex
+// to new, applying zero or more transformations.
+func RenameMetricRegexp(metricRegex string, new string, operations ...map[string]interface{}) map[string]interface{} {
+	out := map[string]interface{}{
+		"match_type": "regexp",
+		"include":    metricRegex,
+		"action":     "update",
+		"new_name":   new,
+	}
+	if len(operations) > 0 {
+		out["operations"] = operations
+	}
+	return out
+}
+
 // UpdateMetric returns a config snippet applies transformations to the given metric name
 func UpdateMetric(metric string, operations ...map[string]interface{}) map[string]interface{} {
 	out := map[string]interface{}{
--- a/confgenerator/confgenerator.go
+++ b/confgenerator/confgenerator.go
@@ -91,28 +91,33 @@ func (uc *UnifiedConfig) GenerateOtelConfig(ctx context.Context, outDir string) 
 		return "", err
 	}
 
-	receiverPipelines["otel"] = AgentSelfMetrics{
-		Version: metricVersionLabel,
-		Port:    otel.MetricsPort,
-	}.MetricsSubmodulePipeline()
+	agentSelfMetrics := AgentSelfMetrics{
+		MetricsVersionLabel: metricVersionLabel,
+		LoggingVersionLabel: loggingVersionLabel,
+		FluentBitPort:       fluentbit.MetricsPort,
+		OtelPort:            otel.MetricsPort,
+		OtelLoggingEnabled:  uc.Logging.Service.OTelLogging,
+		OtelRuntimeDir:      outDir,
+	}
+
+	receiverPipelines["otel"] = agentSelfMetrics.OtelPipeline()
 	pipelines["otel"] = otel.Pipeline{
 		Type:                 "metrics",
 		ReceiverPipelineName: "otel",
 	}
 
-	receiverPipelines["ops_agent"] = OpsAgentSelfMetricsPipeline(ctx, outDir)
+	receiverPipelines["ops_agent"] = agentSelfMetrics.OpsAgentPipeline()
 	pipelines["ops_agent"] = otel.Pipeline{
 		Type:                 "metrics",
 		ReceiverPipelineName: "ops_agent",
 	}
 
-	receiverPipelines["fluentbit"] = AgentSelfMetrics{
-		Version: loggingVersionLabel,
-		Port:    fluentbit.MetricsPort,
-	}.LoggingSubmodulePipeline()
-	pipelines["fluentbit"] = otel.Pipeline{
-		Type:                 "metrics",
-		ReceiverPipelineName: "fluentbit",
+	if !uc.Logging.Service.OTelLogging {
+		receiverPipelines["fluentbit"] = agentSelfMetrics.FluentBitPipeline()
+		pipelines["fluentbit"] = otel.Pipeline{
+			Type:                 "metrics",
+			ReceiverPipelineName: "fluentbit",
+		}
 	}
 
 	otelConfig, err := otel.ModularConfig{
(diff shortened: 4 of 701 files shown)


franciscovalentecastro:
After exploring more the implementation of the monitoring metrics, it seems better to implement the metric `googlecloudlogging/log_count`, which would be similar to `googlecloudmonitoring/point_count`, instead of setting `response_code` manually. See for the already implemented metric : https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/metrics.go#L186. CC @quentinmit .

confgenerator/agentmetrics.go
```diff
@@ -28,11 +27,192 @@ import (
...
+			otel.RenameMetric("otelcol_exporter_send_failed_log_records", "agent/log_entry_count",
+				// change data type from double -> int64
+				otel.ToggleScalarDataType,
+				otel.AddLabel("response_code", "400"),
+				otel.AggregateLabels("sum", "response_code"),
+			),
+			otel.DuplicateMetric("otelcol_process_uptime", "agent/uptime",
+				// change data type from double -> int64
+				otel.ToggleScalarDataType,
+				otel.AddLabel("version", r.LoggingVersionLabel),
```
quentinmit: I don't think we should be producing a second agent uptime stream. For better or worse, the otel collector is the "metrics" agent even though it already handles metrics and traces and now we're adding logging. The only other uptime metric we should have is the one from fluent-bit.

confgenerator/agentmetrics.go
```diff
@@ -28,11 +27,192 @@ import (
...
+			"otelcol_process_uptime",
+			"otelcol_process_memory_rss",
+			"otelcol_exporter_sent_log_records",
+			"otelcol_exporter_send_failed_log_records",
+			"grpc.client.attempt.duration.logging_count",
+			"grpc.client.attempt.duration.monitoring_count",
+			"googlecloudmonitoring/point_count",
+		),
+		otel.MetricsTransform(
+			otel.DuplicateMetric("otelcol_exporter_send_failed_log_records", "agent/log_entry_retry_count",
```
quentinmit: Do you have anything to prevent this from colliding with the stream that comes from fluent-bit?

franciscovalentecastro: Yes, the pipeline that scrapes the `fluent-bit` prometheus endpoint is not created at all if `OtelLogging` is enabled.

https://github.com/GoogleCloudPlatform/ops-agent/blob/b0b1752eaa33c97a83a9793093fc5b8898bc175b/confgenerator/confgenerator.go#L115-L121

It is setup right now in a way that `otelPipelineProcessorsWithOtelLogging` replaces all the fluent-bit metrics.

confgenerator/agentmetrics.go
franciscovalentecastro: The only consideration of not having an `agent/uptime` with `version: logging` is that we would need modified version of the "Expecting Metrics" integration tests when testing with otel logging enabled. 

Keeping an `agent/uptime` for `logging` can be eventually removed, but keeping both would make sure it generates exactly the same set of metrics as fluent-bit does and have "behaviour parity".

quentinmit: But we still run an (empty) fluent-bit when otel logging is enabled, so we should still be reporting its uptime.

confgenerator/agentmetrics.go
quentinmit: I don't think we should do that - if the process is still running, we should still be exporting its metrics. Only when we actually remove fluent-bit entirely can we disable the fluent-bit metric scraping.

confgenerator/agentmetrics.go
franciscovalentecastro: Ok! I implemented this, now fluent-bit `uptime` is reported as is.

confgenerator/agentmetrics.go
franciscovalentecastro: Ok! I implemented a sum of the logging fluent-bit and otel metrics into the same `Cumulative` metrics (with its own Pros and Cons). See more details on the description. We should discuss more design details about this implementation.

confgenerator/agentmetrics.go
```diff
@@ -28,152 +27,351 @@ import (
...
+	OtelPort            int
+	OtelRuntimeDir      string
+	OtelLogging         bool
 }
 
-func (r AgentSelfMetrics) MetricsSubmodulePipeline() otel.ReceiverPipeline {
+var grpcToHTTPStatus = map[string]string{
+	"OK":                  "200",
+	"INVALID_ARGUMENT":    "400",
+	"FAILED_PRECONDITION": "400",
```
quentinmit, line 42: Can you add a comment linking to https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto so it's clear where they came from?

I was about to comment "FAILED_PRECONDITION should be 428", but then I found the canonical definition says it should be 400.

confgenerator/agentmetrics.go
```diff
@@ -28,152 +27,351 @@ import (
...
+	"UNKNOWN":             "500",
+	"INTERNAL":            "500",
+	"DATA_LOSS":           "500",
+	"UNIMPLEMENTED":       "501",
+	"UNAVAILABLE":         "503",
+	"DEADLINE_EXCEEDED":   "504",
+}
+
+func (r AgentSelfMetrics) AddSelfMetricsPipelines(receiverPipelines map[string]otel.ReceiverPipeline, pipelines map[string]otel.Pipeline) {
+	receiverPipelines["prometheus_metrics"] = r.PrometheusMetricsPipeline()
```
quentinmit: Don't we need to worry about these names colliding with the user's configuration? Especially this one seems like it could easily collide.

confgenerator/agentmetrics.go
```diff
@@ -28,152 +27,351 @@ import (
...
+				// remove service.version label
+				otel.AggregateLabels("sum", "version"),
+			),
+			otel.AddPrefix("agent.googleapis.com"),
+		),
+	}
+}
+
+func (r AgentSelfMetrics) LoggingMetricsPipelineProcessors() []otel.Component {
+	if r.OtelLogging {
```
quentinmit: I don't think we need this conditional and the non-otel case - with the OTLP receiver we could already be processing log entries through otel even without this flag enabled.

confgenerator/agentmetrics.go
```diff
@@ -28,152 +27,351 @@ import (
...
-				"fluentbit_uptime",
-				"fluentbit_stackdriver_requests_total",
-				"fluentbit_stackdriver_proc_records_total",
-				"fluentbit_stackdriver_retried_records_total",
+		),
+		otel.MetricsOTTLFilter([]string{}, []string{
+			// Filter out histogram datapoints where the grpc.target is not related to logging.
+			`metric.name == "grpc.client.attempt.duration_count" and (not IsMatch(datapoint.attributes["grpc.target"], "logging.googleapis"))`,
+		}),
+		otel.MetricsFilter(
```
quentinmit, line 208: Why do this filter twice?

confgenerator/agentmetrics.go
```diff
@@ -28,152 +27,351 @@ import (
...
+				ottl.TruncateStartTimeAll("1m"),
+			},
+		),
+		// Some metrics are missing metric unit. Needed to combine metrics.
+		otel.Transform("metric", "metric",
+			[]ottl.Statement{
+				ottl.SetMetricUnitAll("1"),
+			},
+		),
+		// The processors "interval" and "groupbyattrs" batch metrics in a 1 minute interval.
```
quentinmit: Is this guaranteed to be the *same* 1 minute interval, though?

e.g. imagine the interval processor chooses 12:00:30-12:01:29.999 as its 1 minute interval. An otel point could arrive with a timestamp of 12:00:49 and a fluent-bit point could arrive with a timestamp of 12:01:11. They'd be truncated to 12:00 and 12:01 respectively, even though they're both in the 12:00:30-12:01:29.999 interval. So then they wouldn't be aggregated together below, right? Then a minute later we'd have points for 12:01 and 12:02, and the 12:01 point will conflict with the previously-written 12:01 point. Am I missing something?

confgenerator/agentmetrics.go
```diff
@@ -28,152 +27,351 @@ import (
...
+				// change data type from double -> int64
+				otel.ToggleScalarDataType,
+				otel.AddLabel("response_code", "400"),
+				otel.AggregateLabels("sum", "response_code"),
+			),
+			otel.CombineMetrics("^otel_log_entry_count$$", "otel_log_entry_count",
+				otel.AggregateLabels("sum", "response_code")),
+		),
+		// Aggregating delta metrics from fluent-bit and otel logging will.
+		// Keep the initial value so the resulting cumulative metric at the end of the pipeline.
```
quentinmit: I think we need to use `drop` or we'll double-count fluent bit values if otel restarts and fluent-bit does not, no?

confgenerator/agentmetrics.go
```diff
@@ -28,152 +27,351 @@ import (
...
+		otel.CumulativeToDeltaWithOptions("keep",
+			"otel_log_entry_count", "otel_log_entry_retry_count", "otel_request_count",
+			"fluentbit_log_entry_count", "fluentbit_log_entry_retry_count", "fluentbit_request_count",
+		),
+		// Make sure all delta datapoints consist of a 1 minute inteval from "start_time" to "time".
```
quentinmit: What are the start times without this transformation? Does cumulativetodelta not reset the start time already?

confgenerator/otel/ottl/ottl.go
```diff
@@ -216,15 +216,33 @@ func IsNotNil(a Value) Value {
 	return valuef(`%s != nil`, a)
 }
 
+// CopyMetric creates a copy of a given metric with a new name
+func CopyMetric(metricName string, condition string) Statement {
+	return statementf(`copy_metric(%q) where name == "%s"`, metricName, condition)
```
quentinmit: Use `%q` instead of `"%s"` for all of these strings.

confgenerator/otel/ottl/ottl.go
```diff
@@ -216,15 +216,33 @@ func IsNotNil(a Value) Value {
 	return valuef(`%s != nil`, a)
 }
 
+// CopyMetric creates a copy of a given metric with a new name
+func CopyMetric(metricName string, condition string) Statement {
```
quentinmit: `newMetricName, oldMetricName string` so it's clearer what the arguments are for?

confgenerator/otel/processors.go
```diff
@@ -94,15 +94,33 @@ func CastToSum(metrics ...string) Component {
...
-			"include": map[string]interface{}{
-				"metrics":    metrics,
-				"match_type": "strict",
-			},
+	return CumulativeToDeltaWithOptions("", metrics...)
+}
+
+// CumulativeToDeltaWithOptions returns a Component that converts each cumulative metric to delta
+// with the possibility to set the "initial_value" option.
+func CumulativeToDeltaWithOptions(initial_value string, metrics ...string) Component {
```
quentinmit: This doesn't take options, only one specific setting. Maybe name it `CumultiveToDeltaWithInitialValue`?

confgenerator/agentmetrics.go
franciscovalentecastro: > An otel point could arrive with a timestamp of 12:00:49 and a fluent-bit point could arrive with a timestamp of 12:01:11.

This won't happen because `TruncateTime` rounds down the timestamp to a multiple of the given duration (see [1]). All `start_time` and `time` timestamps are going to be of the form `HH:MM:00.0000`, seconds and nanosecond are always *zero*, e.g `12:00:00.000 - 12:01:00.000` .

Edit : `TruncateTime` uses the Go lang `time.Truncate` function.

[1] https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/ottlfuncs#truncatetime
[2] https://pkg.go.dev/time#Time.Truncate

quentinmit: But the time being truncated is the point's original time, right? So the point that originally has a time of `12:00:49` will be truncated to `12:00:00`, but the point that originally has a time of `12:01:11` will be truncated to `12:01:00`, and those aren't the same time.

confgenerator/agentmetrics.go
franciscovalentecastro: There are two cases of `(start_time, time)` time range after `cumulativetodelta` processes a cumulative metric : 

- Case `First / Initial Value` - `Delta Interval (-inf ,  "first scrape time")`: 
  - [Cumulative] The prometheus cumulative metric has the timestamps `start_time = time = "first scrape time"`.
  - [Delta] The resulting delta interval is `start_time = Epoch 0 (1970)`, `time = "first scrape time"`.
  - The value of the delta point is the initial cumulative value.
  - [Edge Case] Fluent-bit restarts and the new "initial value" has a `start_time = Epoch 0 (1970)` which can't be aggregated with Otel data points.
  - Also, if we don't change `start_time = Epoch 0 (1970)` , that will be also the `start_time` of the later `deltatocumulative` metric.

- Case `Previous point stored by processor` - `Delta Interval ("previous scrape",  "current scrape")` : 
  - [Cumulative ] The cumulative metric has the timestamps `start_time = "first scrape time"` , `time = "current scrape time"`.
  - [Delta] The resulting delta point becomes `start_time = "previous scrape"`, `time = "current scrape"`.
  - The value of the delta point is the difference of the datapoints.
  - [Edge Case] A prometheus scrape fails or a point is dropped/missing, the range could be `time - start_time >= 2m`, which may not possible to aggregate with other "normal points".

confgenerator/agentmetrics.go
franciscovalentecastro: This pipeline has an `ExtractCountMetric` processor which creates a new metric. Having two filters enforces that every step of the pipeline only has the required metrics. For example, first filter keeps `grpc.client.attempt.duration`, but the second filter only keeps the new metric `grpc.client.attempt.duration_count`.

This is how it was implemented before this PR. After thinking about it, we don't really need the first filter, so i removed it. This reduces the number of processors in the pipeline, though not sure how much the readability/clarity of the pipeline improves. WDYT ?

confgenerator/agentmetrics.go
franciscovalentecastro: I see! I removed the conditional in `LoggingMetricsPipelineProcessors` so all `logging_metrics` pipelines will sum both metrics. Done.

confgenerator/agentmetrics.go
franciscovalentecastro: I updated the pipeline name to `agent_prometheus_metrics` which seems more specific and less likely to collide. WDYT?

confgenerator/otel/processors.go
franciscovalentecastro: Renamed to `CumultiveToDeltaWithInitialValue`. Done!

confgenerator/otel/ottl/ottl.go
franciscovalentecastro: Updated parameter names to `newMetricName, oldMetricName string`. Done!

confgenerator/otel/ottl/ottl.go
franciscovalentecastro: Done!

confgenerator/agentmetrics.go
franciscovalentecastro: > we'll double-count fluent bit values if otel restarts and fluent-bit does not

This won't happen because the mechanism that keeps the "historic metric count" for the summed metrics is the "deltatocumulative" processor, which will reset if otel restarts.

Note : `drop` could almost work, but we would loose all the "logs sent on startup" count from the first metric datapoint.

confgenerator/agentmetrics.go
franciscovalentecastro: Done!

confgenerator/agentmetrics.go
franciscovalentecastro: > imagine the interval processor chooses 12:00:30-12:01:29.999 as its 1 minute interval. An otel point could arrive with a timestamp of 12:00:49 and a fluent-bit point could arrive with a timestamp of 12:01:11. They'd be truncated to 12:00 and 12:01 respectively, even though they're both in the 12:00:30-12:01:29.999 interval. So then they wouldn't be aggregated together below

I misunderstood your initial example! Yes, this indeed is a problem with this approach to "align/batch" timestamps after using the `interval` processor. The `interval` processor would pick an arbitrary non-regular `1m` window causing this issue. I even experimentally tested this and the `otel` & `fluent-bit` metrics timestamps normally differ around 10 seconds, so this could/will indeed happen.

franciscovalentecastro: ### Solution

With the current pipeline setup, i see three alternatives on how to fix this issue : 

1. `[Currently Implemented in Last Commit]` : "Align" all datapoins within a batch (the resulting resource metrics group after `interval + groupbyattrs`) using `time = TruncateTime(Now(), 1m)`.
    - This has a similar potential downside of what value `Now()` is set too when processing a metrics within a "batch". E.g. One point has `12.00` and other has `12.01` if the processing takes longer than a second. 
    - Experimentally it does work.

2. `[Custom OTTL "Align"]` : We can create custom "Align" function that traverses all the datapoints and sets the `lastest/earliest` for alignment.

3. `[Upstream intervalprocessor logic]` : Updating the `intervalprocessor` is possible, but there are several considerations to create a valuable solution for different users of `otel`. For example, aggregating `delta` points is potentially lossy process and there is no "best" solution only "strategies/heuristics`

confgenerator/agentmetrics.go
quentinmit: Actually I think you hit on something with this. In `PipelineInstance.OTelComponents`, we construct receiver pipeline names of the form `rname` and `rname_0`, and pipeline names of the form `pname_rname`, `pname_rname_0`, `type_pname_rname`, and `type_pname_rname_0`. And the pipeline and receiver name have any embedded `_` turned into `__`. Therefore having a receiver pipeline with two or more underscores separating words and a pipeline with no underscores are both guaranteed to NEVER collide with user config.

confgenerator/agentmetrics.go
quentinmit: I think it's an improvement, thanks.

confgenerator/agentmetrics.go
quentinmit: I think you're now setting `time` and `start_time` to the same value (truncating a second time isn't going to subtract a minute, it's going to produce a noop).

Aside from that, I think what you have is good enough for now. I think you're right that it could still potentially produce different values, but that is a much much smaller race window (microseconds) so I'm not too worried about it.

The only real fix is for the intervalprocessor to adjust the output timestamps, which is why I think 3 is the best long-term strategy. I'm not sure why you say that aggregating delta points is potentially lossy - aside from losing timestamps, of course, but that's the point of the aggregation.

confgenerator/agentmetrics.go
quentinmit: I don't understand why this wouldn't double-count. Consider the following scrapes:

- scrape 1, delta 1 (because `keep`), cumulative 1 (t0-t1)
- scrape 2, delta 1 (2-1 = 1), cumulative 2 (t0-t2)
- scrape 3, delta 1 (3-1 = 1), cumulative 3 (t0-t3)
- otel restarts
- scrape 3, delta 3 (because `keep`), cumulative 3 (t4-t5)

The scrape after the restart should be reporting a cumulative 0 because we already reported those increments, not a cumulative 3 with a fresh start time (which is considered as additional to the previous cumulative 3)

confgenerator/agentmetrics.go
quentinmit: I think I understand why we're resetting the start time now. This would also be fixed if intervalprocessor set start time and end time correctly from the interval window.

confgenerator/otel/ottl/ottl.go
quentinmit: I don't see this change - did you forget to commit it?

franciscovalentecastro: I did commit in https://github.com/GoogleCloudPlatform/ops-agent/pull/1957/commits/d21fcf603aa43a08eeef64647e6df4e3c8c4a748.

confgenerator/agentmetrics.go
franciscovalentecastro: Consider, for contrast, the current released Ops Agent behaviour in the same example. I added some updates in the time ranges assuming `t(n+1) - t(n) ‎ =  1 minute` regularity to simplify the example : 
- scrape at t1, cumulative 1 (t1, t1)
  - first scrape of a prometheus metric has `start_time == time`
- scrape at t2, cumulative 2 (t1, t2)
- scrape at t3, cumulative 3 (t1, t3)
- otel restarts at t4
- scrape at t5, cumulative 3 (t5, t5)
  - first scrape of a prometheus metric has `start_time == time`
  - the scrape gets 3 (because  fluent-bit didn't restart) 
  - reported cumulative 3 with fresh timestamp t5

The current implementation in this PR with `keep` will result in the same behaviour as shown in this example. 

> The scrape after the restart should be reporting a cumulative 0 because we already reported those increments, not a cumulative 3 with a fresh start time (which is considered as additional to the previous cumulative 3)

Following the semantics of a "cumulative metric" the range `(t5, t6)` shouldn't have any data from scrapes `1, 2 or 3`, because there were no observations of an experiment ("logs sent") in that time range. This is "semantically" correct, but enforcing this fails when considering other edge cases. For example, if `otel` takes 3 minutes to start, the first scrape will get "log counts" from an unknown time range since `fluent-bit` started.

The main idea of why using `keep` doesn't double count is that the `initial datapoint value` after a restart only represents data of an "unknown" time range (e.g. `(?, t5)`) that happened before the scrape.

The provided example can be rewritten like this :  

- scrape at t1, delta 1 (because keep), cumulative 1 (t0, t1), 
  - after “aligning” all delta points to a 1m window we have a `start_time = t0`.
- scrape at t2, delta 1 (2-1 = 1), cumulative 2 (t0, t2)
- scrape at t3, delta 1 (3-1 = 1), cumulative 3 (t0, t3)
- otel restarts at t4,
  - deltatocumulative looses in-memory data about the previous datapoints
- scrape at t5, delta 3 (because keep), cumulative 3 (t4, t5),
  - deltatocumulative observes a new metric.
  - after “aligning” all delta points to 1m window we have a `start_time = t4`

confgenerator/agentmetrics.go
franciscovalentecastro: > I'm not sure why you say that aggregating delta points is potentially lossy - aside from losing timestamps, of course, but that's the point of the aggregation.

I was mostly referring to the paragraph in [^1] that says "Aggregating data over an interval is an inherently "lossy" process ...". You are right, the only thing you loose is precision aggregating delta points.

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/intervalprocessor#description


> I think you're now setting time and start_time to the same value (truncating a second time isn't going to subtract a minute, it's going to produce a noop).

I'm not sure at what specific part of the pipeline you refer to here. I refactored the "aligning timestamps truncated 1min window" logic so i think it is now clearer why the `time` and `start_time` are indeed different timestamps.

confgenerator/agentmetrics.go
franciscovalentecastro: > Therefore having a receiver pipeline with two or more underscores separating words and a pipeline with no underscores are both guaranteed to NEVER collide with user config.

I'm seeing the code an most of the explanation makes sense, but the last of part of how to "NEVER" collide i'm not sure it always works.

For example, consider a self metrics receiver pipeline : 

- receiver pipeline `agent__prometheus` (two or more underscores) 
- pipeine `loggingmetrics` (no underscores) 

We could have a collision if a user sets a receiver pipeline named `agent_prometheus` because `_` turns into `__`.  Maybe i'm missing something ?

### New rule 
I think having having a receiver pipeline with one underscore separating words (user receiver pipelines would have 0 or multiple of 2 underscores) and a pipeline with no underscores (user pipelines always have one underscore) are both guaranteed to NEVER collide with user config.

Example 
- receiver pipeline `agent_prometheus` (one underscore) 
- pipeine `loggingmetrics` (no underscores)

quentinmit: The `_` -> `__` conversion only happens for user-supplied pipeline names, so a user receiver named `agent_prometheus` would become the receiver pipeline `agent__prometheus`, but our `agent_prometheus` receiver pipeline would stay as `agent_prometheus` and thus they're not the same name. There's no way for a user config to produce a receiver pipeline named `agent_prometheus`. So your current receiver pipeline name `agent_prometheus_metrics` is correct.

For pipeline names, we concatenate the user's pipeline name with the user's receiver name, so a pipeline named `logging` and a receiver named `metrics` could produce a pipeline named `logging_metrics`. So we would need to name the pipeline `loggingmetrics` (with no underscore as you said) to make it un-collidable. Likewise the existing `ops_agent` pipeline should probably be changed to `opsagent` to ensure it can't collide either.

confgenerator/agentmetrics.go
quentinmit: I think the new logic is now correct. At one point your code said:

```go
                      []ottl.Statement{
                              `set(time, TruncateTime(Now(), Duration("1m")))`,
                              `set(start_time, TruncateTime(start_time, Duration("1m")))`,
```

and I think with that code it's possible for both to end up with the same value. But as I said your code looks good now.

confgenerator/agentmetrics.go
quentinmit: If you're accurately describing the currently released Ops Agent behavior, then that behavior is also buggy. I would have expected us to be correctly using `process_start_time` as the start time for our fluent-bit metrics today.

You're correct that using `drop` would potentially result in undercounting if there were logs uploaded while otel was not running. I think that's the lesser of two evils, since it's only a problem if otel is not running for a long time, whereas overcounting is a problem every single time otel restarts.

confgenerator/agentmetrics.go
franciscovalentecastro: > I think with that code it's possible for both to end up with the same value
Yes, with the previous implementation i agree there could be an issue with the same timestamps while truncating.

Sounds good! Done!

confgenerator/agentmetrics.go
franciscovalentecastro: Ok! Makes sense! Thank you for the explanations! I just updated the receiver pipelines and pipeline names accordingly. I also added some comments describing this restrictions. Done!

confgenerator/agentmetrics.go
franciscovalentecastro: > If you're accurately describing the currently released Ops Agent behavior, then that behavior is also buggy. I would have expected us to be correctly using process_start_time as the start time for our fluent-bit metrics today.

Now i realized I was missing a piece of the (puzzle) pipeline. The OA behaviour example i was describing was from observing the `debugexporter` logs before exporting the metrics.

The `googlecloudexporter` has a feature `cumulative_normalization` [^1] enabled by default. This feature stores "reset" points (`startTime == 0` or  `startTime == endTime` or `currValue < prevValue`) and uses them as "anchor" to substract the subsequent point values from it. This the same behaviour as `auto` [^2] in the `cumulativetodeltaprocessor` (which is very similar to `drop` with additional heuristics).

Note : `fluent-bit` does have `process_start_time`, but `otel` doesn't. None of them are configured to be used in the prometheus receiver.

#### Current OA released behaviour : 

- scrape at t1, cumulative 1 (t1, t1) ->  not exported
  - first scrape of a prometheus metric has start_time == time
  - stored as anchor by `googlecloud`exporter
- scrape at t2, cumulative 2 (t1, t2) -> exported as cumulative 1 (2 - 1) with (t1, t2)
- scrape at t3, cumulative 3 (t1, t3) -> exported as cumulative 2 (3 - 1) with (t1, t3)
- otel restarts at t4
- scrape at t5, cumulative 3 (t5, t5) -> not exported
  - first scrape of a prometheus metric has start_time == time
  - stored as anchor by `googlecloud`exporter

> You're correct that using drop would potentially result in undercounting if there were logs uploaded while otel was not running. I think that's the lesser of two evils, since it's only a problem if otel is not running for a long time, whereas overcounting is a problem every single time otel restarts.

I agree with this! Undercounting is the lesser of two evils. The current OA released behaviour also undercounts. Thank you for point this out!

#### Solution

After all this details its clear to me that using `keep` was double-counting logs. Thank you for pointing this out and providing examples @quentinmit !

I think the best solution is to set `auto` in the `cumulativetodeltaprocessor` which does the same thing as  `cumulative_normalization`. Note that since we are aligning all points to have `endTime - startTime = 1 min`, the export normalization will never be applied.

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/googlecloudexporter#configuration-reference

[^2]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/cumulativetodeltaprocessor/internal/tracking/tracker_test.go#L110-L125

franciscovalentecastro: Discussed offline with @quentinmit . The additional heuristics that "auto" uses only happen for the first point. "drop" is the correct setting for `initial_value` in `cumulativetodelta`.

confgenerator/agentmetrics.go
```diff
@@ -28,152 +27,289 @@ import (
...
+		),
+		// Aggregating as delta metrics isolates data for only the most recent metric cumulative update.
+		// Set `initial_value: drop" to always store the "first point" as "anchor" to set the "start_time" and
+		// calculate next point values as difference with the "first point" value.
+		otel.CumulativeToDeltaWithInitialValue("drop",
+			"otel_log_entry_count", "otel_log_entry_retry_count", "otel_request_count",
+			"fluentbit_log_entry_count", "fluentbit_log_entry_retry_count", "fluentbit_request_count",
+		),
+		otel.TransformationMetrics(
+			[]otel.TransformQuery{
```
quentinmit: Nit: You don't need to create a slice just to pass variadic arguments here.

(Also, yikes, the implementation of `otel.TransformationMetrics` currently performs all metric statements, then all datapoint statements, then all scope statements. Instead of performing them in the order passed. I don't think that's relevant for this processor, but it's a big footgun. And the `Statement` field is a `string` instead of an `ottl.Statement`... Add it to the pile of tech debt, I guess.)

confgenerator/agentmetrics.go
```diff
@@ -28,152 +27,289 @@ import (
...
+				otel.SetName("fluentbit_request_count", "agent/request_count"),
+				otel.SetName("otel_log_entry_count", "agent/log_entry_count"),
+				otel.SetName("otel_log_entry_retry_count", "agent/log_entry_retry_count"),
+				otel.SetName("otel_request_count", "agent/request_count"),
+			}...,
+		),
+		// DeltaToCumulative keeps in memory information of previous delta points
+		// to generate a valid cumulative monotonic metric.
+		otel.DeltaToCumulative(),
+		// The processor "interval" batches metrics into a 1 minute interval.
```
quentinmit: Nit: `The processor "interval" outputs the last point in each 1 minute interval` (it doesn't just batch them up)

confgenerator/agentmetrics.go
franciscovalentecastro: Done ! Addressed `Nit` in https://github.com/GoogleCloudPlatform/ops-agent/pull/1957/commits/d2cbb8f8ba1972dc55428870e0a8762b94d553f4.

confgenerator/agentmetrics.go
franciscovalentecastro: Done ! Addressed `Nit` in https://github.com/GoogleCloudPlatform/ops-agent/pull/1957/commits/d2cbb8f8ba1972dc55428870e0a8762b94d553f4.

PR #1960, opened 2025-06-04
Deprecate instrumentation scope processor

EdwardJXLi:
## Description
Deprecate the existing `modifyscope` processor within ops-agent and migrate to utilizing the transform processor OTTL for achieving the same functionality.

## Related issue
N/A

## How has this been tested?
Unit Tests & Integration Tests

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/apps/active_directory_ds.go
+++ b/apps/active_directory_ds.go
@@ -45,7 +45,10 @@ func (r MetricsReceiverActiveDirectoryDS) Pipelines(_ context.Context) ([]otel.R
 			otel.MetricsTransform(
 				otel.AddPrefix("workload.googleapis.com"),
 			),
-			otel.ModifyInstrumentationScope(r.Type(), "1.0"),
+			otel.TransformationMetrics(
+				otel.SetScopeName("agent.googleapis.com/"+r.Type()),
+				otel.SetScopeVersion("1.0"),
+			),
 		}},
 	}}, nil
 }
--- a/apps/iis.go
+++ b/apps/iis.go
@@ -49,6 +49,8 @@ func (r MetricsReceiverIis) Pipelines(_ context.Context) ([]otel.ReceiverPipelin
 				otel.TransformationMetrics(
 					otel.FlattenResourceAttribute("iis.site", "site"),
 					otel.FlattenResourceAttribute("iis.application_pool", "app_pool"),
+					otel.SetScopeName("agent.googleapis.com/"+r.Type()),
+					otel.SetScopeVersion("2.0"),
 				),
 				// Drop all resource keys; Must be done in a separate transform,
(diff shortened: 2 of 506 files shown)


ridwanmsharif:
failures seem like flakes. Retrying now

apps/active_directory_ds.go
```diff
@@ -45,7 +46,14 @@ func (r MetricsReceiverActiveDirectoryDS) Pipelines(_ context.Context) ([]otel.R
 			otel.MetricsTransform(
 				otel.AddPrefix("workload.googleapis.com"),
 			),
-			otel.ModifyInstrumentationScope(r.Type(), "1.0"),
+			otel.Transform(
```
franciscovalentecastro: Why not just update the implementation of `ModifyInstrumentationScope` with the new use of the `otel.Transform` processor ?

IMO this would be helpful in the long run to reduce code duplication, reduce tech debt and the overall maintenance of 3P app receiver implementations. Also, it will identify this processor instance for future readers with purpose of "changing the instrumentation scope".

EdwardJXLi: cc @ridwanmsharif @avilevy18, I think we discussed this and the idea was deprecating `ModifyInstrumentationScope` entirely would decrease the surface area of `otel.processors`, since `ModifyInstrumentationScope` and `Transform` are almost identical anyways

apps/iis.go
```diff
@@ -49,6 +49,8 @@ func (r MetricsReceiverIis) Pipelines(_ context.Context) ([]otel.ReceiverPipelin
 				otel.TransformationMetrics(
 					otel.FlattenResourceAttribute("iis.site", "site"),
 					otel.FlattenResourceAttribute("iis.application_pool", "app_pool"),
+					otel.SetScopeName("agent.googleapis.com/"+r.Type()),
```
avilevy18, line 52: You already added `SetScopeName` and `SetScopeVersion` on the lines 141 - 144

apps/active_directory_ds.go
EdwardJXLi: Heads up, ended up changing the strategy to use the existing `otel.TransformationMetrics` processor instead.

apps/iis.go
EdwardJXLi: They should be different iirc. One is for the IISv1 receiver and the other is for the IISv2 receiver

avilevy18: Yup. Missed that, sorry

apps/active_directory_ds.go
franciscovalentecastro: Sounds good! The new use of the existing `otel.TransformationMetrics` to set `Scope` looks good to me.

PR #1968, opened 2025-07-03
[transformation_test] Add possibility to test a `receiver` and verify transformations defined within.

franciscovalentecastro:
## Description
Currently "transformations" done as post processing of receiver (e.g. `windows_event_log`, 3P app receivers) can't be tested directly in `transformation tests`, since it only accepts a list of processors. 

This PR adds the possibility to test a `receiver` in the `config.yaml` file of a transformation test. The test will "substitute" the "INPUT" part of a `test_receiver` from the pipeline with a test `files` log receiver while keeping the "post-input" logging processors in the pipeline. The `config.yaml` should also provide an `input_log_parser` to determine how the `input.log` is going to be parsed to mock the "INPUT" data.

Sample `config.yaml` :
```
receiver:
  test_receiver:
    type: windows_event_log
  input_log_parser:
    type: parse_json
    time_key: system_time
    time_format: "%Y-%m-%dT%H:%M:%S.%L%z"
```

## Related issue
b/429241517

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/transformation_test/transformation_test.go
+++ b/transformation_test/transformation_test.go
@@ -97,7 +97,7 @@ var (
 )
 
 type transformationTest struct {
-	Receiver   *testReceiver      `yaml:"receiver,omitempty"`
+	Receiver   *receiverTest      `yaml:"receiver,omitempty"`
 	Processors []loggingProcessor `yaml:"processors,omitempty"`
 }
 
@@ -109,9 +109,13 @@ type loggingProcessor struct {
 	confgenerator.LoggingProcessor
 }
 
-type testReceiver struct {
-	InputReceiver loggingReceiver  `yaml:"input_receiver"`
-	InputParser   loggingProcessor `yaml:"input_parser"`
+// receiverTest describes a test to verify the correctnes of the logging processors
+// implemented within a logging receiver.
+// The test will mock an input log using the provided `input.log` and `input_log_parser`.
+// All the logging processors in `test_receiver` will be added to the pipeline.
+type receiverTest struct {
+	TestReceiver   loggingReceiver  `yaml:"test_receiver"`
+	InputLogParser loggingProcessor `yaml:"input_log_parser"`
 }
 
 func (l *loggingReceiver) UnmarshalYAML(ctx context.Context, unmarshal func(interface{}) error) error {
@@ -300,19 +304,19 @@ func generateFluentBitConfigs(ctx context.Context, name string, transformationTe
 
 	// Only one (or empty) `Receiver` can be set in a transformation test.
 	if transformationTest.Receiver != nil {
-		if transformationTest.Receiver.InputParser.LoggingProcessor == nil {
-			return nil, fmt.Errorf("input_parser is required with input_receiver.")
+		if transformationTest.Receiver.InputLogParser.LoggingProcessor == nil {
+			return nil, fmt.Errorf("input_log_parser is required with test_receiver.")
 		}
 		// Append parser to mock an "input" receiver output.
-		inputParser := transformationTest.Receiver.InputParser.LoggingProcessor
-		components = append(components, inputParser.Components(ctx, flbTag, strconv.Itoa(uidComponentsIndex))...)
+		inputLogParser := transformationTest.Receiver.InputLogParser.LoggingProcessor
+		components = append(components, inputLogParser.Components(ctx, flbTag, strconv.Itoa(uidComponentsIndex))...)
 		uidComponentsIndex += 1
 
-		if transformationTest.Receiver.InputReceiver.LoggingReceiver == nil {
-			return nil, fmt.Errorf("input_receiver is required.")
+		if transformationTest.Receiver.TestReceiver.LoggingReceiver == nil {
+			return nil, fmt.Errorf("test_receiver is required.")
 		}
 		// Append logging processors from receiver "post-processing" step.
-		receiverComponents := transformationTest.Receiver.InputReceiver.LoggingReceiver.Components(ctx, flbTag)
+		receiverComponents := transformationTest.Receiver.TestReceiver.LoggingReceiver.Components(ctx, flbTag)
 		for _, rComponent := range receiverComponents {
 			// We don't add the "INPUT" component to the pipeline.
 			if rComponent.Kind != flbINPUT {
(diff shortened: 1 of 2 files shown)


transformation_test/transformation_test.go
```diff
@@ -62,7 +63,58 @@ var (
 	otelopscolPath = flag.String("otelopscol", os.Getenv("OTELOPSCOL"), "path to otelopscol")
 )
 
-type transformationTest []loggingProcessor
+var (
+	emptyLoggingReceiver = loggingReceiver{}
```
avilevy18: Can we turn this into a constant? Not sure if its possible in go

transformation_test/transformation_test.go
```diff
@@ -62,7 +63,58 @@ var (
...
+			"Setup",
+			"System"},
+		HostInfo: &host.InfoStat{
+			OS:              "windows",
+			Platform:        "win_platform",
+			PlatformVersion: "win_platform_version",
+		},
+		TestGCEResourceOverride: testResource,
+	}
+)
```
avilevy18: Can we serialize this into a yaml file, that can be declared optionally in each test?

Something like, `resource.yaml`

transformation_test/transformation_test.go
```diff
@@ -239,8 +296,20 @@ func generateFluentBitConfigs(ctx context.Context, name string, transformationTe
 
 	components = append(components, input)
```
avilevy18, line 304: Nit: add comment that there always be only 1 receiver per test.

transformation_test/transformation_test.go
```diff
@@ -239,8 +296,20 @@ func generateFluentBitConfigs(ctx context.Context, name string, transformationTe
...
 
-	for i, t := range transformationTest {
-		components = append(components, t.Components(ctx, flbTag, strconv.Itoa(i))...)
+	if transformationTest.Receiver != emptyLoggingReceiver {
+		parseJsonProcessor := confgenerator.LoggingProcessorParseJson{}
+		components = append(components, parseJsonProcessor.Components(ctx, flbTag, strconv.Itoa(0))...)
+
+		receiverComponents := transformationTest.Receiver.LoggingReceiver.Components(ctx, flbTag)
+		for _, rComponent := range receiverComponents {
+			if rComponent.Kind != "INPUT" {
```
avilevy18: NIT: Prefer const instead of `"INPUT"`

transformation_test/transformation_test.go
```diff
@@ -291,7 +360,28 @@ func (transformationConfig transformationTest) generateOTelConfig(ctx context.Co
 		t.Fatal(err)
 	}
 	var components []otel.Component
-	for _, p := range transformationConfig {
+	if transformationConfig.Receiver != emptyLoggingReceiver {
```
avilevy18: NIT: Add comment that there always be 1 receiver, per test

transformation_test/transformation_test.go
franciscovalentecastro: I have considered serializing in a `resource.yaml` format, but since the use case is limited to choosing `linux` or `windows` platforms, i don't think it's worth it to implement here :  

Some details on this : 
- The purpose of mocking a platform here is to unmarshall the config correctly since "Windows" receiver types are not supported in "Linux". There is no other direct usage of this platform.
- The intention is to the "processors" (transformations) set within a "receiver" which are always platform agnostic.
- Setting this `platform` constants is merely a mirror of how `confgenerator_test.go` sets constants for platforms  which doesn't use an artifact like a `resource.yaml` file. 

https://github.com/GoogleCloudPlatform/ops-agent/blob/951f3dcf5b1f3b937c3914472a841eafb7d7de3c/confgenerator/confgenerator_test.go#L54-L147

transformation_test/transformation_test.go
franciscovalentecastro: Addressed in https://github.com/GoogleCloudPlatform/ops-agent/pull/1968/commits/fd41bfaaa501cc7a540c695b129424e2e59f133a. I've also added "test" validation for this. Done!

transformation_test/transformation_test.go
franciscovalentecastro: Addressed in https://github.com/GoogleCloudPlatform/ops-agent/pull/1968/commits/fd41bfaaa501cc7a540c695b129424e2e59f133a. I've also added "test" validation for this. Done!

transformation_test/transformation_test.go
franciscovalentecastro: Added a `const flbINPUT = "INPUT"` for this comparison. See https://github.com/GoogleCloudPlatform/ops-agent/pull/1968/commits/006a9d3077417065677dd8aaae8d3f4fa4aaefd6 . Done!

transformation_test/transformation_test.go
franciscovalentecastro: Now the `transformationTest.Receiver` is a pointer type, so we can just check for `transformationTest.Receiver == nil` instead of checking for an empty struct. Done!

avilevy18 (approved):

transformation_test/transformation_test.go
```diff
@@ -238,9 +300,34 @@ func generateFluentBitConfigs(ctx context.Context, name string, transformationTe
...
+		components = append(components, inputLogParser.Components(ctx, flbTag, strconv.Itoa(uidComponentsIndex))...)
+		uidComponentsIndex += 1
 
-	for i, t := range transformationTest {
-		components = append(components, t.Components(ctx, flbTag, strconv.Itoa(i))...)
+		if transformationTest.Receiver.TestReceiver.LoggingReceiver == nil {
+			return nil, fmt.Errorf("test_receiver is required.")
+		}
+		// Append logging processors from receiver "post-processing" step.
+		receiverComponents := transformationTest.Receiver.TestReceiver.LoggingReceiver.Components(ctx, flbTag)
```
quentinmit, line 319: Manipulating the otel/fluent-bit components after they're already constructed feels like a giant hack. Can't we achieve the same result without any hacks by refactoring the receiver-under-test into a receiver + processor that is only registered in tests, and then using that processor in a transformation test?

franciscovalentecastro: Thanks for the comment! I like the idea of registering `processor`s only for the purpose of testing. This would avoid any need to update `transformation_test.go`. The current change present here seems to need further discussion in the design. I'll close this PR in favour of : https://github.com/GoogleCloudPlatform/ops-agent/pull/1977 .

We can discuss later if adding a `receiver` section is still needed.

franciscovalentecastro:
Closing in favour of https://github.com/GoogleCloudPlatform/ops-agent/pull/1977.

PR #1974, opened 2025-07-10
Windows Binary Signing

avilevy18:
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


kokoro/scripts/build/signing/sign.ps1
```diff
@@ -0,0 +1,47 @@
...
+$files_to_sign = @(
+    'out/bin/fluent-bit.exe',
+    'out/bin/fluent-bit.dll',
+    'out/bin/google-cloud-ops-agent.exe',
+    'out/bin/google-cloud-ops-agent-diagnostics.exe',
+    'out/bin/google-cloud-ops-agent-wrapper.exe',
+    'pkg/goo/maint.ps1'
+)
+
+$amd64_path = 'out/bin/google-cloud-metrics-agent_windows_amd64.exe'
```
jefferbrecht: Any chance we can pipe in the arch to simplify these checks away?

kokoro/scripts/build/signing/sign.ps1
```diff
@@ -0,0 +1,47 @@
...
+
+robocopy "${env:KOKORO_GFILE_DIR}\result" "${env:KOKORO_ARTIFACTS_DIR}\result" /E
+
+Set-Location "${env:KOKORO_ARTIFACTS_DIR}\result"
+
+$timestamp_server = 'http://timestamp.digicert.com'
+Write-Host "Using the timestamp server: '$timestamp_server'"
+
+$files_to_sign = @(
+    'out/bin/fluent-bit.exe',
```
jefferbrecht: Nit: use consistent separators

kokoro/scripts/build/signing/sign.ps1
```diff
@@ -0,0 +1,47 @@
...
+
+# Check which file exists and add it to the list.
+if (Test-Path $amd64_path) {
+    $files_to_sign += $amd64_path
+}
+elseif (Test-Path $x86_path) {
+    $files_to_sign += $x86_path
+}
+else {
+    Throw "ERROR: Could not find the Google Cloud Metrics Agent executable. Checked for '$amd64_path' and '$x86_path'."
```
jefferbrecht: Nit: `throw`

kokoro/scripts/build/signing/sign.ps1
```diff
@@ -0,0 +1,47 @@
...
+    $out = Invoke-Expression $sign_command
+    Write-Host -separator "`n" $out
+    Write-Host "Exit code: $LastExitCode"
+    Write-Host "Verifying: $verify_command"
+    $out = Invoke-Expression $verify_command
+    Write-Host -separator "`n" $out
+    Write-Host "Exit code: $LastExitCode"
+}
+
+exit $LastExitCode
```
jefferbrecht: Shouldn't we return a non-zero code if _any_ of the files failed to sign/verify?

kokoro/scripts/build/signing/sign_windows.sh
jefferbrecht, line 1: Where does this get called?

pkg/goo/presubmit.goospec
jefferbrecht, line 1: Why do we need a separate copy of the goospec?

pkg/goo/google-cloud-ops-agent.goospec
```diff
@@ -12,15 +12,15 @@
...
-    "out\\bin\\msvcp140.dll": "{{$INSTALL_DIR}}/bin/msvcp140.dll",
-    "out\\bin\\vccorlib140.dll": "{{$INSTALL_DIR}}/bin/vccorlib140.dll",
-    "out\\bin\\vcruntime140.dll": "{{$INSTALL_DIR}}/bin/vcruntime140.dll",
-    "out\\bin\\fluent-bit.exe": "{{$INSTALL_DIR}}/bin/fluent-bit.exe",
-    "out\\bin\\fluent-bit.dll": "{{$INSTALL_DIR}}/bin/fluent-bit.dll",
-    "out\\bin\\opentelemetry-java-contrib-jmx-metrics.jar": "{{$INSTALL_DIR}}/bin/opentelemetry-java-contrib-jmx-metrics.jar",
-    "out\\bin\\google-cloud-metrics-agent_{{$GOOS}}_{{$GOARCH}}.exe": "{{$INSTALL_DIR}}/bin/google-cloud-metrics-agent_{{$GOOS}}_{{$GOARCH}}.exe",
-    "out\\bin\\google-cloud-ops-agent.exe": "{{$INSTALL_DIR}}/bin/google-cloud-ops-agent.exe",
-    "out\\bin\\google-cloud-ops-agent-wrapper.exe": "{{$INSTALL_DIR}}/bin/google-cloud-ops-agent-wrapper.exe"
+    "out/bin/msvcp140.dll": "{{$INSTALL_DIR}}/bin/msvcp140.dll",
```
jefferbrecht: Why change the separators here (but not in the presubmit copy)?

kokoro/scripts/build/packaging/package_windows.sh
```diff
@@ -0,0 +1,73 @@
...
+      ;;
+esac
+
+echo "Starting build for Arch: ${Arch} (GoArch: ${GoArch})"
+
+# The working directory is now set by the calling script.
+git config --global --add safe.directory "$(pwd)"
+
+echo "Installing goopack..."
+go install -trimpath -ldflags="-s -w" github.com/google/googet/v2/goopack@latest
```
jefferbrecht: Note the warning in https://github.com/GoogleCloudPlatform/ops-agent/blob/40e46c49c30ffdd1ee3c9257fc553f011ec8f036/Dockerfile.windows#L191

kokoro/scripts/build/signing/sign.ps1
avilevy18: there is alot more work involved with it, I have switch to searching for the executable through regex instead.

kokoro/scripts/build/signing/sign.ps1
```diff
@@ -0,0 +1,64 @@
...
+$files_to_sign = @(
+    'out/bin/fluent-bit.exe',
+    'out/bin/fluent-bit.dll',
+    'out/bin/google-cloud-ops-agent.exe',
+    'out/bin/google-cloud-ops-agent-wrapper.exe',
+    'pkg/goo/maint.ps1'
+)
+
+$metrics_agent_files = Get-ChildItem -Path 'out/bin/' -Filter 'google-cloud-metrics-agent_windows_*.exe' -ErrorAction SilentlyContinue |
+    Where-Object { $_.Name -match '^google-cloud-metrics-agent_windows_(amd64|386)\.exe$' }
```
jefferbrecht: What I mean is that we're defining a kokoro config under a subdirectory named `x86_64_windows`, which is already arch-specific. So there's already a spot where we can define a variable for the arch and avoid having to search for it.

On the other hand, searching is fine too, but as long as we're searching at all then why even require an explicit file list? Is there a scenario where we wouldn't want to sign every executable in `out/bin/` (+ `pkg/goo/maint.ps1`)? Doing that would make it robust against future file renames too, for example.

kokoro/scripts/build/build.ps1
```diff
@@ -93,16 +93,23 @@ if (($env:KOKORO_ROOT_JOB_TYPE -eq 'CONTINUOUS_INTEGRATION') -or ($env:KOKORO_JO
...
-Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out/*.goo" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
-# Copy the .pdb and .dll files from $env:KOKORO_ARTIFACTS_DIR/out/bin to $env:KOKORO_ARTIFACTS_DIR/result.
-# The .pdb and .dll files are saved so the team can use them in the event that we have to debug this Ops Agent build. 
-# They are not distributed to customers.
-Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/*.pdb" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
-Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/*.dll" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
 # Copy Ops Agent UAP Plugin tarball to the result directory.
-(Get-FileHash -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/google-cloud-ops-agent-plugin*.tar.gz" -Algorithm SHA256).Hash.ToLower() | Out-File -FilePath "$env:KOKORO_ARTIFACTS_DIR/result/google-cloud-ops-agent-plugin-sha256.txt" -Encoding ascii
+(Get-FileHash -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/google-cloud-ops-agent-plugin*.tar.gz" -Algorithm SHA256).Hash | Out-File -FilePath "$env:KOKORO_ARTIFACTS_DIR/result/google-cloud-ops-agent-plugin-sha256.txt" -Encoding ascii
```
jefferbrecht: Why remove `.ToLower()`? I vaguely remember that it needed to be strictly lower-case for some reason. CC @jinghan-ma

kokoro/scripts/build/build.ps1
```diff
@@ -93,16 +93,23 @@ if (($env:KOKORO_ROOT_JOB_TYPE -eq 'CONTINUOUS_INTEGRATION') -or ($env:KOKORO_JO
...
 Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/google-cloud-ops-agent-plugin*.tar.gz" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
 
+Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
+Move-Item -Path "./pkg" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
+
+if ($env:PACKAGE -ne $null) {
```
jefferbrecht: We don't need this check: if `PACKAGE` isn't defined, then there won't be a .goo and `Move-Item` will simply move zero files, which achieves the same result.

The check is wrong anyway since it doesn't match the check in `Dockerfile.windows` (e.g. any defined value other than `true` will cause the check to pass even though there wouldn't be a .goo package).

kokoro/scripts/build/build.ps1
```diff
@@ -93,16 +93,23 @@ if (($env:KOKORO_ROOT_JOB_TYPE -eq 'CONTINUOUS_INTEGRATION') -or ($env:KOKORO_JO
...
-# They are not distributed to customers.
-Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/*.pdb" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
-Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/*.dll" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
 # Copy Ops Agent UAP Plugin tarball to the result directory.
-(Get-FileHash -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/google-cloud-ops-agent-plugin*.tar.gz" -Algorithm SHA256).Hash.ToLower() | Out-File -FilePath "$env:KOKORO_ARTIFACTS_DIR/result/google-cloud-ops-agent-plugin-sha256.txt" -Encoding ascii
+(Get-FileHash -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/google-cloud-ops-agent-plugin*.tar.gz" -Algorithm SHA256).Hash | Out-File -FilePath "$env:KOKORO_ARTIFACTS_DIR/result/google-cloud-ops-agent-plugin-sha256.txt" -Encoding ascii
 Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/google-cloud-ops-agent-plugin*.tar.gz" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
 
+Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
+Move-Item -Path "./pkg" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
```
jefferbrecht: Can you add a comment for this? I assume we're moving `pkg` because we want to sign `pkg/goo/build.ps1` later?

Why not just move `pkg/goo/*.ps1`? (And maybe it should be a copy instead of a move so that we're not mutating the source tree...)

kokoro/scripts/build/build.ps1
```diff
@@ -93,16 +93,23 @@ if (($env:KOKORO_ROOT_JOB_TYPE -eq 'CONTINUOUS_INTEGRATION') -or ($env:KOKORO_JO
...
-# The .pdb and .dll files are saved so the team can use them in the event that we have to debug this Ops Agent build. 
-# They are not distributed to customers.
-Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/*.pdb" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
-Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/*.dll" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
 # Copy Ops Agent UAP Plugin tarball to the result directory.
-(Get-FileHash -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/google-cloud-ops-agent-plugin*.tar.gz" -Algorithm SHA256).Hash.ToLower() | Out-File -FilePath "$env:KOKORO_ARTIFACTS_DIR/result/google-cloud-ops-agent-plugin-sha256.txt" -Encoding ascii
+(Get-FileHash -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/google-cloud-ops-agent-plugin*.tar.gz" -Algorithm SHA256).Hash | Out-File -FilePath "$env:KOKORO_ARTIFACTS_DIR/result/google-cloud-ops-agent-plugin-sha256.txt" -Encoding ascii
 Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out/bin/google-cloud-ops-agent-plugin*.tar.gz" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
 
+Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/out" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
```
jefferbrecht: Why are we moving over the entire `out` directory _and_ selectively moving over individual artifacts too?

kokoro/scripts/build/build.ps1
```diff
@@ -93,16 +93,23 @@ if (($env:KOKORO_ROOT_JOB_TYPE -eq 'CONTINUOUS_INTEGRATION') -or ($env:KOKORO_JO
...
+Move-Item -Path "./pkg" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
+
+if ($env:PACKAGE -ne $null) {
+  Move-Item -Path "$env:KOKORO_ARTIFACTS_DIR/result/out/*.goo" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
+}
+
+# Copy the .pdb and .dll files from $env:KOKORO_ARTIFACTS_DIR/out/bin to $env:KOKORO_ARTIFACTS_DIR/result.
+# The .pdb and .dll files are saved so the team can use them in the event that we have to debug this Ops Agent build.
+# They are not distributed to customers.
+Copy-Item -Path "$env:KOKORO_ARTIFACTS_DIR/result/out/bin/*.pdb" -Destination "$env:KOKORO_ARTIFACTS_DIR/result"
```
jefferbrecht: Why are these copies while the others are moves?

kokoro/scripts/build/packaging/package_windows.sh
```diff
@@ -0,0 +1,73 @@
...
+
+echo "Preparing directories and files..."
+mkdir -p out
+mkdir -p "$OutputDir"
+
+# Move pre-built files from the input directory into the current structure.
+mv "${InputDir}/result/out/"* ./out/
+mv "${InputDir}/result/pkg/goo/maint.ps1" ./pkg/goo/maint.ps1
```
jefferbrecht: We should not be overwriting the source tree. The goospec needs to change: it previously assumed we'd always be packaging the original source file, but that's no longer true. It should be treated as a build artifact now. This should also help simplify away some of the special-casing that's been made around it.

pkg/goo/presubmit.goospec
jefferbrecht: Talked offline that this is due to the path separator issue + the fact that it's sometimes invoked from a Linux worker and sometimes from a Windows one.

I'm not a fan of having two copies of this and would prefer we fix this in one of two ways:

1) Change the Linux worker to a Windows one, or
2) Revert back to the older version of goopack that doesn't have the path separator issue

pkg/goo/google-cloud-ops-agent.goospec
avilevy18: reverted to a single goospec.

kokoro/scripts/build/build.ps1
avilevy18: Spoke offline, decided to move maint to `out/bin/` instead. Since its the only thing that needs to be signed.

kokoro/scripts/build/build.ps1
avilevy18: Updated comments

kokoro/scripts/build/packaging/package_windows.sh
avilevy18: everything is inside the `out` directory now. Source tree will remain unaffected.

kokoro/scripts/build/signing/sign.ps1
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    # 5. Verify the signature
+    $verify_command = "& signtool.exe verify /pa /all `"$file_path`" 2>&1"
+    Write-Host "Verifying: $verify_command"
+
+    $out = Invoke-Expression $verify_command
+    $verify_code = $LastExitCode
+    Write-Host -separator "`n" $out
+    Write-Host "Exit code (Verify): $verify_code"
```
jefferbrecht: Nit: delete this line, the exit code is only interesting when it's non-zero, which we're outputting in the next block anyway

kokoro/scripts/build/packaging/package.ps1
```diff
@@ -0,0 +1,148 @@
...
+    # Determine GOPATH
+    $GOPATH = $env:GOPATH
+    if ([string]::IsNullOrWhiteSpace($GOPATH)) {
+        $GOPATH = "$env:USERPROFILE\go"
+    }
+
+    # Run goopack
+    $GoopackPath = "$GOPATH\bin\goopack.exe"
+
+    & $GoopackPath -output_dir "$OutputDir" `
```
jefferbrecht: Any chance we can reuse `pkg/goo/build.ps1`?

kokoro/scripts/build/packaging/package.ps1
```diff
@@ -0,0 +1,119 @@
...
+
+# 4. Change Directory
+Write-Host "Changing directory to git/unified_agents..."
+$TargetDir = Join-Path (Get-Location) "git\unified_agents"
+
+if (Test-Path $TargetDir) {
+    Set-Location $TargetDir
+} else {
+    # Fallback if running relative to root already or path structure differs
+    if (Test-Path "git/unified_agents") { Set-Location "git/unified_agents" }
```
jefferbrecht: Why have a fallback at all here? This signals that we have uncertainty about something that should be hard-coded and deterministic: we should either be in the correct directory already or not.

I'm also not seeing how the fallback is any different from the base case: we haven't called `Set-Location` yet up until this point, so shouldn't `Join-Path (Get-Location) "git\unified_agents"` resolve to the same thing as `git/unified_agents`?

kokoro/scripts/build/signing/sign.ps1
```diff
@@ -0,0 +1,76 @@
...
+
+Set-Location "${env:KOKORO_ARTIFACTS_DIR}/result"
+
+$timestamp_server = 'http://timestamp.digicert.com'
+Write-Host "Using the timestamp server: '$timestamp_server'"
+
+# 1. Start with the specific manual files
+$files_to_process = @()
+
+# 2. Add all .exe and .dll files found in out/bin
```
jefferbrecht: Nit: update comment with `ps1`

PR #1978, opened 2025-07-16
feat: Update nginx receiver to use `LoggingReceiverMacro`

Caleb-Hurshman:
## Description

Refactor the nginx receiver logging to be sub-agent agnostic.
Draft stage, creating this PR to run transformation tests until they are running locally.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


.github/workflows/stale.yml
```diff
@@ -1,3 +1,17 @@
+# Copyright 2025 Google LLC
```
franciscovalentecastro: After the seeing how many files are updated by `make addlicense`, i think it would be better to do this in a separate PR. I'll create it ( https://github.com/GoogleCloudPlatform/ops-agent/pull/1984 ) and work on merging this updates separately.

Let's revert the `addlicense` changes in this PR.

apps/nginx.go
```diff
@@ -123,42 +139,35 @@ func (p LoggingProcessorNginxError) Components(ctx context.Context, tag string,
...
+func loggingReceiverFilesMixinNginxAccess() confgenerator.LoggingReceiverFilesMixin {
+	return confgenerator.LoggingReceiverFilesMixin{
+		IncludePaths: []string{"/var/log/nginx/access.log"},
 	}
-	c := r.ReceiverMixin.Components(ctx, tag)
-	c = append(c, r.LoggingProcessorNginxAccess.Components(ctx, tag, "nginx_access")...)
-	return c
 }
 
 type LoggingReceiverNginxError struct {
```
franciscovalentecastro: The struct `LoggingReceiverNginxError` can be completely removed now since `RegisterLoggingFilesProcessorMacro` with the argument `loggingReceiverFilesMixinNginxError` will take care of registering a `NginxError` receiver.

apps/nginx.go
```diff
@@ -123,42 +139,35 @@ func (p LoggingProcessorNginxError) Components(ctx context.Context, tag string,
...
 				InstrumentationSourceLabel: instrumentationSourceValue(p.Type()),
 			},
-		}.Components(ctx, tag, uid)...,
-	)
-	return c
+		},
+	}
 }
 
 type LoggingReceiverNginxAccess struct {
```
franciscovalentecastro: The struct `LoggingReceiverNginxAccess` can be completely removed now since `RegisterLoggingFilesProcessorMacro` with the argument `loggingReceiverFilesMixinNginxAccess` will take care of registering a `NginxAccess` receiver.

confgenerator/testdata/goldens/logging-receiver_nginx/golden/windows-2012/fluent_bit_main.conf
```diff
@@ -171,7 +171,7 @@
     Match  nginx.nginx_access
     Name   lua
     call   process
-    script b85e520931fd4cdd39bd5a9ab1699d1b.lua
+    script 4a885f78c2b41df75250528bf3ea0c2c.lua
```
franciscovalentecastro: As mentioned in another comment, `fluent_bit_main.conf` files should not change during the 3P app receiver update to use `Logging{Receiver/Processor}Macro`.

confgenerator/testdata/goldens/logging-receiver_nginx/golden/windows/4a885f78c2b41df75250528bf3ea0c2c.lua
```diff
@@ -0,0 +1,12 @@
+
+function process(tag, timestamp, record)
```
franciscovalentecastro: As mentioned in another comment, `*.lua` files should not change during the 3P app receiver update to use `Logging{Receiver/Processor}Macro`. We can also see a difference between the newly generated `.lua` file compared to the old one in the same folder.

apps/nginx.go
```diff
@@ -65,46 +64,63 @@ func init() {
...
 
-func (LoggingProcessorNginxAccess) Type() string {
+func (LoggingProcessorMacroNginxAccess) Type() string {
 	return "nginx_access"
 }
 
-func (p LoggingProcessorNginxAccess) Components(ctx context.Context, tag string, uid string) []fluentbit.Component {
-	return genericAccessLogParser(ctx, p.Type(), tag, uid)
+func (p LoggingProcessorMacroNginxAccess) Expand(ctx context.Context) []confgenerator.InternalLoggingProcessor {
+	return []confgenerator.InternalLoggingProcessor{
```
franciscovalentecastro: ### Missing Processors

The current implementation of the Logging Processors used for `NginxAccess` logs post-processing is not equivalent to the implementation in `genericAccessLogParser`. There are several processors missing from `genericAccessLogParser`.

We can verify this difference after seeing changes in files like `fluent_bit_main.conf` or `*.lua`. These are "golden" files describing the `fluent-bit` config representation of the `nginx` receiver. When updating any of the 3P app receivers none of these files ( `fluent_bit_main.conf`, `*.lua`, `otel.yaml`) should change. The update to use `LoggingReceiverMacro` should preserve `fluent-bit / otel` configs  to keep the same receiver behaviour.

_Note : There may be a rare case when `fluent_bit_main.conf` may need to change, but not in this particular case of the `nginx` receiver._

### Recommended Solution 

My recommended solution is to update the function `genericAccessLogParser` to the following new signature and keep using it to define `LoggingProcessorMacroNginxAccess` : 
```
func genericAccessLogParser(ctx context.Context, processorType, tag, uid string) []confgenerator.InternalLoggingProcessor {
   ... 
}
```

There are several other receivers [^1] (`jetty`, `couchbase`, `varnish`, `apache`, `tomcat`)  that use `genericAccessLogParser`. Since this function is used in many places and reduces a lot of code duplication, IMO preserving `genericAccessLogParser` is important.

I see two alternatives :
1. Updating all 6 receivers in the same PR, which is more straightforward, but involves a bigger PR,
2. Having two versions of the `genericAccessLogParser` with different signatures during the migration and remove the "old" one when its finished.

[^1]: https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fops-agent%20genericAccessLogParser&type=code

confgenerator/testdata/feature/golden.csv
```diff
@@ -265,5 +255,11 @@ App,Field,Override,
 *confgenerator.PrometheusMetrics,config.[].scrape_configs.scrape_timeout
 *confgenerator.PrometheusMetrics,config.[].scrape_configs.static_config_target_groups
 *confgenerator.loggingProcessorMacroAdapter[github.com/GoogleCloudPlatform/ops-agent/apps.LoggingProcessorMacroFlink],confgenerator.ConfigComponent.Type,
+*confgenerator.loggingProcessorMacroAdapter[github.com/GoogleCloudPlatform/ops-agent/apps.LoggingProcessorMacroNginxAccess],confgenerator.ConfigComponent.Type,
```
franciscovalentecastro, line 258: Note : This is just a comment to say that updates in files `golden.csv`, `feature_tracking_otlp.json` and `features.yaml` are expected when updating a receiver.

transformation_test/testdata/logging_processor-nginx-access/output_fluentbit.yaml
```diff
@@ -0,0 +1,123 @@
+- entries:
```
franciscovalentecastro, line 1: Note: One way of verifying the resulting updated implementation of the `ngnix_access` (or any other) receiver is working as expected is that `make transformation_test` should produce exactly the same `output_fluentbit.yaml` before and after the code update to `nginx.go`.

No change in `output_fluentbit.yaml` implies that the behaviour of the receiver/processors is preserved.

Caleb-Hurshman: Very good to know!

confgenerator/testdata/goldens/logging-receiver_nginx/golden/windows-2012/fluent_bit_main.conf
Caleb-Hurshman: I believe the changes to these files were automatically introduced by running `make test_confgenerator_update`. I'll try again after resolving some of the feedback in nginx.go

apps/nginx.go
Caleb-Hurshman: That makes sense. I like option 2 of adding a V2 for `genericAccessLogParser`, that way the individual receivers can have a small update in their respective PRs.

apps/common_logging_processors.go
```diff
@@ -88,3 +88,63 @@ func genericAccessLogParser(ctx context.Context, processorType, tag, uid string)
 	c = append(c, mf.Components(ctx, tag, uid)...)
 	return c
 }
+
```
franciscovalentecastro, line 92: Please add a comment describing the difference between this new function and a `TODO:` above `genericAccessLogParser` to "remove" when all references to it have been updated.

apps/common_logging_processors.go
```diff
@@ -88,3 +88,63 @@ func genericAccessLogParser(ctx context.Context, processorType, tag, uid string)
...
+func genericAccessLogParserV2(ctx context.Context, processorType, tag, uid string) []confgenerator.InternalLoggingProcessor {
```
franciscovalentecastro: The `genericAccessLogParserV2` naming maybe temporary during this migration, but i think it would be better to have a more descriptive naming. 

My suggestion is `genericAccessLogParserAsInternalLoggingProcessor` to describe the return type is `InternalLoggingProcessor`.

Also, let's add a `TODO:` to update the naming when the `old` version is cleaned up.

transformation_test/testdata/ops_agent_test-TestLogEntrySpecialFields/output_otel.yaml
```diff
@@ -1,74 +1,65 @@
...
     rejected_items: 1.0
-    resource:
-      service.instance.id: test-service-instance-id
-      service.name: otelopscol
-      service.version: ""
+    resource: {}
     stacktrace: |-
       go.opentelemetry.io/collector/exporter/exporterhelper/internal.(*BaseExporter).Send
-        go.opentelemetry.io/collector/exporter@v0.130.0/exporterhelper/internal/base_exporter.go:117
+        go.opentelemetry.io/collector/exporter@v0.126.0/exporterhelper/internal/base_exporter.go:116
```
franciscovalentecastro: The updates in this file point out that it was run with an `otelopscol` build that was using `Opentelemetry` version `v0.126.0`. The update in the Ops Agent to  use`v0.130.0` is very recent (https://github.com/GoogleCloudPlatform/ops-agent/pull/1975) so the build may still be using the `v0.126.0` version.

One command that could help is `make rebuild_submodules` to make sure `otelopscol` is built with the latest version.

apps/nginx.go
franciscovalentecastro: Sounds good! I added a separate comment on the naming of the new function `genericAccessLogParserV2`. I will resolve this thread.

apps/common_logging_processors.go
```diff
@@ -88,3 +89,66 @@ func genericAccessLogParser(ctx context.Context, processorType, tag, uid string)
 	c = append(c, mf.Components(ctx, tag, uid)...)
 	return c
 }
+
+// TODO: rename to genericAccessLogParser once the old version is removed
+// genericAccessLogParserAsInternalLoggingProcessor is an internal logging processor that parses access logs.
+// It will eventually replace genericAccessLogParser, as it returns a slice of InternalLoggingProcessor rather than fluentbit.Component, making it more flexible.
+func genericAccessLogParserAsInternalLoggingProcessor(ctx context.Context, processorType, tag, uid string) []confgenerator.InternalLoggingProcessor {
```
franciscovalentecastro: Ohh! I missed this before, `tag` and `uid` are fluent-bit specific parameters which are not used in the new implementation. They can be removed now.

franciscovalentecastro:
This PR is looking good. Thank you for all the updates! I added one last comment. Let's also wait for integration tests to run to see if everything is working correctly.

apps/common_logging_processors.go
```diff
@@ -88,3 +89,66 @@ func genericAccessLogParser(ctx context.Context, processorType, tag, uid string)
 	c = append(c, mf.Components(ctx, tag, uid)...)
 	return c
 }
+
+// TODO: rename to genericAccessLogParser once the old version is removed
+// genericAccessLogParserAsInternalLoggingProcessor is an internal logging processor that parses access logs.
+// It will eventually replace genericAccessLogParser, as it returns a slice of InternalLoggingProcessor rather than fluentbit.Component, making it more flexible.
+func genericAccessLogParserAsInternalLoggingProcessor(ctx context.Context, processorType string) []confgenerator.InternalLoggingProcessor {
```
franciscovalentecastro, line 96: Note: A teammate (@quentinmit) suggested an improvement of this function by converting `genericAccessLogParserAsInternalLoggingProcessor` into an `InternalLoggingProcessor` (1. Creating a struct, 2. Defining this function as its `Expand()` method).

We can do this in a followup PR when updating any of the other receivers (`apache`, `tomcat`, ...) that use this. I just wanted to keep a note on this idea.

franciscovalentecastro (approved):
LGTM! Added a `Note:` of an idea we can address in a followup PR.

franciscovalentecastro:
Mentioned in previous https://github.com/GoogleCloudPlatform/ops-agent/pull/1978#issuecomment-3097399661 about the unrelated test failures to the changes in this PR. Both failures (`Debian 11 frontend lock` and `rabbitmq` in `Ubuntu 24.04` are being addressed separately of this PR). Merging PR.

PR #1987, opened 2025-07-21
feat: Update Elasticsearch json parser to use `LoggingReceiverMacro`

dyl10s:
## Description

Refactored the Elasticsearch json parser to use the `LoggingReceiverMacro`

I did need to add some yaml tags to fix a parsing panic

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


apps/elasticsearch.go
```diff
@@ -149,38 +149,35 @@ func init() {
...
-	// Logs are formatted based on configuration (log4j);
-	// See https://artifacts.elastic.co/javadoc/org/elasticsearch/elasticsearch/7.16.2/org/elasticsearch/common/logging/ESJsonLayout.html
-	// for general layout, and https://www.elastic.co/guide/en/elasticsearch/reference/current/logging.html for general configuration of logging
-
-	jsonParser := &confgenerator.LoggingProcessorParseJson{
-		ParserShared: confgenerator.ParserShared{
-			TimeKey:    "timestamp",
-			TimeFormat: "%Y-%m-%dT%H:%M:%S,%L%z",
+func (p LoggingProcessorMacroElasticsearchJson) Expand(ctx context.Context) []confgenerator.InternalLoggingProcessor {
+	// // sample log line:
```
franciscovalentecastro: Nit: Remove double comments.

apps/elasticsearch.go
```diff
@@ -332,6 +319,6 @@ func (r LoggingReceiverElasticsearchGC) Components(ctx context.Context, tag stri
 }
 
 func init() {
-	confgenerator.LoggingReceiverTypes.RegisterType(func() confgenerator.LoggingReceiver { return &LoggingReceiverElasticsearchJson{} })
```
franciscovalentecastro, line 335: The `ElasticsearchJson` receiver seems to be a special case where the `Processor` is not registered to be externally exposed to users. This means users cannot use `elasticsearch_json` as a processor, only as a receiver.

The most probable reason for this is that it uses `MultilineRules` which need to defined both in the `Receiver` and `Processor` (fluent-bit requires this to avoid separating multi-line logs in different memory "chunks" ). Using the `Processor` separately would break this. 

To keep the previous setup we would need to :
1. Register here only the `Receiver` of `Elasticsearch_json`. See https://github.com/GoogleCloudPlatform/ops-agent/pull/1950 for an example of this.

2. Register the `Processor` for testing only here :  https://github.com/GoogleCloudPlatform/ops-agent/blob/master/transformation_test/transformation_test.go#L595

dyl10s: I believe I have adjusted this correctly by just registering the `Receiver` in `elasticsearch.go` similar to the draft PR you linked and moving processor initialization into `transformation_test.go`. Let me know if I am missing anything!

apps/elasticsearch.go
```diff
@@ -271,15 +268,17 @@ func (p LoggingProcessorElasticsearchGC) Components(ctx context.Context, tag, ui
 	return c
 }
 
-type LoggingReceiverElasticsearchJson struct {
-	LoggingProcessorElasticsearchJson `yaml:",inline"`
-	ReceiverMixin                     confgenerator.LoggingReceiverFilesMixin `yaml:",inline"`
+type LoggingReceiverMacroElasticsearchJson struct {
+	LoggingProcessorMacroElasticsearchJson  `yaml:",inline"`
+	confgenerator.LoggingReceiverFilesMixin `yaml:",inline"`
```
franciscovalentecastro: Let's keep the field `LoggingReceiverFilesMixin` named as `ReceiverMixin`.

apps/elasticsearch.go
```diff
@@ -271,15 +268,17 @@ func (p LoggingProcessorElasticsearchGC) Components(ctx context.Context, tag, ui
...
+	IncludePaths                            []string                      `yaml:"include_paths,omitempty"`
+	MultilineRules                          []confgenerator.MultilineRule `yaml:"multiline_rules,omitempty"`
```
franciscovalentecastro: `IncludePaths` and `MultilineRules` are properties on the `ReceiverMixin` field of this struct, so we can still assign them as before (e.g. using `r.ReceiverMixin.IncludePaths` and `r.ReceiverMixin.MultilineRules`).

Note: In this specific case the refactor doesn't simplify the code as much since we aren't using `RegisterLoggingFilesProcessorMacro`.

confgenerator/testdata/goldens/logging-receiver_elasticsearch/golden/linux-gpu/fluent_bit_main.conf
```diff
@@ -47,22 +47,6 @@
     Tag               elasticsearch.elasticsearch_gc
```
franciscovalentecastro: Note: I believe in this particular receiver there shouldn't be any changes in `fluent_bit_main.conf`. Hopefully addressing the updates in `elasticsearch.go` should remove this changes.

apps/elasticsearch.go
franciscovalentecastro: Thank you for addressing this comment! The general idea of the update is good. I just added some comments above about the new definition of the `LoggingReceiverMacroElasticsearchJson`struct.

confgenerator/testdata/goldens/logging-receiver_elasticsearch/golden/linux-gpu/fluent_bit_main.conf
dyl10s: Yup, that did the trick, thanks for looking into that

transformation_test/transformation_test.go
```diff
@@ -33,6 +33,7 @@ import (
 	"time"
 
 	logpb "cloud.google.com/go/logging/apiv2/loggingpb"
+	"github.com/GoogleCloudPlatform/ops-agent/apps"
 	_ "github.com/GoogleCloudPlatform/ops-agent/apps"
```
franciscovalentecastro: Nit: We can now remove the import of `_ "github.com/GoogleCloudPlatform/ops-agent/apps"`, since we are actually using `github.com/GoogleCloudPlatform/ops-agent/apps` in the `transformation_test` package.

transformation_test/testdata/logging_processor-elasticsearch-json/output_fluentbit.yaml
```diff
@@ -0,0 +1,201 @@
+- entries:
```
franciscovalentecastro, line 1: Note: This is a general comment that i mentioned also in https://github.com/GoogleCloudPlatform/ops-agent/pull/1978#discussion_r2214743140. Another way of verifying that the refactor is correct is that the `output_fluentbit.yaml` should be the same before and after the update to `elasticsearch.go`.

apps/elasticsearch.go
```diff
@@ -271,12 +268,12 @@ func (p LoggingProcessorElasticsearchGC) Components(ctx context.Context, tag, ui
 	return c
 }
 
-type LoggingReceiverElasticsearchJson struct {
-	LoggingProcessorElasticsearchJson `yaml:",inline"`
-	ReceiverMixin                     confgenerator.LoggingReceiverFilesMixin `yaml:",inline"`
+type LoggingReceiverMacroElasticsearchJson struct {
+	LoggingProcessorMacroElasticsearchJson `yaml:",inline"`
+	ReceiverMixin                          confgenerator.LoggingReceiverFilesMixin `yaml:",inline"`
```
franciscovalentecastro: This change should prevent the YAML un-marshalling to validate recursively all fields in `ReceiverMixin`, thus not causing the `panic` with missing yaml tags. 

Suggested change :

```suggestion
	ReceiverMixin                          confgenerator.LoggingReceiverFilesMixin `yaml:",inline" validate:"structonly"`
```

confgenerator/logging_processors.go
```diff
@@ -320,9 +320,9 @@ func (p LoggingProcessorParseRegexComplex) Components(ctx context.Context, tag,
 }
 
 type MultilineRule struct {
-	StateName string
-	Regex     string
-	NextState string
+	StateName string `yaml:"state_name,omitempty"`
+	Regex     string `yaml:"regex,omitempty"`
+	NextState string `yaml:"next_state,omitempty"`
```
franciscovalentecastro: Sorry for bringing this comments so late in the review process. Now i'm realizing that since `MultilineRule` [^1] is a non-configurable feature (not exposed to users), these fields don't need and shouldn't have named yaml tags.

Note : I remember the updates in `elasticsearch.go` were causing some issues with a `panic (missing tags)` when running tests. I haven't been able to reproduce again, but i believe the other suggested change in `elasticsearch.go` should fix that. 

Suggesting the following change :

```suggestion
	StateName string
	Regex     string
	NextState string
```

[^1]: https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/logging_receivers.go#L74

franciscovalentecastro (changes requested):
Sorry for bringing this comments so late in the review process. Added some last suggested changes.

PR #2003, opened 2025-07-23
feat: Update Mongodb receiver to use `LoggingReceiverMacro`

dyl10s:
## Description

Refactored the Mongodb receiver to use `LoggingReceiverMacro`

A few issues came up with this refactor that I will need some guidance on:
- I needed a few more operators (lift/rename) so I implemented them to match existing functionality. Let me know if there is a better direction I can go here.
- There are multiple lua filter components here. I am not sure there exact functionality or how they need to be refactored.
- There is a lot of logic around parsing the timestamp, currently it seems to not parse correctly and I just tried to match the existing logic.
- This is similar to the `ElasticsearchJson` receiver where we just register the processor in the `transformation_test.go` file

## How has this been tested?

Validated there were no changes to the generated fluentbit golden file before and after the refactor

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


transformation_test/testdata/logging_processor-mongodb/output_fluentbit.yaml
```diff
@@ -0,0 +1,1037 @@
...
+      context: main
+      id: 23285.0
+      message: Automatically disabling TLS 1.0, to force-enable TLS 1.0 specify --sslDisabledProtocols 'none'
+      severity: I
+    labels:
+      compute.googleapis.com/resource_name: hostname
+      logging.googleapis.com/instrumentation_source: agent.googleapis.com/mongodb
+    logName: projects/my-project/logs/transformation_test
+    severity: 200.0
+    timestamp: now
```
franciscovalentecastro, line 25: Since all the log entries here seem to have `timestamp: now` (which happens when the log didn't parse a specific timestamp), there may be a bug in the timestamp parsing.

It may predate this PR, but or it could be a mismatched format. Please address this missing timestamp.

apps/mongodb.go
```diff
@@ -230,115 +192,95 @@ func (p *LoggingProcessorMongodb) severityParser(ctx context.Context, tag, uid s
...
-	c := r.ReceiverMixin.Components(ctx, tag)
-	c = append(c, r.LoggingProcessorMongodb.Components(ctx, tag, "mongodb")...)
-	return c
+
+	return &r.ReceiverMixin, r.LoggingProcessorMacroMongodb.Expand(ctx)
 }
 
 func init() {
-	confgenerator.LoggingReceiverTypes.RegisterType(func() confgenerator.LoggingReceiver { return &LoggingReceiverMongodb{} })
+	confgenerator.RegisterLoggingReceiverMacro(func() LoggingReceiverMacroMongodb {
```
franciscovalentecastro: Please register using `RegisterLoggingFilesProcessorMacro` instead. There is no need to only register the `receiver` anymore.

confgenerator/logging_processors.go
```diff
@@ -461,6 +461,58 @@ func (p LoggingProcessorNestWildcard) Components(ctx context.Context, tag, uid s
 	}
 }
 
+type LoggingProcessorRemoveField struct {
+	Field string
+}
+
+func (p LoggingProcessorRemoveField) Components(ctx context.Context, tag, uid string) []fluentbit.Component {
```
franciscovalentecastro, line 541: I've thought about this a bit and realized that the operations done by `LoggingProcessorRemoveField`, `LoggingProcessorHardRename` and `LoggingProcessorNestLift` can be replicated with `LoggingProcessorModifyFields`. Please remove this processors and using modify fields instead.

Here is a list of this operations : 
1. `LoggingProcessorRemoveField` can be replicated with :
```
confgenerator.LoggingProcessorModifyFields{
			Fields: map[string]*confgenerator.ModifyField{
				"jsonPayload.Field": {
					MoveFrom: "jsonPayload.Field",
                     OmitIf: `jsonPayload.Field =~ ".*"`
				},
			},
		}
```

2. `LoggingProcessorHardRename` can be replicated with : 
```
confgenerator.LoggingProcessorModifyFields{
			Fields: map[string]*confgenerator.ModifyField{
				"jsonPayload.dest": {
					MoveFrom: "jsonPayload.src",
				},
			},
		}
```

3. `LoggingProcessorNestLift` can be replicated with : 

```
confgenerator.LoggingProcessorModifyFields{
			Fields: map[string]*confgenerator.ModifyField{
				"jsonPayload.prefix_Field": {
					MoveFrom: "jsonPayload.nested.Field",
				},
			},
		}
```

Using `LoggingProcessorModifyFields` instead of creating processors improves a lot the `Mongodb` processor the following ways : 
- Removes tech debt of having to maintain custom `fluent-bit` configs.
- There is no need to implement each of this processors with Otel (modify fields already has both implementations).
- Improves readability of the Mongodb processor.

apps/mongodb.go
```diff
@@ -230,115 +192,95 @@ func (p *LoggingProcessorMongodb) severityParser(ctx context.Context, tag, uid s
...
+		Field: "message",
+	})
 
 	return c
 }
 
-type LoggingReceiverMongodb struct {
-	LoggingProcessorMongodb `yaml:",inline"`
-	ReceiverMixin           confgenerator.LoggingReceiverFilesMixin `yaml:",inline" validate:"structonly"`
+type LoggingReceiverMacroMongodb struct {
```
franciscovalentecastro: After registering with `RegisterLoggingFilesProcessorMacro` there is no need to have a `LoggingReceiverMacroMongodb`.

transformation_test/testdata/logging_processor-mongodb/output_fluentbit.yaml
dyl10s: In the original implementation there was some special logic to grab a specific fluentbit component from the `LoggingProcessorParseRegix` at a specific index which was just the time parsing section. Because this is now becoming more generic we can not just grab a given index. 

I created a new parser that is just made for parsing the timestamp. 

This all stems from the following comment where the nested timestamp has to be brought to the top level of the JSON structure.
```
	// have to bring $date to top level in order for it to be parsed as timeKey
	// see https://github.com/fluent/fluent-bit/issues/1013
```

I am open to other solutions here but this seems like a reasonable solution that works similarly to the old implementation.

confgenerator/logging_processors.go
dyl10s: Most of the solutions are valid here except there is one condition where `LoggingProcessorHardRename` is needed due to side effects of `MoveFrom`. 

When dealing with a log from `wiredTiger` we want to overwrite the top level `msg` field with the nested `attr.message` field. We only want to do this if the `attr.message` field is present. When using `MoveFrom` it will always overwrite the field with an empty object when `attr.message` is not present. 

From my understanding, I can't think of a solution that allows us to preserve the old `msg` if `attr.message` is not present with the current functionality.

franciscovalentecastro: I think we can use the `OmitIf` feature for this situation. Here the condition `jsonPayload.attr.message :*` checks if the field `attr.message` is present in the log. AFAIU, using "OmitIf: `NOT jsonPayload.attr.message :*`" will skip the `MoveFrom` if the field is not present.


```
confgenerator.LoggingProcessorModifyFields{
			Fields: map[string]*confgenerator.ModifyField{
				"jsonPayload.message": {
					MoveFrom: "jsonPayload..attr.message",
					OmitIf: `NOT jsonPayload.attr.message :*`,
				},
			},
		}
```

Note : I'm partially guiding myself with the [^1] `modify_fields` public documentation, so this may not be fully correct, but i think if this is possible to be done with `ModifyFields` it would use `OmitIf` in some capacity. We can also explore the `ModifyFields` implementation to get more clarity.

[^1]: https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/configuration#logging-processor-modify-fields

dyl10s: I tried that and the `OmitIf` functionality seems to actually omit the destination field if it matches.

From the docs you linked
`If the filter matches the input log record, the output field will be unset.`

so we end up deleting the `jsonPayload.message` field entirely if we don't have the nested message

franciscovalentecastro: Ohhh! Ok! 🤔 Please try further with other variations of the `OmitIf` to see if there is way to make it work. 

### Alternative

If this doesn't work, we can try a solution that uses a custom `LUA` function (or keep using `HardRename`) for `fluent-bit` and a `ModifyFields + CustomConverFunc + OTTL` solution for `Otel` .

I outlined a similar idea here https://github.com/GoogleCloudPlatform/ops-agent/pull/2014#discussion_r2295085664 . In summary you would create a struct `ConditionalMoveFromProcessor` which implements both `InternalLoggingProcessor` (implements `Components()` for fluent-bit) and `InternalOtelProcessor` (implements `Processors()` for otel) and return the custom logic for each subagent. 

You can probably also reuse `LoggingProcessorHardRename` (though i rather have a more specific name) and implement a `Processors()` method that uses `ModifyFields + CustomConverFunc + OTTL`.

dyl10s: I was not able to come up with anything that worked with `OmitIf` and updated the name of `HardRename` to something more clear as well as adding the otel `Processors()` logic. 

Will wait for @quentinmit to see if he has a better solution

PR #2013, opened 2025-07-31
[Draft] Add support for testing `LoggingReceiverFiles` like receivers.

franciscovalentecastro:
## Description
Add support for testing `LoggingReceiverFiles` like receivers. An improve version of https://github.com/GoogleCloudPlatform/ops-agent/pull/1968/files#diff-1be02ca7e3adb3b8074991539a3dc350002bcce6079ba9897a77707786791c88.

## Related issue
b/429241517

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


franciscovalentecastro:
### Why `MultilineRules` were kept in INPUT ?

The main reason i find for why the `MultilineRules` were kept in the `INPUT` component was what it is described in this comment `This is necessary, since using the multiline filter will not work if a multiline message spans between two chunks.`. 
I found some sources that back up this issue  :
- https://github.com/fluent/fluent-bit/issues/3901
- https://github.com/fluent/fluent-bit/issues/4317
- https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/multiline-parsing#limitations
- https://www.cncf.io/blog/2025/04/08/managing-multi-line-logs-with-fluent-bit-and-python/
  - Describing how buffering may interact with multiline parsing.
  - Mentions performance differences between input and filter multiline parsing.


https://github.com/GoogleCloudPlatform/ops-agent/blob/3ed6d7e7e844cad5214470c05dcfa6dcbbe00e7e/confgenerator/logging_receivers.go#L136-L156

### Answer
Re @quentinmit : 

> Why can't these apps just use `LoggingProcessorParseMultilineRegex` instead of making all these changes to hack together support for manipulating file receivers?

I haven't considered so far (though it is indeed possible) to change / refactor the main 3P app processor logic at all. This implies work on "3p app receivers fluent-bit logic" rather than "refactoring 3P receivers for Otel Logging", so it would be additional effort.

Currently, I considered it would be better to support 3P receivers for transformation testing which extends our testing possibilities, rather than considering working on `Fluent-bit` specific questions like `Does changing multiline.parsing not in input will still handle long-multiline logs correctly through multiple chunks and same performance ?`.

With this considerations, do you think its better to work on refactoring `fluent-bit multiline parsing` instead of the approach of this PR ?

franciscovalentecastro:
Re @quentinmit :
> I don't believe my suggestion would change any of the user-visible processor logic.
>
> I think it's worth preserving our abstraction barriers. I don't actually think it would be much of a refactoring - you would keep MultilineRules as it is in LoggingReceiverFilesMixin, but you would add a MergeLoggingProcessor() method that takes a LoggingProcessorParseMultilineRegex and turns it into the MultilineRules field.
>
> The transformation tests would use the processor, but the full config would configure the same parser directly on the input to avoid the chunk size bug.

Thank you for the suggestion of an alternative on how to test receivers with `MultilineRules` in `LoggingReceiverFilesMixin`. I implemented the alternative solution in https://github.com/GoogleCloudPlatform/ops-agent/pull/2025 following the guidelines of `MetricsProcessorMerger`, though with the main difference is the layer of abstraction (ops agent pipeline vs within a receiver) at which this "merges" happen. More details on https://github.com/GoogleCloudPlatform/ops-agent/pull/2025.

PR #2014, opened 2025-07-31
feat: Refactor IIS to use `LoggingProcessorMacro`

Caleb-Hurshman:
## Description

Refactor the IIS receiver logging to be sub-agent agnostic.
I am considering this a "rough draft". I believe the Lua functionality has been preserved. IIS is Windows-only, but the `RegisterLoggingFilesProcessorMacro` function doesn't handle a platform argument yet.

## How has this been tested?

Updated `transformation_test` and `confgenerator_test`, confirmed the output of `output_fluentbit.yaml` was unchanged after refactoring the receiver.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/apps/iis.go
+++ b/apps/iis.go
@@ -210,7 +210,7 @@ func (IISConcatFields) Processors(ctx context.Context) ([]otel.Component, error)
 				OmitIf: `jsonPayload.cs_uri_query = "-"`,
 			},
 			"jsonPayload.http_request_referer": {
-				OmitIf: `jsonPayload.referer = "-"`,
+				OmitIf: `jsonPayload.http_request_referer = "-"`,
 			},
 			"jsonPayload.user": {
 				OmitIf: `jsonPayload.user = "-"`,
--- a/transformation_test/testdata/logging_processor-iis_access/output_otel.yaml
+++ b/transformation_test/testdata/logging_processor-iis_access/output_otel.yaml
@@ -24,7 +24,6 @@
     timestamp: 2022-03-11T01:26:30Z
   - httpRequest:
       protocol: HTTP/1.1
-      referer: -
       remoteIp: "::1"
       requestMethod: GET
       requestUrl: /
@@ -62,7 +61,6 @@
     timestamp: now
   - httpRequest:
       protocol: HTTP/1.1
-      referer: -
       remoteIp: "::1"
       requestMethod: GET
       requestUrl: /forbidden?something=something
--- a/transformation_test/testdata/logging_processor-iis/output_otel.yaml
+++ b/transformation_test/testdata/logging_processor-iis/output_otel.yaml
@@ -24,7 +24,6 @@
     timestamp: 2022-03-11T01:26:30Z
   - httpRequest:
       protocol: HTTP/1.1
-      referer: -
       remoteIp: "::1"
       requestMethod: GET
       requestUrl: /
@@ -47,7 +46,6 @@
     timestamp: 2022-03-11T01:26:30Z
   - httpRequest:
       protocol: HTTP/1.1
-      referer: -
       remoteIp: "::1"
       requestMethod: GET
       requestUrl: /favicon.ico
@@ -70,7 +68,6 @@
     timestamp: 2022-03-11T01:26:32Z
   - httpRequest:
       protocol: HTTP/1.1
-      referer: -
       remoteIp: "::1"
       requestMethod: GET
       requestUrl: /forbidden?something=something


apps/iis.go
```diff
@@ -203,72 +203,191 @@ func (p *LoggingProcessorIisAccess) Components(ctx context.Context, tag, uid str
...
 
-	c = append(c,
-		confgenerator.LoggingProcessorModifyFields{
-			Fields: fields,
-		}.Components(ctx, tag, uid)...,
-	)
-	return c
-}
+	// Field movements happen after transformations
+	moveFields := confgenerator.LoggingProcessorModifyFields{
```
franciscovalentecastro: Why do we need to rewrite this specific `ModifyFields` processor to move fields into `httpRequest` ? The previous implementation seemed to do the same transformations.

apps/iis.go
```diff
@@ -203,72 +203,191 @@ func (p *LoggingProcessorIisAccess) Components(ctx context.Context, tag, uid str
...
-				"Name":    "grep",
-				"Match":   tag,
-				"Exclude": "message ^#(?:Fields|Date|Version|Software):",
-			},
+	}
+
+	// Define the complex transformation fields
+	serverIpTransform := &confgenerator.ModifyField{
+		CopyFrom: "jsonPayload.http_request_serverIp", // Read original value
+		CustomLuaFunc: func(tag string, sourceVars map[string]string) string {
```
franciscovalentecastro: Having two separate fields `CustomLuaFunc` and `CustomOTTLFunc` seems to be a good way of separating the implementations for Fluent-bit and Otel in a clean way. The other solution of adding statements of the form `if subAgent == fluent-bit` would be more cumbersome.

I still think there are some details to consider, for example : 

- Why doesn't `CustomLuaFunc` reuse the already existing `iisMergeRecordFieldsLuaFunction` / `iisMergeRecordFieldsLuaScriptContents` lua scripts ?
- Is it better to construct every field (e.g. `http_request_serverIp`, `http_request_requestUrl`) with a set of individual "statements" (ottl or lua) ? Is it better to only write one script for all fields ?
- Is there a way to create an agnostic `CustomConvertFunc` ? Idea : Instead of writing to separate statements/scripts, we could create a set of statements that would work in both. 

I don't know right now the best answers to all this questions, but i wanted to start discussing the details.

apps/iis.go
```diff
@@ -203,72 +203,191 @@ func (p *LoggingProcessorIisAccess) Components(ctx context.Context, tag, uid str
...
-	c = append(c, r.LoggingProcessorIisAccess.Components(ctx, tag, "iis_access")...)
-	return c
 }
 
 func init() {
-	confgenerator.LoggingReceiverTypes.RegisterType(func() confgenerator.LoggingReceiver { return &LoggingReceiverIisAccess{} }, platform.Windows)
-	confgenerator.LoggingProcessorTypes.RegisterType(func() confgenerator.LoggingProcessor { return &LoggingProcessorIisAccess{} }, platform.Windows)
+	// Note: The new RegisterLoggingFilesProcessorMacro system doesn't yet support
+	// platform restrictions like the previous registration system did.
+	// TODO: Add platform.Windows restriction once the macro system supports it.
```
franciscovalentecastro: Yes, we should add the `platform` parameter to this function. Thanks for adding a TODO!

transformation_test/testdata/logging_processor-iis_access/output_otel.yaml
```diff
@@ -0,0 +1,87 @@
...
+    resource:
+      labels:
+        instance_id: test-instance-id
+        zone: test-zone
+      type: gce_instance
+    timestamp: now
+  partialSuccess: true
+- collector_errors:
+  - caller: ottl@v0.131.0/parser.go:410
+    error: "parsing time \"2022-03-10 17:26:30\" as \" %Y-%m-%d %H:%M:%S\": cannot parse \"2022-03-10 17:26:30\" as \" \""
```
franciscovalentecastro: There seems to be a mismatch between the timestamp format and the log timestamps.

apps/iis.go
```diff
@@ -203,72 +203,191 @@ func (p *LoggingProcessorIisAccess) Components(ctx context.Context, tag, uid str
 				"http_request_status": "integer",
```
franciscovalentecastro: The transformation tests seems to have caught a bug in `TimeFormat: " %Y-%m-%d %H:%M:%S"`. Please remove the "leading space", since this prevents the timestamp parsing to work correctly in the Otel Collector.

transformation_test/testdata/logging_processor-iis_access/output_otel.yaml
```diff
@@ -0,0 +1,87 @@
+- entries:
+  - jsonPayload:
```
franciscovalentecastro: FYI, there is a bug in the Google Cloud Opentelemetry Exporter when setting fields in `httpRequest` which may cause to drop all the fields `httpRequest.*`.

See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/1062.

franciscovalentecastro (commented):
@Caleb-Hurshman This is a good "rough draft". It really brings to light all the technical considerations around refactoring `IIS`. I added some comments. Thank you for the effort in this so far!

confgenerator/logging_modify_fields.go
```diff
@@ -42,10 +42,12 @@ type ModifyField struct {
 	omitVar string `yaml:"-"`
 
 	// Operations to perform
-	Type              string                            `yaml:"type" validate:"omitempty,oneof=integer float"`
-	CustomConvertFunc func(ottl.LValue) ottl.Statements `yaml:"-"`
-	OmitIf            string                            `yaml:"omit_if" validate:"omitempty,filter"`
-	MapValues         map[string]string                 `yaml:"map_values"`
+	Type              string                                                   `yaml:"type" validate:"omitempty,oneof=integer float"`
+	CustomConvertFunc func(ottl.LValue) ottl.Statements                        `yaml:"-"`
+	CustomLuaFunc     func(tag string, sourceVars map[string]string) string    `yaml:"-"`
```
franciscovalentecastro: After further thought, i think we shouldn't add `CustomLuaFunc` / `CustomOTTLFunc` to `ModifyFields`. Some reasons for this : 
- `ModifyFields` current abstraction is to work with the fields `<dest>` , `<source>`, but this new features extend this to an arbitrary list of fields.
- `ModifyFields` is also a feature exposed to users.
- This is required by only the `IIS` processor so adding complexity here is unnecessary for only one case.
- The only required operation missing from ModifyFields is string concatenation. Using `CustomConvertFunc + Concat` [^1] we can do this in Otel.
- For `fluent-bit` we can keep a similar Lua script for the string concatenation and simplify the steps that can be done with ModifyFields (e.g. ommiting fields).

Please remove `CustomLuaFunc` / `CustomOTTLFunc` and do the following instead : 
- Create an `InternalLoggingProcessor` (interface) called `IISConcatFields`  which implements the methods `Components` (fluent-bit) and `Processors` (otel).
- Use `ModifyFields + CustomConvertFunc + Concat` [^1] for the Otel implementation. `CustomConvertFunc` can access any field in the log.
- Use `LuaFilterComponents` for fluent-bit implementation of concatenation.
- Simplify/remove the existing LuaScripts if possible.

Let me know if this is possible or if there any specific limitations.

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/ottlfuncs/README.md#concat

apps/iis.go
franciscovalentecastro: [Optional]

The underlying functions `RegisterType` within `RegisterLoggingFilesProcessorMacro` already have the implementation for the platform field. We could expose this platform argument (see `logging_macros.go`) directly in this PR or do it in a separate PR.

apps/iis.go
franciscovalentecastro: Followup comment : I was referring that the previous implementation did this in a more simplified way with a `or _, field := range ...` , so it doesn't seem necessary to rewrite it.

apps/iis.go
franciscovalentecastro: Followup comment : Disregard the questions / details here. I added some action items in a comment below referring to remove `CustomLuaFunc` / `CustomOTTLFunc`.

apps/iis.go
Caleb-Hurshman: Good catch, the old implementation does seem cleaner, I'll revert. I can't remember if there was a particular reason for this change at the time.

confgenerator/logging_modify_fields.go
Caleb-Hurshman: @franciscovalentecastro Most of this sounds good, but the otel output has not been working properly after changing to `Processors`:
`- config_error: "processor \"processor0\" has invalid configuration: unimplemented"`

I believe there might be a mismatch with the processor interface in OTel, let me know what you think.

franciscovalentecastro: I think the reason of that failure is that the signature of the `Processors` method should return `([]otel.Component, error)` :

https://github.com/GoogleCloudPlatform/ops-agent/blob/c6005d9cbd39e5c7a2897dd026b7311abad1cfe6/confgenerator/config.go#L799-L803

For this particular implementation it seems you would be able to call the `Processors` method of the `modifyFields` processor defined within the `IISConcatFields.Processors` method and "return" the result.

Caleb-Hurshman: I did try the return that way, but encountered the same result. I'll change the return type back, but there's more to this issue. I'll dig in further once the rest of the receivers are up to date.

franciscovalentecastro: Yes, you are right, there is more to this issue. This is a bug i've encountered before, but failed to fix in other PR's. I can fix this in a separate PR or we can do it here. Let me know what do you think its best.

[Optional]

In `logging_macros.go`, all the `Pipelines () / Processors()` methods should use `InternalOtelReceiver` / `InternalOtelProcessor` instead of `OtelReceiver` / `OtelProcessor`. This should fix the `unimplemented` error.

https://github.com/GoogleCloudPlatform/ops-agent/blob/5171236f23b8a2efb4f83ee2cd25cb06fc5d3ee8/confgenerator/logging_macros.go#L64-L66

https://github.com/GoogleCloudPlatform/ops-agent/blob/5171236f23b8a2efb4f83ee2cd25cb06fc5d3ee8/confgenerator/logging_macros.go#L71-L73

https://github.com/GoogleCloudPlatform/ops-agent/blob/5171236f23b8a2efb4f83ee2cd25cb06fc5d3ee8/confgenerator/logging_macros.go#L126-L128

Caleb-Hurshman: I'm glad to hear you know what the issue is! I think a separate PR would be good since a lot is going on in this PR already.

PR #2012, opened 2025-07-31
Updating opentelemetry-operantions-collector to bring googleclientauth

rafaelwestphal:
## Description
The otlphttp exporter needs googleclientauth to be able to send metrics to that endpoint.
## Related issue

b/435217364

## Checklist:
- Unit tests
  - [X] Unit tests do not apply.
  - [ ] Unit tests have been added/modified and passed for this PR.
- Integration tests
  - [X] Integration tests do not apply.
  - [ ] Integration tests have been added/modified and passed for this PR.
- Documentation
  - [X] This PR introduces no user visible changes.
  - [ ] This PR introduces user visible changes and the corresponding documentation change has been made.
- Minor version bump
  - [X] This PR introduces no new features.
  - [ ] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.
  - [ ] This PR bumps the version.

PR #2018, opened 2025-08-06
[Dockerfile] Add `tzdata` in Ubuntu build container setup.

franciscovalentecastro:
## Description
Adding `tzdata` to build containers to avoid failures when `otelopscol` can't find a Time Zone database causing test failures.

Context :
- https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/390
- https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/382
- https://github.com/GoogleCloudPlatform/ops-agent/pull/2012

## Related issue
b/435664151

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -715,7 +715,7 @@ ARG OPENJDK_MAJOR_VERSION
 RUN set -x; apt-get update && \
 		DEBIAN_FRONTEND=noninteractive apt-get -y install git systemd \
 		autoconf libtool libcurl4-openssl-dev libltdl-dev libssl-dev libyajl-dev \
-		build-essential cmake bison flex file libsystemd-dev \
+		build-essential cmake bison flex file libsystemd-dev tzdata \
 		devscripts cdbs pkg-config openjdk-${OPENJDK_MAJOR_VERSION}-jdk zip
 
 SHELL ["/bin/bash", "-c"]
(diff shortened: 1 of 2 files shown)


Dockerfile
jefferbrecht, line 1: AFAICT only the Ubuntu changes are needed. The RH and Debian families already include tzdata OOTB. Still checking the SUSE family

franciscovalentecastro: FWIW, in SUSE family the package is called `timezone` instead of `tzdata` (and it also contains the IANA database).

franciscovalentecastro: I updated the `Dockerfile` to only add `tzdata` in Ubuntu.

jefferbrecht: SUSE does not install `timezone` OOTB though. But are we running the transformation tests there? If not then we can leave it out anyway.

franciscovalentecastro: We don't run transformation tests on SUSE or RH, only in Debian and Ubuntu. Ok, we can leave it out.

jefferbrecht (approved):

quentinmit:
Is there a separate PR to add it as a dependency in the deb/rpm packages? It's a runtime dependency irrespective of whether we're running any tests (and arguably we should write an integration test to make sure this actually works when a user installs our packages)

PR #2021, opened 2025-08-07
[confgenerator] Add `confgenerator` tests for several `otel` 3P app receivers and fix `LoggingReceiverMacro`.

franciscovalentecastro:
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


quentinmit (commented):
I would rather not add all these duplicate test cases. Can we just add otel logging to all of the existing tests instead? (i.e. run the same input.yaml with both cases)

quentinmit:
> > I would rather not add all these duplicate test cases. Can we just add otel logging to all of the existing tests instead? (i.e. run the same input.yaml with both cases)
> 
> Yes, that would be possible. I'm thinking this idea would make `confgenerator_test` closer to how `tranformation_test` works in the sense that there would be two outputs, e.g. `output_otel.yaml` and `output_fluent_bit.yaml`. In the confgenerator case it would folders of goldens.
> 
> > Can we just add otel logging to all of the existing tests instead?
> 
> Would this other approach be to update how `confgenerator_test.go` works ? In a way that :
> 
> * Two subtests `.../otel` and `.../fluentbit` are run.
> * Same `input.yaml` is used, but when insert `OtelLogging =True` in the `../otel` test. Also the `otel_logging` ENV variable is set.
> * Two golden output subfolders (e.g. `golden_fluentbit`, `golden_otel`) are generated for each test folder.

We're already generating 4 different test outputs (`linux`, `linux-gpu`, `windows`, `windows-2012`). I think we can add one or two more outputs (`linux-otel` and `windows-otel`, perhaps) using the existing code, just by adding it to the `platformConfig` struct in the test.

PR #2022, opened 2025-08-07
Generating otlphttp endpoint when feature flag is enabled

rafaelwestphal:
## Description
Created a new feature flag otlp_endpoint that replaces Google managed Prometheus exporter with the otlphttp exporter for Prometheus metrics.

## Related issue
b/435217702

## How has this been tested?
Unit test has been added to verify that the configuration is being generated when the feature flag is enabled.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/confgenerator.go
```diff
@@ -249,7 +262,7 @@ func (p pipelineInstance) otelComponents(ctx context.Context) (map[string]otel.R
 			ReceiverPipelineName: receiverPipelineName,
 		}
 		// Check the Ops Agent receiver type.
-		if receiverPipeline.ExporterTypes[p.pipelineType] == otel.GMP {
+		if receiverPipeline.ExporterTypes[p.pipelineType] == otel.GMP || receiverPipeline.ExporterTypes[p.pipelineType] == otel.Otlp {
```
jefferbrecht: Is the new exporter intended to eventually be re-used for non-Prometheus pipelines as well?

If so, then I think this check is a bit of a time bomb, because there's no restriction on processors for non-Prometheus pipelines. If and when we decide to re-use the exporter for non-Prometheus, it would introduce a bug that we might not catch.

Otherwise, I think the new exporter type should be named something that's specific to Prometheus, i.e. `OtlpGMP` or something.

rafaelwestphal: The new exported is going to be re-used for non-Prometheus pipelines as well.

I changed the check here to prevent that problem from happening. This check should cover us for the scenarios:
* Prometheus receiver with GMP exporter
* Prometheus receiver with OTLP exporter
* OTLP receiver with GMP exporter
*

jefferbrecht: What about OTLP receiver with OTLP exporter, but the OTLP receiver is configured in GMP mode? I think that's still not covered.

What if we add a new map adjacent to `receiverPipeline.ExporterTypes` that identifies the data model (GCM vs. GMP) for each pipeline/signal type? Today we assume that the data model is implied entirely by the exporter type, but with the OTLP exporter it's no longer true. The change you're proposing is potentially going down a rabbit hole where we'd need to exhaustively check every combination of receiver type and exporter type.

jefferbrecht: Or actually, since the data model depends on the configuration in the case of the OTLP receiver, it might have to be a function rather than a map.

rafaelwestphal: I've reworked this part of the code and moved it to the validator.
PTAL

confgenerator/otel/modular.go
```diff
@@ -184,6 +187,14 @@ func (c ModularConfig) Generate(ctx context.Context) (string, error) {
 			telemetryMap["logs"] = logs
 		}
 	}
+	if len(c.Extensions) > 0 {
+		extensionsList := []string{}
```
XuechunHou: Please update `extensionsList := []string{}` to `var extensionList []string`

See: go/go-style/decisions#nil-slices

confgenerator/otel/modular.go
```diff
@@ -37,6 +37,7 @@ const (
 	OTel ExporterType = iota
 	System
 	GMP
+	Otlp
```
braydonk: We should stylize this abbrevation in all-caps similar to GMP

confgenerator/otel/modular.go
```diff
@@ -184,6 +187,14 @@ func (c ModularConfig) Generate(ctx context.Context) (string, error) {
 			telemetryMap["logs"] = logs
 		}
 	}
+	if len(c.Extensions) > 0 {
```
braydonk: This was added here but then is repeated down below, is this intentional?

confgenerator/prometheus.go
```diff
@@ -86,23 +90,26 @@ func (r PrometheusMetrics) Pipelines(ctx context.Context) ([]otel.ReceiverPipeli
...
+	exporter := map[string]otel.ExporterType{
+		"metrics": otel.GMP,
+	}
 	if exp_otlp_exporter {
 		metrics_components = append(metrics_components, otel.MetricUnknownCounter())
 		metrics_components = append(metrics_components, otel.MetricStartTime())
 		metrics_components = append(metrics_components, otel.GCPProjectID(resource.ProjectName()))
+		exporter = map[string]otel.ExporterType{
+			"metrics": otel.Otlp,
+		}
```
XuechunHou: `exporter["metrics"] = otel.Otlp`*

confgenerator/config.go
```diff
@@ -1298,24 +1298,23 @@ func validateComponentKeys[V any](components map[string]V, refs []string, subage
 	return nil
 }
 
+type CustomValidator interface {
```
XuechunHou: Why do you need this interface?

braydonk: I'd recommend renaming this to `CustomProcessorValidator`; the current name makes it sound like something more generic than it is.

confgenerator/config.go
```diff
@@ -1298,24 +1298,23 @@ func validateComponentKeys[V any](components map[string]V, refs []string, subage
...
 	for _, ID := range receiverIDs {
 		receiver, ok := receivers[ID]
 		if !ok {
-			return fmt.Errorf("metric receiver %q is not defined.", ID)
+			return fmt.Errorf("metric receiver %q is not defined", ID)
 		}
-		receiverPipelines, err := receiver.Pipelines(ctx)
-		if err != nil {
-			return err
+		if v, ok := receiver.(CustomValidator); !ok || v.AllowCustomProcessors() {
```
XuechunHou: cant you just do `receiver.AllowCustomProcessors()`?

confgenerator/config.go
```diff
@@ -1298,24 +1298,23 @@ func validateComponentKeys[V any](components map[string]V, refs []string, subage
 	return nil
 }
 
+type CustomValidator interface {
+	AllowCustomProcessors() bool
+}
+
 // Pipelines that export prometheus metrics are not allowed to have Ops Agent processors
-func validateNoCustomGMPProcessors(receivers metricsReceiverMap, receiverIDs, processorIDs []string, subagent string, ctx context.Context) error {
+func validateNoCustomGMPProcessors(receivers metricsReceiverMap, receiverIDs, processorIDs []string) error {
```
braydonk: Since this now checks for receivers to have the validator and is not GMP specific, we should rename this to `validateAllowCustomProcessors`

confgenerator/config.go
XuechunHou: I see.... 

maybe you could include `AllowCustomProcessors()` in  https://github.com/GoogleCloudPlatform/ops-agent/blob/259e38d7ac3fd9a7c8576e20e2be57f7faa37671/confgenerator/config.go#L598 here?

braydonk: This would require every type implementing this interface to also implement the method. The current solution treats the interface not being implemented the same as allowing custom processors.

confgenerator/config.go
braydonk: The code doesn't know a receiver implements `CustomValidator` without this assertion, so it can't call a method from the `CustomValidator` interface.

confgenerator/otel/modular.go
rafaelwestphal: Thanks, bad code merge.

XuechunHou (approved):

PR #2023, opened 2025-08-07
pin cuda version

XuechunHou:
## Description
pin cuda version to 12-9 for ubuntu 2204, 2404 and sles 15. Nvidia driver also needs to be explicitly installed. 

## Related issue
b/436552617
## How has this been tested?
3p integration tests

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/third_party_apps_test/applications/dcgm/debian_ubuntu/install
```diff
@@ -38,7 +38,7 @@ case $DEVICE_CODE in
             # cuda-12-6 is the latest version that supports Debian 11
             sudo apt -y install cuda-12-6
         else
-            sudo apt -y install cuda
+            sudo apt -y install nvidia-driver-580 cuda-12-9
```
jefferbrecht: CUDA 12.x requires driver <580: https://docs.nvidia.com/deploy/cuda-compatibility/minor-version-compatibility.html

This page suggests driver 575: https://developer.nvidia.com/cuda-12-9-0-download-archive

Although there's a versionless package you can try too: `sudo apt-get install -y nvidia-open` or `sudo apt-get install -y cuda-drivers`

XuechunHou: It seems like `sudo apt -y install cuda` would install a meta package that includes an Nvidia driver. However, `sudo apt -y install cuda-12-9` seems to install a smaller bundle and doesn't include an Nvidia driver. This is because I got the `nvidia-smi` command not found error, and that's why i am looking to install nvidia driver explicitly. 

I will try the cmd you shared. Thanks

jefferbrecht:
> Are you sure these particular versions will continue to be installable once new versions are released? Pinning a version feels like a ticking time bomb for the future.
> 
> If the problem is a missing binary from nvidia-driver, let's see if there's a virtual package that can be installed without a version number.

This is intentionally a short term fix to unblock the build. The latest CUDA version (13.x) removes the pre-built demo tool that we're relying on to produce telemetry. Once we decide how we're going to resolve that issue, we can remove the pin.

Note that some distros (e.g. Debian 11) are plain unsupported by the latest CUDA and may need a pin regardless, unless there's some better way to fetch the latest supported version without a pin.

jefferbrecht:
LGTM, I'll open a tracker to migrate to the new [`nvbandwidth`](https://github.com/NVIDIA/nvbandwidth) so that we can remove the pins.

PR #2025, opened 2025-08-08
[confgenerator] Create `MergeInternalLoggingProcessor` in `FilesMixin` to merge processor `MultilineRules`.

franciscovalentecastro:
## Description
This PR creates the interface `InternalLoggingProcessorMerger` to enable `InternalLoggingReceivers` to merge `InternalLoggingProcessors` features into the receiver when possible and supported.

Specifically the new method `MergeInternalLoggingProcessor` of `LoggingReceiverFilesMixin` will merge the processor `LoggingProcessorParseMultilineRegex.Rules []MultilineRules` into the files receiver. 

### Details
- Co-author @quentinmit : All the related `MULTILINE_PARSER` logic was updated so the "merging" only has to be implemented in one place. This reduces a lot of tech debt.

- This solution resembles the already implemented `MetricsProcessorMerger`. This solution differs in the abstraction layer where this "merges" happen. `MetricsProcessorMerger` works for a `processor` in an Ops Agent pipeline. In comparison,  `InternalLoggingProcessorMerger` works for a `processor` within a `LoggingReceiver` definition.

### Updates
- Created `InternalLoggingProcessorMerger` with method `MergeInternalLoggingProcessor`.
- Added `MergeInternalLoggingProcessor` method to `LoggingReceiverFilesMixin`.
  - This will merge `LoggingProcessorParseMultilineRegex.Rules []MultilineRules` into the files receiver when possible.
- Updated `elasticseach_json` receiver and processor with this refactoring.
  -  Transformation test with multiline processing work as expected.

### Context

As described before in https://github.com/GoogleCloudPlatform/ops-agent/pull/2013, there are `fluent-bit` downsides/bugs related to having `MultilineRules` in a separate `FILTER` processor outside of the `INPUT`. This solution enables the `processor` to  merge this `MultilineRules` to the `INPUT` receiver when possible to avoid the downsides/bugs.
 
## Related issue
b/413410475

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/logging_receivers.go
```diff
@@ -74,6 +74,7 @@ type LoggingReceiverFilesMixin struct {
 	MultilineRules          []MultilineRule `yaml:"-"`
 	BufferInMemory          bool            `yaml:"-"`
 	RecordLogFilePath       *bool           `yaml:"record_log_file_path,omitempty"`
+	MergeMultilineRules     bool            `yaml:"-"`
```
quentinmit: Why does this need to be a new field? Why can't we unconditionally do it?

confgenerator/logging_macros.go
```diff
@@ -128,16 +150,21 @@ func (cp loggingProcessorMacroAdapter[LPM]) Processors(ctx context.Context) ([]o
 	return processors, nil
 }
 
-// RegisterLoggingFilesProcessorMacro registers a LoggingProcessorMacro as a processor type and also registers a receiver that combines a LoggingReceiverFilesMixin with that LoggingProcessorMacro.
-func RegisterLoggingFilesProcessorMacro[LPM LoggingProcessorMacro](filesMixinConstructor func() LoggingReceiverFilesMixin) {
-	RegisterLoggingProcessorMacro[LPM]()
+// RegisterLoggingFilesReceiverMacro registers a receiver that combines a LoggingReceiverFilesMixin with that LoggingProcessorMacro.
+func RegisterLoggingFilesReceiverMacro[LPM LoggingProcessorMacro](filesMixinConstructor func() LoggingReceiverFilesMixin) {
```
quentinmit: I take it you're adding this so you can register only the receiver. If we correctly handle the merging, though, can't we just continue to register it as a receiver + processor?

franciscovalentecastro: I did this because all of this `3P app receivers` that have `MultilineRules` in the `INPUT` tail receiver don't register a `processor`. Arguably, after this change, there is no restriction anymore since the `processor` now could work as a standalone without the multiline parsing in the `INPUT`.

confgenerator/logging_receivers.go
franciscovalentecastro: We could. I was just considering all the 3P app receivers that don't have `MultineRules` in the receiver, but they define in the `processor` next to a `Regex` parser.

FWIW, maybe i'm just overly considering on how to keep the same implementation of this `receivers + processors`, and standarizing (all in the INPUT or TAIL) the use of `MultilineRules` would be helpful moving forward..

confgenerator/testdata/goldens/logging-receiver_elasticsearch_custom/golden/linux-gpu/fluent_bit_parser.conf
```diff
@@ -39,17 +39,3 @@
     Name        ops-agent-health.health-checks-json
     Time_Format %Y-%m-%dT%H:%M:%S%z
     Time_Key    time
-
-[MULTILINE_PARSER]
```
franciscovalentecastro, line 43: I'm not sure why here the `elasticsearch_json` confgenerator test this `MULTILINE_PARSER` is removed, but it still works in the `elasticsearch_json` transformation test handling the multiline JSON entries. Is this a bug in the merging logic ?

franciscovalentecastro: Fixed bug in https://github.com/GoogleCloudPlatform/ops-agent/pull/2025/commits/2c59993c69d0556f0966748f1a351431c61b8fc2.

confgenerator/logging_receivers.go
franciscovalentecastro: The need for a field `MergeMultilineRules` was removed after the last updates.

confgenerator/logging.go
```diff
@@ -35,9 +35,11 @@ func setLogNameProcessor(ctx context.Context, logName string) LoggingProcessorMo
 		"logName": {
 			DefaultValue: &logName,
 		},
-		`labels."compute.googleapis.com/resource_name"`: {
+	}
+	if hostName != "" {
+		fields[`labels."compute.googleapis.com/resource_name"`] = &ModifyField{
```
franciscovalentecastro: Note : This change (i don't fully remember if this was intended to be kept or not) adds a lot of `golden` updates.

transformation_test/transformation_test.go
```diff
@@ -261,19 +284,16 @@ func generateFluentBitConfigs(ctx context.Context, name string, transformationTe
...
 
-// generateOTelConfig attempts to generate an OTel config file for the test case.
-// It calls t.Fatal if there is something wrong with the test case, or returns an error if the config is invalid.
-func (transformationConfig transformationTest) generateOTelConfig(ctx context.Context, t *testing.T, name string, addr string) (string, error) {
+func testContext() context.Context {
 	pl := platform.Platform{
 		Type: platform.Linux,
 		HostInfo: &host.InfoStat{
-			Hostname:        "hostname",
+			Hostname:        "", // To remove the resource_name label. TODO: Set to "hostname" instead.
```
franciscovalentecastro: Note on this `TODO`.

franciscovalentecastro (commented):
LGTM! 

There are some test failures (e.g. missing update goldens) which may be solved after removing some debug statements. 

Thank you for all the help in this PR @quentinmit !

confgenerator/logging.go
quentinmit: So it does. We made this change to avoid lots of golden updates in the transformation tests, but I'll put it back the way it was since there are more confgenerator tests than transformation tests.

We still update several of the goldens because we are now correctly stripping the extra newlines in fluent-bit, but that's a meaningful diff.

quentinmit:
Unfortunately we were still getting fluent-bit segfaults on some of the transformation tests. I *think* I might have finally fixed them (by removing the empty `multiline` filter we were generating after moving the rules into the input), so if the checks all pass tonight I'll merge this. Otherwise we need to do more segfault hunting tomorrow.

quentinmit (approved):

quentinmit:
The four remaining test failures all appear to be unrelated flakes due to quota issues in our CI environment, so I'm going to merge this.

PR #2048, opened 2025-08-20
Remove TODO: b/400435104

franciscovalentecastro:
## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test/main_test.go
+++ b/integration_test/ops_agent_test/main_test.go
@@ -2553,7 +2553,7 @@ func TestSystemdLog(t *testing.T) {
 				t.Fatalf("Error writing dummy Systemd log line: %v", err)
 			}
 
-			querySystemdInfoLog := fmt.Sprintf(`severity="INFO" AND jsonPayload.MESSAGE="my_systemd_info_log_message" AND jsonPayload.PRIORITY="6"`)
+			querySystemdInfoLog := `severity="INFO" AND jsonPayload.MESSAGE="my_systemd_info_log_message" AND jsonPayload.PRIORITY="6"`
 			if err := gce.WaitForLog(ctx, logger, vm, "systemd_logs", time.Hour, querySystemdInfoLog); err != nil {
 				t.Error(err)
 			}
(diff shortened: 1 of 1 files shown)


integration_test/ops_agent_test/main_test.go
```diff
@@ -2562,20 +2562,19 @@ func TestSystemdLog(t *testing.T) {
 				t.Fatalf("Error writing dummy Systemd log line: %v", err)
 			}
 
-			querySystemdErrorLog := fmt.Sprintf(`severity="ERROR" AND jsonPayload.MESSAGE="my_systemd_error_log_message" AND jsonPayload.PRIORITY="3"`)
+			querySystemdErrorLog := `severity="ERROR" AND jsonPayload.MESSAGE="my_systemd_error_log_message" AND jsonPayload.PRIORITY="3"`
 			if err := gce.WaitForLog(ctx, logger, vm, "systemd_logs", time.Hour, querySystemdErrorLog); err != nil {
```
jefferbrecht, line 2566: Did the exporter fix cover just NOTICE, or multiple severity levels? It looks like the ERROR test here was not disabled previously, meaning ERROR was somehow working fine while NOTICE wasn't, or the test isn't actually checking what we think it is. I'm guessing it's the former but I'm curious if that's true.

franciscovalentecastro: The fix https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/1031 made sure that all `LogSeverity` set in a log with a string (e.g. `"logging.googleapis.com/severity : NOTICE"`) would be exported correctly to Google Cloud Logging.

### Summary of the issue
This is the upstream issue : https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/1030.

> Did the exporter fix cover just NOTICE, or multiple severity levels?

The issue was that the Otel model severity values (`SeverityText`, `SeverityNumber`) were not being mapped correctly in the exporter to the correct GCL `LogSeverity`. The exporter was missing mappings for `SeverityText` values "DEFAULT", "NOTICE", "WARNING", "CRITICAL", "ALERT", "EMERGENCY" to their correct `LogSeverity`.

The mappings for "INFO", "ERROR", "DEBUG" already existed, because those "levels" exist in both [OTel Logs Data Model](https://opentelemetry.io/docs/specs/otel/logs/data-model/#severity-fields) format and GCL `LogSeverity`.

### Some details
- This issue was in the `googlecloudexporter`, so all otel logging pipelines would be missing some `LogSeverity` mappings.
- I created `logging_processor-severity` transformation test to verify all levels are exported correctly.
- The tests were commented in `TestSystemdLog` with only `NOTICE`, since that's where i caught the bug initially. I don't think adding all levels in this test would be the correct place.

jefferbrecht: Gotcha, thanks!

jefferbrecht (approved):

PR #2049, opened 2025-08-21
Add unknown metric processor

rafaelwestphal:
## Description
This adds the an unknown metric processor for prometheus pipelines with otlphttp endpoint.

## Related issue
b/435217196

## How has this been tested?
Tested this in the Prometheus integration tests.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/otel/processors.go
```diff
@@ -602,3 +602,14 @@ func GCPProjectID(projectID string) Component {
 		map[string]string{"gcp.project_id": projectID}, false,
 	)
 }
+
+func MetricUnknownCounter() Component {
```
ridwanmsharif, line 608: nit: we should leave a comment here for breadcrumbs for why we needed this. Maybe point to a golink to our design doc?

confgenerator/prometheus.go
```diff
@@ -90,6 +90,7 @@ func (r PrometheusMetrics) Pipelines(ctx context.Context) ([]otel.ReceiverPipeli
 	exp_otlp_exporter := experimentsFromContext(ctx)["otlp_exporter"]
 	metrics_components := []otel.Component{otel.GroupByGMPAttrs_OTTL()}
 	if exp_otlp_exporter {
+		metrics_components = append(metrics_components, otel.MetricUnknownCounter())
```
ridwanmsharif, line 93: Lets mention that this is a no-op of the source of the metrics here isn't `prometheus` stuff

ridwanmsharif (approved):

confgenerator/prometheus.go
rafaelwestphal: I'm not sure I follow. This is on the prometheus pipeline, so it would only apply to Prometheus metrics.

PR #2057, opened 2025-09-03
Moving prometheus check to validate step

rafaelwestphal:
## Description
This moves the check for processors in prometheus pipeline to the validate stage.

## Related issue
b/435217702

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/config.go
```diff
@@ -1295,6 +1298,29 @@ func validateComponentKeys[V any](components map[string]V, refs []string, subage
...
 
+func validatePrometheusReceiver(receivers metricsReceiverMap, receiverIDs, processorIDs []string, subagent string, ctx context.Context) error {
+	for _, ID := range receiverIDs {
+		receiver, ok := receivers[ID]
+		if !ok {
+			continue // Some receivers might not be used.
+		}
+		receiverPipelines, err := receiver.Pipelines(ctx)
+		if err != nil {
+			continue
```
ridwanmsharif: Why `continue` here instead of `return err`?

confgenerator/config.go
```diff
@@ -1295,6 +1298,29 @@ func validateComponentKeys[V any](components map[string]V, refs []string, subage
...
+		}
+		for _, receiverPipeline := range receiverPipelines {
+			// Check the Ops Agent receiver type.
+			if receiverPipeline.ExporterTypes[subagent] == otel.GMP {
+				fmt.Println("processorIDs", processorIDs)
```
ridwanmsharif: Remove the println here?

ridwanmsharif (approved):

confgenerator/config.go
```diff
@@ -1295,6 +1298,28 @@ func validateComponentKeys[V any](components map[string]V, refs []string, subage
 	return nil
 }
 
+func validatePrometheusReceiver(receivers metricsReceiverMap, receiverIDs, processorIDs []string, subagent string, ctx context.Context) error {
```
ridwanmsharif: [nit] naming is hard but we should either rename or add a comment about why we do this validation. Or do both.

maybe something like `validateNoCustomGMPProcessors` or sth idk. What we're validating here is that if we are sending to GMP, then Ops Agent users can't configure any processors in this pipeline

confgenerator/config.go
```diff
@@ -1295,6 +1298,29 @@ func validateComponentKeys[V any](components map[string]V, refs []string, subage
 	return nil
 }
 
+// Pipelines that export prometheus metrics are not allowed to have Ops Agent processors
+func validateNoCustomGMPProcessors(receivers metricsReceiverMap, receiverIDs, processorIDs []string, subagent string, ctx context.Context) error {
+	for _, ID := range receiverIDs {
+		receiver, ok := receivers[ID]
+		if !ok {
+			continue // Some receivers might not be used.
```
ridwanmsharif: Actually reading this again - this looks incorrect. If `!ok`, then it looks like a receiver is used in the pipeline but there is no such receiver in the map? 

Does another validator catch this error already or must we also return an error here?

ridwanmsharif: Looks like `validateComponentKeys` will catch this for us. I would call that out and also return an error here but this error would never actually happen. Maybe we can panic too

PR #2069, opened 2025-09-18
Add Support for Elasticsearch v8/v9

schmikei:
## Description
Upstream contrib has added support for metrics to collect against Elasticsearch 9. Between Elasticsearch 7.9 and 9 the log format changed to be ECS format by default. See https://www.elastic.co/docs/deploy-manage/monitor/logging-configuration/elasticsearch-log4j-configuration-self-managed#json-logging

>Elasticsearch version 8.0 marked a significant change in logging configuration, updating to use EcsLayout (which aligns with ECS JSON logging) instead of the previous ESJsonLayout. While previous versions retained the old infrastructure for compatibility, the explicit adoption of EcsLayout for structured JSON logs, adhering to the Elastic Common Schema, became the standard with Elasticsearch 8.0.

Modified the integration to add support for parsing the timestamp from `@timestamp` and similarly the `severity` level change. I could not find a good way to accomdate that Fluentbit needs to know the `TIME_KEY` format in order to parse the JSON and fluentbit does not allow you to retroactively set the time_key once parsed. So a new log format was introduced to cover the ECS schema: `elasticsearch_json_ecs`. Let me know if there's another way to solve this as it feels sort of like a workaround for fluentbit as there is no easy way to support files with different formats of timestamp keys. 

Added an integration_test for elasticsearch 9 as to test the metrics and logs with the new type.

## How has this been tested?
Added an integration test specifically for Elasticsearch 9 and added a transformation_test to account for the extra processing. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


apps/elasticsearch.go
```diff
@@ -255,6 +291,34 @@ func (p LoggingProcessorMacroElasticsearchJson) nestingProcessors() []confgenera
 	return processors
 }
 
+// This is a workaround to support ECS logs from Elasticsearch 8.x alongside Elasticsearch 7.x.
+// It's not apparent how we are supposed to parse the JSON first with indeterminate time keys, so we're creating two separate log
+// types to support being able to read from both formats.
+type LoggingProcessorMacroElasticsearchJsonV8 struct{}
+
+func (p LoggingProcessorMacroElasticsearchJsonV8) Type() string {
+	return "elasticsearch_json_ecs"
```
franciscovalentecastro: [nit] Maybe `elasticsearch_ecs` or `elasticsearch_ecs_json` could be better naming ? I don't have the full context of the use of ES or how some concept are used, but i mostly want to discuss the naming of this new "receiver".

franciscovalentecastro:
Re @schmikei : 

> In version Elasticsearch 8.x they changed the log format to be the [ECS format](https://www.elastic.co/guide/en/elasticsearch/reference/8.18/migrating-8.0.html#breaking_80_logging_changes). This renames the timestamp field of the JSON to be compliant with ECS as @timestamp. So the situation is that anything Elasticsearch 7.x used the old format for the time_key and level, whereas newer versions are going to need the the elasticsearch_json_ecs for proper JSON parsing (specifically the time_key field).
> 
> Yep the big complication for the Ops Agent given these new versions is the new log format listed above. We have different fields for `timestamp` and `severity` which are parsing actions we were accomplishing before. If we were to drop support for ES 7.x logs we could keep it down to one log type again and just ingest the ECS formatted logs:
>
> It seemed to me that the simplest way forward while we still maintain fluentbit was to introduce a separate new log type with its own dedicated parsing while also still targeting the same JSON paths as before. However if we have the flexibility of Otel I could imagine coming back down to just the two log types. Happy to add more context or anything that would help but this was my solution to the different format of logs in a fluentbit fashion.

Given that we have two well defined separate JSON log formats `ES 7.x custom format` and `ECS`, Why don't we target each receiver `elasticsearch_json` and `elasticsearch_json_ecs`, respectively, to those formats ? 

Some details of my proposed solution : 
- Don't filter out logs if they are not conforming to the chosen format. 
  - I think its quite a big decision to choose to "filter out" specific logs depending on their format in way that is not clear to the user.
  - [Edge Case] If a use upgrades their ES version from 7 to 8, but inadvertently keep the same Ops Agent config, they would filter out all their logs. 
- Tailor `elasticsearch_json` and `elasticsearch_json_ecs` to specific log format (`ES 7.x custom format` and `ECS`, respectively).
- Add documentation or change naming to make this distinction clearer.

This would transfer the responsibility to the user of knowing if they have `ECS` or normal JSON and choosing the receiver appropriately. Does this make sense ? This solution has the following pros and cons : 

- Pros : 
  - Simpler to implement and maintain.
  - There is no possibility of "filtering out" logs that were not meant to be filter out. 
  - In general we avoid an overly "restrictive" solution.
  - Even if the chosen JSON receiver "format" is wrong they would get a degraded type of logs (wrong timestamps and severity), instead of fully filtering out logs (Is this right ? I may be misunderstanding the current implementation).
 
- Cons : 
  - Users may not know which JSON log format their ES installed version is using.
  - We could address this with documentation ?

schmikei:
> Given that we have two well defined separate JSON log formats `ES 7.x custom format` and `ECS`, Why don't we target each receiver `elasticsearch_json` and `elasticsearch_json_ecs`, respectively, to those formats ?
> 
> Some details of my proposed solution :
> 
> * Don't filter out logs if they are not conforming to the chosen format.
>   
>   * I think its quite a big decision to choose to "filter out" specific logs depending on their format in way that is not clear to the user.
>   * [Edge Case] If a user upgrades their ES version from 7 to 8, but inadvertently keep the same Ops Agent config, they would filter out all their logs.
> * Tailor `elasticsearch_json` and `elasticsearch_json_ecs` to a specific log format (`ES 7.x custom format` and `ECS`, respectively).
> * Add documentation or change naming to make this distinction clearer.
> 
> This would transfer the responsibility to the user of knowing if they have `ECS` or normal JSON and choosing the receiver appropriately. Does this make sense ? This solution has the following pros and cons :
> 

I think we can probably remove the grep filter if the desired behavior is to submit incorrectly parsed logs (thinking gcp inboud traffic/having immutable logs that are not parsed due to a non ops-agent related upgrade etc). I can imagine a customer getting upset about that fact moreso than not having logs after the Elasticsearch upgrade if we had the information to properly not charge the customer in that case. Both scenarios are not ideal so its moreso just what is the "better" user experience.   

I suppose its more of a product call on that front which I'll defer to you on, but did want to at least say what the motivation is.  If you'd like the filter removed from your side I am happy to make that change, just wanted to highlight the ambiguity of such a move and solidify the decision!

apps/elasticsearch.go
schmikei: I think in context of elasticsearch I think `elasticsearch_ecs` I can support as being intuitive. https://www.elastic.co/docs/reference/ecs/ecs-getting-started

I'll go ahead and make that change 👍

franciscovalentecastro:
Re @dyl10s @schmikei ,

Hello! I found some ideas on how to implement the "automatic field detection" depending on the ElasticSearch version from only one receiver (no need to create a new receiver). We can do different combinations or permutation of this ideas (e.g. use two parse regex for each time field possible name). 

Given the following formats : 

ECS format:
```json
{"@timestamp":"2025-09-19T02:20:17.920Z", "log.level": "INFO", "message":"initialized", "ecs.version": "1.2.0","service.name":"ES_ECS","event.dataset":"elasticsearch.server","process.thread.name":"main","log.logger":"org.elasticsearch.node.Node","elasticsearch.node.name":"es-test-3","elasticsearch.cluster.name":"elasticsearch"}
```

ES 7.x custom format
```json
{"type": "server", "timestamp": "2025-09-18T20:44:01,838-04:00", "level": "INFO", "component": "o.e.n.Node", "cluster.name": "elasticsearch", "node.name": "es-test", "message": "initialized" }
``` 

This is one solution :

1. Use `LoggingProcessorParseJson` to parse the JSON log in any format (`ECS` or `Custom ES 7.x`) without setting a `time_key` : 
```
confgenerator.LoggingProcessorParseJson{ },
```

2. Normalize all fields using `ModifyFields`. Used `CopyFrom` instead of `MoveFrom` to preserve the original fields (this may be valuable to a user) : 

```
return confgenerator.LoggingProcessorModifyFields{
		Fields: map[string]*confgenerator.ModifyField{
			"severity": {
				CopyFrom: "log.level", // ECS format
			},
			"severity": {
				CopyFrom: "level", // Custom ES 7.x format
			},
			
			"temp_time_key": {
				CopyFrom: "@timestamp", // ECS format
			},
			"temp_time_key": {
				CopyFrom: "timestamp", // Custom ES 7.x format
			},
		},
	}
```

3. Use `LoggingProcessorParseRegex` to parse the whole `temp_time_key` as a timestamp (using a regex that matches the whole field). Set `PreserveKey: False` to drop the "temp" time field.
```
confgenerator.LoggingProcessorParseRegex{
	Regex: `(?<timestamp>.*)`,
	Field: "temp_time_key"
	PreserveKey: False
	ParserShared: confgenerator.ParserShared{
		TimeKey:    "timestamp",
		TimeFormat: "%Y-%m-%d %H:%M:%S,%L",
	},
}, 
```

dyl10s:
@franciscovalentecastro I am just getting to this now. I like your recommendation here, but if I am remembering correctly, `ModifyFields` will overwrite the destination with a blank value if the key does not exist. I believe I was having this issue when porting MongoDB. 

So in this case 
```
confgenerator.LoggingProcessorModifyFields{
	Fields: map[string]*confgenerator.ModifyField{
		"severity": {
			CopyFrom: "log.level", // ECS format
		},
		"severity": {
			CopyFrom: "level", // Custom ES 7.x format
		},	
		"temp_time_key": {
			CopyFrom: "@timestamp", // ECS format
		},
		"temp_time_key": {
			CopyFrom: "timestamp", // Custom ES 7.x format
		},
	},
}
```

`temp_time_key` would always either get set to `timestamp` or it would be blank. I should be able to spend some time tomorrow testing this to validate.

apps/elasticsearch.go
```diff
@@ -156,72 +156,115 @@ func (LoggingProcessorMacroElasticsearchJson) Type() string {
...
-					"CRITICAL":    "ERROR",
-					"FATAL":       "FATAL",
-				},
-				MapValuesExclusive: true,
+// LoggingProcessorMacroElasticsearchJsonRemoveTempLevel is a processor that removes the temp_level field from the json payload.
+type LoggingProcessorMacroElasticsearchJsonRemoveTempLevel struct {
+}
+
+// Components returns the fluentbit components for the LoggingProcessorMacroElasticsearchJsonRemoveTempLevel processor which is responsible for removing a temp_level field from the json payload.
+func (l LoggingProcessorMacroElasticsearchJsonRemoveTempLevel) Components(ctx context.Context, tag, uid string) []fluentbit.Component {
```
franciscovalentecastro: I think this processor is not needed if we do `MoveFrom` in the "severity mapping" processor instead of `CopyFrom` : 
```
"severity": {
			MoveFrom: "jsonPayload.temp_level",
			MapValues: map[string]string{
            ...
}
```

Also, we can remove fields with `LoggingProcessorModifyFields` using the `OmitIf` feature.

Note : I mostly kept `CopyFrom` to see all intermediate states of the transformations.

apps/elasticsearch.go
```diff
@@ -284,15 +328,17 @@ func (p LoggingProcessorMacroElasticsearchGC) Expand(ctx context.Context) []conf
 	}
 }
 
+var elasticsearchJSONIncludePaths = []string{
```
franciscovalentecastro: Nit : I don't think this separate method is needed outside of `loggingReceiverFilesMixinElasticsearchJson` which seems a better place to encapsulate all "files" to be read.

apps/elasticsearch.go
```diff
@@ -156,72 +156,115 @@ func (LoggingProcessorMacroElasticsearchJson) Type() string {
 	return "elasticsearch_json"
 }
 
+func loggingProcessorElasticsearchJsonMultilineRules() []confgenerator.MultilineRule {
```
franciscovalentecastro: Nit : Refactor this as a method of `LoggingProcessorMacroElasticsearchJson`.

apps/elasticsearch.go
```diff
@@ -156,72 +156,115 @@ func (LoggingProcessorMacroElasticsearchJson) Type() string {
...
+				"Match":  tag,
+				"Remove": "temp_level",
 			},
-			InstrumentationSourceLabel: instrumentationSourceValue(p.Type()),
 		},
 	}
 }
 
-func (p LoggingProcessorMacroElasticsearchJson) nestingProcessors() []confgenerator.InternalLoggingProcessor {
+func elasticsearchNestingProcessors() []confgenerator.InternalLoggingProcessor {
```
franciscovalentecastro: Nit : Keep as a method of `LoggingProcessorMacroElasticsearchJson`.

apps/elasticsearch.go
schmikei: Ah yep that makes a lot more sense 🤦 getting back into this  after vacation so thank you for the suggestion 👍

transformation_test/testdata/logging_processor-elasticsearch-json-ecs/input.log
```diff
@@ -0,0 +1,6 @@
+{"@timestamp":"2025-09-19T02:20:17.920Z", "log.level": "INFO", "message":"initialized", "ecs.version": "1.2.0","service.name":"ES_ECS","event.dataset":"elasticsearch.server","process.thread.name":"main","log.logger":"org.elasticsearch.node.Node","elasticsearch.node.name":"es-test-3","elasticsearch.cluster.name":"elasticsearch"}
```
franciscovalentecastro, line 1: Please add an example of a "multiline json log" to verify it works in ES 8.x. Note : There is a bug in fluent-bit that causes the last line not to be processed in transformation tests when there is a multi-line processor, so to process the multiline log it should no be the last log line.

apps/elasticsearch.go
```diff
@@ -156,72 +155,95 @@ func (LoggingProcessorMacroElasticsearchJson) Type() string {
...
 			},
-			InstrumentationSourceLabel: instrumentationSourceValue(p.Type()),
 		},
 	}
+	processors = append(processors, p.elasticsearchNestingProcessors()...)
+	return processors
 }
 
-func (p LoggingProcessorMacroElasticsearchJson) nestingProcessors() []confgenerator.InternalLoggingProcessor {
+func (p LoggingProcessorMacroElasticsearchJson) elasticsearchNestingProcessors() []confgenerator.InternalLoggingProcessor {
```
franciscovalentecastro: Nit : This can still be `nestingProcessors`.

apps/elasticsearch.go
```diff
@@ -156,72 +155,95 @@ func (LoggingProcessorMacroElasticsearchJson) Type() string {
 	return "elasticsearch_json"
 }
 
+func (p LoggingProcessorMacroElasticsearchJson) loggingProcessorElasticsearchJsonMultilineRules() []confgenerator.MultilineRule {
```
franciscovalentecastro: Nit : Please update the method name `loggingProcessorElasticsearchJsonMultilineRules` to `parseMultilineRegex` and return the whole `LoggingProcessorParseMultilineRegex` processor instead of only the `MultilineRule`.

apps/elasticsearch.go
franciscovalentecastro: Thank you! Done!

franciscovalentecastro (commented):
Some last comments.

franciscovalentecastro:
Hello @schmikei ! Followup question, what are the differences between the "metrics and logs" collected by ES 7 and ES 8/9. Are all metrics the same ? What are the main differences in the log fiels.
 
I'm asking in the context of updating our Public Documentation for ElasticSearch which lists every log entry field and metric : https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/third-party/elasticsearch#monitored-logs.

Creating a new `metadata.yaml` file (in `integration_test/third_party_apps_test/applications/elasticsearchv9/metadata.yaml`) essentially generates a new page for a separate `ElasticSearch` that would collide in the URL. 

**I wanted to ask your advice of how small or big are the differences ?** If they are small, we could do a "ElasticSearch 9 has some additional fields" (and skip the "elasticsearchv9" page generation), or we would need a separate page to consider all differences. Followup question, is it possible to make the `Log Entry fields` closer to the ES 7 support ?

schmikei:
> what are the differences between the "metrics and logs" collected by ES 7 and ES 8/9 

Logs: 
In elasticsearch 8 they switched the default logging to use ECS format. https://www.elastic.org.cn/docs/8.1/www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.0.html https://www.elastic.co/guide/en/ecs/8.1/index.html

This feature allows us to parse the new format correctly!

Metrics:
Nothing should changed here!

> What are the main differences in the log fields ?

Timestamp/Severity is now derived from the ECS standard fields `@timestamp` and `log.level`.

> I wanted to ask your advice of how small or big are the differences?

Generally I believe these are relatively small changes for the Ops Agent(from the ops agent perspective as its just responsible for parsing the JSON on the JSON logged files). It might be a large effort to map over every field possible in the ECS log fields compared to just linking to the documentation. https://www.elastic.co/docs/reference/ecs/ecs-field-reference. Since the original documentation appears to not capture every format of Elasticsearch log, I'm wondering if its generally more friendly to link out to Elastic's documentation vs trying to match parity of it? Do you have any thought son this front?

> Followup question, is it possible to make the Log Entry fields closer to the ES 7 receiver

I believe that this would be a tall order as you have to accomodate all these different fields in the non-standardized way: https://www.elastic.co/docs/reference/ecs/ecs-field-reference

It might be easier to migrate 7.0 logs to follow ECS format but that seems like a unexpected user expectation as well from my perspective. Either way would be a tall order technically to seemlessly solve.

schmikei:
All that to say is that an ecs log can have these fields present: https://www.elastic.co/docs/reference/beats/filebeat/exported-fields-ecs

Which could be a really extensible integration test if we were to define them in metadata.yaml but the amount of possibilities and combinations/conditions to replicate some of them would approach unreasonable amount of effort to document as say you can have a log with ECS fields and Elasticsearch fields https://www.elastic.co/docs/reference/beats/filebeat/exported-fields-elasticsearch. So I'm not entirely sure how "reasonable" it would be to try and document all the fields aside from maybe the base fields: https://www.elastic.co/docs/reference/ecs/ecs-base

franciscovalentecastro:
@schmikei Thank you for all the answers 🙏 ! This is very helpful. I will merge as is. I'll change the naming of the test folder to `elasticsearch-nine` so it works as separate page in our public docs.

> I believe that this would be a tall order as you have to accomodate all these different fields in the non-standardized way: https://www.elastic.co/docs/reference/ecs/ecs-field-reference

I agree, it would be too much effort to "standardize" all this logging fields or track them at all.

PR #2072, opened 2025-09-23
Import `plugin_comm.proto `from Google Guest Agent Repo 

XuechunHou:
## Description
Remove local copy of `plugin_comm.proto`, and pull from Google Guest Agent repo now that it has been exported publicly.

## Related issue
b/446747142

## How has this been tested?
unit tests + integration tests.
Existing tests should still pass.

## Checklist:
- Unit tests
  - [] Unit tests do not apply.
  - [X ] Unit tests have been added/modified and passed for this PR.
- Integration tests
- Documentation
- Minor version bump


rafaelwestphal (approved):

go.mod
```diff
@@ -1,67 +1,70 @@
 module github.com/GoogleCloudPlatform/ops-agent
 
-go 1.24
+go 1.24.0
```
franciscovalentecastro, line 3: Pinning the `PATCH` version in Golang can cause issues in the future. Please set on `1.24` for the golang version which would pick the latest `1.24.x` go version.

go.mod
```diff
@@ -1,67 +1,70 @@
...
+
+toolchain go1.24.4
```
franciscovalentecastro, line 5: [Nit] This may be non-issue since i don't understand the full story of how `toolchain` works, but this can cause issues with local development.

go.mod
franciscovalentecastro: For context, this is an issue that i've encountered in the past (https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/1036#issuecomment-2961010663) after pinning the PATCH version in `go.mod`.

franciscovalentecastro: It seems other dependencies are using `1.24.0` too now (see https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/detectors/gcp/go.mod). Disregard my comment. This seems to be ok.

PR #2093, opened 2025-10-15
Using otlp exporter for self metrics

rafaelwestphal:
## Description
This updates the self-metrics pipelines to use the otlp exporter. This includes the agent metrics (uptime, api_request_count, point_count) and the feature tracking pipelines.

## Related issue
b/452074867

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/tasks.mak
+++ b/tasks.mak
@@ -52,6 +52,7 @@ otelopscol_local: dist/opt/google-cloud-ops-agent/subagents/opentelemetry-collec
 
 .PHONY: clean_submodules
 clean_submodules:
+	cd submodules/opentelemetry-operations-collector/otelopscol/ && make clean
 	rm -rf dist
 
 ############


integration_test/ops_agent_test/main_test.go
```diff
@@ -2742,7 +2742,7 @@ func testDefaultMetrics(ctx context.Context, t *testing.T, logger *log.Logger, v
 
 func TestDefaultMetricsNoProxy(t *testing.T) {
 	t.Parallel()
-	RunForEachImageAndFeatureFlag(t, []string{OtelLoggingFeatureFlag}, func(t *testing.T, imageSpec string, feature string) {
+	RunForEachImageAndFeatureFlag(t, []string{OtlpHttpExporterFeatureFlag}, func(t *testing.T, imageSpec string, feature string) {
```
franciscovalentecastro: This removes `OtelLoggingFeatureFlag` from the list. Please keep both feature flags.

confgenerator/confgenerator.go
```diff
@@ -57,20 +57,24 @@ func googleCloudExporter(userAgent string, instrumentationLabels bool, serviceRe
...
-		return receiver
+		return pipeline
 	}
-	receiver.ExporterTypes["metrics"] = otel.OTLP
+	pipeline.ExporterTypes["metrics"] = otel.OTLP
 
-	receiver.Processors["metrics"] = append(receiver.Processors["metrics"], otel.GCPProjectID(resource.ProjectName()))
-	return receiver
+	pipeline.Processors["metrics"] = append(pipeline.Processors["metrics"], otel.GCPProjectID(resource.ProjectName()))
+	if systemMetric {
```
franciscovalentecastro: Should this be called `agentSelfMetric` instead of `systemMetric` ?

confgenerator/agentmetrics.go
```diff
@@ -33,6 +34,7 @@ type AgentSelfMetrics struct {
 	OtelPort            int
 	OtelRuntimeDir      string
 	OtelLogging         bool
+	ProjectName         string
```
franciscovalentecastro: Since `ProjectName` is only required for `AgentSelfMetrics` when the OTLP exporter is enabled, **Is there an alternative way to set `gcp.project_id`  instead of hardcoding the value in the config ?**

An alternative could be to add a resourcedetection [1] processor before the exporter. Would that work ?

[1] https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor

confgenerator/confgenerator.go
```diff
@@ -57,20 +57,24 @@ func googleCloudExporter(userAgent string, instrumentationLabels bool, serviceRe
 	}
 }
 
-func ConvertToOtlpExporter(receiver otel.ReceiverPipeline, ctx context.Context) otel.ReceiverPipeline {
+func ConvertToOtlpExporter(pipeline otel.ReceiverPipeline, ctx context.Context, systemMetric bool) otel.ReceiverPipeline {
```
franciscovalentecastro: [Suggestion]
After seeing the followup changes that are required to `ConvertToOtlpExporter` and all its references IMO (e.g. all 3P apps) i have some alternative ideas on how to add the `otlpExporter` into a `otel.ReceiverPipeline`.

Some thoughts :
- From the current `ReceiverPipeline` definition, the design of `AgentSelfMetrics` and `3P app receivers` IMO could more be "declarative" in the sense of only setting `receivers,processors,exporters`, but moving logic like `if exporter == x , then do this` somewhere else.

- Since every `otel.ReceiverPipeline` instantiation call has to be updated to use `ConvertToOtlpExporter`, maybe this logic is better suited into `GenerateOtelConfig()` or `OTelComponents()` in a way that is generic ?

Suggested ideas : 

1. Create an `EnabledExporterTypes` method and use it `otel.ReceiverPipeline` definitions :
`cofgenerator.go`
```
func EnabledExporterTypes(ctx context) map[string]otel.ExporterType {
        expOtlpExporter := experimentsFromContext(ctx)["otlp_exporter"] 
        if expOtlpExporter {
              return ExporterTypes: map[string]otel.ExporterType{
	         "metrics": otel.OTLP,
              }
	}
	return exporters := ExporterTypes: map[string]otel.ExporterType{
	       "metrics": otel.System,
         }
}
```
`active_directory_ds.go`
```
func (r MetricsReceiverActiveDirectoryDS) Pipelines(ctx context.Context) ([]otel.ReceiverPipeline, error) {
	return []otel.ReceiverPipeline(otel.ReceiverPipeline{
		Receiver: otel.Component{...},
		Processors: ...,
		ExporterTypes: EnabledExporterTypes(ctx),
	}, nil
}
```

2.  Move `ConvertToOtlpExporter` logic into `GenerateOtelConfig()` or `OTelComponents()` in a way that is "generic", e.g. works for all receiver pipelines that have the `otel.OTLP` added. One example of this is how we set `otelSetLogNameComponents`.

https://github.com/GoogleCloudPlatform/ops-agent/blob/e4e289266612a443bacd0f1885e8eaf486f1b12f/confgenerator/confgenerator.go#L285-L297

confgenerator/confgenerator.go
```diff
@@ -57,20 +57,24 @@ func googleCloudExporter(userAgent string, instrumentationLabels bool, serviceRe
...
+		return pipeline
 	}
-	receiver.ExporterTypes["metrics"] = otel.OTLP
+	pipeline.ExporterTypes["metrics"] = otel.OTLP
 
-	receiver.Processors["metrics"] = append(receiver.Processors["metrics"], otel.GCPProjectID(resource.ProjectName()))
-	return receiver
+	pipeline.Processors["metrics"] = append(pipeline.Processors["metrics"], otel.GCPProjectID(resource.ProjectName()))
+	if systemMetric {
+		pipeline.Processors["metrics"] = append(pipeline.Processors["metrics"], otel.MetricStartTime())
```
franciscovalentecastro: Is there a "generic" way to set `MetricStartTime` and `MetricsRemoveInstrumentationLibraryLabelsAttributes` that works for all metrics without needing to differentiate explicitly using an extra parameter like `systemMetric` ?

rafaelwestphal: No, this is the equivalent we have today for the different ways we use the GCM exporter. We call is system when we remove the instrumentation labels attributes, and OTEL when we don't. This is just mimicking that same behavior. It's really an XOR behavior here. For the metricstarttime we can just add it everywhere when using the otlphttp exporter

confgenerator/confgenerator.go
rafaelwestphal: This is mimicking the current behavior of the GCM exporter. We have 2 configs for the GCM, one we call otel.System, the other we call otel.OTel. That's  why I chose this name.

confgenerator/agentmetrics.go
rafaelwestphal: I don' t quiet understand what you mean here. We already have the gcp resource detection setup for all pipelines:
https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/otel/modular.go#L208

This gcp.project_id is not populated by that resource detector:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/internal/gcp/documentation.md

This is required by the new endpoint.

confgenerator/confgenerator.go
rafaelwestphal: The problem is that the transformation is different depending if the current exporter is configured as a system or not. If I just declare it as OTLP, I don' t know how to properly setup the pipeline later. It can either be a GMP, GCM system or GCM Otel exporter.

confgenerator/agentmetrics.go
```diff
@@ -297,24 +289,39 @@ func (r AgentSelfMetrics) LoggingMetricsPipelineProcessors() []otel.Component {
 		// DeltaToCumulative keeps in memory information of previous delta points
 		// to generate a valid cumulative monotonic metric.
 		otel.DeltaToCumulative(),
+		otel.MetricStartTime(),
```
franciscovalentecastro: I'm a bit concerned about that this may change the `LoggingMetrics` some unintended way. Some requests : 

- Could we please separate all `MetricStartTime()` changes into a separate PR to makes sure they work individually with the GCM exporter ?
- Could we document and track the effort to fix the upstream issue with `int64` metrics in the `MetricStartTime` processor ?

confgenerator/confgenerator.go
```diff
@@ -57,32 +57,40 @@ func googleCloudExporter(userAgent string, instrumentationLabels bool, serviceRe
...
+func ConvertGCMOtelExporterToOtlpExporter(pipeline otel.ReceiverPipeline, ctx context.Context) otel.ReceiverPipeline {
+	return ConvertToOtlpExporter(pipeline, ctx, false, false)
 }
 
-func ConvertToOtlpExporter(receiver otel.ReceiverPipeline, ctx context.Context, isPrometheus bool) otel.ReceiverPipeline {
+func ConvertGCMSystemExporterToOtlpExporter(pipeline otel.ReceiverPipeline, ctx context.Context) otel.ReceiverPipeline {
+	return ConvertToOtlpExporter(pipeline, ctx, false, true)
+}
+
+func ConvertToOtlpExporter(pipeline otel.ReceiverPipeline, ctx context.Context, isPrometheus bool, isSystem bool) otel.ReceiverPipeline {
```
franciscovalentecastro, line 83: It is great that now all logic to setup the OTLP exporter instead of the GCM exporter is contained within `ConvertToOtlpExporter`! 

#### Question

I want to followup on this comment made above : 

> The problem is that the transformation is different depending if the current exporter is configured as a system or not. If I just declare it as OTLP, I don' t know how to properly setup the pipeline later. It can either be a GMP, GCM system or GCM Otel exporter.

When the migration to the OTLP exporter is fully completed and we remove all the other exporters (System, OTel , GMP) : 

- How are we going to determine which combination of processors (isSystem, isPrometheus) should be applied to each metric pipeline ? 
- Should we add another ReceiverPipeline "property" to differentiate them ?
- When we remove the `GMP` exporter, how is the function  `ConvertPrometheusExporterToOtlpExporter` going to be renamed/refactored ?

I bring up this questions, since i feel the solution of using a "converter"  (`ConvertToOtlpExporter`) is considering only the intermediate state where `GCM` is the main exporter, but will need further refactoring when those are removed. My suggestion here aims to find a solution that won't need any refactoring later.

#### Suggestion
One idea is to have 3 exporter types which help determine the specific processors needed for each pipeline (while keeping the possibility to store the logic in one place ) : 
- System_OTLP
- GMP_OTLP
- OTLP

This keeps the abstraction of a "ReceiverPipeline" while also having a way to differentiate the specific requirements of each metric pipeline.

franciscovalentecastro: Synced offline with @rafaelwestphal. Mentioned eventually we are going to consolidate all `OTLP exporter` processor logic when the migration is complete. Using a `ConvertToOtlpExporter` is considered an intermediate step since it enables development to happen for each receiver separately.

confgenerator/confgenerator.go
```diff
@@ -306,6 +313,14 @@ func (p PipelineInstance) OTelComponents(ctx context.Context) (map[string]otel.R
 			)
 		}
 
+		expOtlpExporter := experimentsFromContext(ctx)["otlp_exporter"]
+		if processors, ok := receiverPipeline.Processors["metrics"]; ok && expOtlpExporter {
+			receiverPipeline.Processors["metrics"] = append(
+				processors,
+				otel.MetricStartTime(),
```
franciscovalentecastro: [Requested change] Create an `otlpExporterComponents` (or `otlpExporterProcessor`) function that contains `MetricStartTime()` and also will serve for future additions of OTLP exporter logic.

confgenerator/confgenerator.go
```diff
@@ -306,6 +313,14 @@ func (p PipelineInstance) OTelComponents(ctx context.Context) (map[string]otel.R
 			)
 		}
 
+		expOtlpExporter := experimentsFromContext(ctx)["otlp_exporter"]
+		if processors, ok := receiverPipeline.Processors["metrics"]; ok && expOtlpExporter {
```
franciscovalentecastro: [Comment] Just as a general comment of our offline discussion, IMO here is the best place to consolidate all `OTLP` exporter logic (e.g. post-processors) since it resembles how `otelSetLogNameComponents` is unilaterally added in all logging pipelines. 

I believe you can remove all uses of `ConvertToOtlpExporter` and add all the "OTLP" logic here. Please keep track of this ideas as the migration moves forward.

confgenerator/testdata/goldens/metrics-exporter_prometheus_otlp/golden/linux-gpu/otel.yaml
```diff
@@ -86,6 +86,10 @@ processors:
         - googlecloudmonitoring/point_count
   interval/loggingmetrics_7:
     interval: 1m
```
franciscovalentecastro, line 86: [Requested changed] Add a `confgenerator/testdata` case for self metrics to track future regressions on "OTLP exporter self metrics" pipelines.

confgenerator/otel/modular.go
```diff
@@ -256,6 +256,12 @@ func (c ModularConfig) Generate(ctx context.Context, expOtlpExporter bool) (stri
 				copyProcessor := CopyHostIDToInstanceID()
 				processorNames = append(processorNames, copyProcessor.name("_global_0"))
 				processors[copyProcessor.name("_global_0")] = copyProcessor.Config
+				// Similar to the resource detector, for any pipeline that is using the
```
franciscovalentecastro: ### **[Requested Change]**

After seeing the implementation of `resourceDetectionProcessorNames` above :

**Why not create a similar pair of  `exporterTypeProcessors` and `exporterTypeProcessorNames` to handle the adding processors at the end of every pipeline ?**

```
exporterType := receiverPipeline.ExporterTypes[pipeline.Type]
if name, ok := exporterTypeProcessorNames[exporterType]; ok {
			processors[name] = exporterTypeProcessors[exporterType].Config
}
```

This also removes the need to have the `expOtlpExporter` flag here, since everything is encoded by using the `otel.OTLP` exporter. Also, the current implementation `if expOtlpExporter && resourceDetectionNames` is adding the processors in pipelines that don't use the OTLP exporter.

https://github.com/GoogleCloudPlatform/ops-agent/blob/d387e0aebae954648c980ed9503f3c85223a0211/confgenerator/testdata/goldens/metrics-exporter_prometheus_otlp/golden/windows/otel.yaml#L887-L900

### **[Context]**

For Otel Logging, I've found a need (see https://github.com/GoogleCloudPlatform/ops-agent/pull/2165) for a separate `otel.Logs` exporter to handle "Log Buffering" settings (which can't be applied to the Metrics Exporters). I also needed to add a `Batch` processor at the end of the pipeline for "log buffering" and overall "logs export performance".

This is my draft of the idea i propose above. IMO, this is extendable and encodes all this "processor logic" depending on the exporter : 

https://github.com/GoogleCloudPlatform/ops-agent/blob/fa9c5ee2a304c34d25c08cec9426e8db7ca2b76a/confgenerator/otel/modular.go#L226-L273

franciscovalentecastro (changes requested):

confgenerator/otel/modular.go
franciscovalentecastro: P.D. `exporterTypeProcessors` could be of type `[]Components` so multiple processors can be added at the end of the pipeline.

franciscovalentecastro (approved):
I'll keep the approval (to not block the PR) modulo addressing the `[Requested Change]` above.

PR #2095, opened 2025-10-16
surface errors via GetStatus

XuechunHou:
## Description
Use `GetStatus` to surface errors thrown from other lifecycle methods. 

## Related issue
b/452138776

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -95,41 +95,61 @@ func (ps *OpsAgentPluginServer) Start(ctx context.Context, msg *pb.StartRequest)
...
+	}
+}
+
+func (ps *OpsAgentPluginServer) logAndSetPluginError(errMsg string, isFatal bool) {
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+
+	ps.pluginError = &OpsAgentPluginError{
+		Message: errMsg,
+		IsFatal: isFatal,
```
jinghan-ma: It's not immediately clear from reading the code when `isFatal` should be true, especially critical errors that render the Ops Agent useless (e.g. bad config) are not considered "Fatal" by UAP standards.

Maybe we change this to something like `needsRestart` or `shouldRestart` to be more explicit?

cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -95,41 +95,61 @@ func (ps *OpsAgentPluginServer) Start(ctx context.Context, msg *pb.StartRequest)
...
+
 	// the subagent startups
-	cancelFunc := func() {
-		ps.Stop(ctx, &pb.StopRequest{Cleanup: false})
-	}
-	go runSubagents(pContext, cancelFunc, pluginInstallDir, pluginStateDir, runSubAgentCommand, ps.runCommand)
+	go runSubagents(pContext, ps.cancelAndSetPluginError, pluginInstallDir, pluginStateDir, runSubAgentCommand, ps.runCommand)
 	return &pb.StartResponse{}, nil
 }
+func (ps *OpsAgentPluginServer) cancelAndSetPluginError(e *OpsAgentPluginError) {
```
jinghan-ma: Is there a reason why we don't log the error when we want to cancel and set plugin error?

cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -95,41 +95,61 @@ func (ps *OpsAgentPluginServer) Start(ctx context.Context, msg *pb.StartRequest)
...
+	if e != nil {
+		ps.pluginError = e
+	}
+}
+
+func (ps *OpsAgentPluginServer) logAndSetPluginError(errMsg string, isFatal bool) {
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+
+	ps.pluginError = &OpsAgentPluginError{
```
jinghan-ma: Should we pre-define `OpsAgentPluginError`s somewhere at the top instead? This way log and set plugin error can just take a `OpsAgentPluginError` or the error's ErrorCode. It also makes it so that the invoker of the error does not need to worry about whether a particular error is fatal (which would be defined already during the definition at the top).

cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -158,8 +179,12 @@ func (ps *OpsAgentPluginServer) GetStatus(ctx context.Context, msg *pb.GetStatus
 	ps.mu.Lock()
 	defer ps.mu.Unlock()
 	if ps.cancel == nil {
-		log.Println("The Ops Agent plugin is not running")
-		return &pb.Status{Code: 1, Results: []string{"The Ops Agent Plugin is not running."}}, nil
+		if ps.pluginError != nil {
```
jinghan-ma: Is it correct to check `ps.pluginError` only if `ps.cancel` is nil? Wouldn't there be cases where the sub-agent processes are still running (`ps.cancel != nil`) but there are still errors that need to be surfaced? Maybe that case doesn't exist currently, but the code should still handle it.

cmd/ops_agent_uap_plugin/service_linux.go
XuechunHou: Done.

cmd/ops_agent_uap_plugin/service_linux.go
XuechunHou: Done

cmd/ops_agent_uap_plugin/service_linux.go
XuechunHou: Done

jinghan-ma (approved):
LGTM

PR #2100, opened 2025-10-21
[confgenerator] Add confgenerator tests for Otel Logging.

franciscovalentecastro:
## Description
This PR adds an individual `otel_logging.yaml` file for each existing confgenerator test to help track the state of Otel Logging features.

~This PR adds two platforms `linux-otel-logging` and `windows-otel-logging` to confgenerator tests to verify the validaty and change of configurations when "otel logging" is enabled.~

Relevant previous PR : https://github.com/GoogleCloudPlatform/ops-agent/pull/2021.

## Related issue
b/454066447

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/confgenerator_test.go
```diff
@@ -264,6 +312,11 @@ func generateConfigs(pc platformConfig, testDir string) (got map[string]string,
 	if err != nil {
 		return
 	}
+
+	if pc.otelLoggingEnabled {
+		mergedUc.Logging.Service.OTelLogging = true
```
quentinmit: My intent was that we would have specific otel logging tests that already have a checked-in config file with `OtelLogging = true` (and an `EXPERIMENTAL_FEATURES` file to enable the experiment).

Do we really need to special-case it here? We don't do that for any of our other global flags.

franciscovalentecastro: > Do we really need to special-case it here?

With this specific PR proposed idea of adding `linux-otel-logging` and `windows-otel-logging` folders to all tests, there has to be a way to enable both the `EXPERIMENTAL_FEATURES` env variable and the `cofig.yaml (experimental_otel_logging: true)`. So `mergedUc.Logging.Service.OTelLogging = true` is indeed needed.

We could modify `config.yaml` files to have this setting too, thought that maybe more complicated. I'll explore the solutions with all this feedback. Thanks for the comment!

franciscovalentecastro:
Discussed offline further with @jefferbrecht and @quentinmit . Closing this PR in favour of a solution that deduplicates `goldens`. That would bring a better experience overall. In the meantime i'll keep adding tests per otel logging feature.

PR #2103, opened 2025-10-24
[confgenerator] Support `LoggingProcessorParseMultilineRegex` in Otel Logging.

franciscovalentecastro:
## Description
Implement `LoggingProcessorParseMultilineRegex` and `LoggingProcessorParseRegexComplex` in Otel Logging.

## Details
- Created `logging-otel-receiver_kafka` confgenerator test to validate resulting config.
- Updated a lot of relevant transformation test goldens.

## Related issue
b/440599473

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/logging_processors.go
```diff
@@ -437,6 +459,60 @@ func (p LoggingProcessorParseMultilineRegex) Components(ctx context.Context, tag
...
+		// The current "recombine" operator multiline support only supports setting a "start_state" ("is_first_entry").
+		// TODO: b/459877163 - Update implementation when opentelemetry supports "state-machine" multiline parsing.
+		if r.StateName == "start_state" {
+			isFirstEntry = append(isFirstEntry, r.Regex)
+		}
+	}
+
+	var exprParts []string
+	for _, r := range isFirstEntry {
+		exprParts = append(exprParts, fmt.Sprintf("body.message matches %q", r))
```
jefferbrecht: Why not build `exprParts` directly in the first loop and eliminate `isFirstEntry` altogether?

confgenerator/logging_processors.go
```diff
@@ -437,6 +459,60 @@ func (p LoggingProcessorParseMultilineRegex) Components(ctx context.Context, tag
...
 }
 
+func (p LoggingProcessorParseMultilineRegex) Processors(ctx context.Context) ([]otel.Component, error) {
+	var isFirstEntry []string
+	for _, r := range p.Rules {
+		// The current "recombine" operator multiline support only supports setting a "start_state" ("is_first_entry").
+		// TODO: b/459877163 - Update implementation when opentelemetry supports "state-machine" multiline parsing.
+		if r.StateName == "start_state" {
+			isFirstEntry = append(isFirstEntry, r.Regex)
+		}
```
jefferbrecht, line 467: I assume that by ignoring some states, there will be some possible log inputs that won't parse properly as multiline. (If that's not true, then we should be able to refactor the receivers to only have `start_state`.)

How do you want to approach testing for that gap? E.g. will you add/change transformation tests later along with b/459877163 to validate whichever edge cases wouldn't work today without a full state machine?

confgenerator/logging_processors.go
franciscovalentecastro: No real reason. In the other `parse_multiline` PR, we have a more complicated expressions setup, so it made more sense to do it in steps. I simplified it. Done!

confgenerator/logging_processors.go
franciscovalentecastro: All the current uses of `LoggingProcessorParseMultilineRegex` [^1] only set two states `start_state` and `cont_state` in a simplified manner such that `start_state = FirstLogLineRegex` and `cont_state = Negation of "FirstLogLineRegex"` (note : i've just double checked [^1], also `git grep -C 10 "start_state"` helps).

[^1]: https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fops-agent%20LoggingProcessorParseMultilineRegex&type=code

https://github.com/GoogleCloudPlatform/ops-agent/blob/ccfedc9126991ebe77ab06ba57d2280e2623ac26/apps/solr.go#L84-L94

This implies all current uses of `LoggingProcessorParseMultilineRegex` can be fully replicated by only setting "is_first_entry" in otel logging. 

### Proposed Refactor
We could refactor `LoggingProcessorParseMultilineRegex` to only be able to set a `start_state` and then set `cont_state` programatically as the "negation" of the "start_state". This will enforce this simplified use of multiline features.

> How do you want to approach testing for that gap? E.g. will you add/change transformation tests later along with b/459877163 to validate whichever edge cases wouldn't work today without a full state machine?

Re @jefferbrecht 

It depends. What do you think of the `Proposed Refactor` ? 

If we refactor `LoggingProcessorParseMultilineRegex` to only set a `start_state`, then there won't be any feature gaps and the 3P app receiver tests are good enough for this.

If we don't refactor it, creating a "transformation_test" would be artificial since we would need to create a "test processor" that uses all the feature of the "state-machine". A "transformation_test" needs a "registered" processor to be able to add to a pipeline.

franciscovalentecastro: NVM, not "all" 3P app receiver implementations use the simplified set of `start_state` and `cont_state`. Though it's only `mysql_slow` and `elasticsearch_json` the ones that use a more complicated (not too much) set of regexes.

> How do you want to approach testing for that gap? E.g. will you add/change transformation tests later along with b/459877163 to validate whichever edge cases wouldn't work today without a full state machine?

The tests of `mysql_slow` and `elasticsearch_json` can serve to compare with the use of all "state-machine" like features.

See draft refactor : https://github.com/GoogleCloudPlatform/ops-agent/pull/2125/commits/dc991f68c6e7a34f928a72995a5fd2bffac33ace#diff-7def08d2dee0c2606af18bb82d03c649a7ece83d1fb913fcfb800c6558eb942e

jefferbrecht: We don't necessarily need to refactor; what I'm mainly after is a set of transformation tests that covers any cases that exercise those complex state transitions, which should pass on fluent-bit today, and may fail/be wrong on OTel, but can be made to pass on OTel later once upstream has better support for state machine-like multiline rules.

franciscovalentecastro: Ok! That makes sense. I thought about this and here is a set of tests that can cover the current differences :

- **[New Test]** I created test test `logging_processor-parse_multiline_regex` that exercises complex state transitions. The outputs in fluent-bit and otel are different.
    - I registered `LoggingProcessorParseMultilineRegex` in `transformation_test.go` in way that can be used only for transformation tests. This still can't be configured by users.

- **[Existing test]** `mysql_slow` and `elasticsearch_json` also have complex "state-machine" definitions that will differ between otel and fluent-bit.

Some additional details : 
- `mysql_slow` and `elasticsearch_json` can't be fully replicated with the current Otel Support, so now i'm sure we need the upstream changes to fully support for this receivers.

franciscovalentecastro: Update : 
- Synced with @jefferbrecht offline. The tests for `mysql_slow` and `elasticsearch_json` are enough since they have complex "state-machine" definitions that will differ between otel and fluent-bit.
- Removed the `[New Test]` named `logging_processor-parse_multiline_regex`.
- I'll do a followup PR to support `YesNoBoolean` casting (used by `mysql_slow`) and `LoggingProcessorNestWildcard` (used by `elasticsearch_json`) to have both tests working.

apps/saphana.go
```diff
@@ -59,11 +59,11 @@ func (p LoggingProcessorMacroSapHanaTrace) Expand(ctx context.Context) []confgen
...
-					"thread_id":             "int",
-					"connection_id":         "int",
-					"transaction_id":        "int",
-					"update_transaction_id": "int",
-					"source_line":           "int",
+					"thread_id":             "integer",
+					"connection_id":         "integer",
+					"transaction_id":        "integer",
+					"update_transaction_id": "integer",
+					"source_line":           "integer",
```
jefferbrecht: What's the behaviour before-and-after for this change? E.g. were the fields accidentally typed as strings before and now they'll be property typed as ints? Is this a breaking change?

franciscovalentecastro: > What's the behaviour before-and-after for this change?

The correct way to set the output field as an "integer" type using `modify_fields` is to set the "integer" not "int".

https://github.com/GoogleCloudPlatform/ops-agent/blob/e97f988308580df1dcbc0429530db77285f3d62e/confgenerator/logging_modify_fields.go#L198-L215

Now all fields are going to show are as integers in the LogEntry instead of "strings". This is true for fluent-bit and otel.

image


> Were the fields accidentally typed as strings before and now they'll be property typed as ints?

Yes. I fixed this since Otel Logging was showing an "OTTL" error due to this bug.

> Is this a breaking change?

It's a bug fix, since the original intention of the code was for those fields to be typed as integers, not strings. Not sure if this will break customers.

confgenerator/logging_processors.go
```diff
@@ -389,6 +390,27 @@ func (p LoggingProcessorParseRegexComplex) Components(ctx context.Context, tag,
 	return components
 }
 
+func (p LoggingProcessorParseRegexComplex) Processors(ctx context.Context) ([]otel.Component, error) {
+	if len(p.Parsers) == 0 {
+		return []otel.Component{}, nil
+	}
```
jefferbrecht: Nit: you can remove this check, it'll return an empty slice anyway if there are no parsers

apps/saphana.go
franciscovalentecastro: Reverted "saphana" fix : https://github.com/GoogleCloudPlatform/ops-agent/pull/2103/commits/e6dc551ec72936ae9a6770a2942beb91345083d3.

confgenerator/logging_processors.go
franciscovalentecastro: Done in : https://github.com/GoogleCloudPlatform/ops-agent/pull/2103/commits/c60aa1d8d45f160ff004c6d48eac21fc6d950a34.

franciscovalentecastro:
All 3P app test failures are related to `Quota 'NVIDIA_L4_GPUS' exceeded` errors. Merging.

PR #2143, opened 2025-11-26
[confgenerator] Add `LoggingReceiverWindowsEventLog` missing features in Otel Logging.

franciscovalentecastro:
## Description
This PR implements the missing V2 and `RenderAsXML` features of the `windows_event_log` receiver. It improves slightly the V1 fields too.

Some details : 
- Created (or improved) the following transformation tests (only for otel processors) : 
  - `logging_processor-windows_event_log_raw_xml`
  - `logging_processor-windows_event_log_v2`
  - `logging_processor-windows_event_log_v1`
- Created the following confgenerator tests :
  - `windows-otel-logging-receiver_winlog2_new_channels`
  - `windows-otel-logging-receiver_winlog2_xml`
- Enabled the following integration tests : 
  - `TestWindowsEventLog`
  - `TestWindowsEventLogV1UnsupportedChannel`
  - `TestWindowsEventLogV2`
  - `TestWindowsEventLogWithNonDefaultTimeZone`

## Related issue
b/413406335

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/logging_receivers.go
```diff
@@ -598,28 +598,35 @@ func (r LoggingReceiverWindowsEventLog) Components(ctx context.Context, tag stri
...
+			"start_at":              "beginning",
+			"poll_interval":         "1s",
+			"ignore_channel_errors": true,
 			// TODO: Configure storage
 		}
-		if r.RenderAsXML {
-			receiver_config["raw"] = true
-			// TODO: Rename to `jsonPayload.raw_xml`
+
+		p, err := windowsEventLogV1Processors(ctx)
```
jefferbrecht: Nit: can we clean up the dispatching a bit here so that we don't (re)assign `p, err` three times and repeat the error check three times?

E.g.:

```go
var p []otel.Component
var err error

if (r.IsDefaultVersion()) {
	p, err = windowsEventLogV1Processors(ctx)
} else if (r.RenderAsXml) {
	receiver_config["include_log_record_original"] = true
	p, err = windowsEventLogRawXMLProcessors(ctx)
} else {
	p, err = windowsEventLogV2Processors(ctx)
}

if err != nil {
	return nil, err
}
```

confgenerator/logging_receivers.go
```diff
@@ -598,28 +598,35 @@ func (r LoggingReceiverWindowsEventLog) Components(ctx context.Context, tag stri
 }
 
 func (r LoggingReceiverWindowsEventLog) Pipelines(ctx context.Context) ([]otel.ReceiverPipeline, error) {
```
jefferbrecht, line 600: IIUC, the fluent-bit version supports [bookmarking](https://github.com/GoogleCloudPlatform/ops-agent/blob/3e5b9daba6a81d07228a0e1fbf5f61ada34de776/confgenerator/logging_receivers.go#L542), and the OTel component supports it as well, but we're not using it here. Are we planning to implement bookmarking later? It's missing from the files receiver too for example.

transformation_test/testdata/logging_processor-windows_event_log_raw_xml/output_fluentbit.yaml
```diff
@@ -0,0 +1,68 @@
+- entries:
```
jefferbrecht, line 1: This looks wrong: the fields are all OTel-ish. Is this actually running fluent-bit?

confgenerator/logging_receivers.go
franciscovalentecastro: I'm planning to add the "bookmarking" in the same task of otel logging "Log Buffering" (b/394636880) using the `filestorage` and `sendingqueue` features ( see [^1] for example ). The bookmarking in the `windowseventlogreceiver` [^2] is set by creating an instance of the `file_storage` extension. I want to add this to the `filelogsreceiver` in the same PR.

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/examples/fault-tolerant-logs-collection

[^2]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.138.0/receiver/windowseventlogreceiver

confgenerator/logging_receivers.go
franciscovalentecastro: Done!

confgenerator/logging_receivers.go
franciscovalentecastro: After discussing offline, i'll separate in two tasks the "Log Buffering" and "Bookmarking" since their goal is fundamentally different (although they use similar `storage`) features. This will be done in a separate PR.

transformation_test/testdata/logging_processor-windows_event_log_raw_xml/output_fluentbit.yaml
franciscovalentecastro: > This looks wrong: the fields are all OTel-ish. Is this actually running fluent-bit?

The `window_event_log` transformation tests (e.g. `logging_processor-windows_event_log_v2`) are worth a comment/explanation. The main difference is that **"they are only testing "OTel Processors" and not Fluent-bit processors**.

Here are some of the reasons this is set this way (and possibly open for feedback to improve the tests) :

1.  "windwos_event_log" transformation tests are only meant to test "processors", not "receivers".
    - The logging processor `LoggingProcessorWindowsEventLogV1` are only created/registered in transformation tests (users can't configure this processor). 
    - This enables us to do tests on the "OTel" processor to make sure the output LogEntry format is similar to the "fluent-bit" format.
    - The logging processor `LoggingProcessorWindowsEventLogV1` has an empty "fluent-bit" implementation. The input from the "winlog" (or "wineventlog") fluent-bit receiver is fundamentally different from the one that otel generates.
    - This enables to test this complex log format before running integration tests.
  
2. This tests mock the input from the otel "windowseventlogreceiver" :
    -  We use sample "json" maps from the "windowseventlogreceiver" standalon. With the "parse_json" and  "modify_fields" processors we mock this receiver output.
    - This input is obtained from the "windowseventlogreceiver" otel log receiver, there is no way to convert the fluent-bit processor implementation / mock input to resemble the "winlog" (or "wineventlog") fluent-bit receivers.

I will add comments and a "static" messages like "this processor only works for 'otel'" in the `output_fluentbit.yaml` for clarity.

confgenerator/logging_receivers.go
```diff
@@ -738,6 +738,150 @@ func windowsEventLogV1Processors(ctx context.Context) ([]otel.Component, error)
...
+			"jsonPayload.Qualifiers":   {CopyFrom: "jsonPayload.event_id.qualifiers"},
+			"jsonPayload.StringInserts": {
+				CopyFrom: "jsonPayload.event_data",
+				CustomConvertFunc: func(v ottl.LValue) ottl.Statements {
+					eventData := ottl.LValue{v.String(), "data"}
+					eventBinary := ottl.LValue{v.String(), "binary"}
+					cacheEventData := ottl.LValue{"cache", "__event_data"}
+					return ottl.NewStatements(
+						cacheEventData.SetIf(ottl.ToValues(eventData), eventData.IsPresent()),
+						cacheEventData.AppendValuesIf(eventBinary, ottl.And(cacheEventData.IsPresent(), eventBinary.IsPresent())),
```
jefferbrecht, line 796: Can you test what happens if the record _only_ has `Binary` and not `Data`? Will `AppendValuesIf` fail if the preceding `SetIf` doesn't end up setting anything on `cacheEventData`?

jefferbrecht: Nevermind, according to the docs it'll create the field if it doesn't exist when appending, and I got curious and tried it myself and it seems to indeed work that way.

PR #2148, opened 2025-12-02
Changing logging metrics to use starttime processor

rafaelwestphal:
## Description
The current implementation implicitly relies on the exporter to set the starttime of the metric.
This makes the behavior explicit. It is also required to move to the OTLP exporter, since it does not adjust the starttime.

## Related issue
b/452074867

## How has this been tested?
Current integration tests.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/confgenerator/agentmetrics.go
+++ b/confgenerator/agentmetrics.go
@@ -285,7 +285,9 @@ func (r AgentSelfMetrics) LoggingMetricsPipelineProcessors() []otel.Component {
 		// DeltaToCumulative keeps in memory information of previous delta points
 		// to generate a valid cumulative monotonic metric.
 		otel.DeltaToCumulative(),
-    otel.MetricStartTime(),
+		otel.MetricStartTime(),
+		// Transfrom from double -> int64 after we apply the metricstarttime processor
+		// This prevents us from hitting b/
 		otel.MetricsTransform(
 			otel.UpdateMetric("agent/log_entry_retry_count",
 				// change data type from double -> int64


confgenerator/agentmetrics.go
```diff
@@ -297,6 +285,23 @@ func (r AgentSelfMetrics) LoggingMetricsPipelineProcessors() []otel.Component {
 		// DeltaToCumulative keeps in memory information of previous delta points
 		// to generate a valid cumulative monotonic metric.
 		otel.DeltaToCumulative(),
+		otel.MetricStartTime(),
+		// Transfrom from double -> int64 after we apply the metricstarttime processor
+		// This prevents us from hitting b/
```
franciscovalentecastro, line 290: Please add a bug tracker and summarize briefly the issue.

confgenerator/agentmetrics.go
```diff
@@ -297,6 +285,23 @@ func (r AgentSelfMetrics) LoggingMetricsPipelineProcessors() []otel.Component {
 		// DeltaToCumulative keeps in memory information of previous delta points
 		// to generate a valid cumulative monotonic metric.
 		otel.DeltaToCumulative(),
+		otel.MetricStartTime(),
```
franciscovalentecastro, line 288: What `strategy` are we using to set the start times ? Is this equivalent to the Google Cloud exporter `strategy` ?

rafaelwestphal: It uses the subtract_initial_point strategy
```
metricstarttime/loggingmetrics_7:
  strategy: subtract_initial_point
```

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/metricstarttimeprocessor/internal/subtractinitial/adjuster.go#L217-L226

This is what is done on the exporter:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/internal/normalization/standard_normalizer.go#L250-L257

From my understanding, they both drop the first point when the starttime is zero and cache it as a reference value for the next points.

franciscovalentecastro: Sounds good ! Thank you! Yes, it makes sense to use that strategy as a replacement for the exporter "standard_normalization".

rafaelwestphal:
This issue has been fixed by https://github.com/GoogleCloudPlatform/ops-agent/pull/2152 and this workaround is no longer needed.

PR #2149, opened 2025-12-02
Add makefile support for fine-grained test runs

jefferbrecht:
## Description
Add makefile support for fine-grained test runs

## Related issue
http://b/462191335

## How has this been tested?
Ran a bunch of local tests for various combinations of the new variables introduced

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/tasks.mak
+++ b/tasks.mak
@@ -148,30 +148,45 @@ transformation_test_update: transformation_test
 # these targets. The Makefile will not ascribe defaults for these, since they
 # are specific to different user environments.
 #
-# Defaults are provided for ZONES and PLATFORMS.
+# Defaults are provided for ZONES and IMAGE_SPECS.
 #
 # If you would like to use a custom build you can provide a REPO_SUFFIX or
 # AGENT_PACKAGES_IN_GCS. These targets function fine without either specified.
+#
+# INTEGRATION_TEST_FILTER gets passed directly to -test.run, and lets you
+# specify which integration tests to run. All tests run by default.
+#
+# THIRD_PARTY_APPS lets you specify which apps to run. All apps run by default.
+#
+# THIRD_PARTY_APPS_ENDPOINT controls which endpoint to use and defaults to
+# googlecloudmonitoring. Allowed values are googlecloudmonitoring, otlphttp,
+# or both using (googlecloudmonitoring|otlphttp).
+
 ZONES ?= us-central1-b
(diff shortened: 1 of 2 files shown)


tasks.mak
```diff
@@ -148,30 +148,45 @@ transformation_test_update: transformation_test
...
 integration_tests:
 	ZONES="${ZONES}" \
-	PLATFORMS="${PLATFORMS}" \
+	IMAGE_SPECS="${IMAGE_SPECS}" \
 	go test -v ./integration_test/ops_agent_test/main_test.go \
 	-test.parallel=1000 \
 	-tags=integration_test \
-	-timeout=4h
+	-timeout=4h \
+	-test.run='$(INTEGRATION_TEST_FILTER)'
```
franciscovalentecastro: Is there a difference with setting the filter with `test.run` or `run` ? Got curious also about `test.parallel`.

jefferbrecht: 🤷 They both seem to work; the docs do just say `-run` and `-parallel`, so I'll change it to those.

braydonk: `-test.run` and `-run` are equivalent; iirc `-test.run` is a backup on the offchance that a user has defined a `-run` flag of their own. (I'm in favour of just using `-run` though)

tasks.mak
```diff
@@ -148,30 +148,45 @@ transformation_test_update: transformation_test
...
 	ZONES="${ZONES}" \
-	PLATFORMS="${PLATFORMS}" \
+	IMAGE_SPECS="${IMAGE_SPECS}" \
 	go test -v ./integration_test/ops_agent_test/main_test.go \
-	-test.parallel=1000 \
+	-parallel=1000 \
 	-tags=integration_test \
-	-timeout=4h
+	-timeout=4h \
+	-run='$(INTEGRATION_TEST_FILTER)'
```
braydonk: nit: we have a mixture of `$()` and `${}`. I'm 99% sure it doesn't actually cause any problems but we maybe should just pick one? In `opentelemetry-operations-collector` I standardized on `$()` for a very vague reason: sometimes I want to format in specifically a bash variable and not a make variable, meaning I do `$${}`, and in that case braces are an additional indicator to myself that I do in fact mean the bash context.

jefferbrecht: Oops, thanks for pointing that out -- fixed

PR #2151, opened 2025-12-03
Update Fluent Bit to v4.0.13

braydonk:
## Description

This PR updates Fluent Bit to v4.0.13 to pick up some CVE fixes.

## Related issue

b/465698843 

## How has this been tested?
The presubmits will be the test.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/transformation_test/transformation_test.go
+++ b/transformation_test/transformation_test.go
@@ -61,23 +61,23 @@ var (
 	otelopscolPath = flag.String("otelopscol", os.Getenv("OTELOPSCOL"), "path to otelopscol")
 
 	multilineTestPatterns = newTestMatchPatterns([]string{
-		".*flink.*",
-		".*elasticsearch.*",
-		".*mysql.*",
-		".*hbase.*",
 		".*cassandra.*",
 		".*couchdb.*",
+		".*elasticsearch.*",
+		".*flink.*",
 		".*hadoop.*",
+		".*hbase.*",
 		".*kafka.*",
+		".*mysql.*",
+		".*oracledb.*",
 		".*postgresql.*",
 		".*rabbitmq.*",
+		".*saphana.*",
+		".*solr.*",
+		".*tomcat.*",
+		".*vault.*",
 		".*wildfly.*",
 		".*zookeeper.*",
-		".*vault.*",
-		".*solr.*",
-		".*oracledb.*",
-		".*saphana.*",
-		".*tomcat.*",
 	})
 )
 
@@ -317,8 +317,14 @@ func generateFluentBitConfigs(ctx context.Context, name string, transformationTe
 				// - https://github.com/fluent/fluent-bit/pull/8545
 				// On newer fluent-bit 4.0.x versions, last log in a file maybe (non-deterministically)
 				// dropped (~%85 retries) or sent (~15% retries) causing flaky tests.
+
 				// Set shutdown "Grace" period to 0s to avoid any unreliable logs to be sent after Exit_On_Eof.
-				// This forces the last log line from an multiline parser to always be dropped.
+				// Set the "Flush" time to 10s, which fixes a race condition in multiline tests that
+				// would sometimes perform a final flush and cause the last line to appear.
+				// (Started in Fluent Bit 4.0.13)
+				//
+				// These settings in combination forces the last log line from any multiline parser
+				// to always be dropped.
 				"Grace": "0",
 				"Flush": "10",
 			},


builds/otel.sh
```diff
@@ -47,11 +47,16 @@ if [ "$SKIP_OTEL_JAVA" != "true" ]; then
...
+fi
+ARCH=$($GO_BIN env GOARCH)
+
 cd submodules/opentelemetry-operations-collector/otelopscol
-ARCH=$(/usr/local/go/bin/go env GOARCH)
-GO_BIN=/usr/local/go/bin/go \
-BUILDARCH=$ARCH \
-TARGETARCH=$ARCH \
+
+BUILDARCH="$ARCH" \
```
franciscovalentecastro, line 57: I'm curious about this update. At some point i had to install directly Go with a newer version in my local dev environment because `ARCH=$(/usr/local/go/bin/go env GOARCH)` was failing.

What is the goal/reasoning for this update ?

tasks.mak
```diff
@@ -128,7 +128,7 @@ dist/opt/google-cloud-ops-agent/subagents/fluent-bit/bin/fluent-bit:
 	bash ./builds/fluent_bit.sh $(PWD)/dist
 
 dist/opt/google-cloud-ops-agent/subagents/opentelemetry-collector/otelopscol:
-	SKIP_OTEL_JAVA=${SKIP_JAVA} bash ./builds/otel.sh $(PWD)/dist
+	SKIP_OTEL_JAVA=${SKIP_JAVA} GO_BIN=$(shell which go) bash ./builds/otel.sh $(PWD)/dist
```
franciscovalentecastro, line 131: Could this potentially pick an "incorrect" Go version from the system? For example, if builds machines have an existing go installation, but we also install a specific version for builds.

transformation_test/transformation_test.go
```diff
@@ -260,22 +298,29 @@ func generateFluentBitConfigs(ctx context.Context, name string, transformationTe
...
+				// The combination of Exit_On_Eof on a tail receiver with a multiline parser causes
+				// the last log in a file to be dropped. See :
+				// - https://github.com/fluent/fluent-bit/issues/8623
+				// - https://github.com/fluent/fluent-bit/issues/8353
+				// - https://github.com/fluent/fluent-bit/issues/3926
+				// Some attempts of a solution have been implemented :
+				// - https://github.com/fluent/fluent-bit/pull/8545
+				// On newer fluent-bit 4.0.x versions, last log in a file maybe (non-deterministically)
+				// dropped (~%85 retries) or sent (~15% retries) causing flaky tests.
+				// Set shutdown "Grace" period to 0s to avoid any unreliable logs to be sent after Exit_On_Eof.
```
franciscovalentecastro, line 321: [nit] Mention `Flush : 10s` in this comment.

builds/otel.sh
braydonk: This script was naively only supporting Go installations at `/usr/local/go/bin/go`. On my machine, Go versions are managed by VSCode and installed in different places. This update is purely to allow for local builds to specify a Go installation to use (it won't affect the Dockerfiles using this script).

tasks.mak
braydonk: This target is only used for local development and won't affect our CI builders. They will still invoke the script without providing a `GO_BIN` which will cause it to default to `/usr/local/go/bin/go` which is where Go is installed in all build containers.

transformation_test/transformation_test.go
```diff
@@ -59,8 +59,42 @@ const (
 var (
 	flbPath        = flag.String("flb", os.Getenv("FLB"), "path to fluent-bit")
 	otelopscolPath = flag.String("otelopscol", os.Getenv("OTELOPSCOL"), "path to otelopscol")
+
+	multilineTestPatterns = newTestMatchPatterns([]string{
```
franciscovalentecastro, line 63: [nit] Please sort the list for future reference.

Using the list of 3p app processors [^1] that use Multiline Regex, i think this is complete.

[^1]: https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fops-agent%20ParseMultilineRegex&type=code

builds/otel.sh
franciscovalentecastro: > On my machine, Go versions are managed by VSCode and installed in different places.

Ahhh! Thanks for the clarification! That was my issue, VSCode was handling the different Go versions, so i installed one main version in that path to avoid the issue.

franciscovalentecastro (approved):
LGTM modulo addressing the open comments.

PR #2166, opened 2025-12-19
[confgenerator] Add `file offset` storage to Otel Logging receivers.

franciscovalentecastro:
## Description
Configuring the `file_storage` extension to be used as `storage` in Otel Logging receivers.

Some details : 
- Added to `systemd`, `files` and `windows_event_log` logging receiver.
- The following folders will be used to store the bookmarks : 
  - Linux : `/var/lib/google-cloud-ops-agent/opentelemetry-collector\file_storage/`
  - Windows : `C:\ProgramData\Google\Cloud Operations\Ops Agent\run\file_storage`

## Related issue
b/469432672

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test/main_test.go
+++ b/integration_test/ops_agent_test/main_test.go
@@ -6005,7 +6005,7 @@ func TestFileOffset(t *testing.T) {
 			t.Error(err)
 		}
 		if len(matchingLogs) != 1 {
-			t.Errorf(`Expected to find only one instance of "first line" log in the backend. Found %d instances.`, len(matchingLogs))
+			t.Errorf(`Expected to find exactly one instance of "first line" log in the backend. Found %d instances.`, len(matchingLogs))
 		}
 		// Verify second log was ingested.
 		if err := gce.WaitForLog(ctx, logger, vm, "files_1", time.Hour, `jsonPayload.message="second line"`); err != nil {


confgenerator/logging_receivers.go
```diff
@@ -606,7 +609,7 @@ func (r LoggingReceiverWindowsEventLog) Pipelines(ctx context.Context) ([]otel.R
 			"start_at":              "beginning",
```
jefferbrecht, line 609: I'm trying to figure out how `start_at` interacts with `storage` but the upstream docs are unclear. I assumed there would be an option for `start_at` that would clearly mean "pick up from where we left off according to the stored offset, rather than starting at the beginning or end", but `beginning` and `end` are the only two options.

How can I be confident that `beginning` is actually going to pick up from the stored file offset on collector restart rather than the beginning of the file? Can we add an integration test for this?

systemd/google-cloud-ops-agent-opentelemetry-collector.service
```diff
@@ -23,7 +23,7 @@ RuntimeDirectory=google-cloud-ops-agent-opentelemetry-collector
 StateDirectory=google-cloud-ops-agent/opentelemetry-collector
 LogsDirectory=google-cloud-ops-agent
 Type=simple
-ExecStartPre=@PREFIX@/libexec/google_cloud_ops_agent_engine -service=otel -in @SYSCONFDIR@/google-cloud-ops-agent/config.yaml -logs ${LOGS_DIRECTORY}
+ExecStartPre=@PREFIX@/libexec/google_cloud_ops_agent_engine -service=otel -in @SYSCONFDIR@/google-cloud-ops-agent/config.yaml -logs ${LOGS_DIRECTORY} -state ${STATE_DIRECTORY}
```
jefferbrecht, line 26: Where does the state directory get created? The `file_storage` extension will not create it automatically: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/storage/filestorage/README.md?plain=1#L20

franciscovalentecastro: From my read of the `systemd` documentation [^1], when you set `StateDirectory=` the folder (and its parents) will be created automatically (with correct permissions i presume). In older `systemd` versions, we did this manually [^2].

For the windows case, i added a step to create the "state" folder.

[^1]: https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectory=

[^2]: https://github.com/GoogleCloudPlatform/ops-agent/blob/master/systemd/google-cloud-ops-agent-opentelemetry-collector.service.d/directories.conf

confgenerator/logging_receivers.go
franciscovalentecastro: > I assumed there would be an option for start_at that would clearly mean "pick up from where we left off according to the stored offset, rather than starting at the beginning or end", but beginning and end are the only two options.

Yeah, the `start_at` is not descriptive. I also assumed "beginning + storage" meant "start at the beginning if there is no bookmark".  

One option is to do a clarification doc PR in the upstream docs for this.

> How can I be confident that beginning is actually going to pick up from the stored file offset on collector restart rather than the beginning of the file? Can we add an integration test for this?

Yeah, we could do an integration test for this since `transformation test` are not meant to test "restarts". 

- How detailed do you think should it be to give us confidence ? 
- Should it test only `files` receiver or also `systemd`, `windowseventlog` ?

We could :
1. Send 5 logs from file.
2. Wait 2 minutes and restarts.
3. Look for duplicate logs in the past 2 mins.

confgenerator/logging_receivers.go
```diff
@@ -207,6 +207,9 @@ func (r LoggingReceiverFilesMixin) Pipelines(ctx context.Context) ([]otel.Receiv
 		"preserve_leading_whitespaces":  true,
 		"preserve_trailing_whitespaces": true,
 	}
+	if !r.TransformationTest {
+		receiver_config["storage"] = fileStorageExtensionID()
```
franciscovalentecastro, line 211: @jefferbrecht I'll respond in this comment to have a threaded message.

> Any thoughts on file_storage vs. db_storage? 

I didn't consider `db_storage` at all since AFAIU it requires to manage a separate DB server ("MySQL") and provide a connection [^1] to the extension. Managing separate SQL db doesn't seem worth it.

> Do you have any concerns about unbounded retention of offsets for scenarios where someone configures a wildcard path?

I haven't thought of this case. Thanks for bringing it up. I assume (have to do code research on this) there is a default behaviour. Right now, the interaction between `exclude_older_than ` and `polls_to_archive` seems [^2] to be helpful to manage this. I'll do some testing.

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/dbstorage#database-storage

[^2]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver#archiving

confgenerator/logging_receivers.go
franciscovalentecastro: I created the `TestLogCursor` integration test to verify this the `cursor` (`bookmark`) is preserved after restart.

I ran this test locally to verify that it indeed fails with an Otel Logging version without the `filestorage` extension
```
=== NAME  TestLogCursor/debian-cloud:debian-11/default
    agents.go:1010: Test logs: /tmp/401636222/TestLogCursor_debian-cloud:debian-11_default
    agents.go:1010: Instance Log: https://console.cloud.google.com/logs/viewer?resource=gce_instance%2Finstance_id%2F5331085291174606091&project=fcovalente-dev
=== NAME  TestLogCursor/debian-cloud:debian-11/otel_logging
    agents.go:1010: Instance Log: https://console.cloud.google.com/logs/viewer?resource=gce_instance%2Finstance_id%2F6330475123837351179&project=fcovalente-dev
    main_test.go:5829: AssertLogMissing(log="jsonPayload.message=\"line #2\""): <nil> failed: unexpectedly found data for log
    main_test.go:5829: AssertLogMissing(log="jsonPayload.message=\"line #1\""): <nil> failed: unexpectedly found data for log
--- FAIL: TestLogCursor (0.00s)
    --- FAIL: TestLogCursor/debian-cloud:debian-11 (0.00s)
        --- PASS: TestLogCursor/debian-cloud:debian-11/default (273.07s)
        --- FAIL: TestLogCursor/debian-cloud:debian-11/otel_logging (310.43s)
```

confgenerator/logging_receivers.go
franciscovalentecastro: I've researched about `exclude_older_than` and `polls_to_archive`, but none of them seem to helpful to handle the file size. 

In my current understanding, there will be an unbounded `filestorage` file size if there is an unbounded number of different files tracked by a "wildcard". 

There are some attempts of adding a `Max File Limit` feature (see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38620), but it's currently "work in progress".

Details : 
- This is only for the `file log receiver offset` storage usage of the extension. For `exporter persistent_queue` you can manage the maximum size from the exporter configuration.
- This is the commit : 3805ae515c77cd1d34b20ab27075948a1e1e8c9b .

jefferbrecht: > @jefferbrecht I'll respond in this comment to have a threaded message.
> 
> > Any thoughts on file_storage vs. db_storage?
> 
> I didn't consider `db_storage` at all since AFAIU it requires to manage a separate DB server ("MySQL") and provide a connection [1](#user-content-fn-1-6c7d26d9e5e612632eceb882a50abc8f) to the extension. Managing separate SQL db doesn't seem worth it.
> 
> > Do you have any concerns about unbounded retention of offsets for scenarios where someone configures a wildcard path?
> 
> I haven't thought of this case. Thanks for bringing it up. I assume (have to do code research on this) there is a default behaviour. Right now, the interaction between `exclude_older_than ` and `polls_to_archive` seems [2](#user-content-fn-2-6c7d26d9e5e612632eceb882a50abc8f) to be helpful to manage this. I'll do some testing.
> 
> ## Footnotes
> 1. https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/dbstorage#database-storage [↩](#user-content-fnref-1-6c7d26d9e5e612632eceb882a50abc8f)
> 2. https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver#archiving [↩](#user-content-fnref-2-6c7d26d9e5e612632eceb882a50abc8f)

It's `sqlite`, not MySQL, so it'd be pretty lightweight -- coincidentally fluent-bit uses sqlite as well for its cursor storage. Management is all done by the extension: all we would do is figure out the right datasource options.

It's just something to think about, we don't have to decide right now, but we'd have to finalize it before turning anything on by default because we can't easily change the storage driver after that point.

integration_test/ops_agent_test/main_test.go
```diff
@@ -5789,6 +5789,96 @@ func TestLogCompression(t *testing.T) {
...
+		if _, err := gce.RunRemotely(ctx, logger, vm, agents.StartCommandForImage(vm.ImageSpec)); err != nil {
+			t.Fatal(err)
+		}
+
+		// Wait 1 min for logs to be ingested after restart.
+		time.Sleep(1 * time.Minute)
+
+		// We should only observe the new logs written after restart.
+		addQueryFuncToWaitGroup(func() error {
+			return gce.AssertLogMissing(ctx, logger, vm, "files_1", 2*time.Minute, `jsonPayload.message="line #1"`)
```
jefferbrecht: This feels overly complicated. All we need to do is verify that one log line is ingested exactly once across an agent restart. Why not just:

1. Write one log line
2. Restart the agent
4. Wait a reasonable amount of time to have ~100% confidence that all lines have been ingested
5. Check that the line was only ingested once
6. And maaaybe write + verify a second line to rule out that the agent didn't crash or fail to read the file after restart

AFAICT that would also achieve the test's goal while simplifying away the intermediate sleeps, the wait groups, etc.

franciscovalentecastro: Yes, thats a simpler test, though i avoided doing a "Only one log line count assertion" due to how `WaitForLog` is implemented. It only returns `found` or `not found`.

We would need to write a `WaitForLogCount` or some other refactoring of `WaitForLog` or `hasMatchingLog` [^1] to implement this. I'll draft the refactoring and see if its worth it.

[^1]: https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/01c257f80f58a4f1fcd9d88aa4e85736bb1c9f83/integration_test/gce-testing-internal/gce/gce_testing.go#L607-L654

franciscovalentecastro: Done! I made the test shorter using suggested steps `1 to 5` mentioned here. The resulting test fails locally without the changes (latest OA release)  in this PR : 

```
=== NAME  TestFileOffset/debian-cloud:debian-11/otel_logging
    main_test.go:6011: Expected to find only one instance of "first line" log in the backend. Found 2 instances.
--- FAIL: TestFileOffset (0.00s)
    --- FAIL: TestFileOffset/debian-cloud:debian-11 (0.00s)
        --- FAIL: TestFileOffset/debian-cloud:debian-11/otel_logging (321.27s)
        --- PASS: TestFileOffset/debian-cloud:debian-11/default (380.59s)
```

cmd/ops_agent_windows/main_windows.go
```diff
@@ -106,6 +106,10 @@ func initServices() error {
 	if err := os.MkdirAll(fluentbitStoragePath, 0644); err != nil {
 		return err
 	}
+	otelStoragePath := filepath.Join(os.Getenv("PROGRAMDATA"), dataDirectory, `run\file_storage`)
```
jefferbrecht:
```suggestion
	otelStoragePath := filepath.Join(os.Getenv("PROGRAMDATA"), dataDirectory, "run", "file_storage")
```

integration_test/ops_agent_test/main_test.go
```diff
@@ -5953,6 +5953,70 @@ func TestLogCompression(t *testing.T) {
...
+		matchingLogs, err := gce.QueryAllLogs(ctx, logger, vm, "files_1", time.Hour, `jsonPayload.message="first line"`, gce.LogQueryMaxAttempts)
+		if err != nil {
+			t.Error(err)
+		}
+		if len(matchingLogs) == 0 {
+			t.Errorf(`Expected to find at least one instance of "first line" log in the backend. Found %d instances.`, len(matchingLogs))
+		}
+		if len(matchingLogs) >= 2 {
+			t.Errorf(`Expected to find only one instance of "first line" log in the backend. Found %d instances.`, len(matchingLogs))
+		}
```
jefferbrecht: Nit: these can be combined into a single `if len(matchingLogs) != 1` check

go.sum
```diff
@@ -85,8 +85,10 @@ github.com/Code-Hex/go-generics-cache v1.5.1/go.mod h1:qxcC9kRVrct9rHeiYpFWSoW1v
 github.com/DataDog/datadog-go v3.2.0+incompatible/go.mod h1:LButxg5PwREeZtORoXG3tL4fMGNddJ+vMq1mwgfaqoQ=
 github.com/GoogleCloudPlatform/google-guest-agent v0.0.0-20250924181420-23412fbd6228 h1:K6GHKAI+PB91Fm+KEwX4q6s6xzZ/+BkOi8Kjxh0tJ6U=
 github.com/GoogleCloudPlatform/google-guest-agent v0.0.0-20250924181420-23412fbd6228/go.mod h1:31SvkAl6ORtir1odRpTl92XpTtn52GnBAHPtCKDSePo=
-github.com/GoogleCloudPlatform/opentelemetry-operations-collector/integration_test/gce-testing-internal v0.0.0-20260119145159-eff68fc7b8ef h1:v2hZ/3wkAB/9TvEqqE3zjlVsqozs4rzZ4zqvW/9YPLQ=
-github.com/GoogleCloudPlatform/opentelemetry-operations-collector/integration_test/gce-testing-internal v0.0.0-20260119145159-eff68fc7b8ef/go.mod h1:hmqK+hsg2z+Xo/Yz16sNxb6TZX+q0KxKsrgYzdpV0P8=
+github.com/GoogleCloudPlatform/opentelemetry-operations-collector/integration_test/gce-testing-internal v0.0.0-20260130171200-8bbc62976af0 h1:QeKYB2SRm5fOk6l91p8kNd6XkDBk/nwaeWnTwwBqwYQ=
```
jefferbrecht: Nit: run `go mod tidy`

integration_test/ops_agent_test/main_test.go
franciscovalentecastro: Done!

go.sum
franciscovalentecastro: Done!

integration_test/ops_agent_test/main_test.go
franciscovalentecastro: Just for completeness, this is the result of running locally the last iteration of this test (with the `!= 1` check) locally (without the `file_storage` update) : 

```
=== NAME  TestFileOffset/debian-cloud:debian-11/otel_logging
    agents.go:1029: Instance Log: [redacted]
    main_test.go:6008: Expected to find exactly one instance of "first line" log in the backend. Found 2 instances.
--- FAIL: TestFileOffset (0.00s)
    --- FAIL: TestFileOffset/debian-cloud:debian-11 (0.00s)
        --- PASS: TestFileOffset/debian-cloud:debian-11/default (305.37s)
        --- FAIL: TestFileOffset/debian-cloud:debian-11/otel_logging (306.46s)
FAIL
```

PR #2179, opened 2026-01-12
Creating tests to verify packages and binaries are signed.

avilevy18:
## Description

Creating tests to verify packages and binaries are signed.

## Related issue

b/474631062

## How has this been tested?

New integration tests, ran locally since they are skipped on presubmits. Please read the bug on how to execute the tests for yourself.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


integration_test/agents/agents.go
```diff
@@ -1283,3 +1287,105 @@ func SetupOpsAgentWithFeatureFlag(ctx context.Context, logger *log.Logger, vm *g
...
+
+	if location.packagesInGCS != "" {
+		return verifyRPMPackageSignedFromGCS(ctx, logger, vm)
+	}
+
+	// Assumes the Ops Agent was set up using the install script (with REPO_SUFFIX).
+	// Since the script has already added the repository to the VM, we only need
+	// to download the package.
+	_, err := gce.RunRemotely(ctx, logger, vm, "dnf download google-cloud-ops-agent")
+	if err != nil {
```
jefferbrecht:
```suggestion
	if _, err := gce.RunRemotely(ctx, logger, vm, "dnf download google-cloud-ops-agent"); err != nil {
```

integration_test/agents/agents.go
```diff
@@ -1283,3 +1287,105 @@ func SetupOpsAgentWithFeatureFlag(ctx context.Context, logger *log.Logger, vm *g
...
+		return fmt.Errorf("failed to get unsigned binaries: %v", err)
+	}
+	output := out.Stdout
+	if output != "" {
+		return fmt.Errorf("found unsigned executables:\n %s", output)
+	}
+	return nil
+}
+
+func verifyIfRPMIsSigned(ctx context.Context, logger *log.Logger, vm *gce.VM, path string) error {
```
jefferbrecht: Nit: not a fan of the name clash between `verifyRPMPackageSigned` and `verifyIfRPMIsSigned`. If this is an implementation detail of the former then just call it something like `verifyRPMPackageSignedImpl`.

integration_test/agents/agents.go
```diff
@@ -1283,3 +1287,105 @@ func SetupOpsAgentWithFeatureFlag(ctx context.Context, logger *log.Logger, vm *g
...
+	// to download the package.
+	_, err := gce.RunRemotely(ctx, logger, vm, "dnf download google-cloud-ops-agent")
+	if err != nil {
+		return fmt.Errorf("failed to run dnf download: %v", err)
+	}
+
+	return verifyIfRPMIsSigned(ctx, logger, vm, "./*.rpm")
+}
+
+func verifyRPMPackageSignedFromGCS(ctx context.Context, logger *log.Logger, vm *gce.VM) error {
```
jefferbrecht: Nit: this is only called once, can we just inline it inside `verifyRPMPackageSigned`? Likewise for `verifyWindowsBinarySignedFromGCS`

avilevy18: I think that would hurt readability, similarly `installWindowsPackageFromGCS` is only called once.

integration_test/agents/agents.go
```diff
@@ -1283,3 +1287,102 @@ func SetupOpsAgentWithFeatureFlag(ctx context.Context, logger *log.Logger, vm *g
...
+	}
+	output := out.Stdout
+	if output != "" {
+		return fmt.Errorf("found unsigned executables:\n %s", output)
+	}
+	return nil
+}
+
+func verifyRPMPackageSignedImpl(ctx context.Context, logger *log.Logger, vm *gce.VM, path string) error {
+	out, err := gce.RunRemotely(ctx, logger, vm, fmt.Sprintf("rpm -q --qf '%%{NAME}-%%{VERSION}-%%{RELEASE} %%{SIGPGP:pgpsig} %%{SIGGPG:pgpsig}\\n' -p %s", path))
```
jefferbrecht: Why not use `rpm --checksig -v`?

integration_test/ops_agent_test/main_test.go
```diff
@@ -5894,6 +5894,28 @@ func TestAppHubLogLabels(t *testing.T) {
...
 }
 
+func TestOpsAgentSigning(t *testing.T) {
+	t.Parallel()
+	gce.RunForEachImage(t, func(t *testing.T, imageSpec string) {
+		t.Parallel()
+		if testing.Short() {
+			t.Skip("skipping test in short mode.")
+		}
+		if !gce.IsRpm(imageSpec) && !gce.IsWindows(imageSpec) {
```
jefferbrecht, line 6120: Why use `gce.IsRpm` here while using `IsRPMBased` everywhere else?

integration_test/agents/agents.go
```diff
@@ -1283,3 +1287,102 @@ func SetupOpsAgentWithFeatureFlag(ctx context.Context, logger *log.Logger, vm *g
...
+	location := LocationFromEnvVars()
+	if !IsRPMBased(vm.ImageSpec) && !gce.IsWindows(vm.ImageSpec) {
+		return fmt.Errorf(fmt.Sprintf("VM image: %s, is not suported for signing", vm.ImageSpec))
+	}
+
+	if IsRPMBased(vm.ImageSpec) {
+		return verifyRPMPackageSigned(ctx, logger, vm, location)
+	}
+
+	return verifyWindowsBinarySigned(ctx, logger, vm, location)
```
jefferbrecht:
```suggestion
	if IsRPMBased(vm.ImageSpec) {
		return verifyRPMPackageSigned(ctx, logger, vm, location)
	} else if gce.IsWindows(vm.ImageSpec) {
		return verifyWindowsBinarySigned(ctx, logger, vm, location)
	} else {
		return fmt.Errorf(fmt.Sprintf("VM image: %s, is not suported for signing", vm.ImageSpec))
	}
```

integration_test/agents/agents.go
jefferbrecht: But why the inconsistency then -- GCS gets its own separate function but non-GCS is inlined? Under the hood those two paths only differ in which path to pass to `verifyRPMPackageSignedImpl` (plus the additional download step for non-GCS which is already inlined) so I'm not seeing why the GCS path warrants its own separate function.

Not a blocker for me but IMO the layering is unnecessary and hurts readability more.

integration_test/ops_agent_test/main_test.go
avilevy18: I implemented `gce.IsRpm` before discovering that `IsRPMBased` already existed. I plan to remove the legacy `IsRPMBased` in a follow-up PR. Since a refactor is coming anyway, it didn't matter which one to use.

integration_test/agents/agents.go
avilevy18: After looking at the differences it makes more sense to go with `--checksig` which actually verifies the digests and signatures.
Added in [94b4b3a](https://github.com/GoogleCloudPlatform/ops-agent/pull/2179/commits/94b4b3a015ceb00f2afab629d0e269bdf3ddbd40)

integration_test/agents/agents.go
avilevy18: You bring up a good point. I have made the change in: [33d6057](https://github.com/GoogleCloudPlatform/ops-agent/pull/2179/commits/33d6057bb7935d8d8a2f188052753afe4f2fec12)

integration_test/agents/agents.go
```diff
@@ -1283,3 +1290,116 @@ func SetupOpsAgentWithFeatureFlag(ctx context.Context, logger *log.Logger, vm *g
...
+}
+
+func verifyRPMPackageSignedImpl(ctx context.Context, logger *log.Logger, vm *gce.VM, path string) error {
+	out, err := gce.RunRemotely(ctx, logger, vm, fmt.Sprintf("rpm --checksig -v %s", path))
+	if err != nil {
+		return fmt.Errorf("failed to run rpm --checksig: %v", err)
+	}
+	output := out.Stdout
+
+	var matchingKeyID string
```
jefferbrecht, line 1365: This could probably be made simpler, e.g.:

```go
	for _, keyID := range approvedRPMKeyIDs {
		expectedOutput := fmt.Sprintf(`    Header V4 RSA/SHA512 Signature, key ID %s: OK
    Header SHA256 digest: OK
    Header SHA1 digest: OK
    Payload SHA256 digest: OK
    V4 RSA/SHA512 Signature, key ID %s: OK
    MD5 digest: OK`, keyID, keyID)
		if strings.Contains(output, expectedOutput) {
			return nil
		}
	}
	return fmt.Errorf("RPM is not signed with any of the approved keys. Output:\n%s", output)
```

But the current implementation does have an advantage of distinguishing between "the right key is there but the template is wrong" vs. "the right key is not there". I'll leave it up to you.

PR #2185, opened 2026-01-21
[confgenerator] Create `otel.Logging` exporter for Otel Logging.

franciscovalentecastro:
## Description
Create `otel.OTelLogs` exporter to have a separate configuration for Otel Logging.

## Details
**Why do we need a separate exporter ?**
  - The `sending_queue` configurations set in this PR and followup changes for "Log Buffering" (b/394636880) are incompatible with metrics exporting (for example, we can't sent point out of order from the queue).

**Why put the `BatchProcessor` at the end of the pipeline ?**
  - We are deferring all details that are "logs exporter" specific and removing them from "ReceiverPipeline" logic. This enables us to swap the exporter without having to configure ReceiverPipelines separately.

**Why do we set `timeout = 60s` to the `OTelLogs` exporter ?**
- The OTel GCP logs exporter doesn't support `retry_on_failure` (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/25900), so the only "retry" logic we can use lives within the `Go Cloud SDK Log` libraries.
- The `timeout` set is used a `Context Timeout` [1] for the underlying log request retry logic [2] which is configured with a `60s` max time limit for retries.

~**Why choose `grpc_pool_size = 10`, `num_consumers = 20` and `send_batch_max_size = 1000`?**~
- ~I chose similar values from the "Performance Tuning" guide in [3] while also validating the exporter will reach 40K/s throughput. We can discuss further.~

[1] https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/logs.go#L385

[2] https://github.com/googleapis/google-cloud-go/blob/main/logging/logging.go#L294

[3] https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/tree/main/exporter/collector#performance-tuning

## Related issue
b/477325045

## How has this been tested?
Existing confgenerator, transformation and integration tests.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/transformation_test/transformation_test.go
+++ b/transformation_test/transformation_test.go
@@ -406,8 +406,8 @@ func (transformationConfig transformationTest) generateOTelConfig(ctx context.Co
 		Exporters: map[otel.ExporterType]otel.ExporterComponents{
 			otel.Logging: {
 				ProcessorsByType: map[string][]otel.Component{
-					// Batch with 1.5s timeout to group all late log entries flushed
-					// from a multiline parser after 1s in the same log request.
+					// Batch with 1.5s timeout to group in the same log request
+					// all late entries flushed from a multiline parser after 1s.
 					"logs": {
 						otel.BatchProcessor(500, 500, "1500ms"),
 					},


confgenerator/otel/processors.go
```diff
@@ -682,7 +682,18 @@ func MetricUnknownCounter() Component {
 	})
 }
 
-func Batch() Component {
+func BatchLogsExporter() Component {
```
quentinmit: Why is this called `BatchLogsExporter` instead of `BatchLogsProcessor`? (Ditto for the other function)

confgenerator/confgenerator.go
```diff
@@ -158,10 +168,11 @@ func (uc *UnifiedConfig) GenerateOtelConfig(ctx context.Context, outDir string)
 		Pipelines:         pipelines,
 		Extensions:        extensions,
 		Exporters: map[otel.ExporterType]otel.Component{
```
quentinmit: I think this is where we should be putting the processors, instead of adding magic conditionals inside `Generate`. That will also make it easy to apply different batching to different signal types. Maybe we can use a type like:

```
type ExporterComponents struct {
  ProcessorsByType map[string][]otel.Component
  Exporter otel.Component
}
```

confgenerator/confgenerator.go
```diff
@@ -34,27 +34,39 @@ import (
...
+	if logsExporter {
+		config["log"] = map[string]any{
+			"grpc_pool_size": 20,
+		}
+		config["sending_queue"] = map[string]any{
+			"enabled":       true,
+			"num_consumers": 40,
+		}
+		config["timeout"] = "60s"
+	}
```
quentinmit: The performance tuning guide says "There is no guaranteed one-size-fits-all combination of settings, and for the majority of use cases the defaults will provide more than sufficient throughput."

Why are the defaults insufficient?

confgenerator/confgenerator.go
```diff
@@ -34,27 +34,39 @@ import (
...
-				// descriptors to be created implicitly with new time series.
-				"skip_create_descriptor": true,
-				// Omit instrumentation labels, which break agent metrics.
-				"instrumentation_library_labels": instrumentationLabels,
-				// Omit service labels, which break agent metrics.
-				"service_resource_labels": serviceResourceLabels,
-				"resource_filters":        []map[string]interface{}{},
-			},
+func googleCloudExporter(userAgent string, instrumentationLabels, serviceResourceLabels, logsExporter bool) otel.Component {
+	config := map[string]interface{}{
```
quentinmit: Having two boolean parameters was already confusing enough; since this is purely additive config, can we just use a separate function that calls `googleCloudExporter` and then adds the additional config?

confgenerator/confgenerator.go
```diff
@@ -158,10 +168,11 @@ func (uc *UnifiedConfig) GenerateOtelConfig(ctx context.Context, outDir string)
...
 		Exporters: map[otel.ExporterType]otel.Component{
-			otel.System: googleCloudExporter(userAgent, false, false),
-			otel.OTel:   googleCloudExporter(userAgent, true, true),
-			otel.GMP:    googleManagedPrometheusExporter(userAgent),
-			otel.OTLP:   otlpExporter(userAgent),
+			otel.System:   googleCloudExporter(userAgent, false, false, false),
+			otel.OTel:     googleCloudExporter(userAgent, true, true, false),
+			otel.OTelLogs: googleCloudExporter(userAgent, true, true, true),
+			otel.GMP:      googleManagedPrometheusExporter(userAgent),
+			otel.OTLP:     otlpExporter(userAgent),
```
quentinmit: Don't you also need to do this for the `otel.OTLP` exporter type?

confgenerator/otel/processors.go
```diff
@@ -682,7 +682,18 @@ func MetricUnknownCounter() Component {
...
+		Type: "batch",
+		Config: map[string]any{
+			"send_batch_max_size": 1000,
+			"send_batch_size":     1000,
+			"timeout":             "200s",
+		},
+	}
+}
+
+func BatchOLTPMetricExporter() Component {
```
quentinmit: Typo (`s/OLTP/OTLP/`)

confgenerator/otel/modular.go
```diff
@@ -221,6 +224,16 @@ func (c ModularConfig) Generate(ctx context.Context) (string, error) {
 		SetIfMissing: resourceDetectionProcessors[SetIfMissing].name("_global_1"),
 	}
 
+	exporterTypeProcessors := map[ExporterType]Component{
+		OTelLogs: BatchLogsExporter(),
+		// The OTLP exporter doesn't batch by default like the googlecloud.* exporters. We need this to avoid the API point limits.
+		OTLP: BatchOLTPMetricExporter(),
```
quentinmit: The `OTLP` exporter type can get metrics, traces, and logs. We shouldn't be applying metrics-tuned batching to traces or logs.

confgenerator/confgenerator.go
franciscovalentecastro: The default config for the exporter only reaches `15K/s logs` throughput in our Soak Tests. Fluent-bit reaches `40K/s logs` throughput. The configs `grpc_pool_size = 20 ` and `num_consumers = 40` achieve this goal. I used the values suggested in the "Performance tuning guide".

`timeout = 60s` is to give the `GCL Go SDK` a longer timeout to do retries. More details in the PR description.

confgenerator/confgenerator.go
franciscovalentecastro: Yes, we do. 

@XuechunHou Is focusing on the OTLP logs exporter. We've synced and we probably are going to do a very similar set of configs for `OTLPLogs` which need to happen separately from `OTLPMetrics`. This distinction will probably happen in a separate PR.

confgenerator/otel/processors.go
franciscovalentecastro: I think `BatchLogsProcessor` is better, i'll change it. I guess i was just overly focused on this processor being related to the `Logs Exporter`.

confgenerator/confgenerator.go
franciscovalentecastro: I created the following type and updated `ModularConfig` accordingly.

```
type ExporterComponent struct {
	Exporter   Component
	Processors []Component
}
```

confgenerator/confgenerator.go
franciscovalentecastro: I realized the "Logging Exporter" config is not only additive. Keeping the "metrics" and "userAgent" is unnecessary. I created `googleCloudLoggingExporter` as a standalone function. Writing it as "derivative" function resulted in more complicated code.

confgenerator/otel/processors.go
franciscovalentecastro: Done!

confgenerator/otel/modular.go
franciscovalentecastro: @XuechunHou Is focusing on the OTLP logs exporter. We've synced and we probably are going to do a very similar set of configs for OTLPLogs which need to happen separately from OTLPMetrics. This distinction will probably happen in a separate PR.

confgenerator/confgenerator.go
franciscovalentecastro: I updated the parameters to `grpc_pool_size = 10`  and `num_consumers = 20` since those also work to get 40K/s throughput.

image

confgenerator/confgenerator.go
```diff
@@ -57,6 +57,58 @@ func googleCloudExporter(userAgent string, instrumentationLabels bool, serviceRe
...
+	return map[otel.ExporterType]otel.ExporterComponent{
+		otel.System: {
+			Exporter: googleCloudExporter(userAgent, false, false),
+		},
+		otel.OTel: {
+			Exporter: googleCloudExporter(userAgent, true, true),
+		},
+		otel.OTelLogs: {
+			Exporter:   googleCloudLoggingExporter(userAgent),
+			Processors: otelLogsExporterProcessors(),
```
XuechunHou: I think inlining the processor list provides better readability; creating a separate function tends to disrupt the reading flow.

confgenerator/confgenerator.go
quentinmit: I was suggesting the *type* be called `ExporterComponents` (because it is the set of components needed to implement the exporter), not that the map be renamed to `ExporterComponents`.

Also, you have a single slice of processors, instead of a different slice of processors for each type. This means you can't apply different batching for metrics, logs, and trace.

confgenerator/confgenerator.go
quentinmit: I know you can get higher throughput out of both fluent-bit and otelopscol by changing the number of workers; why isn't 15K/s already sufficent? When we looked at this in April 2024 we concluded that both fluent-bit and otelopscol were constrained by worker count/batch size/API latency, but that both 15K/s and 40K/s were well above the target throughput, so we didn't need to make any changes here.

confgenerator/otel/modular.go
```diff
@@ -240,21 +248,30 @@ func (c ModularConfig) Generate(ctx context.Context) (string, error) {
...
 			processors[name] = processor.Config
 		}
+		// Processors specific to an exporter type.
+		if exporterComponents, ok := c.ExporterComponents[exporterType]; ok {
+			for i, processor := range exporterComponents.Processors {
+				name := processor.name(fmt.Sprintf("%s_%d", prefix, i))
+				processorNames = append(processorNames, name)
+				processors[name] = processor.Config
+			}
+		}
```
quentinmit: These processors should be added after the resource detection processors, not before. And they shouldn't have `prefix` in their names; they need to be shared across all pipelines that use the exporter (so that e.g. logs and metrics can be batched across receivers). You can do both of these by adding the processors like ~this:

```go
exporter := c.ExporterComponents[exporterType]
if _, ok := exporterNames[exporterType]; !ok {
	name := exporter.Exporter.name(exporterType.Name())
	exporterNames[exporterType] = name
	exporters[name] = exporter.Config
}
for i, processor := range exporter.Processors[pipeline.Type] {
  name := processor.name(fmt.Sprintf("%s_%s_%d", exporterNames[exporterType], pipeline.Type, i))
  if _, ok := processors[name]; !ok {
    processors[name] = processor.Config
  }
}
```

confgenerator/otel/processors.go
```diff
@@ -689,7 +689,18 @@ func MetricUnknownCounter() Component {
 	})
 }
 
-func Batch() Component {
+func BatchLogsProcessor() Component {
```
quentinmit: These only ever get used once, no? Why don't we have both of their definitions inline in `modular.go` instead of defining new functions?

confgenerator/confgenerator.go
quentinmit: I agree, although I would probably resolve it by pushing it up into these functions and having them return the `ExporterComponent` struct directly, so you can them keep this map where it was in `GenerateOTelConfig` below.

confgenerator/otel/modular.go
quentinmit: Since you're introducing the `ExporterComponents` struct, you are touching the code and you can fix this while you're there. :)

confgenerator/otel/modular.go
rafaelwestphal: +1 to adding it after the resourcedetection. They should be appended to end of the pipeline.

confgenerator/otel/processors.go
franciscovalentecastro: I created the following processor that is reused in several places with different arguments.
```
func BatchProcessor(sendBatchSize, sendBatchMaxSize int, timeout string) Component {
	return Component{
		Type: "batch",
		Config: map[string]any{
			"send_batch_size":     sendBatchSize,
			"send_batch_max_size": sendBatchMaxSize,
			"timeout":             timeout,
		},
	}
}
```

confgenerator/otel/modular.go
franciscovalentecastro: Done!

confgenerator/confgenerator.go
franciscovalentecastro: I moved everything to `GenerateOTelConfig` after the `ExporterComponents` struct update. PTAL.

confgenerator/otel/modular.go
franciscovalentecastro: You're right :) I added a separate batching for OTLP "logs" and "metrics" in this PR. Done!

confgenerator/confgenerator.go
franciscovalentecastro: Simplified further the `googleCloudLoggingExporter` implementation to only set `timeout = 60s`.

confgenerator/confgenerator.go
franciscovalentecastro: @quentinmit After syncing offline we decided the target throughput will be 10 K/s logs which can be achieved with the current defaults. I removed `grpc_pool_size`  and `num_consumers` config and only kept `timeout`.

As we discussed, if we want a higher default throughput we can configure the defaults upstream [^1].

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.144.0/exporter/googlecloudexporter/README.md?plain=1#L214

confgenerator/confgenerator.go
franciscovalentecastro: Thank you for adding more details, now its clear for me the goal of your first `ExporterComponents` proposed idea. This is now the implemented `ExporterComponents` :

```
type ExporterComponents struct {
	Exporter         Component
	ProcessorsByType map[string][]Component
}
```

confgenerator/otel/processors.go
franciscovalentecastro: I created only one `BatchProcessor(sendBatchSize, sendBatchMaxSize int, timeout string)` now.

rafaelwestphal (approved):

confgenerator/confgenerator.go
franciscovalentecastro: I added a separate batching for OTLP "logs" and "metrics" in this PR. Done!

transformation_test/transformation_test.go
```diff
@@ -403,18 +403,26 @@ func (transformationConfig transformationTest) generateOTelConfig(ctx context.Co
...
-						"default_log_name": "my-log-name",
-						"endpoint":         addr,
-						"use_insecure":     true,
+				},
+				Exporter: otel.Component{
+					Type: "googlecloud",
+					Config: map[string]any{
+						"project": "my-project",
+						"sending_queue": map[string]any{
+							"enabled": false,
```
quentinmit, line 420: Why are we disabling the sending queue? Won't this result in effectively num_workers = 1?

transformation_test/transformation_test.go
```diff
@@ -403,18 +403,26 @@ func (transformationConfig transformationTest) generateOTelConfig(ctx context.Co
...
-			otel.OTel: {
-				Type: "googlecloud",
-				Config: map[string]any{
-					"project": "my-project",
-					"sending_queue": map[string]any{
-						"enabled": false,
+		Exporters: map[otel.ExporterType]otel.ExporterComponents{
+			otel.Logging: {
+				ProcessorsByType: map[string][]otel.Component{
+					// Batch with 2s timeout to accomodate for multiline parsing flush period.
```
quentinmit: I don't understand this comment. multiline's flush period should be entirely separate from the batch processor (it's not going to pull partially-parsed entries from within the multiline processor, regardless of what the timeout is set to).

confgenerator/confgenerator.go
```diff
@@ -171,11 +179,39 @@ func (uc *UnifiedConfig) GenerateOtelConfig(ctx context.Context, outDir string)
...
+			otel.GMP: {
+				Exporter: googleManagedPrometheusExporter(userAgent),
+			},
+			otel.OTLP: {
+				Exporter: otlpExporter(userAgent),
+				ProcessorsByType: map[string][]otel.Component{
+					// The OTLP exporter doesn't batch by default like the googlecloud.* exporters.
+					// We need this to avoid the API point limits.
+					"metrics": {
+						otel.BatchProcessor(200, 200, "200ms"),
```
quentinmit, line 217: Isn't this also going to result in waiting until there are 200 points before we send our first batch? I think that might cause some write errors if we include multiple points for the same time series in the same batch as a result. If I understand the config right, I think `1, 200` would be closer to the existing behavior of the existing `googlecloudexporter`.

(Feel free to punt this discussion later since the problem already exists.)

transformation_test/transformation_test.go
franciscovalentecastro: The goal here is to keep all test log lines in the same logging request. Currently the "multiline parser" `flush_timeout` is set to `1s` (which is configured the same way as the fluent-bit flush argument). 

https://github.com/GoogleCloudPlatform/ops-agent/blob/a6937e12c8f04becab929f01b17318933f3ae5fb/confgenerator/logging_processors.go#L487-L492

This processor batches all the "last log entries" in a "multiline parser" that may be waiting (until the flush timeout) for a possibly new entry to be appended. This log lines are usually sent in a separate log request. 

*This is mostly for presentation purposes, but it can avoid "race condition" tests failures/flakiness with the transformation test goldens.*

I will improve the comment.

transformation_test/transformation_test.go
franciscovalentecastro: > Why are we disabling the sending queue? 

I didn't have a reason, since this was implemented in the initial Otel Logging support PR. I was mostly copying the exporter config to the new format.

One of my hypothesis is that "multiple workers" could potentially send the log entries to the mock server in different order, causing changing "goldens" in our transformation tests.

> Won't this result in effectively num_workers = 1?

I think that is true.

confgenerator/confgenerator.go
franciscovalentecastro: > Isn't this also going to result in waiting until there are 200 points before we send our first batch?

Yes, `send_batch_size = send_batch_max_size = 200` will wait until 200 points are added to the batch or `200ms` are passed for the batch to be sent.

> If I understand the config right, I think `1, 200` would be closer to the existing behavior of the existing `googlecloudexporter`.

I'm not fully sure the current behaviour of the exporter without batching. `send_batch_size = 1` will trigger the "batch to be sent" (but possibly add more points during a time window), while `send_batch_max_size = 200` sets a hard limit.

> Feel free to punt this discussion later since the problem already exists.

Sure, i'll punt this dicussion, i'll sync with @rafaelwestphal about this.

PR #2186, opened 2026-01-23
otlp: switch to UTR endpoint

ridwanmsharif:
Adds option to send traces to UTR endpoint instead. No-op by default.

### DIFF
--- a/integration_test/ops_agent_test/main_test.go
+++ b/integration_test/ops_agent_test/main_test.go
@@ -64,6 +64,8 @@ import (
 	"cloud.google.com/go/monitoring/apiv3/v2/monitoringpb"
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
 	"cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
+	trace "cloud.google.com/go/trace/apiv1"
+	cloudtrace "cloud.google.com/go/trace/apiv1/tracepb"
 	"github.com/GoogleCloudPlatform/opentelemetry-operations-collector/integration_test/gce-testing-internal/gce"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/fluentbit"
 	"github.com/GoogleCloudPlatform/ops-agent/confgenerator/otel"
@@ -5192,7 +5194,7 @@ traces:
 	})
 }
 
-func TestOTLPTraces(t *testing.T) {
+func TestOTLPTracesLegacy(t *testing.T) {
 	t.Parallel()
 	gce.RunForEachImage(t, func(t *testing.T, imageSpec string) {
 		t.Parallel()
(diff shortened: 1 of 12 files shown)


XuechunHou:
Can you merge this PR with https://github.com/GoogleCloudPlatform/ops-agent/pull/2292? 


In https://github.com/GoogleCloudPlatform/ops-agent/pull/2292 Rafael added a global config knob to turn on and off UTR for metrics and logs. We could reuse that config for trace. 

Can you add a few commits to Rafa's PR, adding the processors needed for trace UTR, the OTLP exporter configuration for the trace endpoint (https://github.com/GoogleCloudPlatform/ops-agent/pull/2186/changes#diff-bcf4db7b1ead9873183e1c4b42f5022443169df5866d48e713032a239e9d74f6), and the integration tests. Leave out the trace-only config in this PR. Thanks

integration_test/ops_agent_test/main_test.go
```diff
@@ -5241,6 +5243,111 @@ metrics:
...
+		// Verify that none of the spans have the legacy g.co attribute.
+		// Also verify span and resource attributes.
+		var foundResourceAttr, foundSpanAttr bool
+		for _, span := range fullTrace.Spans {
+			t.Logf("Span %q labels: %v", span.Name, span.Labels)
+			if span.Labels != nil {
+				if _, ok := span.Labels["g.co/r/gce_instance/instance_id"]; ok {
+					t.Errorf("Unexpected legacy attribute g.co/r/gce_instance/instance_id found in span %q", span.Name)
+				}
+				if val, ok := span.Labels["custom.resource.attribute"]; ok {
```
XuechunHou, line 5330: I dont think `foundResourceAttr`,  `foundSpanAttr`  variables are needed here. 
the logic can be simplified to 

```
val, ok := span.Labels["custom.resource.attribute"]

if !ok 
{ t.Errorf(t.Error("Resource attribute 'custom.resource.attribute' not found in trace spans"))}

if val !=  "my-resource-value" {
t.Errorf("Expected custom.resource.attribute to be 'my-resource-value', got %q", val)

}
```

integration_test/ops_agent_test/main_test.go
```diff
@@ -5241,6 +5243,111 @@ metrics:
...
+					if val == "my-resource-value" {
+						foundResourceAttr = true
+					} else {
+						t.Errorf("Expected custom.resource.attribute to be 'my-resource-value', got %q", val)
+					}
+				}
+				if val, ok := span.Labels["custom.span.attribute"]; ok {
```
XuechunHou, line 5337: Same here i think we can simplify this logic and avoid the `foundSpanAttr` variable.

XuechunHou (approved):

rafaelwestphal (approved):
LGTM

PR #2196, opened 2026-02-04
Adding api_request_count self-metric when running otlp exporter

rafaelwestphal:
## Description
This adds api_request_count when using the otlp exporter instead of the GCM/GMP.
The metric used by the previous exporters is not present in the otlp exporter.

## Related issue
b/481750163

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/confgenerator/agentmetrics.go
+++ b/confgenerator/agentmetrics.go
@@ -75,6 +75,7 @@ var grpcToStringStatus = map[string]string{
 	"13": "INTERNAL",
 	"14": "UNAVAILABLE",
 	"15": "DATA_LOSS",
+	"16": "UNAUTHENTICATED",
 }
 
 func (r AgentSelfMetrics) AddSelfMetricsPipelines(receiverPipelines map[string]otel.ReceiverPipeline, pipelines map[string]otel.Pipeline, ctx context.Context) {
--- a/confgenerator/testdata/goldens/combined-receiver_otlp_otlphttp/golden/linux-gpu/otel.yaml
+++ b/confgenerator/testdata/goldens/combined-receiver_otlp_otlphttp/golden/linux-gpu/otel.yaml
@@ -605,6 +605,8 @@ processors:
           value: "14"
         - new_value: DATA_LOSS
           value: "15"
+        - new_value: UNAUTHENTICATED
+          value: "16"
         - new_value: UNKNOWN
           value: "2"
         - new_value: INVALID_ARGUMENT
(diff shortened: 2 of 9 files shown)


confgenerator/agentmetrics.go
```diff
@@ -57,6 +57,26 @@ var grpcToHTTPStatus = map[string]string{
 	"DEADLINE_EXCEEDED":   "504",
 }
 
+// Following reference : https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+var grpcToStringStatus = map[string]string{
+	"0":  "OK",
+	"1":  "CANCELLED",
```
XuechunHou, line 63: It's missing `UNAUTHENTICATED` https://github.com/googleapis/googleapis/blob/81dba2bd2c1f40c796e461a400298368851012db/google/rpc/code.proto#L102

confgenerator/agentmetrics.go
```diff
@@ -133,7 +153,9 @@ func (r AgentSelfMetrics) PrometheusMetricsPipeline() otel.ReceiverPipeline {
 }
 
 func (r AgentSelfMetrics) OtelPipelineProcessors(ctx context.Context) []otel.Component {
```
franciscovalentecastro, line 156: Please simplify the logic of `OtelPipelineProcessors` since it's hard read/follow/understand what is going to be resulting pipeline after all the conditionals are processed.

Suggested update : 
- Write only one pipeline that processes all metrics unconditionally, both GCM exporter and OTLP exporter metrics. All transformations processors will be in the pipeline.

- Only use a conditional at the top to the determine which metrics will be allowed into the pipeline. :
```
filteredMetrics := []string{
		"grpc.client.attempt.duration_count",
		"googlecloudmonitoring/point_count",
	}

expOtlpExporter := experimentsFromContext(ctx)["otlp_exporter"]
if expOtlpExporter {
		filteredMetrics = []string{
			"otelcol_exporter_sent_metric_points",
			"otelcol_exporter_send_failed_metric_points",
			"rpc.client.duration_count",
		}
}
```

rafaelwestphal: This current implementation makes it so that there are no changes in the current Otel.yaml files, except when the feature flag is enabled.

franciscovalentecastro: >  This current implementation makes it so that there are no changes in the current Otel.yaml files, except when the feature flag is enabled.

I see! That's a good thing to preserve. I'd still suggest to consider ways to improve the readability (comments, unify transformation blocks).

PR #2197, opened 2026-02-05
Creating point-count metric when using otlp exporter

rafaelwestphal:
## Description
This PR creates the point_counter metric when the otlp exporter is used. The underlying metric used by the GCM exporter is not available, so we use the corresponding otel self-metric to calculate it.

## Related issue
b/468059325

## Checklist:
- Unit tests
- Integration tests
  - [] Integration tests do not apply.
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test/main_test.go
+++ b/integration_test/ops_agent_test/main_test.go
@@ -2579,7 +2579,7 @@ func TestSystemLogByDefault(t *testing.T) {
 	})
 }
 
-func testDefaultMetrics(ctx context.Context, t *testing.T, logger *log.Logger, vm *gce.VM, window time.Duration, otlpHTTPExperiment bool) {
+func testDefaultMetrics(ctx context.Context, t *testing.T, logger *log.Logger, vm *gce.VM, window time.Duration) {
 	if !gce.IsWindows(vm.ImageSpec) {
 		// Enable swap file: https://linuxize.com/post/create-a-linux-swap-file/
 		// We do this so that swap file metrics will show up.
@@ -2619,10 +2619,6 @@ func testDefaultMetrics(ctx context.Context, t *testing.T, logger *log.Logger, v
 
 		if err != nil {
 			t.Fatal(err)
-		}
-		// b/468059325: This bug prevents us from having accurate labels for point_count. Disabling this check for now.
-		if metric.Type == "agent.googleapis.com/agent/monitoring/point_count" && otlpHTTPExperiment {
-			continue
 		}
 		err = metadata.AssertMetric(metric, series)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


confgenerator/agentmetrics.go
```diff
@@ -131,7 +132,60 @@ func (r AgentSelfMetrics) PrometheusMetricsPipeline() otel.ReceiverPipeline {
...
+			"otelcol_exporter_send_failed_metric_points",
+		}
+		extraTransforms = []map[string]interface{}{
+			otel.UpdateMetric("otelcol_exporter_sent_metric_points",
+				otel.ToggleScalarDataType,
+				otel.AddLabel("status", "OK"),
+			),
+			otel.UpdateMetric("otelcol_exporter_send_failed_metric_points",
+				otel.ToggleScalarDataType,
+				otel.AddLabel("status", "UNKNOWN"),
```
XuechunHou, line 159: Change this status to 'ERROR'?

rafaelwestphal: We can't. The values for this label are preset and need to pass a check.You can find the values for the label at https://github.com/GoogleCloudPlatform/ops-agent/blob/master/integration_test/ops_agent_test/agent_metrics/metadata.yaml#L48

Those values are possible grpc response values found at 
https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto

XuechunHou: Since otelcol_exporter_sent_metric_points and otelcol_exporter_send_failed_metric_points do not currently reflect accurate counts for successful and failed points, the status labels applied to them are also inaccurate.

We should drop the `status` label until the point-counting logic is fixed, which will likely be a post-public preview task.

XuechunHou: Let me know if the metric points could be written without a status label.

jefferbrecht: How bad is the inaccuracy, how often do we get partial successes, etc.? Dropping the label turns off _all_ information about passed vs. failed points, and we need to weigh whether that's a better UX than keeping a label that has an error margin of +-N%.

XuechunHou: Looking at our test project: https://screenshot.googleplex.com/YLZDN2GbWNmjr6M the `CreateTimeSeries` call returns OK MOST of the time.

XuechunHou: Sure we can display inaccurate status label for now. status:OK for sent points, and status:ERROR for failed points. 


Fixing sent/failed metric points counting is in progress, ETA unknown, likely will finish post public preview. 
We also cannot display granular status codes (e.g: status:INVALID_ARGUMENT, status:DEADLINE_EXCEEDED) in the `status` label currently; It's fixable though in the otlp exporter, but will also likely be a post public preview feature.  

public preview ETA I have in mind is: EO Q1 or early Q2(April).

XuechunHou: https://github.com/open-telemetry/opentelemetry-collector/pull/14247 Upstream added the detailed error code for failed points recently.

PR #2200, opened 2026-02-06
[UAP] Fix UAP Windows unit test flake

XuechunHou:
## Description

This PR fixes the UAP unit test `TestStart()` flakiness. Sample stacktrace dump: https://github.com/GoogleCloudPlatform/ops-agent/actions/runs/20041243958/job/57475601078?pr=1998 (windows)

`windowsEventLogger.Close()` is called within the `cancel` function when either of the two sub-agent processes exits. These sub-agent processes are spawned as separate goroutines. The cancel function logs any error messages before calling `Close()` on the Windows event logger. The crash occurs when one subprocess exits and closes the eventLogger while the other subprocess is still attempting to access it to log errors. 

The fix to it is to handle windows event logger creation and closing inside main.go function. The entire UAP plugin is essentially an RPC server, the windows event logger is now created before the server starts listening for requests (blocking), and closed when server stops serving. 

## Related issue
b/465497407

## How has this been tested?
UAP windows unit tests should not flake anymore

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/cmd/ops_agent_uap_plugin/service_linux_test.go
+++ b/cmd/ops_agent_uap_plugin/service_linux_test.go
@@ -206,7 +206,7 @@ func TestStart_subagentsRunning(t *testing.T) {
 				t.Errorf("%v: got nil cancel function, want non-nil", tc.name)
 			}
 			if ps.pluginError != nil {
-				t.Errorf("%v: got pluginError: %v, want nil, because sub-agent processes are running", tc.name, ps.pluginError)
+				t.Errorf("%v: got pluginError: %v, want nil", tc.name, ps.pluginError)
 			}
 		})
 	}
@@ -249,7 +249,7 @@ func TestStart_subagentsExitedSuccessfully(t *testing.T) {
 	ps.mu.Lock()
 	defer ps.mu.Unlock()
 	if ps.cancel != nil {
-		t.Errorf("got cancel function: %v, want cancel function to be reset to nil, because sub-agent processes exited with errors", ps.cancel)
+		t.Errorf("got cancel function: %v, want cancel function to be reset to nil, because sub-agent processes exited successfully", ps.cancel)
 	}
 	if ps.pluginError != nil {
 		t.Errorf("got pluginError: %v, want nil pluginError, because sub-agent processes exited successfully", ps.pluginError)
--- a/cmd/ops_agent_uap_plugin/service_windows_test.go
+++ b/cmd/ops_agent_uap_plugin/service_windows_test.go
@@ -221,8 +221,7 @@ func TestStart_subagentsRunning(t *testing.T) {
 		mockRunCommandFunc RunCommandFunc
 	}{
 		{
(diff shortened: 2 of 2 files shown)


cmd/ops_agent_uap_plugin/plugin_windows.go
```diff
@@ -0,0 +1,113 @@
...
+			// being reused again. If file already exist bind can fail.
+			log.Fatalf("Failed to remove %q: %v\n", address, err)
+		}
+	}
+
+	logger, err := createWindowsEventLogger()
+	if err != nil {
+		log.Fatalf("Failed to create windows event logger: %v", err)
+	}
+	defer logger.Close()
```
jinghan-ma: Is this the only difference between plugin_linux and plugin_windows? If so, do we need to separate them? Or can we just keep `pulgin_common.go` and conditionally set the logger to WindowsEventLogger if we're on windows?

jinghan-ma: i.e. Create a function that returns the default logger on linux and windows event logger on windows and sets that as the Logger to use.

cmd/ops_agent_uap_plugin/service_linux_test.go
```diff
@@ -202,6 +202,8 @@ func TestStart(t *testing.T) {
 			t.Parallel()
 			ps := &OpsAgentPluginServer{cancel: tc.cancel, runCommand: tc.mockRunCommandFunc}
 			ps.Start(context.Background(), &pb.StartRequest{})
+			ps.mu.Lock()
+			defer ps.mu.Unlock()
```
jinghan-ma, line 204: Why are we trying to acquire `ps.mu` here? I would understand if it's some other mutex, so we avoid tests inteferring with with each other, but wouldn't the test changing the state of `ps.mu` potentially affect the operations of the Plugin Server?

cmd/ops_agent_uap_plugin/plugin_windows.go
XuechunHou: Done

cmd/ops_agent_uap_plugin/service_linux_test.go
XuechunHou: The lock must be acquired before accessing the ps.cancel function. This prevents a race condition where ps.cancel is read while being modified by sub-agent processes.

I've updated the logic of the `TestStart()` test, essentially it makes sure that the cancel function is `non-nil` while sub-processes are running and `nil` when the sub-processes have exited (with or without errors).

cmd/ops_agent_uap_plugin/service_linux_test.go
jinghan-ma: Should we split this into two cases? IIUC, when Start() should have a non-nil cancel function, but we detected `ps.cancel` to be nil, this will incorrectly report the error as `Start() got cancel function: want cancel function to be reset to nil`  (e.g. when the happy path fails)

XuechunHou: Done

PR #2208, opened 2026-02-13
[confgenerator ] Set `block_on_overflow` and exporter batching in Otel Cloud Logging exporter.

franciscovalentecastro:
## Description
This PR sets the following configurations in the Otel Google Cloud Logging exporter : 
- `block_on_overflow: true`
- Exporter batching instead of `batchprocessor`

## Related issue
b/394636880

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/confgenerator.go
```diff
@@ -58,14 +58,29 @@ func googleCloudExporter(userAgent string, instrumentationLabels bool, serviceRe
 }
 
 func googleCloudLoggingExporter() otel.Component {
-	return otel.Component{
-		Type: "googlecloud",
-		Config: map[string]interface{}{
-			// Set to mirror the 60s max limit of default retry window in Google Cloud Logging apiv2 go client :
-			// https://github.com/googleapis/google-cloud-go/blob/logging/v1.4.2/logging/apiv2/logging_client.go#L78-L90
-			"timeout": "60s",
+	config := map[string]interface{}{
```
quentinmit: Nit: Can you keep the original structure (Config inline in the return statement) so the diff is more readable?

confgenerator/confgenerator.go
```diff
@@ -223,13 +238,8 @@ func (uc *UnifiedConfig) GenerateOtelConfig(ctx context.Context, outDir, stateDi
...
 			otel.Logging: {
-				Exporter: googleCloudLoggingExporter(),
-				ProcessorsByType: map[string][]otel.Component{
-					// Batching logs improves log export performance.
-					"logs": {
-						otel.BatchProcessor(1000, 1000, "200ms"),
-					},
-				},
+				Exporter:         googleCloudLoggingExporter(),
+				ProcessorsByType: map[string][]otel.Component{},
```
quentinmit: `nil` (or just not specifying the field at all) is the same as an empty map.

confgenerator/confgenerator.go
```diff
@@ -58,14 +58,29 @@ func googleCloudExporter(userAgent string, instrumentationLabels bool, serviceRe
...
+			// Blocks the "sending_queue" on overflow to reduce log loss.
+			"block_on_overflow": true,
+			// Set batch in "sending_queue" is recommended instead of batch processor.
+			"batch": map[string]interface{}{
+				"flush_timeout": "200ms",
+				"min_size":      1000,
+				"max_size":      2000,
+				"sizer":         "items",
+			},
+			"queue_size": 2000,
```
quentinmit: AIUI this will allow 2000 parallel batches in memory, each of which is allowed to have up to 2000 log records. That seems like a crazy amount of memory that might be used by in-flight logs.

I don't think there's any advantage to setting the queue size larger than num_consumers + 1, which with the defaults would be 11. (This means every consumer can be active with a batch, and we can also have the next batch ready when a consumer becomes available.) Would you mind testing to see what that does to our throughput and memory usage? (I expect the throughput to be identical, with lower memory usage.)

confgenerator/confgenerator.go
franciscovalentecastro: Done.

confgenerator/confgenerator.go
franciscovalentecastro: Done. 

I was going to add the `db_storage` in the same change initially, so i needed a conditional for "If Non-Disk Buffered Receiver" -> "Then enabled sendind_queue persistence.". Not necessary in this PR.

confgenerator/confgenerator.go
franciscovalentecastro: > AIUI this will allow 2000 parallel batches in memory, each of which is allowed to have up to 2000 log records.

I don't think the `sending_queue` or `queue_size` refer to parallel batches in memory. From reading [^1], the `queue_size` refers to the elements set in `sizer`, which currently is set to `items` (logs in this case). The default value is `queue_size: 1000`, so i don't think there is too much additional memory usage (since we were already keeping 1000 logs in memory in queue).

I set to `2000` to match the `batch` parameters. All the performance experiments where done with this values.

I could go back to batch using `1000`, i was mostly experimenting to see if it helped to add `2000`.

> I don't think there's any advantage to setting the queue size larger than num_consumers + 1, which with the defaults would be 11. (This means every consumer can be active with a batch, and we can also have the next batch ready when a consumer becomes available.) Would you mind testing to see what that does to our throughput and memory usage? (I expect the throughput to be identical, with lower memory usage.)

Reducing the queue size to `num_consumers + 1` would definitely reduce memory usage, but i think it may impact the throughput.

[^1]: https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/exporterhelper#sending-queue

quentinmit: Just to close the loop on this, we decided that the default `sizer: requests` meant that the `queue_size` is intepreted as *number of incoming batches*, not number of outgoing batches. So with `queue_size: 2000`, the default `sizer: requests`, and the default behavior of the file receiver, it would limit to 2000 batches of 100 entries (the default file receiver batch size) each. That's nearly useless since we have no control over how the receiver happens to batch log entries, so we've switched to using `sizer: items` at both levels of the sending queue.

quentinmit (approved):

PR #2211, opened 2026-02-18
[integration test] Enable `TestSystemLogByDefault` test for Otel Logging.

franciscovalentecastro:
## Description
Enable the `TestSystemLogByDefault` integration test for Otel Logging.

Details : 
- `TestSystemLogByDefault` fails in `Debian 11` sometimes (~50%) when `/var/log/message` and `/var/log/syslog` file `fingerprint` is the same, e.g. The first 10 log lines are the same in both files, possibly done by `rsyslogd`.
    - Otel Logging only tails on of the files (e.g. `/var/log/messages` ), in this case causing `/var/log/syslog` to not be ingested. This results in a test failure.
    - Setting `fingerprint_size: 5kb` seems to alleviate/fixed the issues (i compared when the differences start).
    - This change causes more memory use.
    - See : https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver

## Related issue
b/413406335

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/agents/agents.go
+++ b/integration_test/agents/agents.go
@@ -238,8 +238,7 @@ func getOpsAgentLogFilesList(imageSpec string) []string {
 		}
 	}
 	return []string{
-		"/var/log/messages",
-		"/var/log/syslog",
+		gce.SyslogLocation(imageSpec),
 		OpsAgentConfigPath(imageSpec),
 		"/var/log/google-cloud-ops-agent/health-checks.log",
 		"/var/log/google-cloud-ops-agent/subagents/logging-module.log",


integration_test/ops_agent_test/main_test.go
```diff
@@ -2560,11 +2560,11 @@ func TestSystemdLog(t *testing.T) {
 
 func TestSystemLogByDefault(t *testing.T) {
 	t.Parallel()
-	gce.RunForEachImage(t, func(t *testing.T, imageSpec string) {
+	RunForEachImageAndFeatureFlag(t, []string{agents.OtelLoggingFeatureFlag}, func(t *testing.T, imageSpec string, feature string) {
 		t.Parallel()
 		ctx, logger, vm := setupMainLogAndVM(t, imageSpec)
 
-		if err := agents.SetupOpsAgent(ctx, logger, vm, ""); err != nil {
+		if err := agents.SetupOpsAgentWithFeatureFlag(ctx, logger, vm, "", feature); err != nil {
```
quentinmit, line 2567: Just to check my understanding, this test works because we're relying on `SetupOpsAgentWithFeatureFlag` to also enable the `experimental_otel_logging` setting?

confgenerator/logging_receivers.go
```diff
@@ -206,14 +206,14 @@ func (r LoggingReceiverFilesMixin) Pipelines(ctx context.Context) ([]otel.Receiv
 		"include_file_name":             false,
 		"preserve_leading_whitespaces":  true,
 		"preserve_trailing_whitespaces": true,
+		"fingerprint_size":              "5kb",
```
quentinmit, line 209: We're just sort of guessing that there will be a difference in the first 5kb, right? I feel like we should also report this as a bug upstream, since this isn't a crazy-uncommon scenario. (It could actually happen with any pair of log files, on any distro; after rotation, I think it would be easy for e.g. `auth.log` and `syslog` to contain the same first N log entries)

In the meantime, this seems like a reasonable workaround.

franciscovalentecastro: > We're just sort of guessing that there will be a difference in the first 5kb, right?

Yes, I've explored the "diff" of `/var/log/messages` and `/var/log/syslog` in Debian 11, and found that the first differences start showing in the range of the first 4kb to 5kb. Of course, this is not an statistical average from a sample, but it has shown to mitigate the failures.

> I feel like we should also report this as a bug upstream, since this isn't a crazy-uncommon scenario.

I will do that, i guess, since the `fingerprint` is a feature of the `filelogreceiver` it's more of unfavourable consequence of this file tailing strategy. Finding a way to mitigate this (e.g. "disable fingerprint checking") would be helpful.

integration_test/ops_agent_test/main_test.go
franciscovalentecastro: Yes, `SetupOpsAgentWithFeatureFlag ` sets both the environment variable and `experimental_otel_logging:true` in the config file.

confgenerator/logging_receivers.go
jefferbrecht: Wouldn't the two files have different inodes? Why is the fingerprint getting checked at all -- shouldn't the collector see that they are different files and process them independently & fully?

franciscovalentecastro: I found the design doc for this the `Fingerprint` feature [^1] which does not mention identifying logs by its `inodes`. It seems it guides fully by the fingerprint.

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.146.0/pkg/stanza/operator/input/file/design.md#deduplication-of-files

franciscovalentecastro: Created an upstream issue : https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46496.

PR #2214, opened 2026-02-20
Update IIS regex to also detect new format

jinghan-ma:
The Feb Update for Windows made a [change to the log format for IIS](https://learn.microsoft.com/en-us/iis/configuration/system.applicationhost/sites/sitedefaults/logfile/), where `BytesRecv` and `BytesSent` are included by default in the logExtFileFlags attribute on Windows 11 and Windows Server 2019 and later.

Updating the regex with an optional group at the end to catch this.

### DIFF
--- a/confgenerator/agentmetrics.go
+++ b/confgenerator/agentmetrics.go
@@ -57,9 +57,30 @@ var grpcToHTTPStatus = map[string]string{
 	"DEADLINE_EXCEEDED":   "504",
 }
 
+// Following reference : https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+var grpcToStringStatus = map[string]string{
+	"0":  "OK",
+	"1":  "CANCELLED",
+	"2":  "UNKNOWN",
+	"3":  "INVALID_ARGUMENT",
+	"4":  "DEADLINE_EXCEEDED",
+	"5":  "NOT_FOUND",
+	"6":  "ALREADY_EXISTS",
+	"7":  "PERMISSION_DENIED",
+	"8":  "RESOURCE_EXHAUSTED",
+	"9":  "FAILED_PRECONDITION",
+	"10": "ABORTED",
+	"11": "OUT_OF_RANGE",
+	"12": "UNIMPLEMENTED",
+	"13": "INTERNAL",
+	"14": "UNAVAILABLE",
+	"15": "DATA_LOSS",
+	"16": "UNAUTHENTICATED",
+}
+
 func (r AgentSelfMetrics) AddSelfMetricsPipelines(receiverPipelines map[string]otel.ReceiverPipeline, pipelines map[string]otel.Pipeline, ctx context.Context) {
 	// Receiver pipelines names should have 1 underscore to avoid collision with user configurations.
-	receiverPipelines["agent_prometheus"] = r.PrometheusMetricsPipeline()
+	receiverPipelines["agent_prometheus"] = r.PrometheusMetricsPipeline(ctx)
 
 	// Pipeline names should have no underscores to avoid collision with user configurations.
 	pipelines["otel"] = otel.Pipeline{
@@ -80,15 +101,15 @@ func (r AgentSelfMetrics) AddSelfMetricsPipelines(receiverPipelines map[string]o
 		Processors:           r.LoggingMetricsPipelineProcessors(),
 	}
 
-	receiverPipelines["ops_agent"] = r.OpsAgentPipeline()
+	receiverPipelines["ops_agent"] = r.OpsAgentPipeline(ctx)
 	pipelines["opsagent"] = otel.Pipeline{
 		Type:                 "metrics",
 		ReceiverPipelineName: "ops_agent",
 	}
 }
 
-func (r AgentSelfMetrics) PrometheusMetricsPipeline() otel.ReceiverPipeline {
-	return otel.ReceiverPipeline{
+func (r AgentSelfMetrics) PrometheusMetricsPipeline(ctx context.Context) otel.ReceiverPipeline {
+	return ConvertGCMSystemExporterToOtlpExporter(otel.ReceiverPipeline{
 		Receiver: otel.Component{
 			Type: "prometheus",
 			Config: map[string]interface{}{
@@ -129,11 +150,13 @@ func (r AgentSelfMetrics) PrometheusMetricsPipeline() otel.ReceiverPipeline {
 				),
 			},
 		},
-	}
+	}, ctx)
 }
 
 func (r AgentSelfMetrics) OtelPipelineProcessors(ctx context.Context) []otel.Component {
+	durationMetric := "grpc.client.attempt.duration"
 	filteredMetrics := []string{
+		"grpc.client.attempt.duration_count",
 		"googlecloudmonitoring/point_count",
 	}
 	pointCountMetric := otel.RenameMetric("googlecloudmonitoring/point_count", "agent/monitoring/point_count",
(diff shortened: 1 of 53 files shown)


apps/iis.go
```diff
@@ -196,7 +196,7 @@ func (p *LoggingProcessorIisAccess) Components(ctx context.Context, tag, uid str
...
 		// sample line: 2022-03-10 17:26:30 ::1 GET / - 80 - ::1 Mozilla/5.0+(Windows+NT+10.0;+WOW64;+Trident/7.0;+rv:11.0)+like+Gecko - 200 0 0 352
 		// sample line: 2022-03-10 17:26:32 ::1 GET /favicon.ico - 80 - ::1 Mozilla/5.0+(Windows+NT+10.0;+WOW64;+Trident/7.0;+rv:11.0)+like+Gecko - 404 0 2 49
-		Regex: `^(?<timestamp>\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}:\d{2})\s(?<http_request_serverIp>[^\s]+)\s(?<http_request_requestMethod>[^\s]+)\s(?<cs_uri_stem>\/[^\s]*)\s(?<cs_uri_query>[^\s]*)\s(?<s_port>\d*)\s(?<user>[^\s]+)\s(?<http_request_remoteIp>[^\s]+)\s(?<http_request_userAgent>[^\s]+)\s(?<http_request_referer>[^\s]+)\s(?<http_request_status>\d{3})\s(?<sc_substatus>\d+)\s(?<sc_win32_status>\d+)\s(?<time_taken>\d+)$`,
+		Regex: `^(?<timestamp>\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}:\d{2})\s(?<http_request_serverIp>[^\s]+)\s(?<http_request_requestMethod>[^\s]+)\s(?<cs_uri_stem>\/[^\s]*)\s(?<cs_uri_query>[^\s]*)\s(?<s_port>\d*)\s(?<user>[^\s]+)\s(?<http_request_remoteIp>[^\s]+)\s(?<http_request_userAgent>[^\s]+)\s(?<http_request_referer>[^\s]+)\s(?<http_request_status>\d{3})\s(?<sc_substatus>\d+)\s(?<sc_win32_status>\d+)\s(?<time_taken>\d+)(?:\s(?<sc_bytes>\d+)\s(?<cs_bytes>\d+))?$`,
```
jefferbrecht: `time_taken` should still be the final column:

```
jefferbrecht@IISTEST C:\inetpub\logs\LogFiles\W3SVC1>type u_ex260217.log
#Software: Microsoft Internet Information Services 10.0
#Version: 1.0
#Date: 2026-02-17 19:52:53
#Fields: date time s-ip cs-method cs-uri-stem cs-uri-query s-port cs-username c-ip cs(User-Agent) cs(Referer) sc-status sc-substatus sc-win32-status sc-bytes cs-bytes time-taken
2026-02-17 19:52:53 ::1 GET /forbidden something=something 80 - ::1 Mozilla/5.0+(Windows+NT;+Windows+NT+10.0;+en-US)+WindowsPowerShell/5.1.26100.32370 - 404 0 2 5035 184 192
```

apps/iis.go
```diff
@@ -270,6 +273,6 @@ func (r LoggingReceiverIisAccess) Components(ctx context.Context, tag string) []
 }
 
 func init() {
-	confgenerator.LoggingReceiverTypes.RegisterType(func() confgenerator.LoggingReceiver { return &LoggingReceiverIisAccess{} }, platform.Windows)
-	confgenerator.LoggingProcessorTypes.RegisterType(func() confgenerator.LoggingProcessor { return &LoggingProcessorIisAccess{} }, platform.Windows)
+	confgenerator.LoggingReceiverTypes.RegisterType(func() confgenerator.LoggingReceiver { return &LoggingReceiverIisAccess{} })
```
franciscovalentecastro: I've encountered this in the past, it's because we only use a "Linux" platform in tests. I've explored in the past adding Windows platform for transformation tests, but there was some push-back from the reviewers. 

Registering for both Linux and Windows will break some of the expected functionality (e.g. You shouldn't be able to enable `iis` in Linux)

I'd suggest doing the same we did in ActiveDirectory and register the `iis_access` processor here just for the transformation tests : 

https://github.com/GoogleCloudPlatform/ops-agent/blob/master/transformation_test/transformation_test.go#L716

apps/iis.go
```diff
@@ -191,12 +191,15 @@ const (
 
 func (p *LoggingProcessorIisAccess) Components(ctx context.Context, tag, uid string) []fluentbit.Component {
 	c := confgenerator.LoggingProcessorParseRegex{
+		// There's now 2 different default formats.
+		// The new format has new fields sc_bytes and cs_bytes added right before time_taken
 		// Documentation:
 		// https://docs.microsoft.com/en-us/windows/win32/http/w3c-logging
```
franciscovalentecastro, line 198: Is this documentation link still relevant to describe there is now 2 formats (old and new) ? 

Please update the documentation link accordingly in way that we have a clear reference for both old and new.

confgenerator/config.go
```diff
@@ -427,6 +427,20 @@ func (r *componentTypeRegistry[CI, M]) RegisterType(constructor func() CI, platf
...
+// Helper for Transformation Tests
+func (r *componentTypeRegistry[CI, M]) UnregisterType(constructor func() CI) {
+	if r.TypeMap != nil {
+		name := constructor().Type()
+		delete(r.TypeMap, name)
+	}
+}
+
+// Helper for Transformation Tests
+func (r *componentTypeRegistry[CI, M]) ReregisterType(constructor func() CI, platforms ...platform.Type) {
```
franciscovalentecastro: I like this solution, because this new method for `Unregister` and `Reregister` are only going to be used in transformation tests without having to add additional testing workarounds.

I'd suggest to consider an alternative name over `Reregister`. 

Some alternatives (i've got from Gemini 😆 ) are `OverrideType`, `ReplaceType` or `UpdateType` which could be better. I like `ReplaceType` because it's clear we are removing the old registration.

franciscovalentecastro (commented):
Added two comments. You may need to run `make test_confgenerator_update` to fix the current "Build" failures.

confgenerator/config.go
```diff
@@ -427,6 +427,20 @@ func (r *componentTypeRegistry[CI, M]) RegisterType(constructor func() CI, platf
 	r.TypeMap[name] = &componentFactory[CI]{constructor, platformsValue}
 }
 
+// Helper for Transformation Tests
+func (r *componentTypeRegistry[CI, M]) UnregisterType(constructor func() CI) {
```
jefferbrecht: Nit: this doesn't need to be exported

apps/iis.go
```diff
@@ -191,12 +191,15 @@ const (
 
 func (p *LoggingProcessorIisAccess) Components(ctx context.Context, tag, uid string) []fluentbit.Component {
 	c := confgenerator.LoggingProcessorParseRegex{
+		// There's now 2 different default formats.
```
jefferbrecht: Nit: there is only 1 default format; we're only retaining the old one for backwards compatibility Just In Case

apps/iis.go
jinghan-ma: Yeah, it's still the same "format", it's just that they added the new `sc_bytes` and `cs_bytes` before time taken. I don't think there's a snapshot of the documentation that we can refer to for the old "format"

apps/iis.go
jinghan-ma: True. Let me update the wording.

confgenerator/config.go
jinghan-ma: Yes. `ReplaceType` is a good name.

apps/iis.go
```diff
@@ -191,12 +191,16 @@ const (
...
+		// sample line old format: 2022-03-10 17:26:30 ::1 GET /iisstart.png - 80 - ::1 Mozilla/5.0+(Windows+NT+10.0;+WOW64;+Trident/7.0;+rv:11.0)+like+Gecko http://localhost/ 200 0 0 18
+		// sample line old format: 2022-03-10 17:26:30 ::1 GET / - 80 - ::1 Mozilla/5.0+(Windows+NT+10.0;+WOW64;+Trident/7.0;+rv:11.0)+like+Gecko - 200 0 0 352
+		// sample line old format: 2022-03-10 17:26:32 ::1 GET /favicon.ico - 80 - ::1 Mozilla/5.0+(Windows+NT+10.0;+WOW64;+Trident/7.0;+rv:11.0)+like+Gecko - 404 0 2 49
+		// sample line new format: 2026-02-19 10:28:49 ::1 GET /forbidden something=something 80 - ::1 Mozilla/5.0+(Windows+NT;+Windows+NT+10.0;+en-US)+WindowsPowerShell/5.1.26100.32370 - 404 0 2 5035 184 189
+		Regex: `/^(?<timestamp>\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}:\d{2})\s(?<http_request_serverIp>[^\s]+)\s(?<http_request_requestMethod>[^\s]+)\s(?<cs_uri_stem>\/[^\s]*)\s(?<cs_uri_query>[^\s]*)\s(?<s_port>\d*)\s(?<user>[^\s]+)\s(?<http_request_remoteIp>[^\s]+)\s(?<http_request_userAgent>[^\s]+)\s(?<http_request_referer>[^\s]+)\s(?<http_request_status>\d{3})\s(?<sc_substatus>\d+)\s(?<sc_win32_status>\d+)(?:\s(?<sc_bytes>\d+)\s(?<cs_bytes>\d+))?\s(?<time_taken>\d+)$/`,
```
jefferbrecht, line 203: Gemini tells me that the introduction of leading and trailing `/` will cause Fluent Bit to look for those characters literally. The transformation tests pass though. Is Gemini... wrong?

jefferbrecht: gemini wrong https://github.com/fluent/fluent-bit/blob/master/src/flb_regex.c#L132-L135

PR #2217, opened 2026-02-23
[Testing] Update GPU driver to CUDA 13.1.1 for Ubuntu/Debian

LujieDuan:
## Description
The new Ubuntu 2404 image has a newer kernel, so we need to upgrade the CUDA/Nvidia driver. 
- Upgrade from 12.9 to the newest 13.1.1;
- With CUDA 13.1, it has a new method to pin to a specific version of the driver; 
- With CUDA 13+, the demo app we use to generate load is no longer available from the repo; replaced with a simple gpu app;
- Keep other distros unchanged for now to first unblock tests - only updated Ubuntu/Debian. 

## Related issue
[b/486139680](http://b/486139680)

## How has this been tested?
Integration tests passing. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/third_party_apps_test/applications/dcgm/exercise
+++ b/integration_test/third_party_apps_test/applications/dcgm/exercise
@@ -1,6 +1,22 @@
 set -e
 
-# Run the bandwidthTest demo with a large range to create a process that uses
-# GPU for a period that is longer than default collection interval of 60s
-/usr/local/cuda/extras/demo_suite/bandwidthTest --memory=pinned --mode=range \
-  --start=1024 --end=20480 --increment=1
+# Run a CUDA program to create a process that uses GPU for a period that is 
+# longer than default collection interval of 60s
+if [ -x "/usr/local/cuda/extras/demo_suite/bandwidthTest" ]; then
+  # bandwidthTest from the CUDA demo package is available with CUDA 12 and earlier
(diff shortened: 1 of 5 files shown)


integration_test/third_party_apps_test/applications/dcgm/exercise
```diff
@@ -1,6 +1,22 @@
...
+    --start=1024 --end=20480 --increment=1
+else
+  # For CUDA 13+: 
+  # demos are no longer available as a package but require compiling from 
+  # GitHub source.
+  # bandwidthTest is also replaced by https://github.com/nvidia/nvbandwidth 
+  # which has dependencies such as Boost.
+  # Use gpu-burn instead - easy to compile and run. 
+  echo "CUDA 13+ environment detected. Using gpu-burn..."
+  git clone --depth 1 https://github.com/wilicc/gpu-burn
```
jinghan-ma, line 18: Does GPU burn work for Cuda 12 and earlier? If so, can we use it to test earlier versions as well?

LujieDuan: This app should work for all GPU models and CUDA versions, but it requires a different make flags to be compatible with some older GPUs. Will differ it to another PR and also want to upgrade to CUDA 13 for all distros as well.

PR #2239, opened 2026-03-18
Enable OTel Logging for unmodified default pipelines

quentinmit:
## Description
Use OTel Logging for `default_pipeline` iff the default pipeline has not been modified and the user does not specify `experimental_otel_logging: false`.

## Related issue
b/494017653

## How has this been tested?
Integration tests and new confgenerator test

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/confgenerator/testdata/goldens/logging-default_no_otel/golden/linux-gpu/otel.yaml
+++ b/confgenerator/testdata/goldens/logging-default_no_otel/golden/linux-gpu/otel.yaml
@@ -51,7 +51,7 @@ processors:
   filter/loggingmetrics_1:
     metrics:
       datapoint:
-      - metric.name == "grpc.client.attempt.duration_count" and (not IsMatch(datapoint.attributes["grpc.target"], "logging.googleapis"))
+      - metric.name == "grpc_client_attempt_duration_count" and (not IsMatch(datapoint.attributes["grpc_target"], "logging.googleapis"))
   filter/loggingmetrics_2:
     metrics:
       include:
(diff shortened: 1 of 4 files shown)


confgenerator/confgenerator.go
```diff
@@ -412,6 +411,10 @@ func (p PipelineInstance) OTelComponents(ctx context.Context) (map[string]otel.R
 	if err != nil {
 		return nil, nil, fmt.Errorf("receiver %q has invalid configuration: %w", p.RID, err)
 	}
+	gceMetadataAttributesProcessors, err := addGceMetadataAttributesProcessor(ctx).Processors(ctx)
+	if err != nil {
+		panic("Failed to generate static ModifyFields")
```
braydonk, line 416: Should we be surfacing the error here? Just wondering why the generic panic rather than returning the error or formatting the error into the panic.

confgenerator/config.go
```diff
@@ -1084,16 +1089,25 @@ func (uc *UnifiedConfig) loggingPipelines(ctx context.Context) ([]PipelineInstan
 	if err != nil {
 		return nil, err
 	}
+	platformDefaultConfig := BuiltInConfStructs[platform.FromContext(ctx).Name()].Logging
 	exp_otlp := experimentsFromContext(ctx)["otlp_logging"]
-	exp_otel := l.Service.OTelLogging
+	// N.B. Temporarily gate the "auto" otel logging behind an experiment flag.
```
braydonk, line 1094: What does `N.B.` stand for?

confgenerator/confgenerator.go
quentinmit: This condition indicates a code bug in Ops Agent; if it ever happens, it wouldn't be a condition that the user could resolve (since we're using a hard-coded input to the ModifyFields processor). I guess we could format `err` into the panic, but given that this is effectively a `never` condition, I figured it's fine to leave that for whoever debugs a future panic.

confgenerator/config.go
quentinmit: [Nota Bene](https://en.wikipedia.org/wiki/Nota_bene). Roughly "take notice" or "pay special attention to this".

submodules/opentelemetry-operations-collector
rafaelwestphal, line 1: What are you pointing to here?
I just updated this submodule

quentinmit: Great question. I think Git helpfully included a revert with the golden update. I've updated the commit to remove the submodule change.

PR #2262, opened 2026-04-07
Normalize Prometheus metric names for OTLP exporter and update tests

rafaelwestphal:
## Description

The original behavior of prometheus metric name is to replace dots and slashes with _. Currently otlp_exporter implementation breaks that behavior. Adding a processor to fix the breaking change.

## Related issue
b/476109839

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


confgenerator/otel/processors.go
```diff
@@ -689,6 +689,14 @@ func MetricUnknownCounter() Component {
 	})
 }
 
+// PrometheusMetricNormalize replaces dots and slashes with underscores in metric names.
+// This is to maintain behavior for Prometheus metrics when using the OTLP exporter.
+func PrometheusMetricNormalize() Component {
+	return Transform("metric", "metric", []ottl.Statement{
+		`replace_pattern(metric.name, "[./]", "_")`,
```
quentinmit: This does not come close to matching the existing behavior of the googlemanagedprometheusexporter, unfortunately. It replaces *all* unsupported characters with underscore, using https://github.com/prometheus/otlptranslator/blob/v1.0.0/metric_namer.go#L158

I *think* the equivalent OTTL regex might be `replace_pattern(metric.name, "[^a-zA-Z0-9:_]+", "_")`.

AFAICT, any string of one-or-more invalid characters is converted to a single underscore.

We should consider using the same upstream library (I guess via a custom OTTL function) to make sure we actually have the same behavior.

confgenerator/confgenerator.go
```diff
@@ -110,10 +110,9 @@ func ConvertToOtlpExporter(pipeline otel.ReceiverPipeline, ctx context.Context,
...
 		}
 
-		// b/476109839: For prometheus metrics using the OTLP exporter. The dots "." in the metric name are NOT replaced with underscore "_".
-		// This is diffrent from the GMP endpoint.
 		if isPrometheus {
 			pipeline.Processors["metrics"] = append(pipeline.Processors["metrics"], otel.MetricUnknownCounter())
+			pipeline.Processors["metrics"] = append(pipeline.Processors["metrics"], otel.PrometheusMetricNormalize())
 			// If a metric already has a domain, it will not be considered a prometheus metric by the UTR endpoint unless we add the prefix.
 			// This behavior is the same as the GCM/GMP exporters.
 			pipeline.Processors["metrics"] = append(pipeline.Processors["metrics"], otel.MetricsTransform(otel.AddPrefix("prometheus.googleapis.com")))
```
quentinmit, line 120: Nit: These three can all be a single call to `append` with four arguments.

quentinmit (commented):
What is our migration plan for cleaning up this tech debt in the future? What will we do when customers ask us to support utf-8 metrics or when GMP starts supporting them by default?

confgenerator/otel/processors.go
rafaelwestphal: Yeah, i'm still working on this PR. It can get messy very quickly with how the upstream translation is done.

rafaelwestphal: Fixed the implementation to match the behavior of GMP exporter.

confgenerator/otel/processors.go
```diff
@@ -689,6 +689,22 @@ func MetricUnknownCounter() Component {
 	})
 }
 
+// PrometheusMetricNormalize normalizes metric names to match Prometheus conventions.
+func PrometheusMetricNormalize() Component {
+	return Transform("metric", "metric", []ottl.Statement{
+		// 1. Replace all non-compliant characters (except :) with underscore
+		`replace_pattern(metric.name, "[^a-zA-Z0-9:]", "_")`,
+		// 2. Collapse multiple consecutive underscores
+		`replace_pattern(metric.name, "_+", "_")`,
```
quentinmit, line 698: AFAICT from reading the upstream code, multiple consecutive underscores are only collapsed when `WithMetricSuffixes` is true. That's [set to the `AddMetricSuffixes` config](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/googlemanagedprometheusexporter/config.go#L60) from the exporter, [which we set to `false`](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/confgenerator/confgenerator.go#L187).

So instead, I think we're in [this case](https://github.com/prometheus/otlptranslator/blob/v1.0.0/metric_namer.go#L185-L206):

```go
	// Simple case (no full normalization, no units, etc.).
	metricName := strings.Join(strings.FieldsFunc(name, func(r rune) bool {
		return !isValidCompliantMetricChar(r) && r != '_'
	}), "_")

	// Namespace?
	if mn.Namespace != "" {
		namespace := strings.Join(strings.FieldsFunc(mn.Namespace, func(r rune) bool {
			return !isValidCompliantMetricChar(r) && r != '_'
		}), "_")
		normalizedName = namespace + "_" + metricName
		return
	}

	// Metric name starts with a digit? Prefix it with an underscore.
	if metricName != "" && unicode.IsDigit(rune(metricName[0])) {
		metricName = "_" + metricName
	}

	normalizedName = metricName
	return
```

This does not collapse multiple underscores, nor does it remove a leading or trailing underscore.

I think you want, instead of 1-4,

```
replace_pattern(metric.name, "[^a-zA-Z0-9:_]+", "_").
```

which replaces consecutive invalid characters with a single underscore, without touching any existing underscores.

apps/otlp.go
```diff
@@ -142,6 +143,7 @@ func (r ReceiverOTLP) Pipelines(ctx context.Context) ([]otel.ReceiverPipeline, e
 		converter = confgenerator.ConvertPrometheusExporterToOtlpExporter
 	}
 	return []otel.ReceiverPipeline{converter(otel.ReceiverPipeline{
+		AllowUTF8: (r.AllowUTF8 != nil && *r.AllowUTF8),
```
quentinmit, line 146: Why is this a separate field in the `ReceiverPipeline` struct instead of the logic just being added to `metricsProcessors`?

apps/otlp.go
```diff
@@ -37,6 +37,7 @@ type ReceiverOTLP struct {
 
 	GRPCEndpoint string `yaml:"grpc_endpoint" validate:"omitempty,hostname_port" tracking:"endpoint"`
 	MetricsMode  string `yaml:"metrics_mode" validate:"omitempty,oneof=googlecloudmonitoring googlemanagedprometheus" tracking:""`
+	AllowUTF8    *bool  `yaml:"allow_utf8" validate:"omitempty"`
```
quentinmit, line 40: I believe this option also applies to the Prometheus receiver. But either way, we should think hard about whether and how we want to add a config knob for this. (If we want to eventually have UTF-8 on for everyone, any config knob should probably be experimental.)

PR #2305, opened 2026-05-19
Remove otel_logging feature flag from integration tests

rafaelwestphal:
## Description
OTel logging is now the default backend in confgenerator. The `otel_logging` feature flag (experiment) is redundant. This commit removes it from integration tests to simplify the test matrix and avoid redundant test runs.

- Removed `OtelLoggingFeatureFlag` and related helpers from `integration_test/agents/agents.go`.
- Updated `integration_test/ops_agent_test/main_test.go` to remove `OtelLoggingOTLPExporterFeatureFlag` and simplify `RunForEachImageAndFeatureFlag` calls.
- Fixed tests that assumed `default` was Fluent Bit to correctly expect OTel logging behavior.
- Fluent-bit is still present and will fallback to it when the pipeline is not fully supported by Otel Logging.
-
## Related issue
b/517100244

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


jinghan-ma:
> Fluent-bit is still present and will fallback to it when the pipeline is not fully supported by Otel Logging.

Why are we keeping Fluentbit as a fallback? Is it because we plan to remove it in a future PR?

rafaelwestphal:
> > Fluent-bit is still present and will fallback to it when the pipeline is not fully supported by Otel Logging.
> 
> Why are we keeping Fluentbit as a fallback? Is it because we plan to remove it in a future PR?

Short answer: yes. 

We still need to remove the things that fluent-bit is being used to fallback. There are 2:
- multiline
- logs via TCP

confgenerator/confgenerator_test.go
```diff
@@ -268,17 +268,32 @@ func generateConfigs(pc platformConfig, testDir string) (got map[string]string,
...
 	}
+	hasFluentBit := false
+	for _, p := range pipelines {
+		if p.Backend == confgenerator.BackendFluentBit {
+			hasFluentBit = true
+			break
+		}
+	}
+
+	if hasFluentBit {
```
jinghan-ma, line 283: Do we need to do this check? Wouldn't `GenerateFluentBitConfigs` return an empty map even if `hasFluentBit` is false and still have the same result (i.e. nothing gets added to `got`)

confgenerator/config.go
```diff
@@ -1162,8 +1154,32 @@ func (uc *UnifiedConfig) loggingPipelines(ctx context.Context) ([]PipelineInstan
...
+					for _, pr := range expandedProcessors {
+						if _, ok := pr.(InternalOTelProcessor); !ok {
+							pipelineSupportsOTel = false
+							break
+						}
+					}
+				}
+				if pipelineSupportsOTel {
+					for _, pr := range processors {
+						if _, ok := pr.Component.(OTelProcessor); !ok {
```
jinghan-ma, line 1174: Why is it that we have to check the `InternalOTelProcessor` for `expandedProcessors` but `OTelProcessor` here?

jinghan-ma (approved):

jinghan-ma (approved):
LGTM. Left two small comments.

confgenerator/confgenerator_test.go
rafaelwestphal: If we do not have this check, GenerateFluentBitConfigs will NOT return an empty map.

Even when there are no user logging pipelines that require Fluent Bit, the config generator still always appends the core self-metrics input/output and self-logs collection components (service.Component(), fluentbit.MetricsInputComponent(), generateSelfLogsComponents(), etc.) to the Fluent Bit configuration.

As a result, GenerateFluentBitConfigs would still return a minimal fluent_bit_main.conf configuration containing these self-logs/metrics pipelines. If we didn't have this hasFluentBit bypass in the test runner, we would be forced to keep a minimal fluent_bit_main.conf file inside every single OTel logging golden test directory in testdata/goldens/. By using the hasFluentBit check, we can cleanly omit Fluent Bit goldens from OTel-only test directories.

jinghan-ma: I see. Makes sense.

confgenerator/config.go
rafaelwestphal: InternalOTelProcessor only requires the Processors() method. Internally expanded processors (like the ones generated under-the-hood by Windows iis and mssql receiver macros) are generated dynamically in Go code and are never parsed from the user's config.yaml. Because they are purely internal, they do not need to be registered as user-facing components, so checking InternalOTelProcessor is sufficient and avoids requiring redundant Component registration boilerplate for internal macros.

OTelProcessor is a full user-facing interface that extends Component (which requires Type() and UnmarshalYAML() methods to be registered as a valid user-facing YAML component) and InternalOTelProcessor (which generates the final OTel collector pipeline representation via Processors()). User-facing pipeline processors configured in the service.pipelines block must implement OTelProcessor.

PR #2313, opened 2026-05-27
fix uap network check hang

XuechunHou:
## Description
This PR fixes a bug where the UAP plugin hangs and gets killed by the guest agent with a 'context deadline exceeded' error. This happens when a firewall blocks egress traffic to dl.google.com, causing the Start() execution to hang on the health check and preventing it from returning a response (reported in b/516831603).

### Root Cause
1. No HTTP Timeout in Network Healthchecks: The network checks use the default Go http.Get(), which does not configure a connection timeout. If a firewall silently drops egress packets to dl.google.com, the TLS handshake blocks indefinitely.

### Solution
1. **Timeout Configured**: Replaced http.Get with an http.Client that enforces a 10 * time.Second timeout per attempt. Capped at 1 retry limit. 
2. Timebound HealthChecks: Use context.WithTimeOut to time bound every HealthCheck.
4. **Concurrent Execution**: Refactored NetworkCheck.RunCheck to run all network requests concurrently using sync.WaitGroup. 
5. **Integration Test Timout Protection**: Added the `-max-time 60` flag to grpcurl commands in integration tests to  enforce a timeout on a grpc request. This matches production where context is configured to timeout (30 seconds as reported customer case). 

## Related issue
b/516831603

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


cmd/ops_agent_uap_plugin/service_linux.go
```diff
@@ -101,10 +103,17 @@ func (ps *OpsAgentPluginServer) Start(ctx context.Context, msg *pb.StartRequest)
...
+
+	// Trigger Healthchecks asynchronously.
+	go func() {
+		healthCheckFileLogger := healthchecks.CreateHealthChecksLogger(filepath.Join(pluginStateDir, LogsDirectory))
+		runHealthChecks(healthCheckFileLogger)
+	}()
+
 	// Subagent config generation
 	if err := generateSubagentConfigs(pContext, ps.runCommand, pluginInstallDir, pluginStateDir); err != nil {
 		ps.cancelAndSetPluginError(&OpsAgentPluginError{Message: fmt.Sprintf("Start() failed to generate subagent configs: %s", err), ShouldRestart: false})
```
jinghan-ma: Can we actually run the health checks asynchronously? If we trigger it here, and the runSubagents subroutine gets called shortly later, wouldn't there be the possibility of a false positive when it comes to things like port check? i.e. Ops Agent starts and binds to the target ports. Health Check detects that the ports are in use and reports an error (incorrectly, because the Ops Agent is what's using those ports).

Is it necessary to make Health Checks asynchronous if we have timeouts? Would it be more reasonable to have a global timeout for Health Checks instead?

internal/healthchecks/network_check.go
```diff
@@ -80,9 +81,12 @@ func (r networkRequest) SendRequest(logger logs.StructuredLogger) error {
 	bf := backoff.NewExponentialBackOff()
 	bf.MaxElapsedTime = MaxRequestElapsedTime
 	expTicker := backoff.NewTicker(bf)
+	client := &http.Client{
+		Timeout: 5 * time.Second,
```
jinghan-ma: Why did we choose 5 seconds? This seems on the low end to me?

internal/healthchecks/network_check.go
```diff
@@ -18,14 +18,15 @@ import (
...
+	"sync"
 	"time"
 
 	"github.com/GoogleCloudPlatform/ops-agent/internal/logs"
 	"github.com/GoogleCloudPlatform/ops-agent/internal/platform"
 	"github.com/cenkalti/backoff/v4"
 )
 
-const MaxRequestElapsedTime = 30 * time.Second
+const MaxRequestElapsedTime = 15 * time.Second
```
jinghan-ma: This assumes that each request timeout is handled by go right away after the timeout. If there's any gaps inbetween (e.g. due to scheduling), this would not be accurate.

I think it's more robust and easier to read to use `WithMaxRetries` instead.

cmd/ops_agent_uap_plugin/service_linux.go
XuechunHou: Good catch. reverted to synchronous execution

internal/healthchecks/network_check.go
XuechunHou: done.

internal/healthchecks/network_check.go
XuechunHou: guest agent would timeout the Start() call in 30 seconds. 

But I could change the default timeout to 1 min.

jinghan-ma (approved):

jinghan-ma (approved):

PR #1969, opened 2025-07-07
Timestamp format transformation in windowsEventLogV1Processors 

XuechunHou:
## Description
Transform RFC3339Nano timestamp into Cloud Logging compatible format: `%Y-%m-%d %T.%s +0000`

## Related issue

b/424474019 

## How has this been tested?

Transformation test

## Checklist:
- Unit tests
  - [X] Unit tests do not apply.
  - [ ] Unit tests have been added/modified and passed for this PR.
- Integration tests
  - [X] Integration tests do not apply.
  - [ ] Integration tests have been added/modified and passed for this PR.
- Documentation
  - [X] This PR introduces no user visible changes.
  - [ ] This PR introduces user visible changes and the corresponding documentation change has been made.
- Minor version bump
  - [X] This PR introduces no new features.
  - [ ] This PR introduces new features, and there is a separate PR to bump the [minor version](https://github.com/GoogleCloudPlatform/ops-agent/blob/master/VERSION) since the last [release](https://github.com/GoogleCloudPlatform/ops-agent/releases) already.
  - [ ] This PR bumps the version.

PR #2388, opened 2026-07-17
[chore] Update otelopscol to v0.159.0

franciscovalentecastro:
## Description
Update otelopscol to v0.159.0 and fix some `timestamp parsing` bugs in 3rd party apps.

Details : 
- Added opt-in cpu labels : https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49162
- Disable newly on by default metric : https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49325
- Added new possible values to `mysql.commands` metric : https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49863
- Field verification was improved in : https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48919
- Fixed timestamp parsing in some receivers.

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump


apps/couchbase.go
```diff
@@ -256,7 +256,7 @@ func (p LoggingProcessorMacroCouchbase) Expand(ctx context.Context) []confgenera
 						Regex: `^\[(?<type>[^:]*):(?<level>[^,]*),(?<timestamp>\d+-\d+-\d+T\d+:\d+:\d+.\d+Z),(?<node_name>[^:]*):([^:]+):(?<source>[^\]]+)\](?<message>[\s\S]*)$`,
 						Parser: confgenerator.ParserShared{
 							TimeKey:    "timestamp",
-							TimeFormat: "%Y-%m-%dT%H:%M:%S.%L",
+							TimeFormat: "%Y-%m-%dT%H:%M:%S.%LZ",
```
quentinmit: I think you want `%z` to parse this as UTC, or else it will be parsed as a local time, no?

apps/postgresql.go
```diff
@@ -134,7 +134,7 @@ func (p LoggingProcessorMacroPostgresql) Expand(ctx context.Context) []confgener
 						Regex: `^(?<time>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}.\d{3,} \w+)\s*\[(?<tid>\d+)\](?:\s+(?<user>\S*)@(?<database>\S*))?\s*(?<level>\w+):\s+(?<message>[\s\S]*)`,
 						Parser: confgenerator.ParserShared{
 							TimeKey:    "time",
-							TimeFormat: "%Y-%m-%d %H:%M:%S.%L %z",
+							TimeFormat: "%Y-%m-%d %H:%M:%S.%L %Z",
```
quentinmit, line 137: This is a fun one. strptime's `%z` is documented as "An RFC-822/ISO 8601 standard timezone specification. RFC-822 allows `"UT"  / "GMT" /  "EST" / "EDT" /  "CST" / "CDT" /  "MST" / "MDT" /  "PST" / "PDT" / 1ALPHA`. ISO 8601 allows only `Z` or `±nn:nn`. Notably absent from this list is `UTC`.

I added a test using `UT`, and glibc's strptime actually does not parse `UT` or `GMT` as valid `%z` specifiers.

So I'm convinced that `%Z` is indeed the correct specifier for these logs, and the fact that `%z` worked with Fluent Bit was a happy accident.

confgenerator/logging_receivers.go
```diff
@@ -958,7 +958,8 @@ func noFluentBitImplementation(ctx context.Context, tag, uid string) []fluentbit
 }
 
 func formatSystemTime(v ottl.LValue) ottl.Statements {
-	return v.Set(ottl.FormatTime(ottl.ToTime(v, "%Y-%m-%dT%T.%sZ"), "%Y-%m-%d %T.%s +0000"))
+	formattedTime := ottl.FormatTime(ottl.ToTime(v, "%Y-%m-%dT%T.%sZ"), "%Y-%m-%d %T.%s")
+	return v.Set(ottl.Concat([]ottl.Value{formattedTime, ottl.StringLiteral("+0000")}, " "))
```
quentinmit: Hm, is this guaranteed to actually produce a UTC time? Should we actually just be formatting with `%z` instead?

franciscovalentecastro: We were creating timestamps for Windows Event Logs using the output format `"%Y-%m-%d %T.%s +0000"`, which i assume because the `fluent-bit` windows event logs were always using `UTC` in the format `+0000`.

I set `%z` (which translates to Golang `Z0700`) which generates `Z` instead. I think this is better. We could somehow hack the format to `%z -> 0700` to get `+0000` but it doesn't seem worth it.

The original PR (https://github.com/GoogleCloudPlatform/ops-agent/pull/1969) i assume was trying to replicate exactly what Windows Event Logs API returns (AFAIU, xml logs are stored always with UTC [^1] ) and how fluent-bit event logs timestamps are displayed.

[^1]: https://superuser.com/questions/1531591/what-time-zone-does-windows-event-logs-show

apps/postgresql.go
franciscovalentecastro: Great!

apps/couchbase.go
franciscovalentecastro: Yes, you are right.

This also gives me insight that fluent-bit doesn't "fail to parse" a timestamp that has a "suffix" that is not within the described format. Being a bit more lenient that way.

LujieDuan:
fyi: found another issue with 0.156.0 that we will have to update: [PR 2395](https://github.com/GoogleCloudPlatform/ops-agent/pull/2395) - [Adding aggregation_type for combine](https://github.com/GoogleCloudPlatform/ops-agent/pull/2395/changes/8d3069165d9325bb168d974c890ac054f76ddca4)

franciscovalentecastro:
> fyi: found another issue with 0.156.0 that we will have to update: [PR 2395](https://github.com/GoogleCloudPlatform/ops-agent/pull/2395) - [Adding aggregation_type for combine](https://github.com/GoogleCloudPlatform/ops-agent/pull/2395/changes/8d3069165d9325bb168d974c890ac054f76ddca4)

Acknowledge! I was intending to update to 0.158.0 anyways. We can't update to 156 since it's missing the timestamp parsing fix. This PR hacks the fix, but it was released in 158.

quentinmit (approved):

PR #2394, opened 2026-07-27
test: fix TestRestartVM timeouts by preventing systemctl 10-line log truncation

rafaelwestphal:
## Description
Replaces `systemctl status google-cloud-ops-agent` with `journalctl -b 0 -u google-cloud-ops-agent --no-pager` in `getRecentServiceOutputForImage`.

In integration tests like `TestRestartVM`, `waitForExpectedHealthCheckResults` periodically polls `getRecentServiceOutputForImage` to confirm that `Network`, `Ports`, and `API` healthchecks have logged `[PASS]`.
1. **Fixes Log Truncation:** Previously, `systemctl status` capped output to the last 10 log lines. On service startup, after healthchecks printed `[PASS]`, sub-daemons (Fluent Bit & OTel Collector) frequently output >7 lines of startup logs, pushing the `[PASS]` messages out of `systemctl status`'s 10-line window. This caused false timeout failures even though healthchecks had passed.
2. **Prevents Pre-Reboot False Positives:** Using `-b 0` (`--boot=0`) ensures `journalctl` queries logs **strictly from the current post-reboot boot session**, ignoring any healthcheck pass messages generated prior to `gce.RestartInstance`.
3. 

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/integration_test/ops_agent_test/main_test.go
+++ b/integration_test/ops_agent_test/main_test.go
@@ -5331,13 +5331,13 @@ func waitForExpectedHealthCheckResults(ctx context.Context, logger *log.Logger, 
 
 	backoffPolicy := backoff.WithContext(backoff.NewConstantBackOff(5*time.Second), ctx)
 	return backoff.Retry(func() error {
-		cmdOut, err := gce.RunRemotely(ctx, logger, vm, getRecentServiceOutputForImage(vm.ImageSpec))
+		output, err := getHealthCheckResultsForImage(ctx, logger, vm)
 		if err != nil {
 			return err
 		}
 		for name, expected := range checks {
-			if !strings.Contains(cmdOut.Stdout, healthCheckResultMessage(name, expected, "")) {
-				return fmt.Errorf("expected %s check to %s in service output:\n%s", name, expected, cmdOut.Stdout)
+			if !strings.Contains(output, healthCheckResultMessage(name, expected, "")) {
+				return fmt.Errorf("expected %s check to %s in service output:\n%s", name, expected, output)
 			}
 		}
 		return nil
--- a/integration_test/ops_agent_test/main_test.go
+++ b/integration_test/ops_agent_test/main_test.go
@@ -5353,7 +5353,7 @@ func getRecentServiceOutputForImage(imageSpec string) string {
 		}, ";")
 		return cmd
 	}
-	return "sudo systemctl status google-cloud-ops-agent"
+	return "sudo journalctl -b 0 -u google-cloud-ops-agent --no-pager"
 }
 
 func getHealthCheckResultsForImage(ctx context.Context, logger *log.Logger, vm *gce.VM) (string, error) {


internal/healthchecks/network_check.go
```diff
@@ -55,12 +56,6 @@ var (
 			successMessage:   "Request to packages.cloud.google.com was successful.",
 			healthCheckError: PacApiConnErr,
 		},
-		{
-			name:             "dl.google.com",
-			url:              "https://dl.google.com/cloudagents/add-google-cloud-ops-agent-repo.sh",
-			successMessage:   "Request to dl.google.com was successful.",
-			healthCheckError: DLApiConnErr,
-		},
 	}
```
quentinmit, line 64: All of Google's frontends support `/generate_204` as a healthcheck endpoint. That is probably lighter-weight than fetching these actual files.

internal/healthchecks/network_check.go
```diff
@@ -77,7 +72,21 @@ func (r networkRequest) SendRequest(logger logs.StructuredLogger) error {
 		Timeout: 10 * time.Second,
 	}
 
-	response, err := client.Get(r.url)
+	backoffPolicy := backoff.WithMaxRetries(backoff.NewConstantBackOff(2*time.Second), 3)
```
quentinmit: This backoff is going to also delay Ops Agent startup by up to 6s, isn't it? Is that really considered acceptable?

internal/healthchecks/network_check.go
rafaelwestphal: I need to revert this change.. we were discussing changeing the endpoint to something else

internal/healthchecks/network_check.go
rafaelwestphal: Hmm it' s a good point.. but does it make sense to have Ops Agent up if we cant reach logs/metric endpoint? This is just  once on VM start

quentinmit: Yeah, if Ops Agent is up and running it can start collecting and buffering logs and metrics until the network comes up and they can be flushed.

rafaelwestphal: We don't buffer metrics iirc, but I get your point :) Any suggestions on how we can prevent this from being flake/false positive?

quentinmit: I believe the exporter will buffer metrics until the gRPC channel is healthy. (There might be a fairly short timeout?)

Do you have a specific distro on which you've seen `network-online.target` fail? Per [systemd's documentation](https://systemd.io/NETWORK_ONLINE/), `it indicates a configured, routable IP address of some kind.`, and on GCE that pretty much only happens when the network is up and a DHCP lease has been received.

Maybe we can fork off a separate process that will wait longer (60s?) for network checks, without blocking the main startup of OTel or Fluent Bit?

internal/healthchecks/network_check.go
rafaelwestphal: I reverted this change

internal/healthchecks/network_check.go
rafaelwestphal: I decided to drop the retry on this for now and only changed it so it gets all the logs from boot time.

integration_test/ops_agent_test/main_test.go
```diff
@@ -5353,7 +5353,7 @@ func getRecentServiceOutputForImage(imageSpec string) string {
 		}, ";")
 		return cmd
 	}
-	return "sudo systemctl status google-cloud-ops-agent"
+	return "sudo journalctl -b 0 -u google-cloud-ops-agent --no-pager"
```
avilevy18, line 5356: I don't think we want the `-b 0` not very familiar with it, but from my understanding it won't include the logs before the reboot, which I think we want in this case.

rafaelwestphal: We never want logs before the reboot. We only want logs for the current run/boot.

LujieDuan (approved):
LGTM

PR #2399, opened 2026-08-06
transformation_test: allow pipeline to drain before interrupting otelopscol

rafaelwestphal:
## Description

In `transformation_test`, `otelopscol` is signaled to terminate (`os.Interrupt`) as soon as the `fileconsumer` receiver logs the `"Consuming files"` log message 3 times.
On slower runners and emulated CPU environments (such as ARM64 `aarch64` under QEMU during Debian package builds), the signal was being delivered within 2ms of file consumption starting—shutting down the receiver and pipeline before the OTTL transformation statements and gRPC export batch could finish processing. This caused occasional empty payload drops and test failures (e.g. `TestTransformationTests/logging_processor-couchdb/otel-otlpexporter`).

## Related issue
b/543358344

## Checklist:
- Unit tests
- Integration tests
- Documentation
- Minor version bump

### DIFF
--- a/transformation_test/transformation_test.go
+++ b/transformation_test/transformation_test.go
@@ -28,6 +28,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


transformation_test/transformation_test.go
```diff
@@ -611,10 +611,14 @@ func (transformationConfig transformationTest) runOTelTestInner(t *testing.T, na
 			if strings.HasPrefix(msg, "Consuming files") {
 				consumingCount += 1
 				if consumingCount == 3 {
-					// We've processed the entire input file. Signal the collector to stop.
-					if err := cmd.Process.Signal(os.Interrupt); err != nil {
-						t.Errorf("failed to signal process: %v", err)
-					}
+					// We've processed the entire input file. Give the pipeline a brief window to finish
+					// transforming and exporting logs before signaling the collector to stop.
+					go func() {
```
franciscovalentecastro, line 618: Could we try other options before relying on a `time.Sleep(*)`? 

I've read that the `batch` processor as well as the `interval` processor do "gracefully" drain [^1] the signals on shutdown. I see two options we can try first :

- Set a `batchprocessor` at the end of the transformation test pipeline.
- Try with other `process signals` like `os.Stop`, since maybe `os.Interrupt` forces the process a lot. Idk about this :P stuff, so maybe it's already doing what we expect. Maybe trying to send `syscall.SIGTERM`.

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47238

jinghan-ma: I agree that we should at least try to do a graceful shutdown. Maybe we send a `SIGTERM` first. If the process fails to stop after some time, we can send a kill signal, but likely something went wrong if we can't gracefully shutdown in time.

jinghan-ma (approved):

transformation_test/transformation_test.go
rafaelwestphal: We already have a batch processor
https://github.com/GoogleCloudPlatform/ops-agent/blob/588c4b962526c18cc104d00a87ca83fcb9c19f94/transformation_test/otlp_helpers_test.go#L54

Processes cannot catch sigstop, only sigint or sigterm.

franciscovalentecastro: Maybe we can try with the `batchprocessor` set directly in the exporter, maybe that works better in shutdown. I remember the batch process was going to be deprecated in the favour of the exporter batch cofnig.

